PR #1, opened 2024-11-04
fixed extraction

gregpr07:


Resanso:
Error 401: No auth credentials found when trying to use the DeepSeek model with OpenRouter API. Please check if the API key is correctly set and accessible in the environment variables.

Rahwulkumar (commented):
noop

PR #26, opened 2024-11-18
Add ReturnTextAction

tgaldes:
Not sure if this is wanted in the main repo or the right way to do it, but my fantasy football team has lost 4 in a row and nothing you can say will hurt me more than that.

I could certainly use some direction regarding the test fixture that created a default Controller instance for me, adding a post instantiation 'set_output_func' that is only used in a test is a red flag.

### DIFF
--- a/browser_use/controller/service.py
+++ b/browser_use/controller/service.py
@@ -18,17 +18,23 @@ from browser_use.controller.views import (
 	ScrollDownAction,
 	SearchGoogleAction,
 	SwitchTabAction,
+	ReturnTextAction,
 )
 from browser_use.utils import time_execution_sync
+import pdb
 
 logger = logging.getLogger(__name__)
 
 
 class Controller:
-	def __init__(self, keep_open: bool = False):
+	def __init__(self, keep_open: bool = False, output_func: callable = None):
 		self.browser = Browser(keep_open=keep_open)
 		self.registry = Registry()
 		self._register_default_actions()
+		self._output_func = output_func
+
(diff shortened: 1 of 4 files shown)


MagMueller:
I think it's a great idea to give the agent the capability of outputting text.

So lets add the return_text to the Controller. 

Only add a better description into it in the registry - so that the agent knows for what to call this function.

And remove this output_func which you pass in - you don't need that - you can directly log the text in return text.

tgaldes:
My thought with the more generic output function is that the creator of the Controller instance can decide what to do with the output rather than being forced to come back and parse the log output. The Agent/Controller pair that is created can then communicate back to the calling process in a more structured way (or just a simple print if that's what you want)

How about I
- Default the output to a print instead of None
- Clean up the need for the having a setter function outside the constructor. Probably by using a different test fixture than the default in the test I added

emileindik:
I agree with @tgaldes that we need a way to pass in an optional callback in the cases where we want to do something with the output text other than just printing it to stdout. Like if the user wants to save the output text to a file for example.

MagMueller:
We just launched a feature where you can write custom functions 

e.g. in this example [example](https://github.com/gregpr07/browser-use/blob/main/examples/who_starred_my_repo.py)

you can give the agent the tool to save to a file. And you can just define the structure you want.

tgaldes:
now THAT is what I'm talking about! you see my flexible action and raise me flexible action CREATION. Let me play around with it and I'll probably close my PR

PR #39, opened 2024-11-21
Update example in README.md to be copy-paste-go

pietrozullo:
I had to google what an async function is in python, now one can just copy paste.


gregpr07:
Thanks Pietro!

PR #41, opened 2024-11-21
PromptVariables: Added custom defined variables to prompts for form input values

BedroomDev:
Added custom defined variables to prompts for form input values

```
variables = {
    "username": {
        "name": "username",
        "description": "The username which should be input",
        "value": "Test123"
    }
}

# Initialize controller
controller = Controller(variables=variables)

```

`test.py --task="go to http://localhost:3000 and input the @username"`

### DIFF
--- a/browser_use/controller/service.py
+++ b/browser_use/controller/service.py
@@ -20,13 +20,15 @@ from browser_use.controller.views import (
 	SwitchTabAction,
 )
 from browser_use.utils import time_execution_sync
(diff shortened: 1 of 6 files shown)


BedroomDev:
### Im sorry but **this does contain** the LogLevel Result changes

browser_use/controller/service.py
BedroomDev, line 1: PromptVariable changes only in this file

PR #43, opened 2024-11-21
Allow headless option when initializing controller

jeanmw:
Simple change: allow headless option by making it an optional argument to the Controller class. Headless option is already implemented in the Browser class, so all we have to do is pass that down.


MagMueller:
Great idea - does it work like expected for you?

jeanmw:
> Great idea - does it work like expected for you?
yes!

PR #53, opened 2024-11-24
Hot fix/max token message manager

MagMueller:


MagMueller:
This lets users specify max input tokens.

If the messages get too long, we first delete old messages. Then we remove the image of the current state. Then we cut the current state until we are in the limit.

For now, if models don't have a tokenizer I assume 1 token is approx 3 characters.

For images, I assume they have 800 tokens.

If you still get token errors with your model, decrease the limit.


When you use models with small context this can lead to that we lose important messages from the past.


One problem that could arise:
If a page is very very long and we are over the limit - we will delete first old messages.
Maybe it's smarter to first remove text from the current state.

MagMueller:
I changed it now so that the first the image and then the state is cut. Old messages are never cut out.

PR #61, opened 2024-11-26
Fix the task validation in agent service (10X efficiency!)

Shahar-Y:
Fixes Issue identified by @mahlertom in issue https://github.com/gregpr07/browser-use/issues/56. 

The issue adresses the problem with the function `get_next_action` in `Agent` where the AI invoke does not fully understand that it completed a task and does not ultimately call the `done` action. 
**It actually became much more efficient in solving some tasks!**
Instead of being stuck in a loop, it understands whether or not it has already solved a given task and continues running.

In order to solve this problem, I have inserted a prompt into the end of `input_messages` that emphasizes the agent's task and reminds it to call `done` when completed. 
I have tested it on the code that @mahlertom submitted and on dosens of my own tests over the past weeks since I opened the PR - it works perfectly. 

Cheers.


gregpr07:
Ah shit, sorry I completely missed this wonderful PR. @MagMueller we merged this without merging it right?

MagMueller:
Yes @gregpr07 We need this.

MagMueller:
Fixed in #235

MagMueller:
Thank you very much guys and sorry for the massive delay on this.

PR #66, opened 2024-11-28
Add the ability to click on iFrames and shadow elements

gregpr07:
This is hugely requested feature.
Also, increased DOM processing speed by 10x.


cheungseol:
iframes without same origin seems not work

PR #82, opened 2024-12-03
Fix support to remote browsers, add documentation and example

ramanidan:
As described in title, small bugfix in connection to wss based remote browser connection, along with clarification in the README and example


gregpr07:
I will merge this, but first I have to redesign the Browser service a little bit, I will switch back to only 1 context per 1 browser. I though it was gonna be a huge improvement, but well turns out it's pretty bad.

ramanidan:
@gregpr07 
That's great - We also tend to prefer having a single context

PR #106, opened 2024-12-20
fix headless behavior according to comment

yubinhu:
Comment: 
```
Default values:
		headless: True
			Whether to run browser in headless mode
```


gregpr07:
Now fixed in https://docs.browser-use.com/customize/browser-settings#browser-configuration

Issue #102, opened 2024-12-18
Clear git history to remove large file?


yubinhu:
Hi Gregor, would you like to clear git history to remove previously committed large files in .git? This is creating issues whenever someone clones the repo. 

The big file seems to be: ./.git/lfs/objects/ab/32/ab32dca74aff21e80c1d05457e61204216fd50109c6c8bdd158de4231c3cdaf0 

@TheInventorMan

gregpr07:
Will handle it, you can also push the PR if you want :)

RutamBhagat:
> Will handle it, you can also push the PR if you want :)

Hey @gregpr07 I would like to work on this issue issue

Jashwanth537:
hey @gregpr07 , I'm new to the codebase. Can I pick this up, if not solved yet?

Cfomodz:
This has been resolved - the large files were removed in PRs #2104 and #2105 (merged June 2025).

Current repository is only 39M, confirming the large file issues are fixed and cloning is now fast. Thanks to @yubinhu for raising this, to @RutamBhagat for volunteering to work on it, and @Jashwanth537 for bumping it and jumping in ready to help!

PR #107, opened 2024-12-22
Fix: Remove large file from Git history and implement Git LFS for test data

RutamBhagat:
Fixes: #102 

The main issue was that the file `tests/mind2web_data/processed.json` was directly committed to the Git history, causing the repository to be large and slow to clone.

To resolve this, I took the following steps:

1. **Removed `tests/mind2web_data/processed.json` from the Git history:** I used the `git filter-repo` tool to rewrite the history and remove this file from all past commits. This significantly reduces the repository size and improves cloning speed.

2. **Implemented Git LFS for `tests/mind2web_data/processed.json`:**  To ensure that this file (and any future large files) are managed efficiently, I have added it to Git Large File Storage (LFS). This means that the actual file content is stored separately, and Git only tracks a small pointer in the repository.

- The `.gitattributes` file was updated to track `tests/mind2web_data/processed.json` with LFS.
- The `tests/mind2web_data/processed.json` file itself now contains the Git LFS pointer.

**Important Note Regarding Force-Push:**

Rewriting Git history to remove the large file required a force-push. This force-push was performed on my forked repository during the history rewriting process. The current state of the `remove-large-file-from-git-history` branch reflects this rewritten history.


RutamBhagat:
Hi @gregpr07, about the issue #102, fixing it will need a force-push. It might be best if you handle this part yourself

PR #150, opened 2025-01-03
Encode select options properly for reliable AI selection

shyam-king:
I noticed a behavior where the `agent` was attempting to choose `select` options based on custom text instead of the exact text value in the option, hence failing to pick the correct option and/or failing completely because no options would match the exact text by AI. 

By returning the text and the value (encoded it to account for any encoding problems) and asking the AI to pick the correct value, enforces a match and thus improves the reliability of native select.


shyam-king:
Upon further I think a better thing to do is to encode the test content properly so that the AI understands the exact option properly, because the "value" attribute can have any random string that the devs of the website may use which might not be relevant for the AI (just like how it shouldn't be relevant for a typical use browsing the website)

Will revert the `value` thing and instead just encode the text content properly so that AI returns it properly for the select option. The problem I was facing was basically some text boxes have extra content which AI tends to ignore sometimes. Since, it is not easy to reproduce on demand I'm having trouble cooking up a quick example.

shyam-king:
Have showcased the issue in the [test file](https://github.com/browser-use/browser-use/blob/dcc3b3de290cc4c5375cfa13e753bc34c78484ee/tests/test_dropdown_complex.py)

PR #152, opened 2025-01-03
Added support for Browserbase to browser-use

warfaj:
Support for Browserbase's headless browser infra.
Added configuration options and example.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,6 +1,10 @@
 OPENAI_API_KEY=
 ANTHROPIC_API_KEY=
 
+# Optional Browserbase API key and project ID 
+BROWSERBASE_API_KEY=
+BROWSERBASE_PROJECT_ID=
+
 # Set to false to disable anonymized telemetry
 ANONYMIZED_TELEMETRY=true
 
--- a/README.md
+++ b/README.md
@@ -178,6 +178,8 @@ The most important options are:
 - `minimum_wait_page_load_time`: Minimum time to wait before getting the page state for the LLM input
 - `wait_for_network_idle_page_load_time`: Time to wait for network requests to finish before getting the page state
 - `maximum_wait_page_load_time`: Maximum time to wait for the page to load before proceeding anyway
+- `browserbase_api_key`: API key for Browserbase integration
+- `browserbase_project_id`: Project ID for Browserbase integration
 
 ## More examples
 
--- a/browser_use/browser/browser.py
+++ b/browser_use/browser/browser.py
@@ -5,6 +5,7 @@ Playwright browser on steroids.
 import asyncio
 import logging
 from dataclasses import dataclass, field
+from typing import Optional
 
 from playwright._impl._api_structures import ProxySettings
 from playwright.async_api import Browser as PlaywrightBrowser
@@ -39,6 +40,12 @@ class BrowserConfig:
 		chrome_instance_path: None
 			Path to a Chrome instance to use to connect to your normal browser
 			e.g. '/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome'
+
+		browserbase_api_key: None
+			API key for Browserbase integration
+
+		browserbase_project_id: None
+			Project ID for Browserbase integration
 	"""
 
 	headless: bool = False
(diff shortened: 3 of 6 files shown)


gregpr07:
I think it's nicer on the library level if Browserbase code is separate. You can achieve the same thing by passing in `wss_url`. Is there any other benefit than eliminating 5 lines of code? (there is also a bunch of products that are the same as Browserbase (anchorbase, steel, etc) so it's also less restricting. Let me know what you think!

waleedlatif1:
I tried using browser-use with browserbase via the wss_url approach, but encountered issues. Specifically, browser-use uses `.connect()` internally while browserbase requires `.connect_over_cdp()` for its websocket connections. This results in a KeyError: 'error' in Playwright's message handling.

Here's a minimal reproduction:

```
from browser_use import Agent, BrowserConfig, Browser

browser_config = BrowserConfig(
    wss_url='wss://connect.browserbase.com?apiKey=YOUR_API_KEY',
    disable_security=True
)

# This fails with KeyError: 'error' in Playwright's message handling
browser = Browser(config=browser_config)
```

The same connection works fine when using Playwright directly with `.connect_over_cdp()`. Perhaps we could support an optional connect_method parameter in BrowserConfig to handle these cases, or proceed with the implementation from the PR.

reustle:
Whether using `wss_url` or `cdp_url` I have issues using Browserbase. When using `wss_url`, I get the error described at the top. When using `cdp_url`, it tries to open a bunch of new tabs in the session and then fails. I don't see any log files being generated by Browser-use like it does when the run is successful.

I agree that we probably don't need to integrate BB specifically into this library, assuming the `wss_url` is repaired.

image
image

billyj854 (changes requested):
Help me get rid of hackers

e4stwood:
Also seem to be having issues with CDP across browserless & browserbase.
Same as @reustle - I get `Error occurred in event listener `with wss://connect.browserbase.com/?apiKey= (then immediately get 429 too many requests because browser-use keeps retrying. 

Also having trouble with browerless but unsure if thats a redherring. 

Any work arounds to use a cloud service?

mkuts12:
I had similar issues as @e4stwood and added a small delay in the `_initialize_session` function,
```
		page = await context.new_page()


		# Let Browser to finish connecting so we don't get 409
		await asyncio.sleep(2)

		# Instead of calling _update_state(), create an empty initial state

		initial_state = BrowserState(
```


Regarding the PR, I use 2 services for external browser, Browserbase among them, and it works with the CDP connection:
Here's the code to get the connection string:
```
async def get_browserbase_session() -> str | None:
	project_id = os.environ.get('BROWSERBASE_PROJECT_ID', '')
	browserbase_token = os.environ.get('BROWSERBASE_TOKEN', '')
	browserbase_stealth = os.environ.get('BROWSERBASE_STEALTH', '')

	url = "https://api.browserbase.com/v1/sessions"

	payload = {
		"projectId": project_id,
		"browserSettings": {
			"advancedStealth": browserbase_stealth == 'true'
		},
		"proxies": True
	}
	headers = {
		"X-BB-API-Key": browserbase_token,
		"Content-Type": "application/json"
	}

	response = await asyncio.to_thread(requests.post, url, json=payload, headers=headers)
	status_code = response.status_code
	if status_code == 429:
		raise HTTPException(
			status_code=status.HTTP_429_TOO_MANY_REQUESTS,
			detail=response.json()
		)
	elif status_code >= 400:
		raise HTTPException(
			status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
			detail={"source":"browserbase", "detail":response.json()}
		)
	return response.json()['connectUrl']
```

Here's how I use it with the agent,
```
	chat = ChatOpenAI(api_key=SecretStr(os.environ.get('OPENAPI_KEY', '')), model="gpt-4o")
	cdp_url = await get_browserbase_session()
	browser = Browser(config=BrowserConfig(
		cdp_url=cdp_url,
	))
	agent = Agent(
		task=task,
		browser=browser,
		llm=chat
       )
```
Something like that.

I think this PR is not needed anymore

pirate:
Closing this for now as it's now as many improvements have been made to the CDP connection flow since it was opened. If anyone is still encountering issues with remote browser providers please open new issues/PRs. In general we want to keep the codebase provider-agnostic and avoid adding specific branded config, most browser providers support connection by a normal WSS or CDP URL, so we shouldn't need platform-specific API connection handling and auth keys. Thanks all.

Issue #163, opened 2025-01-06
[BUG] select_dropdown_option does not work with react apps


shyam-king:
## Description

The JS used for dealing with drop downs (`<select>` elements) does not seem to handle react apps well.

### Example prompt:

```plaintext
go to https://codepen.io/shyam-king/pen/ByBJoOv and select tiger dropdown
```
### Expectation:
<img width="330" alt="image" src="https://github.com/user-attachments/assets/b9a9edd5-8f9a-4ffb-bd46-40a41584e243" />

The agent selects "Tiger" and the text element also displays "Tiger"

### Current Behaviour:
<img width="350" alt="image" src="https://github.com/user-attachments/assets/5ca4ae07-acea-4ad9-abdc-098fa4318217" />

The agent selects "Tiger" correctly but the corresponding changes to text element does not happen.

Cfomodz:
This was first but 370 was more generic, so deduplicating -> #370

PR #164, opened 2025-01-06
Fix: react handlers do not handle existing select implementation

shyam-king:
The current implementation of `Controller.select_dropdown_option` does not handle React applications well. Current implementation dispatches a "change" event on the select element. This is generally okay, but react does not attach listeners directly to the DOM elements. Instead react has listeners on the root element, and every event is expected to bubble to the root element for it to be handled. 

I have added `bubbles: true` to fix this issue.

Fixes issue described in #163  (have created a basic test case as well)

**EDIT [2024-05-07 11:18 IST]:** Just adding bubbles:true still did not work for few React websites (example: https://www.autotrader.co.uk/). Replaced the implementation with Playwright's [select_option()](https://playwright.dev/docs/api/class-locator#locator-select-option). This seems to work across frameworks as far as I have tested.


gregpr07:
@shyam-king wait, does this work or not? Also, could you please remove the unnecessary `pipfiles`.

gregpr07:
Is this a different approach to https://github.com/browser-use/browser-use/pull/150/?

shyam-king:
> Is this a different approach to #150?

Nope, these are independent

shyam-king:
> @shyam-king wait, does this work or not? Also, could you please remove the unnecessary `pipfiles`.

Works, removed Pipfiles

PR #190, opened 2025-01-08
Add Extract Element HTML action

franperezlopez:
This change allows saving some memory in the context, in cases we don't need to dump the whole page

### DIFF
--- a/browser_use/controller/views.py
+++ b/browser_use/controller/views.py
@@ -45,3 +45,8 @@ class ScrollAction(BaseModel):
 
 class SendKeysAction(BaseModel):
 	keys: str
+
+class ExtractElementHtmlAction(BaseModel):
+	index: int
+	format: Literal['text', 'markdown', 'html'] = 'html'
+	xpath: Optional[str] = None
--- a/browser_use/agent/prompts.py
+++ b/browser_use/agent/prompts.py
@@ -50,7 +50,9 @@ class SystemPrompt:
        {"go_to_url": {"url": "https://example.com"}},
        {"extract_page_content": {}}
      ]
-
+   - Extract element HTML: [
+       {"extract_element_html": {"index": 1}}
+     ]
 
 3. ELEMENT INTERACTION:
    - Only use indexes that exist in the provided element list
--- a/browser_use/controller/service.py
+++ b/browser_use/controller/service.py
@@ -18,6 +18,7 @@ from browser_use.controller.views import (
 	SearchGoogleAction,
 	SendKeysAction,
 	SwitchTabAction,
+	ExtractElementHtmlAction
 )
 from browser_use.utils import time_execution_async, time_execution_sync
 
@@ -47,6 +48,37 @@ class Controller:
 			msg = f'🔍  Searched for "{params.query}" in Google'
 			logger.info(msg)
 			return ActionResult(extracted_content=msg, include_in_memory=True)
+
+		@self.registry.action(
+			'Extract element HTML', param_model=ExtractElementHtmlAction, requires_browser=True
+		)
+		async def extract_element_html(params: ExtractElementHtmlAction, browser: BrowserContext):
+			session = await browser.get_session()
+			state = session.cached_state
+
+			if params.index not in state.selector_map:
+				raise Exception(
+					f'Element with index {params.index} does not exist - retry or use alternative actions'
+				)
+
+			element_node = state.selector_map[params.index]
+
+			try:
+				page = await browser.get_current_page()
+				element_handle = await browser.get_locate_element(element_node)
+				html_content = await page.evaluate('(element) => element.outerHTML', element_handle)
+				content = MainContentExtractor.extract(  # type: ignore
+					html=html_content,
+					output_format=params.format,
+				)
+				msg = f'📝 Extracted HTML for index {params.index}:\n{content}'
+				logger.info(msg)
+				return ActionResult(extracted_content=content, include_in_memory=True)
+			except Exception as e:
+				logger.warning(
+					f'Failed to extract HTML for element with index {params.index} - most likely the page changed'
+				)
+				return ActionResult(error=str(e))
 
 		@self.registry.action(
 			'Navigate to URL in the current tab', param_model=GoToUrlAction, requires_browser=True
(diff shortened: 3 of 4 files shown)


ankurt:
Swag, thanks for adding this. I'll give this branch a try.

franperezlopez:
great .. let me know if you have any problem

browser_use/controller/views.py
```diff
@@ -45,3 +45,8 @@ class ScrollAction(BaseModel):
 
 class SendKeysAction(BaseModel):
 	keys: str
+
+class ExtractElementHtmlAction(BaseModel):
+	index: int
+	format: Literal['text', 'markdown', 'html'] = 'html'
+	xpath: Optional[str] = None
```
shyam-king, line 52: why `xpath` though? I don't see it being used anywhere either.

franperezlopez: you're right ... this model is a mixture of `ClickElementAction` and `ExtractPageContentAction` models, so I kept the same philosophy. I noticed that the service methods using those models, only use the `xpath` for logging (`logger.debug(f'Element xpath: {element_node.xpath}')`), so I'm fine about removing the `xpath`

gregpr07:
Wait, what is the use case for this? Like, imo it's quite hard for the model to "know" which element it should extract. Do you have some examples which work really well?

ankurt:
> Do you have some examples which work really well?

I'm experiencing this now.

I found shadeform.ai works great, but it struggled here: https://solanacompass.com/validators in detecting the next pagination button.

gregpr07:
This makes me think -> is there a better way to do this? Like, could model kinda do the "tree search" to find proper html haha. Do you have any ideas?

ankurt:
There definitely are, but needs some creativity and experimenting. I think this is the crux of building a good browser agent. Decision making is important, but with graph states, and sufficient decision trees, teams can even hard code 70% of useful cases. Make sure it executes those decision trees consistently on ANY webpage it finds is the hard part.

Then moving from the 70% case => 99% can require higher order intelligent systems. But we're not even close to the 70% case yet.

Cfomodz:
@ankurt could a vector database be useful (and could the same possibly assist with local models being able to handle providing valid responses (by not filling up their context windows - at least that's what I'm assuming is ending up happening most of the time, I have not seen the logs)?

ankurt:
> @ankurt could a vector database be useful (and could the same possibly assist with local models being able to handle providing valid responses (by not filling up their context windows - at least that's what I'm assuming is ending up happening most of the time, I have not seen the logs)?

I'm not sure there's a need for any kind of RAG system at this point. If it's needed, we could just store an embedding in a python dictionary 'embedding' => 'content', and just loop to find closest embeddings. 

It's an interesting idea though. Maybe vector search with after pooling + encoding attributes from the DOM elements? Perhaps the underlying tokens in the DOM might contain some semantic fragments of their "purpose" automatically?

"search box" component, is likely an "input" tag, and will have a place holder of: "Search ..."

Vector search would nail that. But it will be useless for icon based buttons which frankly has been the issue for me so far.

Cfomodz:
Are you then passing those buttons to 4o (or another vision enabled model) to parse the DOM? I guess that is where I was thinking it could come in handy. If each call stored data/info about each element that it interacted with (or got info back from a vision model). I don't actually know that much about RAG beyond a user perspective; I just know it can store larger amounts of data than a models context window and then return the applicable bits. 

I know we're already doing that with the steps, in a way (doing something and then passing its status to the next call instead of passing the entire step), so maybe adding additional abstraction is unnecessary, I just can't stop thinking about some of the selenium scripts I've written and how sometimes the pages are thousands of lines of code, and the only reason it knows where to even be looking is because I told it the div to work within, or the type of the element or whatever it may be.

If it's able to keep things segmented by something like url, then it could easily store data about every element for every site, and return the data for the elements that seem as though they may apply?

I guess to stay with your search example, it would store - and again - keep in mind that I don't know what I am talking about:
website (base_domain, page_name, page_summary, page_purpose, page_elements=[list])
elements (id, xpath, class, purpose, type, noteworthy_css)
So then it becomes a perfect example where, if what I'm suggesting is possible, it would have 2 [or more] results for 'search' and then, comparing them with its memory would be able to know, without making another call to a vision model, that one is a search bar and one is offering search as a service.

Regarding icons, again, this seems particularly helpful, storing either id, filename, xpath and associating that with what it is (which it gets back from the vision model the first time it visits that page, or makes separate calls to an LLM to understand the DOM, put that into the vector store, and then, and only then, actually make the real task request to the llm (with the data from the vector store that it's already worked out the details of)). I often find myself shoving large chunks of HTML at 4o and telling it to write selenium code to do x and then y, so long as a, b, or c. It does fine, because it's only within that context. However, if I were to do what I did years ago (wild that I can say that) and go select the elements myself, and pass:
what I want done (click, copper .text, etc)
what element I want it done to (id, xpath, selector)
{repeated for each action}
then it could write my entire test in one call instead of one step at a time.

shyam-king:
For my particular use case, I added an action to extract hrefs of all interactive nodes

franperezlopez:
> For my particular use case, I added an action to extract hrefs of all interactive nodes [originally for the same purpose, the model would try to extract whole page content that'd go way beyond the token limit]. What's the purpose of extracting the whole HTML anyways?
> 
> You can have a look at the action [here](https://github.com/shyam-king/browser-use/commit/b1ea19b23c8c0cd3aaffe61212f5cef9fd65fe2f#diff-af6f6bc87d2d09d2a9f08f03fc959eb25b5637f37e959ec655daeba030ede225R340).
> 
> I did not make a PR yet, since I'm still unsure about this approach: I'm currently extracting href's of only the nodes that are getting highlighted since I expect the AI to potentially filter down the list and return (or memorize) only those links relevant to the task.

My case was very similar .. this is the prompt that I used:
"""
1. Open twitter bookmarks
2. From the first post, the date of the post contains a link to the post itself
3. Extract the link and add it to the database
"""
There is no need to parse the href in python, as the model already knows how to deal with html markup. so this action, can be used in tons of similar cases. The good thing you can try to iterate this for all the elements in the page that you need, without using the default page extract method, which can exceed the context window.

MagMueller:
In the current state we exclude links to save tokens. You can simply include them by passing this to the agent:

```python
include_attributes: list[str] = [
			'title',
			'href',
			'type',
			'name',
			'role',
			'tabindex',
			'aria-label',
			'placeholder',
			'value',
			'alt',
			'aria-expanded',
		],
``` 


The HTML can be super big (that's why we extract the elements) so I would not include this in the default actions.

MagMueller:
If you need this - you can just register this function.

PR #195, opened 2025-01-09
Added ollama example in  ollama.py

jbyerline:
Pretty self explanatory, adding in a working example for the use case of pointing ti local or remote ollama instance. Created with the help of this [Reddit](https://www.reddit.com/r/LocalLLaMA/comments/1huau1d/browser_use_running_locally_on_single_3090/) post.


gregpr07:
Thanks!

PR #201, opened 2025-01-10
Feature: Disallow Google AI Overview on search

Michaelgathara:
- Added Google param that turns off the Google AI overview feature
Removed the overview since browser use agents have a tendency to always use the data from them, thus defeating the purpose of digging deeper into multiple pages. This is especially useful when trying to get data within a certain time period, ie: 2025 where the Google AI overview will cite pages from far before 2020


Michaelgathara:
Could potentially make this a flag the user passes in when they initialize the agent as well, let me know if that is a more preferred approach

PR #210, opened 2025-01-10
fix: highlight elements arg in build-dom-tree script

pratikpakhale:
Currently the flag passed in the javascript evaluation does not work because the argument expected in javascript function is a `boolean` but the value passed while evaluating is a `list`.


Cfomodz:
Funny, I was wondering why that didn't work while solving #225 and didn't even notice this bug even though I looked right at that line. I ended up getting hacky so they could get their recording, but am glad someone caught the actual issue.

I have just tested this and can confirm is it working, and it also closes #225

PR #215, opened 2025-01-11
Example for deepseek

MagMueller:


123asdfghjkli:
i don't work  Failed to deserialize the JSON body into the target type: response_format: response_format.type `json_schema` is unavailable now at line 1 column 12440

PR #227, opened 2025-01-12
Added capability to show total token usage based on flag

saurabh2905:
This pull request introduces support to see the total token usage by Agent. 

**Key Changes**

- Added show_token_usage flag support to Agent class
- If the flag is set to true, then It collects the token usage from each LLM response

Sample Info log if flag is set to True:
   Total Token Usage: {'input_tokens': 8876, 'output_tokens': 593, 'total_tokens': 9469}

### DIFF
--- a/browser_use/agent/service.py
+++ b/browser_use/agent/service.py
@@ -19,6 +19,7 @@ from langchain_core.language_models.chat_models import BaseChatModel
 from langchain_core.messages import (
 	BaseMessage,
 	SystemMessage,
+	AIMessage,
 )
 from openai import RateLimitError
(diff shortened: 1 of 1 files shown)


MagMueller:
Awesome thank you very much.

The problem is that currently, our calculation is wrong of token usage.

E.g. we need to check if this works for 
- different models 
- with(out) vision
- with(out) tool_call_in_content=True, because currently we do not count the tokens inside tool_calls

MagMueller:
If this is urgent to you you can use lang smith

micheal888666:
I simply tested this code and it works, but it seems that the latest version has not been released?

PR #229, opened 2025-01-12
Added capability to use value from env if input text is like env vars

saurabh2905:
Added capability to look for value from env If the text value follows the format of environment variables. It can be used to input password from env. It will avoid sharing the password with the LLM/Model.

Example: 
   Task: Go to some website and login with username 'john.doe@website.com' and password 'SOME_WEBSITE_PASSWORD'

### DIFF
--- a/browser_use/controller/service.py
+++ b/browser_use/controller/service.py
@@ -1,5 +1,6 @@
 import asyncio
 import logging
+import os
 
 from main_content_extractor import MainContentExtractor
 from playwright.async_api import Page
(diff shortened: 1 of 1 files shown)


MagMueller:
To make it more robust can we wrap it inside 

```
<secret>WEBSITE_PASSWORD</secret>
```

Parse it with regex, and change the system prompt to tell the model how to handle passwords.

MagMueller:
I think the best way is to have a dictionary with the secrets, pass it to the agent, then to the multi_act function, then to act and replace inside the param dict all secrets. 

Then it also works for other functions then only input_text

pirate:
Closing for now because as it stands, this PR effectively gives the LLM access to env secrets, which would allow users to leak our LLM API keys and other secrets used in our cloud backend.

We need to give much more thought to how we handle this securely, I think a solution like what @MagMueller suggested is the right approach but it's going to take a bit of work to do properly.

Thanks for your work so far on this though @saurabh2905.

Issue #233, opened 2025-01-12
[Enhancement] Improve input state information about the content we cut out


MagMueller:
We do not include the entire website state, because its too big. We only include whats visable on screen and cut the rest.

Therefore, we should include information inside the state what we cut out. E.g. in the end of the state, we could write things like
- .... 100 words more  or
- ... 5 pages more or
- <cut content> use scroll to see more or
- for more content call extract content

shyam-king:
+1 on the scroll part

PR #234, opened 2025-01-13
Enhancement/improve input state information about the content we cut out

MagMueller:


MagMueller:
Fixed #233 by always saying that we cut something out.

To actually be certain that we cut something we need to include coordinates and check if unprocessed nodes are below current screen.

PR #302, opened 2025-01-18
implement highlight function

katiue:
help highlight the interacting elements that the agent is interacting with like filling forms, click button, etc.

### DIFF
--- a/browser_use/browser/context.py
+++ b/browser_use/browser/context.py
@@ -788,7 +788,7 @@ class BrowserContext:
 			tag_name = element.tag_name or '*'
 			return f"{tag_name}[highlight_index='{element.highlight_index}']"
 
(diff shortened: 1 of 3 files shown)


MagMueller:
awesome - for what are you using this?

to take a screenshot between model output and action execution? If we save that - we could take cool screenshots for the gif.

katiue:
it just highlight the interacting element so the user can keep track of the browser actions easier since it sometime causes confusion about what the agent did

katiue:
but your idea could be really cool if the the highlighted was screenshot but let's merge small things first so we can make it easier to work with

katiue:
@MagMueller this i what i mean

https://github.com/user-attachments/assets/cad66351-8326-47bc-831b-f428f3908437

MagMueller:
Awesome! lets improve the gif with the next pr.

PR #315, opened 2025-01-19
Hot fix/tool calling method

MagMueller:


MagMueller:
This is to automatically separate between different libraries to use function_calling or json mode

PR #317, opened 2025-01-19
Fix action names if no action is there

MagMueller:


MagMueller:
This solves the error when there is no action from the model returned

PR #353, opened 2025-01-22
added docs to monorepo

gregpr07:
previously the docs were at [in a separate repo][(https://github.com/browser-use/docs). Makes much more sense and easier to develop if they are inside a monorepo!


gregpr07:
@MagMueller could you also please update the docs to support
- pause and stop functionality
- Set highlight_elements False if you don't want to see the highlights
- support for gemini, qwen and other models like llama and granite3.1

Issue #238, opened 2025-01-13
Ollama example broken


arsen3d:
```
import os

# Optional: Disable telemetry
# os.environ["ANONYMIZED_TELEMETRY"] = "false"

# Optional: Set the OLLAMA host to a remote server
os.environ["OLLAMA_HOST"] = "http://localhost:11434"

import asyncio
from browser_use import Agent
from langchain_ollama import ChatOllama


async def run_search() -> str:
    agent = Agent(
        task="Search for a 'browser use' post on the r/LocalLLaMA subreddit and open it.",
        llm=ChatOllama(
            model="qwen2.5:32b-instruct-q4_K_M",
            num_ctx=32000,
        ),
    )

    result = await agent.run()
    return result


async def main():
    result = await run_search()
    print("\n\n", result)


if __name__ == "__main__":
    asyncio.run(main())
````
Terminal Log:
```
(base) PS C:\Users\arsen\repos\tools> python .\qwen.py   
INFO     [browser_use] BrowserUse logging setup complete with level info
INFO     [root] Anonymized telemetry enabled. See https://github.com/gregpr07/browser-use for more information.
INFO     [agent] 🚀 Starting task: Go to fahdmirza.com and return me the title of first blog post from home page.
INFO     [agent]
📍 Step 1
response {'raw': AIMessage(content='<tool_call>\n{"name": "go_to_url", "args": {"url": "https://fahdmirza.com"}}\n</tool_call>', additional_kwargs={}, response_metadata={'model': 'qwen2.5:latest', 'created_at': '2025-01-13T06:07:43.2262085Z', 'done': True, 'done_reason': 'stop', 'total_duration': 5119936400, 'load_duration': 4080449500, 'prompt_eval_count': 1798, 'prompt_eval_duration': 471000000, 'eval_count': 28, 'eval_duration': 232000000, 'message': Message(role='assistant', content='<tool_call>\n{"name": "go_to_url", "args": {"url": "https://fahdmirza.com"}}\n</tool_call>', images=None, tool_calls=None)}, id='run-a2d06cfe-f85d-4c93-ab65-e757a9893976-0', usage_metadata={'input_tokens': 1798, 'output_tokens': 28, 'total_tokens': 1826}), 'parsed': None, 'parsing_error': None}
ERROR    [agent] ❌ Result failed 1/3 times:
 Could not parse response.
INFO     [agent]
📍 Step 1
[...]

PR #373, opened 2025-01-24
Fix: ollama.py example with AgentHistoryList type

esthor:
**Fixes this issue**: https://github.com/browser-use/browser-use/issues/238

**Problem**: running the ollama example as is leads to an error:
```
ERROR    [agent] ❌ Result failed 1/3 times:
 Could not parse response.
```

**Solution**: Simply converting from a `str` type to the `AgentHistoryList` type gets it working.

**Clarifying Question & Proposed Future Work**: I'm not a python expert, and assumed types were optional, but is the reason this was failing due to some type checking? If so, perhaps a better error message regarding a type error could help in future cases.

**Shoutout**: Huge shoutout to @arsen3d for opening the issue and supplying working code!

### DIFF
--- a/examples/ollama.py
+++ b/examples/ollama.py
@@ -1,4 +1,4 @@
-import os
+# import os
 
 # Optional: Disable telemetry
 # os.environ["ANONYMIZED_TELEMETRY"] = "false"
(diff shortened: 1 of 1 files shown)


rjambrecic:
I don't think this solved the problem. 
This function does return `AgentHistoryList` and you have corrected the typing.

But the problem is still here. These models are flaky and if you run the test 10 times, it will fail at least 50% with:
```console
ERROR    [agent] ❌ Result failed 1/3 times:
 Could not parse response.
```

PR #374, opened 2025-01-24
Dont log an error when go back is not fully done

MagMueller:


MagMueller:
This pull request includes changes to the `browser_use/browser/context.py` file to modify the behavior of the `go_back` and `go_forward` methods. The changes primarily focus on adjusting the timeout settings and handling exceptions differently.

Changes to navigation methods:

* [`browser_use/browser/context.py`](diffhunk://#diff-02d6625e62a1cd7252a850b065c545a42620cf0aebfc595ccbe51f065aee9947L567-R581): Updated the `go_back` method to use a 10 ms timeout and changed the logging level from `warning` to `debug` for exceptions. The method no longer waits for the page and frames to load after navigating back.
* [`browser_use/browser/context.py`](diffhunk://#diff-02d6625e62a1cd7252a850b065c545a42620cf0aebfc595ccbe51f065aee9947L567-R581): Updated the `go_forward` method to use a 10 ms timeout and added exception handling with `debug` level logging. The method no longer waits for the page to load after navigating forward.

PR #375, opened 2025-01-24
Add button class to click

MagMueller:


MagMueller:
This pull request includes a small but important change to the `browser_use/dom/buildDomTree.js` file. The change adds a check for a specific class to determine if an element should be considered interactive.

* [`browser_use/dom/buildDomTree.js`](diffhunk://#diff-8f70f7a52b8e76e1c2669b77e09c25410507572b842d6503a101290ef135793fR159-R164): Added a check for the `address-input__container__input` class to the `hasInteractiveRole` condition to ensure elements with this class are recognized as interactive.

agrudkow:
@MagMueller could you elaborate on why such a hard-coded check was necessary i.e. why this class?

btw. this PR added a redundant whitespace

PR #387, opened 2025-01-25
Feature/add rag concept

MeshkatShB:
I added the RAG feature, so if there is a plus button on the cloud, you can now give additional information to the model as a file. Check example for more clarification. 

In the example folder, `read_from_file.py` file has been added. via `ChatOpenAI` call, you can create your llm. Afterwards, you can set your grocery list via a file (`grocery_list.txt`) and the model can do a prompt of:
```
"Go to instacart.com, search through my grocery list, add them to cart"
```
With this feature, you can give the model additional information.


MagMueller:
Thank you very much.

Is the query just the task?

MagMueller:
The use case of this is to upload large files, and the model retrieves the most relevant parts into the context.
This could be interesting to run agent 1, pause it and for agent 2 add the message history of agent 1 into the rag context.

MeshkatShB:
> Thank you very much.
> 
> Is the query just the task?

Yes, it's just the `self.task` but with retrieving it from the file as a query and adding the result to the prompt. But it can be modified. I wanted it to be a draft for another line of features. We can discuss more details if you're available.

MeshkatShB:
> The use case of this is to upload large files, and the model retrieves the most relevant parts into the context. This could be interesting to run agent 1, pause it and for agent 2 add the message history of agent 1 into the rag context.

Yeah. Good idea. It can have multiple use-cases and I'd be glad to discuss how it can be fitted into `browser-use`.

I was thinking about how in the Operator Demo, they gave a list of groceries (from image) and then prompted it to buy (or add to cart) from instacart.

sauravpanda:
@MeshkatShB We have moved away from langchain, and will have to close this PR. Feel free to create new one with any ideas you might have!

PR #391, opened 2025-01-25
Make telemetry opt-in.

jmccardle:
I was very disappointed to encounter this line in the console while trying out demo code:

`INFO     [root] Anonymized telemetry enabled. See https://github.com/browser-use/browser-use for more information.`

Yet there's not really any information in the description of the repo that would describe this; I had to search the files for "telemetry" and work backwards from there.

I don't have a relationship with `https://eu.i.posthog.com` and I don't want my software communicating with this service by default. Please accept this PR and make the default behavior to not collect telemetry.

### DIFF
--- a/browser_use/telemetry/service.py
+++ b/browser_use/telemetry/service.py
@@ -36,11 +36,14 @@ class ProductTelemetry:
 	_curr_user_id = None
 
 	def __init__(self) -> None:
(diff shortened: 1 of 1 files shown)


MagMueller:
I've updated the link for information to:
https://docs.browser-use.com/development/telemetry 

Just remove it if you don't like it. We will leave it like that to improve the product.

PR #413, opened 2025-01-26
updated path for example use-cases

FotieMConstant:
Hi i noticed the paths to the example code leads to 404. Most of the paths that led to 404 have been updated to link to the file on the main repo


MagMueller:
Thank you very much - saw this pr too late already merged:)

Issue #358, opened 2025-01-23
how to download a file?


agn-7:
### Bug Description

I wrote a controller to download some files, but it isn't invoked by the browser-use:

### Reproduction Steps

Clicks on the download buttons, but nothing happens

### Code Sample

```python
from dotenv import load_dotenv
from langchain_openai import ChatOpenAI
from browser_use import Agent, Browser, BrowserConfig
from browser_use.controller.service import Controller
import asyncio

load_dotenv()

controller = Controller()

@controller.action('Download PDF and CV, and use the task_id for the filename')
def download_file(url: str, filename: str):
    print(11111111111111111111111111111111111111111111111111111111111111111111111111111)
    try:
        import requests

        # Send GET request to download the file
        response = requests.get(url, stream=True)
        response.raise_for_status()  # Raise exception for bad status codes

        # Open file in binary write mode
        with open(filename, 'wb') as file:
            # Write the content in chunks to handle large files
            for chunk in response.iter_content(chunk_size=8192):
                if chunk:
                    file.write(chunk)

        return True, f"Successfully downloaded file to {filename}"

    except requests.exceptions.RequestException as e:
        return False, f"Failed to download file: {str(e)}"
    except IOError as e:
        return False, f"Failed to save file: {str(e)}"


async def main():
    browser_config = BrowserConfig(
        headless=False,
        disable_security=True,
        extra_chromium_args=["--enable-automation"]
    )

    # Set download path in new_context_config
    browser_config.new_context_config.accept_downloads = True
    # browser_config.new_context_config.downloads_path = DOWNLOADS_DIR

    browser = Browser(config=browser_config)

    agent = Agent(
        task="go to https://joblookup.com/uk/employer/login website, login with test@test.com username and test.com(2025) password. click on application, click on download button and download file, if there are many buttons for download click on all of them one by one",
        llm=ChatOpenAI(model="gpt-4o"),
        browser=browser,
        controller=controller,
    )
    result = await agent.run()
    print(result)

asyncio.run(main())
```

### Version

0.1.23

### LLM Model

GPT-4o

### Operating System

Ubuntu 22.04

### Relevant Log Output

```shell

```

agn-7:
Here is the recording video to show the issue:

https://github.com/user-attachments/assets/24f03530-035d-4f47-a4ba-92606d96dbca

j0yk1ll:
@agn-7 I added the functionality to my branch https://github.com/j0yk1ll/open-operator

[...]

PR #416, opened 2025-01-26
fix: enable downloads

j0yk1ll:
fix / solution for https://github.com/browser-use/browser-use/issues/358#issuecomment-2613983742

### DIFF
--- a/browser_use/browser/context.py
+++ b/browser_use/browser/context.py
@@ -76,6 +76,9 @@ class BrowserContextConfig:
 		save_recording_path: None
 			Path to save video recordings
 
+		save_downloads_path: None
+	        Path to save downloads to
+
 		trace_path: None
 			Path to save trace files. It will auto name the file with the TRACE_PATH/{context_id}.zip
 
@@ -108,6 +111,7 @@ class BrowserContextConfig:
 	no_viewport: Optional[bool] = None
 
 	save_recording_path: str | None = None
+	save_downloads_path: str | None = None
 	trace_path: str | None = None
 	locale: str | None = None
 	user_agent: str = (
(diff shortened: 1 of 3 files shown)


MagMueller:
Awesome thank you - is it possible to directly read the document content and return it via
`return ActionResult(extracted_content=content)` So the model knows whats inside

sanath-nova:
Is there an ETA for this to be merged? Seems like a great add-on that will reduce my workload significantly! 💯

sanath-nova:
@j0yk1ll - I tried your version. Seems to be working well, but while downloading multiple files in sequence, if the file name is same, it seems to be getting overwritten to the existing file rather than being downloaded as a new file. I think this is Chromium's behaviour and can be fixed too.

j0yk1ll:
> So the model knows whats inside

Oh I missed you comment. I believe this could become problematic because it totally depends on the file type. For certain predefined filetypes we could add such functionality.

j0yk1ll:
> Is there an ETA for this to be merged? Seems like a great add-on that will reduce my workload significantly! 💯

Until it is merged you can try my fork https://github.com/j0yk1ll/openoperator would be greatly appreciated

j0yk1ll:
> @j0yk1ll - I tried your version. Seems to be working well, but while downloading multiple files in sequence, if the file name is same, it seems to be getting overwritten to the existing file rather than being downloaded as a new file. I think this is Chromium's behaviour and can be fixed too.

I will look into it. Otherwise we could just append or prepend a timestamp to make it unique.

sanath-nova:
> I will look into it. Otherwise we could just append or prepend a timestamp to make it unique.

Yeah I made a workaround using watchdog already. This was just an fyi.

PR #422, opened 2025-01-26
deepseek-r1 compatibility

adsharma:
These changes improve compatibility with deepseek-r1:7b running on ollama.

Even with these changes, I'm not able to complete simple tasks. But hopefully it takes us closer.

### DIFF
--- a/browser_use/agent/message_manager/service.py
+++ b/browser_use/agent/message_manager/service.py
@@ -295,7 +295,10 @@ class MessageManager:
 			if isinstance(message, HumanMessage):
 				streak += 1
 				if streak > 1:
-					merged_messages[-1].content += message.content
+					if isinstance(message.content, list):
+						merged_messages[-1].content += message.content[0]['text']
+					else:
+						merged_messages[-1].content += message.content
 				else:
 					merged_messages.append(message)
 			else:
@@ -307,7 +310,7 @@ class MessageManager:
 		"""Extract JSON from model output, handling both plain JSON and code-block-wrapped JSON."""
 		try:
 			# If content is wrapped in code blocks, extract just the JSON part
-			if content.startswith('```'):
+			if '```' in content:
 				# Find the JSON content between code blocks
 				content = content.split('```')[1]
 				# Remove language identifier if present (e.g., 'json\n')
@@ -316,5 +319,5 @@ class MessageManager:
 			# Parse the cleaned content
 			return json.loads(content)
 		except json.JSONDecodeError as e:
-			logger.warning(f'Failed to parse model output: {str(e)}')
+			logger.warning(f'Failed to parse model output: {content} {str(e)}')
 			raise ValueError('Could not parse response.')
--- a/browser_use/agent/service.py
+++ b/browser_use/agent/service.py
@@ -352,8 +352,7 @@ class Agent:
 	@time_execution_async('--get_next_action')
 	async def get_next_action(self, input_messages: list[BaseMessage]) -> AgentOutput:
 		"""Get next action from LLM based on current state"""
-
-		if self.model_name == 'deepseek-reasoner':
+		if self.model_name == 'deepseek-reasoner' or self.model_name.startswith('deepseek-r1'):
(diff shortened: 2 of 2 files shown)


alan008:
Upvote for merging this. We need local run via ollama, where model is called "deepseek-r1:8b", not 'deepseek-reasoner' as hardcoded in service.py

browser_use/agent/message_manager/service.py
```diff
@@ -295,7 +295,10 @@ def merge_successive_human_messages(self, messages: list[BaseMessage]) -> list[B
 			if isinstance(message, HumanMessage):
 				streak += 1
 				if streak > 1:
-					merged_messages[-1].content += message.content
+					if isinstance(message.content, list):
+						merged_messages[-1].content += message.content[0]['text']
+					else:
+						merged_messages[-1].content += message.content
```
MagMueller, line 301: for what is this change?

adsharma: In my testing, `deepseek-r1:7b` doesn't follow the prompt and gives you text even if you asked for a list and vice-versa. This is defensive coding based on dynamic examination of what I got from the model in pdb.

Mistral is now advocating using pydantic based validation instead of this type of `ininstance()` checks.

alan008:
@adsharma could you please write, what you mean by "even with these changes, I'm not able to complete simple tasks". Is Browser interaction working with deepseek-r1:7b or not ? (with these changes merged to trunk)

adsharma:
Doesn't work for me (yet). Here's the error I get:

```
INFO     [agent] 🚀 Starting task: Compare the price of housing between Miami and New York
INFO     [agent] 
📍 Step 1
ERROR    [agent] ❌ Result failed 1/3 times:
 browser_use.agent.views.AgentOutput() argument after ** must be a mapping, not list
INFO     [agent]
```

repro: https://gist.github.com/adsharma/c0fc29b8c444900a00c8a7391e9aa29c

adsharma:
The problem is [this code](https://gist.github.com/adsharma/a132f1e1dedb0bd251db284eefc67329
) expects a dict. But `deepseek-r1:7b` outputs a list:

https://gist.github.com/adsharma/a132f1e1dedb0bd251db284eefc67329

The type checker catches this problem about `output.content`:

```
Argument of type "str | list[str | dict[Unknown, Unknown]]" cannot be assigned to parameter "text" of type "str" in function "_remove_think_tags"
  Type "str | list[str | dict[Unknown, Unknown]]" is not assignable to type "str"
    "list[str | dict[Unknown, Unknown]]" is not assignable to "str"
```

adsharma:
Sounds like a prompt engineering problem to me. I tried a few different prompts. Got something that's closer to the expected schema. But can't get past pydantic validation. Can share the prompts I tried in another PR.

adsharma:
Updated prompts here: https://github.com/adsharma/browser-use/commits/more_fixes/

Working better for me.

adsharma:
Have you considered `lm-format-enforcer` or similar models to enforce the pydantic schema on models that are not very good at conforming?

alan008:
There are two small distilled DeepSeek models on Ollama:

DeepSeek-R1-Distill-Qwen-7B
ollama run deepseek-r1:7b

DeepSeek-R1-Distill-Llama-8B
ollama run deepseek-r1:8b

Maybe try with second of them (based on Llama, not Qwen) ?

adsharma:
llama distilled 8b works a bit better, but still has trouble conforming to schema. Examples:

```
Failed to parse model output: {
	"current_state": {
		"evaluation_previous_goal": "Unknown - No previous actions to evaluate.",
		"memory": "",
		"next_goal": "Check weather for both cities"
	},
	"action": [
		{"name": "simulate_weather_check", {"Miami_temp": 85, "New_York_temp": 70}}
	]
}
```

```
WARNING  [message_manager] Failed to parse model output: {
  "current_url": "https://www.weather.com" // Example URL that provides weather data
  "action": {
    "name": "go_to_url",
    "args": {
      "url": "https://www.weather.com"
    }
  },
  "next_goal": "Extract temperature values from the website"
}
 Expecting ',' delimiter: line 2 column 44 (char 45)
 ```

PR #427, opened 2025-01-27
Add Azure OpenAI UI and model example

somethingwentwell:


SolsticeSpectrum:
Something went wrong, maybe don't use azure by default, perhaps an if check that uses openai if azure key cannot be found?

MagMueller:
I think its fine to have the azure model example, but I don't think we need this in the gradio ui

somethingwentwell:
> I think its fine to have the azure model example, but I don't think we need this in the gradio ui

Got it, just keep the Azure OpenAI example now

CakeRepository:
I disagree here it would be nice to have the Gradio example I'm confused

Issue #297, opened 2025-01-17
[Enhancement] Highlight elements with python


MagMueller:
Currently, we draw the colorful bounding boxes with Java script in buildDomTree.js, but we should store the coordinates of each DOM node and draw the bounding boxes later in Python to have more control over that.

This lets us highlight single elements to make much better GIFs and videos afterward. 
Also, users can see the screenshot after the run with and without the bounding boxes.

MagMueller:
This would solve e.g. #225

PR #444, opened 2025-01-28
added highlight manager in browser_use/dom

PaperBoardOfficial:
This PR resolves https://github.com/browser-use/browser-use/issues/297


MagMueller:
Hi @PaperBoardOfficial, thank you very much! 

We still execute JS code inside Python. What do you think about saving the coordinates of each element inside the domNodeElement? Then we can take the screenshot without bounding boxes and later, in Python, draw the specific boxes around the coordinates without the need to execute JS code. This way, we can, for example, modify the bounding boxes in the GIF based on the coordinates and the plain screenshot after the run. 

Then we could, for instance, show all bounding boxes in the GIF first, and then only the interacted element. Do you see the benefit of that? We would not be dependent on the browser being on that specific page.

PaperBoardOfficial:
@MagMueller I modified the code, now one can create gif with/without highlights
[gif_without_highlight](https://drive.google.com/drive/folders/1-8Taz1Hr2VxqfDsgGBPegoK9_AW4Baog?usp=sharing)
[gif_with_highlight](https://drive.google.com/drive/folders/1pB5NJ5uW80YZ1SPg2yke3BQ9TRA_sF1b?usp=sharing)

**changes done**: 
1. floated position data of elements from buildDomTree.js to DOMElementNode to DOMHistoryElement. Then this DOMHistoryElement is used in create_history_gif to highlight only the interacted element.
2. create gif_highlights_enabled flag in Agent class to enable/disable highlighting.
3. removed highlightElement code from buildDomTree.js
4. add gif_highlights_enabled=True in gemini.py just to give an example.

**questions:**
1. should i remove the VISUAL PROMPT (prompts.py) and remove_highlights() (context.py) because now we will not be creating highlight overlays in js code, hence we will not send highlighted screenshot to llm.
2. do you need **all** the elements highlighted like before or are you fine with the highlights on interacted elements only?
3. what else do you recommend?

MagMueller:
Awesome thank you very much ! I just pushed something which caused conflicts; I can fix them.

MagMueller:
@PaperBoardOfficial 
Great idea. The bounding boxes help the llm to understand the image. So in the best case we have them.

I just realized if we do it this way, the user cant see the bounding boxes inside the browser while the model is running. Many people love that.

PaperBoardOfficial:
Hey! could you tell me what should be done now? should i purge this PR? or should we go on with what was decided before i.e. merge the PR?

PaperBoardOfficial:
@MagMueller Pls tell me what should be done now, because this PR is becoming stale.

PaperBoardOfficial:
@Nivg The task was to shift the highlighting logic away from buildDomTree.js to create_history_gif() in python. This was done to save time during agent run, so that we can highlight the screenshots later on. So, benefit is to save time during llm communication.
But the downside is: if the user is using use_vision then he/she will send plain screenshot to llm,so llm won't see the highlighted boxes hence accuracy will decrease.

Nivg:
> @Nivg The task was to shift the highlighting logic away from buildDomTree.js to create_history_gif() in python. This was done to save time during agent run, so that we can highlight the screenshots later on. So, benefit is to save time during llm communication. But the downside is: if the user is using use_vision then he/she will send plain screenshot to llm,so llm won't see the highlighted boxes hence accuracy will decrease.

Maybe it's better to introduce highlight control separated from the dependency of creating the gif's? 
A general question (maybe it's not the place for that...), how much does the screenshots adds to accuracy in contract to just reading the DOM or just reding the DOM with the highlights added to it?

pirate:
@PaperBoardOfficial don't put hours into fixing these conflicts yet, we are still discussing the architecture needs for this one.

PR #446, opened 2025-01-28
fix: update a misleading description of the scroll_to_text action

agrudkow:
Fix misleading description of one of the scroll_to_text action


MagMueller:
This function is meant to scroll to something that is not inside the viewport. However, the model may not know that it exists, so it has not found it yet.

PR #484, opened 2025-01-30
Fixed instructions to install Playwright in README.md

chirag64:


Cfomodz:
At risk of exposing myself as a complete imposter: I don't get it...
You have to pip install and then playwright install, right? Also, when did npm come into this?

PR #491, opened 2025-01-31
Extend custom functions with user interaction and advanced browser interactions

shouryamaanjain:
Add new functions for user interaction and advanced browser interactions.

* **User Interaction Functions**
  - Add `ask_for_user_input` function to prompt the user for input and return the response.
  - Add `display_notification` function to show notifications to the user.
  - Register these functions using the `@controller.action` decorator with `requires_browser=True` parameter.

* **Advanced Browser Interaction Functions**
  - Add `handle_file_upload` function to upload files to a website.
  - Add `interact_with_iframes` function to interact with elements inside iframes.
  - Register these functions using the `@controller.action` decorator with `requires_browser=True` parameter.

* **Examples**
  - Create `user_interaction.py` to demonstrate the usage of `ask_for_user_input` and `display_notification` functions.
  - Create `advanced_browser_interactions.py` to demonstrate the usage of `handle_file_upload` and `interact_with_iframes` functions.

### DIFF
--- a/browser_use/controller/service.py
+++ b/browser_use/controller/service.py
@@ -434,6 +434,74 @@ class Controller:
 				logger.error(msg)
 				return ActionResult(error=msg, include_in_memory=True)
 
+		@self.registry.action(
+			description='Ask for user input',
+			requires_browser=True,
+		)
+		async def ask_for_user_input(prompt: str, browser: BrowserContext) -> ActionResult:
+			page = await browser.get_current_page()
+			user_input = await page.evaluate(f'prompt("{prompt}")')
+			msg = f'User input received: {user_input}'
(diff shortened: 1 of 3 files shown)


MagMueller:
Hey thank you - we dont want ask human in default actions, because some tasks should be completely autonomes.

For the other functions:
the model does not know the selectors, it only knows the index of the element so the parameters should be the index

I think those are great functions which you can register your self or your use-case

pirate:
See discussion here: https://github.com/browser-use/browser-use/pull/1486#issuecomment-2841945217

We are ok adding an example of asking the human for help as a custom action, but we do not want to add it to the default set of actions on the main controller.

PR #492, opened 2025-01-31
Update README.md with minimum Python version 3.11

mladen1nf:
There is a error that says:

```
ERROR: Could not find a version that satisfies the requirement browser-use (from versions: none)
ERROR: No matching distribution found for browser-use
```

if you don't have Python 3.11

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,9 +22,9 @@
 
 
 # Quick start
-Minimum Python requirement of 3.11
 
-With pip:
+
+With pip (Python>=3.11):
 
 ```bash
 pip install browser-use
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,7 +20,9 @@
 
 🌩️ Skip the setup - try our hosted version for instant browser automation! [Try it now](https://cloud.browser-use.com).
 
+
 # Quick start
+Minimum Python requirement of 3.11
 
 With pip:
```
MagMueller, line 27: can you write that here With pip (python 3.11)

PR #515, opened 2025-02-02
Add "wait_for_element" capability.

jersobh:
This PR should add the "wait_for_element" action and initial_action, with the optional timeout which defaults to 10 seconds.

It's useful for lazy-loaded websites, loading overlays or elements that only appears after a certain time. The current behavior is to scroll the page until the element is spotted, and it may miss the desired elements that will be no longer visible after scrolling.


MagMueller:
How does the model know the selector?

I like the idea of giving the model the option to wait longer

jersobh:
@MagMueller it knows the selector by the selector param:

```
async def wait_for_element(self, selector: str, timeout: float) -> None:
		"""
		Waits for an element matching the given CSS selector to become visible.
		Args:
		    selector (str): The CSS selector of the element.
		    timeout (float): The maximum time to wait for the element to be visible (in milliseconds).
		Raises:
		    TimeoutError: If the element does not become visible within the specified timeout.
		"""
		page = await self.get_current_page()
		await page.wait_for_selector(selector, state="visible", timeout=timeout)
```


```
await browser.wait_for_element(params.selector, params.timeout)
msg = f'👀  Element with selector "{params.selector}" became visible within {params.timeout}ms.'
```

Animaxx:
It looks great, are we going to merge this PR?

jersobh:
@pirate :)

pirate (approved):

pirate:
thanks!

PR #520, opened 2025-02-02
Add hover and evaluate actions

jersobh:
This PR adds hover and evaluate abilities to browser-use. It's now possible to run javascript functions on webpages (evaluate), which is very useful for scrolling, triggering specific functions or interacting with dynamic elements.

Also adds the hover ability, so it's abe to hover elements.

### DIFF
--- a/browser_use/browser/context.py
+++ b/browser_use/browser/context.py
@@ -11,7 +11,7 @@ import re
 import time
 import uuid
 from dataclasses import dataclass, field
-from typing import TYPE_CHECKING, Optional, TypedDict
+from typing import TYPE_CHECKING, Any, Optional, TypedDict
 
 from playwright.async_api import Browser as PlaywrightBrowser
 from playwright.async_api import (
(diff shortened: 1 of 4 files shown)


MagMueller:
@jersobh 
this is very powerful. Thank you very much.

Are there any downsides of such powerful code - e.g. if people run this in production - whats the worst the agent could do?

jersobh:
> @jersobh this is very powerful. Thank you very much.
> 
> Are there any downsides of such powerful code - e.g. if people run this in production - whats the worst the agent could do?

No, I don't think so. The evaluate gives the user some low-level power, as the user can inject specific functions or get specific outputs. And the hover option is really necessary for many situations. Scrolling is more prompt to break the extraction.

MagMueller:
I think this is good as an additional custom function, but not in the default function set.

jersobh:
@MagMueller sure, it's up to you. I, at least, use both functions daily basis.

PR #524, opened 2025-02-02
Add google spreadsheet support

### LINKED ISSUE #508: Google sheet support
### Problem Description

Currently, it's hard for the agent to navigate in Google Sheets. However, it would enable many new applications.

Image

### Proposed Solution

Iframes only work when we do not use the real browser, so first, you need to log in to Google, for example, with your cookie file. Then, find a clever way that browser use can utilize sheets. 

For example, only via the action send_keys (but this requires a lot of prompting). The best thing would be if we manage to detect each cell as a button (for that, you need to edit buildDomTree.js).

### Alternative Solutions

Maybe custom functions work for sheets, e.g. with

def input_in_cell(row, col, text, browser):
... and inside you hardcode how to access a row or col. 

### Additional Context

Check e.g. with langchain smith or directly our `input_messages` what information goes to the model - to understand what the current state in google sheets looks like.

jersobh:
Fixes #508 

 /claim #508

Basic operations, such as opening a google sheet, inserting values to cells, adding and deleting rows and inserting functions to cells.

It takes a spreadsheet url to get started.


saibalaji717 (approved):

j0yk1ll:
@jersobh 

From experience the agents are not good at respecting the order of actions that depend on each other, e.g. OpenGoogleSpreadsheetAction before InsertValueAction

Did you test the implementation on some tasks that involve spreadsheets? Is it relieable?

vinayakathavale15:
any update on this?

DroidNinja:
Can we merge this if its stable?

jersobh:
I can addore features, if needed.

Em sáb., 8 de mar. de 2025 às 20:05, Arun Sharma ***@***.***>
escreveu:

pirate:
We want to try to do everything through the browser, instead of using site-specific API adapter workarounds. We've added sheets support with keyboard shortcuts for now: https://github.com/browser-use/browser-use/pull/1101 but we may revisit this approach in the future if it doesn't work well.

Thanks for your work here, feel free to publish this in a repo of your own as it may still be useful for people!

PR #546, opened 2025-02-04
LiteLLM-like Facade

j0yk1ll:
Hey I added a litellm like facade that makes it simpler to change models. 

Before you always had to adjust the langchain imports. Now you just have to adjust the model string. I find this a big quality of life change.


PaperBoardOfficial:
@j0yk1ll Thanks for your contribution!
Could you revert the formatting changes. Its making it difficult to review the PR. Also, pls fix the merge conflicts.

Issue #541, opened 2025-02-03
[browser] Failed to update state: Failed to parse HTML to dictionary (eval of buildDomTree.js)


ticapix:
### Bug Description

Hello,

I'm having a weird behavior I can't debug.
How to enable logging debug or verbose to find why the browser can't parse the page "https://www.coretech.it/it/service/home/home.php" ?

Other webpages are working fine.

### Reproduction Steps

My `pyproject.toml` is rather simple

```ini
[project]
name = "browser-agent"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.11"
dependencies = [
    "browser-use>=0.1.35"
]
```

I'm running on Windows11 (powershell, not WSL2) , Python 3.11.

I've tried with OpenAI and Nebius and got the same error output.

### Code Sample

```python
#!/usr/bin/env python3
from pydantic import SecretStr
from browser_use.browser.context import BrowserContext, BrowserContextConfig
from browser_use.browser.browser import Browser, BrowserConfig
from browser_use import Agent
from langchain_openai import ChatOpenAI
from browser_use import Agent
import asyncio
from dotenv import load_dotenv
import os
import logging

load_dotenv()
NEBIUS_API_KEY = os.environ['NEBIUS_API_KEY']


log_level = logging.DEBUG
logging.basicConfig(
    level=log_level, format='%(asctime)s - %(levelname)s - %(message)s')
logger = logging.getLogger(__name__)


async def main(url: str):
    context = BrowserContext(
        browser=Browser(
            config=BrowserConfig(
                headless=True
            )),
        config=BrowserContextConfig(
            locale='en-US'
        )
    )

    # llm = ChatOpenAI(model="gpt-4o")
    # small_llm = ChatOpenAI(model='gpt-4o-mini', temperature=0.0)
    llm = ChatOpenAI(base_url='https://api.studio.nebius.ai/v1/',
                     model='meta-llama/Llama-3.3-70B-Instruct', api_key=SecretStr(NEBIUS_API_KEY),
                     temperature=0)
    small_llm = llm
    initial_actions = [
        {'open_tab': {'url': url}},
    ]
    task = """Browse the website and find the company name."""
    agent = Agent(task=task, initial_actions=initial_actions,
                  llm=llm, browser_context=context, use_vision=False, page_extraction_llm=small_llm)
    history = await agent.run()
    # print(history.urls())              # List of visited URLs
    # print(history.screenshots())       # List of screenshot paths
[...]

PR #582, opened 2025-02-06
detect if page.evaluate() works properly

ticapix:
error handling for https://github.com/browser-use/browser-use/issues/541


MagMueller:
fixed

PR #603, opened 2025-02-07
Fix Documentation . Added missing documentation to Gemini under supported models.

GODCREATOR333:


GODCREATOR333:
Made changes to documentation to correct pass the gemini_api_key to the SecretStr class using dotenv.

GODCREATOR333:
Made changes to documentation to correct pass the gemini_api_key to the SecretStr class using dotenv. To the file docs/customize/supported-models.mdx under the section create agent using gemini api.

Issue #601, opened 2025-02-07
Agent not closing intermediate tabs


rodrigobdz:
### Bug Description

The logs show that the agent closes tabs, but none is closed.

### Reproduction Steps

1. Run the code snippet
2. Expect `browser-use` to close intermediate tabs, before proceeding to open new ones
3. No intermediate tabs are closed

### Code Sample

```python
import asyncio
import os
from pathlib import Path

from browser_use import Agent, Browser, BrowserConfig
from langchain_openai import AzureChatOpenAI
from pydantic import SecretStr

# Configure the browser to connect to your Chromium instance
# Source: https://docs.browser-use.com/customize/real-browser
browser = Browser(
    config=BrowserConfig(
        # Specify the path to your Chromium executable
        chrome_instance_path="/Applications/Chromium.app/Contents/MacOS/Chromium",
    )
)

# Initialize the model
# Source: https://docs.browser-use.com/customize/supported-models#azure-openai
llm = AzureChatOpenAI(
    model="gpt-4o",
    api_version="2024-10-21",
    azure_endpoint=os.getenv(
        "AZURE_OPENAI_ENDPOINT"),
    api_key=SecretStr(os.getenv("AZURE_OPENAI_KEY")),
)


async def main():
    # Create the agent with your configured browser
    agent = Agent(
        task="Visit https://platform.openai.com/settings/organization/billing/history. Then, download invoices for the last month.",
        llm=llm,
        browser=browser,
    )
    await agent.run()


if __name__ == "__main__":
    asyncio.run(main())
```

### Version

0.1.36

### LLM Model

GPT-4o

### Operating System

macOS 15.2

### Relevant Log Output

```shell
INFO     [agent] 🎯 Next goal: Close the tab after download and revisit the main billing page tab to initiate the next action.
INFO     [agent] 🛠️  Action 1/1: {"send_keys":{"keys":"Control+W"}}
INFO     [controller] ⌨️  Sent keys: Control+W
```

iamarsenibragimov:
## Issue: Browser-use agent gets stuck in infinite loop when trying to close tabs

### Description
I'm experiencing a critical issue with browser-use where the agent cannot properly close tabs, even with extremely simple instructions. Instead of closing tabs as instructed, the agent gets stuck in an infinite loop switching between tabs.

### Steps to Reproduce
I'm using very basic instructions like:
1. Open a new tab with site X
2. Remember some text from the page
3. Close the tab
4. Navigate to site Y

[...]

PR #613, opened 2025-02-07
Fix Issue #601 - add function to close intermediate tabs

k-w-lee:
###  **Resolves:** #601   Fix: Ensure intermediate tabs are properly closed before opening new ones

This PR resolves https://github.com/browser-use/browser-use/issues/601
---

### **Issue Description**  
The agent was expected to close intermediate tabs before opening new ones, but logs showed that no tabs were actually closed. This caused multiple unused tabs to remain open, potentially leading to performance issues and resource leaks.  

#### **Reproduction Steps**  
1. Run the agent with multiple tabs opening in sequence.  
2. Observe that intermediate tabs are not being closed.  
3. Expectation: The agent should close all but the last opened tab before proceeding.  
4. Actual Result: Tabs remain open despite logs indicating closures.  

---

### **Fix Details**  
- Implemented `close_intermediate_tabs` method.  
- Retrieves the current browser session and context.  
- Ensures all but the last opened tab are closed.  
- Added a safeguard to check if a tab is already closed before attempting to close it.  
- Logs successful closures and warnings on failures.  

### **Affected File**
-  browser-use/browser_use/browser/context.py

Looking forward to feedback! 🚀

### DIFF
--- a/browser_use/browser/context.py
+++ b/browser_use/browser/context.py
@@ -585,6 +585,31 @@ class BrowserContext:
 			await self.switch_to_tab(0)
 
 		# otherwise the browser will be closed
+				
+	async def close_intermediate_tabs(self):
+		"""Closes all but the last opened tab."""
+		session = await self.get_session()
+		context = session.context
+
+		# Check if there are multiple pages
+		if len(context.pages) <= 1:
+			return  # No intermediate tabs to close
+
+		# Keep track of the last page
+		last_page = context.pages[-1] if context.pages else None
+
+		# Close all other pages except the last one
+		for page in context.pages[:-1]:
+			try:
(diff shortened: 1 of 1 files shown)


MagMueller:
Awesome, and then do you use custom functions to call this? What happens if the model opens 5 tabs, then switches to tab 3, and then we close tabs with your function? Does it break or work as intended? You want to close all tabs except the current one, right? Maybe you could have a parameter to specify which tabs to close, or all except the current one.

pirate:
Thanks for your work on this @k-w-lee, I added a simpler solution here: #1162 to just expose tab closing to the agent as an action, so if you want it to close tabs put it in your task prompt.

Issue #544, opened 2025-02-04
Tab Groups in Chrome


Freffles:
### Problem Description

browser use does not interact with chrome tab groups nor other UI elements that belong to the browser.

![Image](https://github.com/user-attachments/assets/cb5e6b28-4bef-4588-9d85-8529dcbf7c91)

### Proposed Solution

Would be great if chrome tab groups were able to be recognised and opened/closed. For example, Open the tab group "API Balance" and go to the first tab in the group and return the remaining balance.

Possibly interact with other UI elements.

https://developer.chrome.com/docs/extensions/reference/api/tabGroups?hl=en#description

SwapnilSonker:
want to know more about this issue @Freffles

pirate:
Blocked by https://github.com/browser-use/browser-use/issues/1563

#633 does nothing it's a red-herring

Cfomodz:
Chrome tab groups are part of the browser UI that's outside the page context, whether accessed through Playwright or CDP. 

Tab group manipulation would require Chrome extension APIs or OS-level automation, which is beyond the scope of web automation tools like browser-use.

PR #633, opened 2025-02-08
#544 issue resolved , used chrometabs api

SwapnilSonker:
use chrometabs api to group the tabs and altered controller and browser's context.py part

### DIFF
--- a/browser_use/browser/browser.py
+++ b/browser_use/browser/browser.py
@@ -193,6 +193,9 @@ class Browser:
 				'--no-default-browser-check',
 				'--no-startup-window',
 				'--window-position=0,0',
+                '--enable-experimental-extension-apis',
+                '--disable-extensions-except=/path/to/your/extension',  # If using an extension
+                '--load-extension=/path/to/your/extension'
 				# '--window-size=1280,1000',
 			]
 			+ self.disable_security_args
--- a/browser_use/browser/browser.py
+++ b/browser_use/browser/browser.py
@@ -194,8 +194,6 @@ class Browser:
 				'--no-startup-window',
 				'--window-position=0,0',
                 '--enable-experimental-extension-apis',
-                '--disable-extensions-except=/path/to/your/extension',  # If using an extension
-                '--load-extension=/path/to/your/extension'
 				# '--window-size=1280,1000',
 			]
 			+ self.disable_security_args
--- a/browser_use/browser/browser.py
+++ b/browser_use/browser/browser.py
@@ -212,7 +212,7 @@ class Browser:
 				'--no-default-browser-check',
 				'--no-startup-window',
 				'--window-position=0,0',
-        '--enable-experimental-extension-apis',
+				'--enable-experimental-extension-apis',
 				# '--window-size=1280,1000',
 			],
 			'firefox': [
--- a/browser_use/browser/context.py
+++ b/browser_use/browser/context.py
@@ -1120,39 +1120,6 @@ class BrowserContext:
 		session.cached_state = self._get_initial_state()
 		session.current_page = await session.context.new_page()
     
-	async def group_tabs(self, tab_ids: list[int] , title: str, color: str = "blue"):
-		"""Reset the browser session
-		Call this when you don't want to kill the context but just kill the state
-		"""
-		# close all tabs and clear cached state
-		page = await self.get_current_page()
-
-		js = f"""
-			chrome.tabs.group({{ tabIds: {tab_ids} }}, (groupId) => {{
-				chrome.tabGroups.update(groupId, {{ 
-					title: "{title}",
-					color: "{color}"
-				}});
-			}});
-			"""
-
-		await page.evaluate(js)
-
-	async def ungroup_tabs(self, tab_ids: list[int]):
-		"""Reset the browser session
-		Call this when you don't want to kill the context but just kill the state
-		"""
-		# close all tabs and clear cached state
-		page = await self.get_current_page()
-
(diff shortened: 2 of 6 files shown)


MagMueller:
I think we should not keep this: '--disable-extensions-except=/path/to/your/extension' in the default arguments with this path. You could pass that in as additional arguments.

Could you make it work with an ad blocker, which we could use by default?

MagMueller:
and i think this function is a good custom function, but not for all users

MagMueller:
Okey because if the browser does not open, then retry normally does not help right?

SwapnilSonker:
@MagMueller if the browser does not then the retry does work , if there is any issue i will try this and let you know, yeah we can try for the ad blocker cause it will be a type of mandatory feature to bypass the ad blocker.

MagMueller:
So, you say there are cases where retry works here?

But let's say there is an error; normally, we do the retry in the agent service.py.

MagMueller:
can you make a custom_function example with ungroup / group

SwapnilSonker:
changes have been made @MagMueller review it.

browser_use/browser/browser.py
```diff
@@ -193,6 +193,9 @@ async def _setup_standard_browser(self, playwright: Playwright) -> PlaywrightBro
 				'--no-default-browser-check',
 				'--no-startup-window',
 				'--window-position=0,0',
+                '--enable-experimental-extension-apis',
+                '--disable-extensions-except=/path/to/your/extension',  # If using an extension
```
PaperBoardOfficial, line 197: this will lead to error because path is not passed here.

browser_use/browser/context.py
```diff
@@ -1119,6 +1119,39 @@ async def reset_context(self):
 
 		session.cached_state = self._get_initial_state()
 		session.current_page = await session.context.new_page()
+    
+	async def group_tabs(self, tab_ids: list[int] , title: str, color: str = "blue"):
```
PaperBoardOfficial, line 1123: These functions can be removed from context.py and shifted to examples/custom-functions/group_ungroup.py because context.py contains default actions only. In the group_ungroup.py, once can easily access the playwright page using : `page = await browser.get_current_page()`

browser_use/browser/views.py
```diff
@@ -15,7 +15,13 @@ class TabInfo(BaseModel):
 	url: str
 	title: str
 
+class GroupTabsAction(BaseModel):
```
PaperBoardOfficial, line 18: These actions can be moved to examples/custom-functions/group_ungroup.py

browser_use/controller/service.py
```diff
@@ -433,6 +435,35 @@ async def select_dropdown_option(
 				logger.error(msg)
 				return ActionResult(error=msg, include_in_memory=True)
 
+		@self.registry.action(
+			description="Visually group browsers tab in chrome",
+			param_model=GroupTabsAction
+		)
+		async def group_tabs(self, params: GroupTabsAction, browser: BrowserContext):
```
PaperBoardOfficial, line 442: these can be removed from here, as these will be shifted to examples/custom-functions/group_ungroup.py.

browser_use/controller/views.py
```diff
@@ -42,6 +42,16 @@ class ScrollAction(BaseModel):
 class SendKeysAction(BaseModel):
 	keys: str
 
+class GroupTabsAction(BaseModel):
```
PaperBoardOfficial, line 45: Remove these as well, as they will be put in examples/custom-functions/group_ungroup.py

browser_use/browser/views.py
SwapnilSonker: imported them here, just for the alignment, will cause no error.

browser_use/controller/views.py
SwapnilSonker: is it necessary to shift them ??

SwapnilSonker:
@PaperBoardOfficial  @MagMueller , review it please.

browser_use/controller/views.py
PaperBoardOfficial: Yes, because this file contains models of only default actions.

browser_use/browser/views.py
PaperBoardOfficial: putting action model here is unnecessary, it might confuse developers later working on this file.

PaperBoardOfficial:
@SwapnilSonker I have added comment. Please check them.

SwapnilSonker:
@PaperBoardOfficial fulfilled all the comments , lets close this.

pirate:
wait this doesn't actually implement anything... where is the browser_context.group_tabs code?

SwapnilSonker:
> wait this doesn't actually implement anything... where is the browser_context.group_tabs code?
> 
> afaik the tab grouping API is not exposed over CDP so this is impossible without loading a custom browser extension to act as a bridge

I have to look on this , cause the last time I made the changes I used API key to implement that but , I can add a fallback if required.

pirate:
what do you mean? tab grouping is not something available via API key... it's a low level chrome JS API only exposed to extensions

SwapnilSonker:
> what do you mean? tab grouping is not something available via API key... it's a low level chrome JS API only exposed to extensions

Oh, it's like this. I apologise for the accidental red herring. I will look for a better alternative.

pirate:
Don't bother it's actually a very difficult issue because it requies a bridge extension, we're already developing it and we'll add this feature back when it's ready.

Issue #640, opened 2025-02-09
Empty <a> tags being sent to llm


PaperBoardOfficial:
### Bug Description

Currently, empty <a> tags are being sent to llm and this is causing two issues:
1. it is increasing unnecessary cost by using more tokens.
2. it is confusing the llm.

### Reproduction Steps

NA

### Code Sample

```python
NA
```

### Version

0.1.36

### LLM Model

Other (specify in description)

### Operating System

Windows 11

### Relevant Log Output

```shell
logs: https://drive.google.com/file/d/1yPav2S2eWmYOyLCq0jMaEFPiDaorYC7U/view?usp=sharing
```

PR #641, opened 2025-02-09
Fix redundant <a> tag

PaperBoardOfficial:
This PR resolves https://github.com/browser-use/browser-use/issues/640

The issue is that the <a> tags have a single child (the text node) but if the text node is not visible (isTextNodeVisible() returns false) the textnode is returned as null in builDomTree.js and later on when the user prompt is formed, the empty "<a><a/>" is sent to llm. This is increasing the token count and confusing the llm.

I fixed it by adding a check that if tag is <a> and child is null, then don't add the <a> to nodeData in buildDomTree.js
This change won't hamper other cases when <a> has img as child or anything else as child.

[before logs](https://drive.google.com/file/d/1yPav2S2eWmYOyLCq0jMaEFPiDaorYC7U/view?usp=drive_link)
[after logs](https://drive.google.com/file/d/1pOPZd20Uzl8UZQx0x0HI1Mxae6ay_zmU/view?usp=sharing)


MagMueller:
Awesome thanks - could you quickly resolve the conflict?

PR #688, opened 2025-02-12
Fixes #669 - adjusted exception type 

### LINKED ISSUE #669: Adding save_downloads_path in BrowserContextConfig hides indexes/dropdowns
### Bug Description

When creating a BrowserContextConfig, if you add save_downloads_path it will reduces indexing capabilities and close dropdowns automatically, decreasing quality in tasks that needs dropdowns. 

### Reproduction Steps

1. Install browser-use
2. Create a simple Browser
3.  Add a BrowserContextConfig with save_downloads_path
4. Run a task that needs dropdowns interaction

### Code Sample

```python

import asyncio
import os
from browser_use import Agent
from browser_use.browser.browser import Browser, BrowserConfig
from browser_use.browser.context import BrowserContextConfig

async def main():
    config = BrowserConfig(
                    headless=False,
                    disable_security=True,
                    # Comment the next 3 lines to make it work
                    new_context_config=BrowserContextConfig(
                        save_downloads_path="/home/tom/git-reps/RPA/test_workflows/docs",
                    ),
                )

    browser = Browser(config=config)

    prompt = """
        1. Enter https://formstone.it/components/dropdown/demo/
        2. You will see some examples labels and dropdowns. Your job is to select the second option 'Two', in the 'Basic' dropdown.
        """

    llm = ChatOpenAI()

    agent = Agent(
                task=prompt,
                llm=llm,
                browser=browser,
            )

    result = await agent.run()

    return

if __name__ == "__main__":
    asyncio.run(main())

```

### Version

0.1.36

### LLM Model

### Operating System

Ubuntu 24.04 LTS

### Relevant Log Output

```shell

```

tomlavez:
Exception was not expecting `TimeoutError`. Changed for a generic `Exception`

### DIFF
--- a/browser_use/browser/context.py
+++ b/browser_use/browser/context.py
@@ -959,7 +959,7 @@ class BrowserContext:
 						await download.save_as(download_path)
 						logger.debug(f'Download triggered. Saved file to: {download_path}')
 						return download_path
-					except TimeoutError:
+					except Exception:
 						# If no download is triggered, treat as normal click
 						logger.debug('No download triggered within timeout. Checking navigation...')
 						await page.wait_for_load_state()


MagMueller:
Mh whats the error - we dont want to catch here all exceptions, only if its timeout

tomlavez:
When no download is triggered, it isn't a TimeoutError Exception that is generated, so the exception is not caught and normal clicks are not treated as normal clicks

PR #712, opened 2025-02-13
Cleanup Code

maticzav:
This PR cleans up code by removing shallow functions and unused functions from the codebase. I've moved some stateless functions to utility files so that the linter can check whether they are dead code or still used.

### DIFF
--- a/browser_use/agent/message_manager/service.py
+++ b/browser_use/agent/message_manager/service.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import json
 import logging
 from typing import Dict, List, Optional, Type
 
@@ -8,7 +7,6 @@ from langchain_core.messages import (
 	AIMessage,
 	BaseMessage,
 	HumanMessage,
-	SystemMessage,
 	ToolMessage,
 )
 
@@ -34,6 +32,7 @@ class MessageManager:
 		max_actions_per_step: int = 10,
 		message_context: Optional[str] = None,
 		sensitive_data: Optional[Dict[str, str]] = None,
+		available_file_paths: Optional[List[str]] = None,
 	):
 		self.max_input_tokens = max_input_tokens
 		self.history = MessageHistory()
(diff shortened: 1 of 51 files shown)


browser_use/controller/service.py
```diff
@@ -433,67 +429,23 @@ async def select_dropdown_option(
...
-			results.append(await self.act(action, browser_context, page_extraction_llm, sensitive_data, available_file_paths))
-
-			logger.debug(f'Executed action {i + 1} / {len(actions)}')
-			if results[-1].is_done or results[-1].error or i == len(actions) - 1:
-				break
-
-			await asyncio.sleep(browser_context.config.wait_between_actions)
-			# hash all elements. if it is a subset of cached_state its fine - else break (new elements on page)
-
-		return results
```
MagMueller, line 490: I think we should not delete multi_act

browser_use/agent/message_manager/service.py
```diff
@@ -64,7 +57,9 @@ def __init__(
 			context_message = HumanMessage(content='Context for the task' + self.message_context)
 			self._add_message_with_tokens(context_message)
 
-		task_message = self.task_instructions(task)
+		task_message = HumanMessage(
+			content=f'Your ultimate task is: """{task}""". If you achieved your ultimate task, stop everything and use the done action in the next step to complete the task. If not, continue as usual.'
+		)
```
MagMueller, line 62: i think we should keep it as a function so that we can call it from outside for new tasks

browser_use/dom/views.py
```diff
@@ -152,11 +152,6 @@ def get_file_upload_element(self, check_siblings: bool = True) -> Optional['DOME
 
 		return None
 
-	def get_advanced_css_selector(self) -> str:
-		from browser_use.browser.context import BrowserContext
-
-		return BrowserContext._enhanced_css_selector_for_element(self)
-
```
MagMueller, line 159: i think we should keep this

PR #716, opened 2025-02-13
Feat: Add support for secondary PostHog telemetry configuration

PaulLockett:
I want to use Posthog for my telemetry, so I added a second client defined by env variables that extend the current telemetry logging.

### DIFF
--- a/browser_use/telemetry/service.py
+++ b/browser_use/telemetry/service.py
@@ -41,6 +41,7 @@ class ProductTelemetry:
 
 		if telemetry_disabled:
 			self._posthog_client = None
+			self._secondary_posthog_client = None
 		else:
 			logging.info(
 				'Anonymized telemetry enabled. See https://docs.browser-use.com/development/telemetry for more information.'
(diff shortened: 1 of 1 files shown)


browser_use/telemetry/service.py
```diff
@@ -56,11 +57,29 @@ def __init__(self) -> None:
...
+			if secondary_api_key and secondary_host:
+				logging.info(
+					"Secondary telemetry enabled. Metrics will also be sent to the secondary PostHog account."
+				)
+				self._secondary_posthog_client = Posthog(
+					project_api_key=secondary_api_key,
+					host=secondary_host,
+					disable_geoip=False,
+				)
+				if not self.debug_logging:
```
PaperBoardOfficial, line 72: @PaulLockett Thanks for the contribution!
Wouldn't it be better to have separate debug_logging flag for secondary posthog account. Because a user may want to silence the primary posthog but enable the secondary posthog's log.
Also, I think there should be separate ANONYMIZED_TELEMETRY flag for primary and secondary posthog, because one may want to disable primary and enable the secondary.

PR #723, opened 2025-02-14
Browser Task Automation with Gemini AI

falahgs:
add new example tool Browser Task Automation with Gemini AI


PaperBoardOfficial:
@falahgs Thanks for your contribution! But, Browser Use already has a gradio based web-ui : https://github.com/browser-use/web-ui. Also, Gemini is already supported by browser-use.

Issue #629, opened 2025-02-08
Visible element gets WARNING  [controller] Element not clickable with index 12 - most likely the page changed and gets back to the previous page


idunskiy:
### Bug Description

I'm testing browser use to help to automate the mailbox creation process. 
When I go to one of the pages where the model fills the fields, I then need to click the Continue button.
I gave it specific instructions, however everytime model clicks Back button instead of Continue.

### Reproduction Steps

1. Install browser use. 
2. Give the following prompt:
"You need to go to https://www.anytimemailbox.com/ and Choose Wyoming, United States in the field with Address or city placeholder. It will be a dropdown, and you choose Wyoming, United States from the list. Then you need to click the button to proceed. "
             "On the next page you need to choose a Casper plan. "
             "On the next screen you select Bronze, US $9.99 / month plan. "
             "On the next screen you enter Nina in the field with f_firstname id or First name, ignore Middle Initial field, enter Kobaladze in the field with f_lastname id and text Last name, and Best LLC in the field with f_company id and Company (Optional) placeholder, then you click CONTINUE button, index 12 and wait. If you receive Element not clickable with index 12 - most likely the page changed - just wait."

Result: model goes to the previous page.

### Code Sample

```python
async def agent_open_llc():
    agent = Agent(
        task="You need to go to https://www.anytimemailbox.com/ and Choose Wyoming, United States in the field with Address or city placeholder. It will be a dropdown, and you choose Wyoming, United States from the list. Then you need to click the button to proceed. "
             "On the next page you need to choose a Casper plan. "
             "On the next screen you select Bronze, US $9.99 / month plan. "
             "On the next screen you enter Nina in the field with f_firstname id or First name, ignore Middle Initial field, enter Kobaladze in the field with f_lastname id and text Last name, and Best LLC in the field with f_company id and Company (Optional) placeholder, then you click CONTINUE button, index 12 and wait. If you receive Element not clickable with index 12 - most likely the page changed - just wait. "
        llm=ChatOpenAI(model="gpt-4o"),
    )
    result = await agent.run()
    print(result)
```

### Version

0.1.36

### LLM Model

GPT-4o

### Operating System

macOS 12.0.1

### Relevant Log Output

```shell
INFO     [controller] ⌨️  Input Nina into index 4
INFO     [controller] ⌨️  Input Kobaladze into index 7
INFO     [controller] ⌨️  Input Best LLC into index 10
WARNING  [controller] Element not clickable with index 12 - most likely the page changed
[...]

PR #726, opened 2025-02-14
Made click element action more robust

PaperBoardOfficial:
Issue link: https://github.com/browser-use/browser-use/issues/629
The issue primarily occurred because the click element didn't acknowledge that the page navigated and it still continued to click on the existing element. Also, the code was not waiting for the page to be loaded completely. Also, it didn't have proper retry mechanism. So, without retry, it was just calling the llm again and again with below errors, which in turn was increasing token count as well.

These were the warnings and errors:
`WARNING  [controller] Element not clickable with index 12 - most likely the page changed`
`Error: Failed to click element: Page.evaluate: Execution context was destroyed, most likely because of a navigation`

So, these are the changes I did:
- created a new ElementClickHandler class to shift the click handling logic there.
- added code to handle retry (default is 3 tries).
- added code to handle popups.
- added code to handle navigation.
- all the timeouts and retries are configurable.

This is the prompt i used to test the code (I improved the prompt given by the user in corresponding issue):
[prompt](https://drive.google.com/file/d/19uiI8uJpdcJ5VfmebOtk8_ZIAfrS-dyn/view?usp=sharing)

These are additional file for reference:
[logs](https://drive.google.com/file/d/1I6NKlMzMdPeVA7ICgRQ9UYcITIl5IUcT/view?usp=sharing)
[gif](https://drive.google.com/file/d/1SUzT6jRmoG99V-QwGNRrIDcLxtDK-FjI/view?usp=sharing)

way forward:
the ElementClickHandler can be further improved to handle unnecessary overlays (other than popups) which sometimes cover the clickable elements and hamper the clicking functionality. This can be put in roadmap.


PaperBoardOfficial:
I updated the code.
This is the RCA: The issue was occurring due to this line: return await perform_click(lambda: element_handle.click(timeout=1500))
some websites like https://www.anytimemailbox.com/ take too much time to load after navigation (more than 1500 sec in our case), which was causing a timeout.
The solution was just increase the timeout, but that is not a good solution. So, i made the whole click element action as robust as possible.

I have a descriptive video on this issue: https://youtu.be/sqkk0Bpe8Bc

any suggestions are welcome.

PaperBoardOfficial:
@gregpr07 @MagMueller pls review this PR

all-mute:
up!

Shahar-Y:
This is a serious bug for me as well, would love to see it fixed

PaperBoardOfficial:
@Shahar-Y 
I already submitted a PR for it. But it is upto the repo owners to merge it. A lot of the prs are left to become stale even without reviewing. There's not even a slight bit of enthusiasm for external contributions. I am not going to come again and again to my old PRs to resolve merge conflicts.
Even now I am thinking that either:
1. one should fork browser-use or
2. create a fresh new repo/library (probably in typescript cuz it is more native to browser than python and it should be rest api based not a sdk client so that multiple languages can use it) or 
3. start contributing to nano-browser (web extension) or create your extension own using wxt.

pirate:
@PaperBoardOfficial we are reviewing this PR now and are still interested in it. There was a 1 month gap while the founders raised money and did some hiring, but coding is resuming now.

A few notes on the code:

- Can you move all the click functions back to the main browsercontext class, we will refactor them out in the future but for now we'd like to keep it in one place and avoid passing the getters like `get_current_page` to other classes. It should simplify it a bit as there will be no need for the class init and getter proxying logic.

- the async/await style of awaiting potential navigations/downloads is great, I like that design 👌 we should use it other places too many

- should `POPUP_OPENED` be a possible click result as well? or are we handling that like a navigation event?

Thanks, let me know your thoughts. If you decide you don't want to work on it anymore that's ok too, but it would be a shame because there are some good ideas here 🥲

PaperBoardOfficial:
Thanks for the encouraging words! I wonder why you talk so _sweetly_ 🤔
I refactored the code:

- moved the models and enum to views.py, moved the element_click_handler to context.py.
- deleted the models,enums folder and deleted elementclickhandler class.
- also now one can pass the timeout configuration to the browser context config. see examples/features/custom_click_config.py file. (all the timeouts are configurable and are in seconds)
- for popups, there's already a line in the system prompt which tells the llm to close a popup if it appears, but I felt that if we can handle it without calling the llm by just using js, it would be awesome, that's why I added it. On a side note, I added a default adblocker so that we can save time in popup ads (refer this PR https://github.com/browser-use/browser-use/pull/801).

pirate:
Great, this is getting a lot closer, thanks for those changes.

> for popups, there's already a line in the system prompt which tells the llm to close a popup if it appears, but I felt that if we can handle it without calling the llm by just using js, it would be awesome, that's why I added it. On a side note, I added a default adblocker so that we can save time in popup ads (refer this PR https://github.com/browser-use/browser-use/pull/801).

For now we prefer to not hardcode handling of popups, it's better to let the Agent do it. The reason is that `browser-use` is an extremely general tool, some users are using to QA their sites and test that the popup works for example, so we don't want to hardcode too much in python.

Could you remove that manual popup handling for now? Or if you prefer, instead of deleting you can move it to a new example with a custom action called `close_popups` that the agent can use as a tool if it sees a popup.

PaperBoardOfficial:
deleted the popup function

PaperBoardOfficial:
@pirate Any update on getting this merged? Should I resolve the conflicts again?

PaperBoardOfficial:
@pirate I have resolved the merge conflicts again. Will this PR get merged?

pirate:
We're actively working on this part of the codebase, will keep you posted.

PaperBoardOfficial:
up!

Issue #741, opened 2025-02-16
Wrong link for `extra_chromium_args` in Browser Settings page


RishabhKodes:
### Type of Documentation Issue

Broken link

### Documentation Page

https://docs.browser-use.com/customize/browser-settings#additional-settings

### Issue Description

The link to the `extra_chromium_args` points to a line of code which doesn't give the arguments https://github.com/browser-use/browser-use/blob/main/browser_use/browser/browser.py#L154, instead it should point to https://github.com/browser-use/browser-use/blob/main/browser_use/browser/browser.py#L180.

### Suggested Changes

The link to the chromium args should be changed from https://github.com/browser-use/browser-use/blob/main/browser_use/browser/browser.py#L154 to https://github.com/browser-use/browser-use/blob/main/browser_use/browser/browser.py#L180

Varun270:
I would like to work on this issue, Kindly assign this to me!!

Cfomodz:
This documentation issue has been resolved through codebase restructuring.

Specifically, the original browser.py file referenced in the issue no longer exists - the codebase was migrated to CDP architecture.

PR #746, opened 2025-02-17
Doc fix: Fixed broken link to extra_chromium_args

RishabhKodes:
The link to the `extra_chromium_args` in the documentation https://docs.browser-use.com/customize/browser-settings#additional-settings, points to a line of code which doesn't give the arguments https://github.com/browser-use/browser-use/blob/main/browser_use/browser/browser.py#L154. 

Instead, it should point to https://github.com/browser-use/browser-use/blob/main/browser_use/browser/browser.py#L180.

### DIFF
--- a/docs/customize/browser-settings.mdx
+++ b/docs/customize/browser-settings.mdx
@@ -28,21 +28,21 @@ config = BrowserConfig(
 
 ## Core Settings
 
-- **headless** (default: `False`)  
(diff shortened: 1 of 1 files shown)


RishabhKodes:
Fixes: https://github.com/browser-use/browser-use/issues/741#issuecomment-2662384312

PR #766, opened 2025-02-18
Dev

MagMueller:


browser_use/controller/service.py
```diff
@@ -25,16 +26,18 @@
 from browser_use.utils import time_execution_sync
 
 logger = logging.getLogger(__name__)
-from langchain_core.language_models.chat_models import BaseChatModel
 
 
-class Controller:
+Context = TypeVar('Context')
```
pirate, line 32: @maticzav what is this used for? I don't see anywhere in the codebase whwere we use the `context: Context` kwarg, attr, or TypeVar

PR #795, opened 2025-02-21
Adding parameter to Agent.run() in order to allow the developer to execute a custom Python function after Agent.step()

carlosplanchon:
I'm implemented the solution of the problem I described on this issue.

### DIFF
--- a/browser_use/browser/context.py
+++ b/browser_use/browser/context.py
@@ -1084,10 +1084,16 @@ class BrowserContext:
 				pass
 
 			# Get element properties to determine input method
+			tag_handle = await element_handle.get_property("tagName")
+			tag_name = (await tag_handle.json_value()).lower()
 			is_contenteditable = await element_handle.get_property('isContentEditable')
+			readonly_handle = await element_handle.get_property("readOnly")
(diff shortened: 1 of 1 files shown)


carlosplanchon:
I left my personal .python-version sorry.

rroa:
You might get a runtime error if `after_step_func` is None. I would suggest you add a check like this:

```
if after_step_func is not None:
    await after_step_func(self)

**Comment by MagMueller** [comment]
The agent has the callback 'register_new_step_callback' and 'register_done_callback' is that enough for your use-case?

**Comment by carlosplanchon** [comment]
ss-2025-02-24_19-22-08
ss-2025-02-24_19-21-45

Okay, I tried register_new_step_callback(). 
That may work, but I need to work with the Agent object itself.

A solution may be to rewrite the call to  register_new_step_callback() as
` register_new_step_callback(self)  # Agent object is directly passed to register_new_step_callback()
`

That would allow me to do (in my custom function):

```
    website_html: str = await agent_obj.browser_context.get_page_html()
    website_screenshot: str = await agent_obj.browser_context.take_screenshot()
```

Issue #727, opened 2025-02-14
Idea: Install ublock origin on playwright browser instance


PaperBoardOfficial:
### Problem Description

I was running one of the examples and it ran into unnecessary ad pages. It increases calls to llm and also bloats the user prompt sent to llm.
One solution is that one can use their normal chrome browser which one uses on daily basis (which will have adblock extension installed in it).
But, for the playwright instance, I wanted that there should be python code that would install an adblock extension  on the playwright instance and then only it should start processing the tasks. This feature is important because on a remote server (like Ubuntu), where I don't have a physical display, I can't download a extension. Hence, it would be nice if my python code can provide me a playwright instance installed with the extension.

### Proposed Solution

something like this could be used:
```
UBLOCK_ORIGIN_PATH = "path/to/extracted/ublok-origin"
async with async_playwright() as p:
    browser = p.chromium.launch_persistent_context(
                user_data_dir="user_data",  # Needed to persist extension settings
                headless=False,  # Extensions don't work in headless mode
                args=[
                    f"--disable-extensions-except={UBLOCK_ORIGIN_PATH}",
                    f"--load-extension={UBLOCK_ORIGIN_PATH}"
                ]
            )
```

MagMueller:
Do you know an addblocker that could work?

PaperBoardOfficial:
I prefer "ublock origin". But, need to test it out.

rohan-choudharyy:
I think "ublock-origin" might not work in chrome , because chrome is deprecating manifest v2 support in favor of manifest v3.
It works fine for other browsers like "FireFox".

The alternative could be to use uBOL which is manifest v3 supported, but it might not perform with same capabilities as the parent .

Zhaobudaoyuema:
uBOL The full interception requires user confirmation to be enabled. Does anyone know if there are any other plugins that can directly load and filter ads without requiring confirmation?

PR #801, opened 2025-02-21
Added extension support in playwright built in browser

PaperBoardOfficial:
Issue link: https://github.com/browser-use/browser-use/issues/727

The task was to add adblocker in the playwright instance. I thought of using Ublock Origin, but Chrome is soon deprecating it because it doesn't follow Manifest V3. So, I added Ublock Lite.
If the flag enable_adblock is True, the chrome extension will be downloaded from official github release only during first run. We will not re-download the extension again and again, instead we will reuse the previously download one after the first run.

I tried using normal context of playwright, but I found out that extensions require persistent context. So that is why the code of creating persistent context is in context.py.

These are the [logs](https://drive.google.com/file/d/1m2lfo8JkI9_WObOuRuaSK6g0M7iWGkGY/view?usp=sharing).


PaperBoardOfficial:
@gregpr07 @MagMueller pls review this PR

browser_use/browser/chrome.py
```diff
@@ -147,7 +147,7 @@
 	'--no-pings',
 	'--no-first-run',
 	'--no-default-browser-check',
-	'--no-startup-window',
+	# '--no-startup-window',
```
pirate, line 150: out of curiosity, why remove this?

pirate:
This looks almost ready to go @PaperBoardOfficial nice work! I will merge this once I test it and then we can close all the other extension PRs. If you see anyone else working on extensions please warn them we are merging this PR so they don't waste a lot of effort!

FYI I am working on my own startup [ArchiveBox](https://github.com/ArchiveBox/ArchiveBox) for the next 2.5 weeks starting tomorrow, and then I am coming back to browser-use afterwards (and staying). Since I am the one usually reviewing and merging browser-use PRs, there may be a gap again in responses for a couple weeks. Please don't be offended if the founders don't respond during that time, they are still working hard on the cloud codebase, hiring, ops, etc. 😁 

Afterwards I will resume right where I left off in the PR queue, and this PR is in my top 4 to merge next.

browser_use/browser/chrome.py
PaperBoardOfficial: When I was testing the builtin browser with extension, it wouldn't start. So, I searched which flag maybe causing an issue and this was the one. I tried turning off/on multiple flags but this was the one causing issue.
Also the offical docs of [playwright](https://playwright.dev/python/docs/chrome-extensions) mentions this
>Extensions only work in Chrome / Chromium launched with a persistent context. Use custom browser args at your own risk, as some of them may break Playwright functionality.

pirate: hmm interesting, I suspect it has something to do with the background extension contexts that are spawned when the startup window appears?

PaperBoardOfficial: probably this could be the reason.
this is what claude gave me:
>Background Script Limitations: While background scripts (service workers) can run without an active window, they have limited functionality without a window context. Many Chrome APIs require an active tab or window to function properly.

PaperBoardOfficial:
>then I am most likely coming back to browser-use afterwards (and staying)

I hope you stay, there's so much to learn from you 🏴‍☠️.

pirate:
Blocked by https://github.com/browser-use/browser-use/pull/1218

Then will come back to this.

PaperBoardOfficial:
@pirate Thanks for the reply.
But I don't have maintainer access. I don't think I can push code to somebody else's PR.

pirate:
Ok I will read your other two PRs next. I am going to start full-time work back on browser-use in 2 days so I'll have more time to dig in then. Figuring out a persistent user data dir + extensions solution is my first priority so I will read fully and take inspiration from both your open PRs and do my own work on top to merge Browser + BrowserContext.

halllo:
Looking forward to seeing this pr merged 👍

PaperBoardOfficial:
> Ok I will read your other two PRs next. I am going to start full-time work back on browser-use in 2 days so I'll have more time to dig in then. Figuring out a persistent user data dir + extensions solution is my first priority so I will read fully and take inspiration from both your open PRs and do my own work on top to merge Browser + BrowserContext.

I have already implemented user data dir and extension in this PR. Will you create your own solution for extension and user data dir?  will this PR not get merged? should I close this PR?

PR #805, opened 2025-02-21
use camoufox for bot evasion

iloveitaly:
The existing bot evasion doesn't work at all. This is slightly better:

https://github.com/tinyfish-io/tf-playwright-stealth

But still fails on many sites. Camoufox is a impressive piece of engineering and
it looks to be compatible with the playwright API.

I *just* got this working, so before putting more effort I want to kick the tires and see what breaks.
Wanted to post this here in case others were running into issues as well.

### DIFF
--- a/browser_use/browser/browser.py
+++ b/browser_use/browser/browser.py
@@ -18,6 +18,8 @@ from browser_use.browser.context import BrowserContext, BrowserContextConfig
 
 logger = logging.getLogger(__name__)
 
+CAMOUFOX = True
+
 
 @dataclass
 class BrowserConfig:
@@ -179,6 +181,12 @@ class Browser:
 
 	async def _setup_standard_browser(self, playwright: Playwright) -> PlaywrightBrowser:
 		"""Sets up and returns a Playwright Browser instance with anti-detection measures."""
+		if CAMOUFOX:
+			from camoufox.async_api import AsyncCamoufox
+
+			browser = await AsyncCamoufox(proxy=self.config.proxy, geoip=True).start()
+			return browser
+
 		browser = await playwright.chromium.launch(
 			headless=self.config.headless,
 			args=[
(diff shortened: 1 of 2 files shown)


gaurav-cointab:
Hi is it working and is it better at bot detection evasion? @iloveitaly

iloveitaly:
@gaurav-cointab use https://camoufox.com/ instead. This is working for me, at least better than the stealth plugins.

browser_use/browser/context.py
```diff
@@ -264,6 +264,10 @@ async def _create_context(self, browser: PlaywrightBrowser):
 			# Connect to existing Chrome instance instead of creating new one
 			context = browser.contexts[0]
 		else:
+			# camofoux
+			if True:
+				return await browser.new_context()
```
pirate, line 269: why the `if True`?

pirate, line 269:
```suggestion
```

Returning a browser without the config params below would break a bunch of behavioral stuff, they're not just there for stealth.

browser_use/browser/browser.py
```diff
@@ -18,6 +18,8 @@
 
 logger = logging.getLogger(__name__)
 
+CAMOUFOX = True
+
```
pirate, line 22:
```suggestion
CAMOUFOX = True

```

this should be an env var instead of hardcoded

pirate: ah ok no worries. Support for non-chromium browsers did just get merged recently https://github.com/browser-use/browser-use/pull/950, so there may be some interest in this. I suspect it's too much work to get this working properly though unless you're willing to put in weeks of work. There are tons of raw CDP calls all over the codebase that would need to be switched to webdriver bidi equivalents without breaking anything.

pirate:
I think we're going to focus all our effort on patchright right now via #857. While I'd love to support more browsers, in practice it adds a lot more surface area for things to break subtly and I think patchright will get us most of the way there.

If you feel really strongly that this is mistake / you thing camoufox can beat patchright feel free to comment back here and we can continue the discussion.

PR #813, opened 2025-02-22
Fix issue #762: Complete the system prompt

### LINKED ISSUE #762: [Bug] Looks the system prompts incomplete
see: https://github.com/browser-use/browser-use/blob/f0b9522ef403e6d8d644ec0ceb8709457c5f1d16/browser_use/agent/prompts.py#L96

mannaandpoem:
Added prompt aligns with the existing system prompt structure, ensuring the agent can diagnose issues, adapt its strategy, and continue making progress.

Fix this issue:
https://github.com/browser-use/browser-use/issues/762

### DIFF
--- a/browser_use/agent/system_prompt.md
+++ b/browser_use/agent/system_prompt.md
@@ -63,6 +63,7 @@ Common action sequences:
 
 8. Long tasks:
 - Keep track of the status and subresults in the memory. 
+- If stuck, try new strategies and avoid repeating ineffective paths already attempted.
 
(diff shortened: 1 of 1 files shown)


MagMueller:
Can you add it to the new system prompt file and reduce the token count a bit? I think the most important thing is to remember failed paths.

mannaandpoem:
Thank you. I've updated the system prompt file with a shorter version, keeping the key point about remembering failed paths.

PR #816, opened 2025-02-22
[Fix] Handle RuntimeError: Event loop is closed (#219)

### LINKED ISSUE #219: [Test Coverage] Pytest Coverage
#### **Overview**
Our current Pytest suite is outdated and messy, and it needs improvement to ensure robust testing of `browser-use` agents.

---

#### **Proposed Tasks**
- Update and clean up existing Pytests.
- Add tests for different models to ensure compatibility and robustness.
- Design a framework or strategy to effectively test `browser-use` agents for complex tasks.

---

#### **Why This is Important**
Improved testing ensures stability, scalability, and reliability for `browser-use` agents, especially as the system evolves to support more models and complex use cases.

---

#### **Contribute**
If you're interested, please comment or submit a PR.

We can provide API-Credits to contributors! 🚀

achatap:
### Overview
This PR fixes the issue `RuntimeError: Event loop is closed` by ensuring proper cleanup of the event loop when the script is interrupted.

### Changes Made:
- Wrapped `agent.run()` inside `try...finally` to ensure browser resources are closed.
- Handled `KeyboardInterrupt` and other exceptions gracefully.

### Fixes Issue:
Fixes #219

### Testing:
- Verified with Pytest that all test cases pass.
- Manually tested agent shutdown.

Let me know if any modifications are needed! 🚀

### DIFF
--- a/browser_use/utils.py
+++ b/browser_use/utils.py
@@ -2,6 +2,11 @@ import logging
 import time
 from functools import wraps
 from typing import Any, Callable, Coroutine, ParamSpec, TypeVar
+import asyncio
+import functools
+from contextlib import asynccontextmanager
+from typing import Optional, Callable
+from browser_use import Agent
 
 logger = logging.getLogger(__name__)
 
@@ -52,3 +57,32 @@ def singleton(cls):
 		return instance[0]
 
 	return wrapper
+
+
+class BrowserSessionManager:
+    @staticmethod
+    @asynccontextmanager
+    async def manage_browser_session(agent: Agent):
+        """Context manager for browser session handling with proper cleanup"""
+        try:
+            yield agent
+        finally:
+            if agent and getattr(agent, 'browser', None):
+                await agent.browser.close()
+
+def with_error_handling(cleanup_callback: Optional[Callable] = None):
+    """Decorator for handling common errors in browser automation scripts"""
+    def decorator(func):
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            try:
+                return asyncio.run(func(*args, **kwargs))
+            except KeyboardInterrupt:
+                print("\nScript interrupted by user")
+            except Exception as e:
+                print(f"An error occurred: {str(e)}")
+            finally:
+                if cleanup_callback:
+                    cleanup_callback()
+        return wrapper
+    return decorator
--- a/examples/amazon_search.py
+++ b/examples/amazon_search.py
@@ -6,6 +6,7 @@ Simple try of the agent.
 
 import os
 import sys
+from browser_use.utils import BrowserSessionManager, with_error_handling
 
 sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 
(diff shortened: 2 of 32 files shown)


pirate:
This seems like a nice improvement.

There are many other examples that we'd have to add the try/catches to as well, is there a more general solution we can use that would improve the Ctrl+C and teardown process for all the examples?

achatap:
Please let me know if there's anything else I need to do. This is my first time contributing to open source. Thanks a lot for your guidance, @pirate 

### Proposal for Improved Error Handling

As suggested by @pirate, I recommend implementing a more robust and reusable error-handling solution that can be applied across all examples. Here's the proposed approach:

1. Create a dedicated utilities module (`browser_use/utils.py`) that provides:
   - A context manager for browser session lifecycle management
   - A decorator for standardized error handling
   - Graceful handling of keyboard interrupts (Ctrl+C)
   - Consistent cleanup processes

2. Key benefits:
   - Centralized error handling logic
   - Guaranteed browser cleanup on all exit paths
   - Reduced code duplication across examples
   - Flexible design allowing custom cleanup callbacks
   - Improved maintenance and consistency

3. Implementation would require:
   - Moving existing error handling utilities to a dedicated module
   - Updating all example scripts to use the new utilities
   - Adding proper documentation for the new utilities

4. Example usage would look like:
```python
from browser_use.utils import BrowserSessionManager, with_error_handling

async def main():
    async with BrowserSessionManager.manage_browser_session(agent) as managed_agent:
        await managed_agent.run()

@with_error_handling()
async def run_script():
    await main()

pirate:
Nice work @achatap! I think this will help improve the CLI experience a lot.

A few questions:

- whether or not the browser gets closed at the end is actually sort of already controlled by a config option `config._force_keep_context_alive`, we dont want to ignore that and force-close the browser every time. I think the most urgent part is the clean error handling and better Ctrl+C experience, we can think about logic to close/not close the browser in a separate PR. My apologies when I said "cleanup" originally I was just referring to the asyncio runloop, not necessarily the whole browser 😅 

- if we only handle error catching for now, is it possible to do this with just the decorator and not the context manager? if so, that will make this PR much simpler with only 1 line for the import and 1 line for the decorator needed to change across all the examples. the repo is moving very fast currently and big PRs with many lines of changes are hard to merge without introducing conflicts in many other PRs

- I don't think the `cleanup_callback` is necessary at the moment. It's a good idea but until we have a need lets not add too much extra logic. 

The core problems the repo needs help solving are:

- Ctrl+C inside any of the examples doesn't work the first time, you have to press several times
- Ctrl+C inside the examples triggers a wall of text of `KeyboardInterrupt` error traceback when it finally does work 
image
- it should print just one line at the end instead unless it exited due to a real error: `Got Ctrl+C (SIGINT), stopping agent and exiting gracefully.`. If there was an `Exception` it should print the traceback and then `Got {exception} error, stopping agent and exiting.` at the end.

pirate:
I have implemented this in another PR here: https://github.com/browser-use/browser-use/pull/1178

I added pause/resume functionality with Ctrl+C so it should be much easier now.

achatap:
Hi @pirate ,

I apologize for the delay in responding, I was occupied with other urgent work. Thank you for your prompt feedback and for addressing the previous issue. I plan to implement additional improvements to the browser-use project and will submit a new pull request shortly.

Thank you for your continued support.

PR #845, opened 2025-02-24
Added support of right click on elements.

gaurav-cointab:
Added support of right click on elements.


gregpr07:
why is this useful?

gaurav-cointab:
I had a scenario, where the website had menu opens on right click on elements. like shown in the image attached

image

PR #886, opened 2025-02-27
add html download action in the registry

ai-naymul:
- if users want to download html of the current page they can save that into their preferred location
@MagMueller @gregpr07 @maticzav

### DIFF
--- a/browser_use/controller/service.py
+++ b/browser_use/controller/service.py
@@ -182,6 +182,30 @@ class Controller(Generic[Context]):
 				msg = f'📄  Extracted from page\n: {content}\n'
 				logger.info(msg)
 				return ActionResult(extracted_content=msg)
+		
+		# HTML Download
+		@self.registry.action(
+			'Download the full HTML content of the current page',
+			param_model=NoParamsAction,
+		)
+		async def download_html(_: NoParamsAction, browser: BrowserContext) -> ActionResult:
+			"""Retrieves and returns the full HTML content of the current page"""
+			try:
+				page = await browser.get_current_page()
+				html_content = await page.content()
+				msg = f'📄 Page HTML content\n: {html_content}\n'
+				logger.info(f'Successfully retrieved page HTML: {msg}')
+				return ActionResult(
+					extracted_content=html_content,  # Return raw HTML
+					is_done=True,
+					include_in_memory=False
+				)
+			except Exception as e:
+				logger.error(f'Failed to get HTML content: {str(e)}')
+				return ActionResult(
+					error=f'Error retrieving HTML: {str(e)}',
+					extracted_content=''
+				)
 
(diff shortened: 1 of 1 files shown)


pirate (approved):

pirate:
this seems useful. seems worth it to add singlefile, MHTML, WARC, HAR, etc. support in the future too

MagMueller:
The problem is that HTML files are quite large; therefore, the LLM's context is too small, so the agent cannot handle them. What would you like to do with the HTML? If you just need to save it, we could do it similarly to saving a PDF. However, I think with the current context of the LLM (without RAG), this is too large and not yet useful.

pirate:
Yes the intent is not to pass the HTML file to an LLM, this action is just so the LLM can download HTML as an opaque output file if the user asks for it, similar to PDF.

e.g. `task='follow all the links on example.com and download the raw HTML for each'`

MagMueller:
@ai-naymul  
We should not pass the entire HTML to the LLM. This seems to me like a custom function for your use-case. If you need it, you can just register it.

ai-naymul:
> @ai-naymul We should not pass the entire HTML to the LLM. This seems to me like a custom function for your use-case. If you need it, you can just register it.

@MagMueller @pirate We are not passing the entire html to the LLM. The flow work like the following: go to a site and after loading the site it will just save the the html from the browser source.

For example: I was trying to aggregrate a site where there are many pagination as you mentioned llm model context window are limited although gemini has million in context window but if is there many paginations out there it will eventually out the context limit as I faced also and its also time taking process to extract information using agent in large amount. Thats where extracting data from html is efficient and now I am using it for that task I thought this would be good addidation for that task like aggregating with raw html similar to #1095

pirate:
@MagMueller pointed out to me that it actually is passing it to the LLM currently because every `ActionResult(extracted_content=html_content, ...)` is passed during the next action in order to summarize the previous action. Even though it's not persisted in memory, it still can use up a ton of context out of the next action.

Better way to do it would be like how PDFs are handled, as a file saved directly to the current working directory, and return the path in an `ActionResult` https://github.com/browser-use/browser-use/pull/1095

ai-naymul:
> @MagMueller pointed out to me that it actually is passing it to the LLM currently because every `ActionResult(extracted_content=html_content, ...)` is passed during the next action in order to summarize the previous action. Even though it's not persisted in memory, it still can use up a ton of context out of the next action.
> 
> Better way to do it would be like how PDFs are handled, as a file saved directly to the current working directory, and return the path in an `ActionResult` #1095
> 
> However even then I think this is best moved to the `examples/use-cases` subdirectory so people can use it when they need it. There are many other better scraping tools out there if you really just need to download raw html.

Thanks for pointing it out. patched it here #1117

Issue #687, opened 2025-02-12
No Playwright's http_credentials context in BrowserContextConfig


ntthaibk:
### Bug Description

I'm trying to bypass a Http Credentials pop up check when open the Web Page, and it required to enter the username/password every time I open that web page, after searching around I found that we can include the Http Credentials using Playwright's context like mentioned here

https://playwright.dev/python/docs/network#http-authentication

but in the BrowserContextConfig of Browser Use, there's no parameter to set. So I'm stuck there, is there any way to work around this? Or should we add http_credentials as a new parameter for BrowserContextConfig, I'm happy to send an MR

Thansk & Best Regards

### Reproduction Steps

1. Run the task of Browser Use
2. Can not authorize since it's a Pop Up
3. Browser Use cancelled and then go to google and search for our internal web Page name => which is nothing to find

### Code Sample

```python
import asyncio
import os

from dotenv import load_dotenv
from langchain_google_genai import ChatGoogleGenerativeAI
from pydantic import SecretStr

from browser_use import Agent,Browser, BrowserConfig
from playwright.async_api import Browser as PlaywrightBrowser
from browser_use.browser.context import BrowserContextConfig, BrowserContext

load_dotenv()
api_key = os.getenv('GEMINI_API_KEY')
if not api_key:
	raise ValueError('GEMINI_API_KEY is not set')

llm = ChatGoogleGenerativeAI(model='gemini-2.0-flash-exp', api_key=SecretStr(api_key))


async def run_search():
	agent = Agent(
		# task=(
		# 	'Go to url r/LocalLLaMA subreddit and search for "browser use" in the search bar and click on the first post and find the funniest comment'
		# ),
		
        task=("sample"
		),
		llm=llm,
		max_actions_per_step=4,
		use_vision=False,
		# browser_context=context
	)

	await agent.run(max_steps=25)


if __name__ == '__main__':
	asyncio.run(run_search())
```

### Version

0.1.36

### LLM Model

Gemini 1.5 Pro

### Operating System

macOS 14.5

### Relevant Log Output

```shell

```

pirate:
The workaround is to add the credentials in the URL in your task: `task="Visit 'https://username:password@example.com' and click the button xyz..."`

Cfomodz:
browser-use has since migrated from Playwright to CDP (Chrome DevTools Protocol), making BrowserContextConfig obsolete. HTTP authentication is now handled differently.

PR #893, opened 2025-02-27
Add http_credentials to browser context.

b0rgcube:
In the corporate environment browser-use has to usually use the AD authorization. To make it work with Playwright, browser-use should include a way to set the http_credentials in the browser context.

This PR is fixing the following issue:

No Playwright's http_credentials context in BrowserContextConfig #687
https://github.com/browser-use/browser-use/issues/687

example usage:
```
config = BrowserContextConfig(
        http_credentials={"username": "USER@CORP.com", "password": "aVeryImportantPassword"},
)
```


pirate:
hmm cant the user just add the credentials to the url they ask the agent to visit if needed 'https://username:password@example.com'? I would rather not configure credentials on the context since they could be different for different URLs.

b0rgcube:
@pirate Unfortunately that doesn't work for AD authorization. We are trying to use Browser Use in the large corporate environment to interact with internal apps. When I try your workaround I'm still presented with the browser pop-up to enter the credentials. Since this pop-up is not part of the page Browser Use can't interact with it. The only way to go through the authorization is to use http_credentials.

browser_use/browser/context.py
```diff
@@ -105,6 +105,10 @@ class BrowserContextConfig(BaseModel):
 
 	    include_dynamic_attributes: bool = True
 	        Include dynamic attributes in the CSS selector. If you want to reuse the css_selectors, it might be better to set this to False.
+		
+		  http_credentials: None
+          Dictionary with HTTP authentication credentials, e.g.
```
pirate, line 110:
```suggestion
          Dictionary with HTTP basic authentication credentials for corporate intranets (only supports one set of credentials for all URLs at the moment), e.g.
```

pirate:
Ah that's a royally annoying design design by playwright, oh well. I guess this is fine for now. We can figure out unique per-URL credentials later. Thanks for contributing!

PR #894, opened 2025-02-27
If chrome user data directory not given, then take guest as default

gaurav-cointab:
This PR enables the agent to directly take the guest profile if the user data directory is not specified.

### DIFF
--- a/browser_use/browser/browser.py
+++ b/browser_use/browser/browser.py
@@ -50,6 +50,7 @@ class BrowserConfig:
 	disable_security: bool = True
 	extra_chromium_args: list[str] = field(default_factory=list)
 	chrome_instance_path: str | None = None
+	is_chrome_guest: bool | None = True
 	wss_url: str | None = None
 	cdp_url: str | None = None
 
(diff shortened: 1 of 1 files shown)


gregpr07:
I like the idea, but why do we need a new parameter for this?

gaurav-cointab:
What happened with me is that when I ran without the guest profile option, and without the chrome user data directory path, it showed me a list of profiles (which is a lot in my case), and even if I had only 1 profile, but if the guest profile is enabled in chrome, then also it will show me that screen to select a profile, which I think will become a bottle neck when doing the automation, so I thought if a user is using chrome.exe, and is not mentioning the chrome user data directory, then in that case only take the user to the guest profile. I added the variable so that if in future someone wants to do the selecting of profile, then they can turn off the flag, and run with it. In that case the user will see the profile selection screen, and then after the profile selection is done by the user, then the automation can proceed.

gaurav-cointab:
I am closing this as it is always better to specify a user data directory in the temp folder rather than specifying the guest mode. using user data directory helps solve other problems also.

PR #895, opened 2025-02-27
fix: close httpx connections to avoid memory leaks

adamdelezuch89:
Fixed an issue where the event loop was closing before all HTTPX connections were properly terminated, causing resource warnings and potential memory leaks. The problem occurred because browser-use was shutting down threads before Playwright had fully closed its HTTP connections. Now, before shutting down the agent, all HTTPX processes are explicitly closed, ensuring a clean shutdown and preventing unclosed transport warnings.

### DIFF
--- a/browser_use/agent/service.py
+++ b/browser_use/agent/service.py
@@ -628,11 +628,7 @@ class Agent(Generic[Context]):
 				)
 			)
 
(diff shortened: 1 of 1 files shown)


pirate (approved):

pirate:
cc @MagMueller

Issue #439, opened 2025-01-27
Error: 'Failed to Complete Task in Maximum Steps' Not Visible in Agent History List


mustafasahin0:
### Bug Description

"When my agent fails to complete a task within the defined maximum number of steps, it stops with an error message saying, 'Failed to complete task in maximum steps.' However, when I check the Agent History List, I see that the list is empty. It seems that after throwing this error, we are not saving the issue in the agent history.

### Reproduction Steps

Add an error entry in self.history whenever we hit the else block.

### Code Sample

```python
async def run(self, max_steps: int = 100) -> AgentHistoryList:
		"""Execute the task with maximum number of steps"""
		try:
			self._log_agent_run()

			# Execute initial actions if provided
			if self.initial_actions:
				result = await self.controller.multi_act(self.initial_actions, self.browser_context, check_for_new_elements=False)
				self._last_result = result

			for step in range(max_steps):
				if self._too_many_failures():
					break

				# Check control flags before each step
				if not await self._handle_control_flags():
					break

				await self.step()

				if self.history.is_done():
					if self.validate_output and step < max_steps - 1:
						if not await self._validate_output():
							continue

					logger.info('✅ Task completed successfully')
					if self.register_done_callback:
						self.register_done_callback(self.history)
					break
			else:
				logger.info('❌ Failed to complete task in maximum steps')

			return self.history
		finally:
			self.telemetry.capture(
				AgentEndTelemetryEvent(
					agent_id=self.agent_id,
					success=self.history.is_done(),
					steps=self.n_steps,
					max_steps_reached=self.n_steps >= max_steps,
					errors=self.history.errors(),
				)
			)

			if not self.injected_browser_context:
				await self.browser_context.close()

			if not self.injected_browser and self.browser:
				await self.browser.close()

			if self.generate_gif:
				output_path: str = 'agent_history.gif'
				if isinstance(self.generate_gif, str):
					output_path = self.generate_gif

				self.create_history_gif(output_path=output_path)
```

### Version

main

### LLM Model

GPT-4

### Operating System

macOs 15.2

### Relevant Log Output

```shell

```

Shyanil:
@mustafasahin0 @Cfomodz  Hi, I've made changes to address issue #439 based on my understanding. As someone new to open source, I wanted to ensure the fix properly addresses the problem of the missing error message in the Agent History List.

Shyanil:
Hi @mustafasahin0 ,

[...]

PR #919, opened 2025-03-02
Fix/agent history max steps #439

Shyanil:
# **Fix: Store "Failed to complete task in max steps" in Agent History**  #439 

## **Overview**  
This update ensures that when the agent **fails due to exceeding max steps**, the error is properly **stored in history** instead of just being logged.  

### **Before (Issue)**  
```python
else:
    logger.info('❌ Failed to complete task in maximum steps')
```
🔴 **Issue**: The error was not recorded in `self.state.history.history`, leaving history empty.  

### **After (Fix)**  
```python
else:
    error_message = 'Failed to complete task in maximum steps'
    
    self.state.history.history.append(
        AgentHistory(
            model_output=None,
            result=[ActionResult(error=error_message, include_in_memory=True)],
            state=BrowserStateHistory(url="", title="", tabs=[], interacted_element=[], screenshot=None),
            metadata=None
        )
    )

    logger.info(f'❌ {error_message}')
```
✅ **Fix**:  
- Stores error in **Agent History List**.  
- Prevents history from being **empty on failure**.


gregpr07:
Do we actually want this? What's the point?

I think we have to solve this in some other way - it's very confusing if the last output is just empty, we are working on making the output memory also include `is_complete` (self eval by agent) so people can use that to see if it was complete or not.

Shyanil:
@gregpr07 Sorry, I think I'm late. No problem. I see that the issue is closed. By the way, thanks for your kind words.

Issue #903, opened 2025-02-28
Shadowroots not working properly


danielgobona:
### Bug Description

I am trying to browse page with multiple nested shadowroots. It worked on older versions of browser-use (for example 0.1.20 was ok), but current version(0.1.40) does not work. I can not exactly share webpage where i am trying to use it, because i am testing it on proprietary internal system.

I tried to investigate what was happening in `buildDomTree.js` because i suspected this would be the problem, because i could not see any elements inside shadowRoots highlighted by this js code. Specificaly i found the problem in the current implementation  of `buildDomTree` function.

When i investigated shadowRoots of the internal system, i discovered that just calling `node.shadowRoot` does not work, because there is only one `slot` element. But all the nested elements inside `slot` are part of node.childNodes. Something like that - https://mdn.github.io/web-components-examples/slotchange/ where you have shadowRoots, but elements are outside of shadowRoot in node.

In 0.1.20, https://github.com/browser-use/browser-use/blob/0.1.20/browser_use/dom/buildDomTree.js#L396 here it processed shadowRoots, but then it processed also just childNodes. In the new implementation https://github.com/browser-use/browser-use/blob/0.1.40/browser_use/dom/buildDomTree.js#L917 it is processing just shadowRoots or childNodes, not both. We are missing elements next to shadowRoot this way.

I am not super familiar with shadowRoots (i did not implement that system which i am trying to browse) and i am not sure what was the intent of this new implementation change by the browser-use team,  so i want to discuss before suggesting some change, but i think if it is better to process shadowRoots and childNodes together, or to resolve this other way.

### Reproduction Steps

Browse page with multiple nested shadowRoots with slot elements.

### Code Sample

```python
...
```

### Version

pip 0.1.40

### LLM Model

Gemini 2.0 Flash

### Operating System

Windows 11

### Relevant Log Output

```shell

```

PR #928, opened 2025-03-03
process elements next to shadowroot along with shadowroot

danielgobona:
This adresses issue https://github.com/browser-use/browser-use/issues/903 
With the current implementation, if a node contains both a shadowRoot and other elements, only the shadowRoot is processed, while the other elements inside the node are ignored. I was using an older version (0.1.20) of browser-use on some pages that heavily rely on shadowRoots, and it was working back then.

### DIFF
--- a/browser_use/dom/buildDomTree.js
+++ b/browser_use/dom/buildDomTree.js
@@ -927,6 +927,15 @@
     if (node.tagName) {
       const tagName = node.tagName.toLowerCase();
 
+      // Handle shadow DOM
+      if (node.shadowRoot) {
+        nodeData.shadowRoot = true;
+        for (const child of node.shadowRoot.childNodes) {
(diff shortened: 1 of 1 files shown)


gregpr07:
Could you please send me an example where it fails for you?

danielgobona:
@gregpr07 Okay try https://www.ibm.com/us-en
With current browser-use, nothing is highlighted at all on this website.

pirate:
@danielgobona are you and @cheewba possibly fixing the same bug here?

[`dfd5b12` (#998)](https://github.com/browser-use/browser-use/pull/998/commits/dfd5b124f2e2b89724cdf51e9dee980955f8193a)

pirate:
I'm closing this for now, it's "merged" via different PR, multiple submitted the same fix via different PRs.

PR #932, opened 2025-03-04
Fix: Unable to Retrieve Interactive Elements Bound with addEventListener

Jonathanqjs:
Since window.getEventListeners is not available in Playwright, it is unable to retrieve some interactive elements that have event listeners bound via addEventListener. This fix injects JavaScript code to intercept the addEventListener function and manually simulate window.getEventListeners each time a BrowserSession is initialized, resolving the issue. An example can be seen on the homepage of https://en.fofa.info/.

Before fix
Pasted Graphic 1

After fix
FQFA

### DIFF
--- a/browser_use/browser/context.py
+++ b/browser_use/browser/context.py
@@ -136,11 +136,54 @@ class BrowserContextConfig:
 	_force_keep_context_alive: bool = False
 
 
-@dataclass
 class BrowserSession:
-	context: PlaywrightBrowserContext
-	cached_state: BrowserState | None
+	
+	def __init__(
+			self,
+			context: PlaywrightBrowserContext,
+			cached_state: BrowserState | None=None
+		):
+		init_script = """
+			(() => {
+				if (!window.getEventListeners) {
+					window.getEventListeners = function (node) {
+						return node.__listeners || {};
+					};
 
+					// Save the original addEventListener
+					const originalAddEventListener = Element.prototype.addEventListener;
+
+					const eventProxy = {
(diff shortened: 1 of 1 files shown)


pirate:
This is a good fix, thanks @Jonathanqjs. Can you sign the CLA so we can merge? Thanks

pirate (approved):

Jonathanqjs:
I’ve already signed the CLA. Please go ahead with the merge. Thanks!

pirate:
thanks!

shawyang0304:
I have also noticed this issue recently. I saw that this issue was said to have been resolved, but in fact it wasn't. Could you please tell me if it hasn't been merged into the master branch yet? Then I also tried to solve it. The following solution is also feasible.Adjustments related to getEventListeners:

    // Enhanced event listener check
    if (window._eventListeners && window._eventListeners.has(element)) {
      const map = window._eventListeners.get(element);
      return Object.keys(map).length > 0;
    }
    const eventAttrs = [
      'onclick', 'onmousedown', 'onmouseup', 'ondblclick',
      'onmouseover', 'onmouseout', 'onfocus', 'onblur',
      'onkeydown', 'onkeyup', 'onchange', 'onsubmit'
    ];
    for (const attr of eventAttrs) {
      if (element.hasAttribute(attr) || typeof element[attr] === 'function') {
        return true;
      }
    }

pirate:
this PR was already merged and released @shawyang0304:
image

PR #945, opened 2025-03-05
[bug][multi_act] Detect index change caused by page change in multi_act

AlinJiang:
This PR fixed an issue in multi_act, in some cases the element index can change after previous action, because the page just changed.

One example is when changing language on amazon.com:
iShot_2025-03-05_00 58 22

After selecting another language, before clicking submit button, the index changed:
iShot_2025-03-05_00 58 33

Sample log after this fix:
```
INFO     [agent] 🎯 Next goal: Select English language option
INFO     [agent] 🛠️  Action 1/2: {"click_element":{"index":25}}
INFO     [agent] 🛠️  Action 2/2: {"click_element":{"index":36}}
INFO     [controller] 🖱️  Clicked button with index 25: 
INFO     [agent] Element index changed after action 1 / 2, because page changed.

INFO     [agent] 🛠️  Action 1/2: {"click_element":{"index":25}}
INFO     [agent] 🛠️  Action 2/2: {"click_element":{"index":34}}
INFO     [controller] 🖱️  Clicked button with index 25: 
INFO     [controller] 🖱️  Clicked button with index 34: 
```

### DIFF
--- a/browser_use/agent/service.py
+++ b/browser_use/agent/service.py
@@ -659,7 +659,20 @@ class Agent(Generic[Context]):
 		for i, action in enumerate(actions):
 			if action.get_index() is not None and i != 0:
 				new_state = await self.browser_context.get_state()
-				new_path_hashes = set(e.hash.branch_path_hash for e in new_state.selector_map.values())
+				new_selector_map = new_state.selector_map
+
+				# Detect index change after previous action
+				orig_target = cached_selector_map.get(action.get_index())  # type: ignore
+				orig_target_hash = orig_target.hash.branch_path_hash if orig_target else None
+				new_target = new_selector_map.get(action.get_index())  # type: ignore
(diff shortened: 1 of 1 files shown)


AlinJiang:
@pirate Also following up on this PR, this PR verifies if the element index mapping has changed during a multi_act sequence, which can occur when a previous action alters the page layout or content. Here's a concise overview:
1. The LLM’s view of the element is based on `cached_selector_map`, which provides the original target element.
2. After we finished an action in the `multi_act` sequence, we use the latest `selector_map` with the original index to fetch the new target element.
3. If the original target and the new target differ, it indicates that the mapping has shifted. The sequence is then halted, and the LLM is prompted to retry with the updated page.

pirate (approved):
good idea, thanks!

PR #946, opened 2025-03-05
[bug][DOM] Detect <label> as interactive elements

AlinJiang:
This PR fixed an issue in interactive element detection, in some cases the interactive element should be <label>, because its associated <input> element was overlaied and not the top-most element, ended up being skipped. One example is the language choice options on amazon.com .


Mustafiz04 (approved):

pirate:
https://github.com/browser-use/browser-use/pull/1031 should have massively improved this. Have you pulled `main` recently?

AlinJiang:
@pirate Hi I have pulled and tested the latest commit and the issue still exists.

PR #956, opened 2025-03-06
Enabled downloads over CDP session, tested in browserbase.com.

manojsundayai:


pirate (approved):

pirate:
Can change this to check if downloads are already configured first, and leave the existing config if so? Some browser backends already have downloads enabled with their own path and we don't want to overwrite existing config.

manojsundayai:
> @manojsundayai Can you plz change this to check if downloads are already configured first, and leave the existing config if so? Some browser backends we use already have downloads enabled with their own path and we don't want to overwrite any existing config.

@pirate Lemme work on it more, still in WIP. I'll take your feedback into account.

Thanks!

pirate:
I don't know if checking if downloads are already configured is possible, there is no equivalent `browser.getDownloadBehavior` API. Perhaps this could be an opt-in config option then like `config.headless` to avoid breaking existing envs?

pirate:
I think the best solution for now is to only run this for browsers launched and managed by browser-use. If connecting to a remote browser that's already running we should not change its existing settings.

So move this code to only execute during `_setup_browser_with_instance` and `_setup_standard_browser`.

Mustafiz04 (approved):

pirate:
we added download & upload support in some other PRs, thank you for your work here though!

Issue #966, opened 2025-03-07
HAR file recording


jlogtenberg:
### Problem Description

To be able to use browser-use for network traffic analysis, it should be able to create a HAR file of all the requests and responses that it processes during the run.

### Proposed Solution

Since Browser-Use makes use of Playwright under the hood, and Playwright supports creating a HAR file this can be easily implemented. To accomplish this, the variable `record_har_path = FILE_PATH` should be added to browser-use\browser\context.py in the function `async def _create_context(self, browser: PlaywrightBrowser):`. If this path can be provided in BrowserContextConfig, this will allow users to create their own browser with HAR file recording enabled.

### Additional Context

PR #967, opened 2025-03-07
HAR file recording

jlogtenberg:
These changes will allow for the recording of the HAR file (a JSON-formatted archive file format for logging of a web browser's interaction with a site) while the agent is running, along with specifying where this HAR file should be saved.

Added a `save_har_path` to class BrowserContextConfig to store the path where the HAR file should be saved.
Added `record_har_path = self.config.save_har_path` to context initialization to enable HAR file recording itself.

Solution for #966.


pirate (approved):

pirate:
thanks!

PR #980, opened 2025-03-09
Trying to click on missing element after a load timeout

mkuts12:
sometimes a click happens but the page fails to load in time, in this line,
```
					await page.wait_for_load_state()
```
then the except will try to click on a non-existing element and fail anyways. This change makes sure that we retry the click only when the click fails.

### DIFF
--- a/browser_use/browser/context.py
+++ b/browser_use/browser/context.py
@@ -1142,16 +1142,22 @@ class BrowserContext:
 					await self._check_and_handle_navigation(page)
 
 			try:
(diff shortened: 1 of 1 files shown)


Shahar-Y:
This is a serious bug in my end as well, would love to see it fixed

pirate:
@mkuts12 we're ready to review/merge but we need the CLA signed please 🙏

mkuts12:
Signed.
Sorry, I've signed it in another PR so I didn't know I need to sign it again

PR #981, opened 2025-03-09
Make the context in `extract_content` action smaller

mkuts12:
Adding the `strip` option to Markdownify will keep the text but not add the url to the output.

from their [docs](https://github.com/matthewwithanm/python-markdownify?tab=readme-ov-file#options):
```
from markdownify import markdownify as md
md('<b>Yay</b> <a href="http://github.com">GitHub</a>', strip=['a'])  # > '**Yay** GitHub'
```

This has lowered the character count of a page like [Amazon](https://www.amazon.com/s?k=clocks&crid=1SLPSQOSK7JDP&sprefix=clocks%2Caps%2C226&ref=nb_sb_noss_1) from 367277 characters to 48969.

Quick fix to a more deep problem.

### DIFF
--- a/browser_use/controller/service.py
+++ b/browser_use/controller/service.py
@@ -179,11 +179,15 @@ class Controller(Generic[Context]):
 		@self.registry.action(
 			'Extract page content to retrieve specific information from the page, e.g. all company names, a specifc description, all information about, links with companies in structured format or simply links',
 		)
-		async def extract_content(goal: str, browser: BrowserContext, page_extraction_llm: BaseChatModel):
+		async def extract_content(goal: str, should_strip_link_urls: bool, browser: BrowserContext, page_extraction_llm: BaseChatModel):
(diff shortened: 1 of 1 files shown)


CloneOfAlex (approved):

MagMueller:
Great idea - I think sometimes people want to include links, can you let the llm choose if links are included or not?

mkuts12:
Not sure how to tackle it so I've added a parameter to the function so the LLM can add a parameter.
I ran the task 
```
Go to the URL: 'https://www.amazon.com/', look for clocks and get inside the first item they offer then, extract the products name and price
```
and got False there. I think it makes sense so we don't break current uses.

So I guess that the task will need to somehow make it clearer should it strip the URLs? Or should I add a custom type with a description on the parameter?

pirate:
LGTM thanks!

PR #998, opened 2025-03-11
Bugfix: error when trying to create empty dir.

cheewba:
If the conversation path doesn't contain any directory path, `save_conversation` tries to create an empty directory structure, which raises an error.

### DIFF
--- a/browser_use/dom/buildDomTree.js
+++ b/browser_use/dom/buildDomTree.js
@@ -955,16 +955,16 @@
           if (domElement) nodeData.children.push(domElement);
         }
       }
-      // Handle shadow DOM
-      else if (node.shadowRoot) {
-        nodeData.shadowRoot = true;
-        for (const child of node.shadowRoot.childNodes) {
-          const domElement = buildDomTree(child, parentIframe);
-          if (domElement) nodeData.children.push(domElement);
-        }
-      }
-      // Handle regular elements
       else {
+        // Handle shadow DOM
+        if (node.shadowRoot) {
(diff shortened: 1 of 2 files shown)


pirate (approved):
have not tested it personally but both fixes look good to me in theory

browser_use/dom/buildDomTree.js
```diff
@@ -955,16 +955,16 @@
...
       else {
+        // Handle shadow DOM
+        if (node.shadowRoot) {
+          nodeData.shadowRoot = true;
+          for (const child of node.shadowRoot.childNodes) {
+            const domElement = buildDomTree(child, parentIframe);
+            if (domElement) nodeData.children.push(domElement);
+          }
+        }
+        // Handle regular elements
```
MagMueller, line 967: Why did you move this outside of the else if?  - now we do the loop twice

cheewba: there're different loops, the first one is for shadow root children and the next one is for regular children.

pirate: I believe open shadow roots can have both normal dom node children and shadow children, you see this pattern on https://trailhead.salesforce.com/content/learn/modules/starting_force_com/starting_intro?trail_id=force_com_admin_beginner for example. I tested this fix and it works, in fact it was necessary for me to get salesforce automation working, and I included it in my branch [`73369f9` (#1114)](https://github.com/browser-use/browser-use/pull/1114/commits/73369f910c387a34152d764aab9e3c39338a1411)

PR #1015, opened 2025-03-13
 fix: empty interacted element in history when interacted index is 0

NvRom:
potiential value of `index` could be None and numeric. If the value is `0`, it always skips css selector appending.

### DIFF
--- a/browser_use/agent/views.py
+++ b/browser_use/agent/views.py
@@ -166,7 +166,7 @@ class AgentHistory(BaseModel):
 		elements = []
 		for action in model_output.action:
 			index = action.get_index()
-			if index and index in selector_map:
+			if index != None and index in selector_map:
 				el: DOMElementNode = selector_map[index]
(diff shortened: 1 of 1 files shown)


browser_use/agent/views.py
```diff
@@ -166,7 +166,7 @@ def get_interacted_element(model_output: AgentOutput, selector_map: SelectorMap)
 		elements = []
 		for action in model_output.action:
 			index = action.get_index()
-			if index and index in selector_map:
+			if index != None and index in selector_map:
```
gaurav-cointab, line 169: make it "is not" instead of "!="

NvRom: thanks Gaurav, has updated

NvRom:
hi @gaurav-cointab, could you help review the PR? thanks

NvRom:
hi @MagMueller could you help review this slight change? Thanks~

pirate (approved):

PR #1021, opened 2025-03-14
Added a Chinese version of the README

eagle666:
Added a Chinese version of the README, you can click to choose between viewing the Chinese or English version of the README.


eagle666:
sorry, I should use eagle666 account to commit PR not the geekea66 account, I have already change commit account to eagle666 in another PR

PR #1028, opened 2025-03-15
Adding parameter to Agent.run() in order to allow the developer to execute a custom Python function after Agent.step() (updated)

carlosplanchon:
Today I updated this PR to match upstream.


pirate:
Can you share the use case? What problem are you trying to solve with these hooks?

carlosplanchon:
Hello! Yes, of course. I'm using hooks to run a function that captures the website's HTML, a screenshot, and other browsing history details, which I then send to an external service.

But I think this could also be useful for other purposes, especially in cases where a developer needs to run custom logic while the browser is in use.

The function I'm hooking:

```python
async def record_activity(agent_obj):
    website_html: str = await agent_obj.browser_context.get_page_html()
    website_screenshot: str = await agent_obj.browser_context.take_screenshot()
    
    ...
    # Send data to an external service.
    requests.post(url, json=...)
```

browser_use/agent/service.py
```diff
@@ -591,7 +591,12 @@ async def take_step(self) -> tuple[bool, bool]:
 
 	# @observe(name='agent.run', ignore_output=True)
 	@time_execution_async('--run (agent)')
-	async def run(self, max_steps: int = 100) -> AgentHistoryList:
+	async def run(
+		self,
+		max_steps: int = 100,
+		before_step_func: Callable | None = None,
+		after_step_func: Callable | None = None
```
pirate, line 598:
```suggestion
		on_step_start: Callable | None = None,
		on_step_end: Callable | None = None
```

pirate:
Ok, seems reasonable.

- Can you add a type signature for the hook function and rename them to `on_step_start`/`on_step_end`
- Can you add an example showing how to use this new feature under: `examples/custom-functions/custom_hooks_before_after_step.py`

Thanks for contributing!

carlosplanchon:
Ready, thx! :D

carlosplanchon:
It may be needed to add the following paramenters in the file agent/service.py, in the __init__ of the class Agent:

```python
    before_step_func: Callable | None = None,
    after_step_func: Callable | None = None,
    max_steps: int = 30,
 ```
 
In my fork, I’m running it with these parameters because it throws an error if they’re not included in the __init__.

I didn’t include this in the PR since I wasn’t sure about your stance on having these parameters in the __init__, but I’m happy to add them if you’d like.

pirate:
> In my fork, I’m running it with these parameters because it throws an error if they’re not included in the init.

Why does that happen? We're not using `self.on_step_start` anywhere. Does the current example work as-is in the PR?

carlosplanchon:
Oh sorry, there was an error in the import in the example.
It's fixed now and works as is!

pirate:
awesome, thanks!

pirate:
If you would like to, it would also be super helpful to document these new hooks in the `docs/` folder somewhere! You can probably get `claude` code to do it even.

There are already people beginning to use it 😄 https://github.com/browser-use/browser-use/issues/1183

pirate:
@carlosplanchon I was skeptical of the use at first but I have to say this has been already used many times since you contributed it! Thanks again for adding an awesome feature! 🚀

PR #1029, opened 2025-03-15
Update README.md

eldyvoon:
this line confused many javascript developer!

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ pip install browser-use
 install playwright:
 
 ```bash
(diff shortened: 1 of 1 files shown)


MagMueller:
playwright install is correct. We install playwright this way

pabshazon:
Hey @eldyvoon , makes sense the confusion.
Installing with pip installs the python library so you can use it in your python code.
But playwright install installs the CLI tool provided by Playwright which browser use needs.

PR #1031, opened 2025-03-15
Performance Improvement: Efficient Detection of Interactive Elements

Rahul-Sharma-1729:
## Background
In most web applications, developers or frameworks provide visual cues to indicate interactive elements. One common approach is changing the cursor style when hovering over such elements. This helps users quickly identify elements they can interact with.

Upon analyzing widely used websites such as Amazon, Google, eBay, Salesforce, Reddit, Google Docs  etc., I observed a consistent pattern: interactive elements typically have cursor styles that differ from the default pointer, signaling interactivity.

## Implementation
To enhance this behavior, I have implemented the doesElementHaveInteractivePointer function. This function determines whether an element has an interactive cursor by inspecting its computed cursor style via window.getComputedStyle().

Key advantages of this approach:

- Performance Optimization: It eliminates the need for more expensive DOM traversal operations (e.g., closest()) and avoids calls to window.getEventListeners().
- Efficiency: By directly checking the computed styles, it removes the need to create additional data structures such as sets, improving execution speed.
- Additionally removed the unused call to window.getComputedStyle() in the same function.

This implementation ensures a lightweight and efficient way to detect interactive elements, aligning with best practices in web development.

### DIFF
--- a/browser_use/dom/buildDomTree.js
+++ b/browser_use/dom/buildDomTree.js
@@ -462,6 +462,23 @@
       return false;
     }
 
+    function doesElementHaveInteractivePointer(element) {
+      if (element.tagName.toLowerCase() === "html") return false;
+      const style = window.getComputedStyle(element);
+
+      let interactiveCursors = ["pointer", "move", "text", "grab", "cell"];
+
+      if (interactiveCursors.includes(style.cursor)) return true;
+
+      return false;
+    }
+
+    let isInteractiveCursor = doesElementHaveInteractivePointer(element);
+
+    if (isInteractiveCursor) {
+      return true;
+    }
+
     // Special handling for cookie banner elements
     const isCookieBannerElement =
       (typeof element.closest === 'function') && (
@@ -571,9 +588,6 @@
     )) {
       return true;
     }
-
-    // Get computed style
-    const style = window.getComputedStyle(element);
 
     // Check for event listeners
     const hasClickHandler =


Rahul-Sharma-1729:
Upon Testing the new changes on some commonly used websites internationally, these are the results I found:

1. The original function misses some interactive elements on the page.
2. The newer function captures more elements on a granular level, with better average time per element and a better "slowest time"
3. The test was done by getting all the elements on the page using document.querySelectorAll("*") and traversing through the array of elements to run the function and find the metrics attached.
4. Manual verification was also done for each interactive element captured by logging it and mapping with the actual website.

@MagMueller  @gregpr07 
image

teddybear082:
I adapted this code change to a similar project and seems to work great.  Good job OP.

Rahul-Sharma-1729:
@teddybear082  Hey thanks teddy, This code is tried and tested on production environments as well, hopefully it can be merged to browser-use

browser_use/dom/buildDomTree.js
```diff
@@ -462,6 +462,23 @@
       return false;
     }
 
+    function doesElementHaveInteractivePointer(element) {
+      if (element.tagName.toLowerCase() === "html") return false;
+      const style = window.getComputedStyle(element);
+
+      let interactiveCursors = ["pointer", "move", "text", "grab", "cell"];
+
+      if (interactiveCursors.includes(style.cursor)) return true;
```
pirate, line 471: wow this is a really smart approach, nice

gregpr07:
Oh wow, that's really smart!! Thanks man

pirate:
@Rahul-Sharma-1729 thanks again for this. Out of curiosity what is your day job? Do you work on browser automation?

Rahul-Sharma-1729:
@pirate Hey, I'm a software engineer at an AI startup.
Yeah, I do work on browser automation—most of my day-to-day involves browsers, web extensions, data scraping, and similar things. I spend a lot of time dealing with XPaths, shadow DOMs, Iframes, and all that fun stuff on daily basis.

PR #1035, opened 2025-03-16
Devops - Removed hatch

pabshazon:
## Description

Removed hatch, setuptools, and updated docs

## Motivation

Smaller memory footprint, simpler devops workflows, code as documentation

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install build hatch
+          pip install build
       - name: Build package
         run: python -m build
       - name: Publish package
--- a/docs/development/local-setup.mdx
+++ b/docs/development/local-setup.mdx
@@ -32,6 +32,10 @@ uv venv --python 3.11
 uv pip install -e ".[dev]"
 ```
 
+3. Install hatch:
+
+Follow their [installation page](https://hatch.pypa.io/latest/install/#command-line-installer)
+
 <Note>
   The `-e` flag installs the package in "editable" mode, which means your local code changes
   will be reflected immediately without requiring reinstallation. The `[dev]` part installs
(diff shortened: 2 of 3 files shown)


MagMueller:
Hi we use uv to manage dependencies and environments, Hatch might be redundant - what do you think?

pabshazon:
Sounds good.
I have extended the PR to include all hatch dependencies I found.
Building and running here.

MagMueller:
@gregpr07

MagMueller:
@pabshazon We load the buildDomTree.js as a resource in the pip package inside dom/service.py - do we need it for that inside the pip package?

pabshazon:
@MagMueller , imo you do not need hatch nor setuptools as dependencies. Not in runtime, and not in dev.

With 'python -m build' you will use hatchling (the minimal build package of hatch), as is configured so in the pyproject.toml file, and it is enough to package the .js file correctly into the dist package without the rest of hatch installed.

As I am looking at dom/service.py, you are managing that using importlib.resources (the native python lib to manage static resources into packages): is cool and efficient imo, and does not require hatch nor setuptools, because hatchling will correctly bundle the package with the .js file in its place.

✅  I did run a test in my local and is working without hatch nor setuptools deps. Committing it here for you to test - I am still unfamiliar with local env cache for testing.

pabshazon:
Committed, I have added also this piece
```
[tool.hatch.build]
include = [
  "browser_use/dom/buildDomTree.js",
]
```

While it is not needed, I am leaving it because serves as documentation as code - whenever we need to build differently, the design contract is explicit as code here and we can refactor quickly including edge and key cases.

pabshazon:
After building, you can unzip the wheel package and check the .js is inside the package.

Screenshot 2025-03-17 at 20 39 10

pabshazon:
I think there might be an issue with the build, in which you need to dist/file to not exist, I have been using 'rm -rf dist' in my local, not sure if you have local_dev scripts (they can help as documentation and automation at the same time).

pabshazon:
✅  Just tested with a fresh .venv and dist and worked well.
I wonder if you have dev scripts or runbooks to ensure all cache is deleted before new builds.

pirate (approved):

PR #1042, opened 2025-03-16
docs:Added a Chinese version of the README

eagle666:
Added a Chinese version of the README, you can click to choose between viewing the Chinese or English version of the README.


pirate:
without any core team members that speak chinese fluently (我能说一点，但不够流利 😢) I am worried that this will go out of date over time. is this somthing you are willing to commit to helping us maintain for years?

PR #1046, opened 2025-03-17
Add dolphin browser driver support

ProphetDiceBot:


jersobh:
There's a ChatGoogleGenerativeAI example already. It makes no sense to have one example per model version.

pirate:
@ProphetDiceBot thanks for contributing and working on this!

Unfortunately we cannot accept the instagram example for legal reasons (we cannot endorse using browser-use to do fake engagement on social media) but the dolphin browser driver is great, I've tweaked the PR to focus on that part instead.

Issue #1020, opened 2025-03-13
blank chromium page at step 1, nothing happens


illtellyoulater:
### Bug Description

Running on Ubuntu 22.04.
Installed everything as described in the README.
Saved OpenAI API Key in a .env file within the example project dir.

When I try to run example.py the chromium browser is launched, but it does nothing and just shows a blank page. 
At the same time this is what the console is printing:

```
INFO     [agent] 🚀 Starting task: Go to https://www.google.com/, then compare the price of gpt-4o and DeepSeek-V3
INFO     [agent] 📍 Step 1
INFO     [agent] 📍 Step 1
INFO     [agent] 📍 Step 1
INFO     [agent] 📍 Step 1
INFO     [agent] 📍 Step 1
INFO     [agent] 📍 Step 1
CINFO     [agent] 📍 Step 1
CINFO     [agent] 📍 Step 1

```
I also tried the example in examples/simple.py from this repo, but the result was exactly the same.
How can I fix this?

### Reproduction Steps

1. Install browser-use and all its deps in a conda env with python=3.11
2. Install playwright 
3. Save OpenAI API key in a .env file (OPENAI_API_KEY="my_secret_key")
4. run example code in README

### Code Sample

```python
from langchain_openai import ChatOpenAI
from browser_use import Agent
import asyncio
from dotenv import load_dotenv
load_dotenv()

async def main():
    agent = Agent(
        task="Go to http://www.google.com/, then compare the price of gpt-4o and DeepSeek-V3",
        llm=ChatOpenAI(model="gpt-4o"),         
        save_conversation_path="logs/conversation"
    )
    await agent.run()
  
asyncio.run(main())
```

### Version

0.1.40

### LLM Model

GPT-4o

### Operating System

Ubuntu 22.04

### Relevant Log Output

```shell
INFO     [browser_use] BrowserUse logging setup complete with level info
INFO     [root] Anonymized telemetry enabled. See https://docs.browser-use.com/development/telemetry for more information.
browser-use/lib/python3.11/site-packages/browser_use/agent/message_manager/views.py:59: LangChainBetaWarning: The function `load` is in beta. It is actively being worked on, so the API may change.
  value['message'] = load(value['message'])
INFO     [agent] 🚀 Starting task: Find the founders of browser-use and draft them a short personalized message
INFO     [agent] 📍 Step 1
INFO     [agent] 📍 Step 1
INFO     [agent] 📍 Step 1
```

[...]

PR #736, opened 2025-02-15
Update service.py

YonderUrik:
This pull request focuses on enhancing the handling of tool call responses in the `get_next_action` method of the `browser_use/agent/service.py` file. The changes ensure that tool call responses are properly converted into the `AgentOutput` format, improving the robustness of the agent's response handling.

Enhancements to tool call response handling:

* [`browser_use/agent/service.py`](diffhunk://#diff-3736f406057606f65f3ea79952b7f0ec8f8a22f25a49d7a0771f4cfbf43cbb92L452-R482): Added logic to handle tool call responses by checking for parsing errors and converting tool calls to the `AgentOutput` format if present. This includes creating the current state and action from the tool call details.

Related iusses:
#656

PR #1048, opened 2025-03-17
fix: Adapts all models that don't support tool_call (not just deepseek-resoner)

WonderYours:
The program only has separate support for deepseek-reasoner, but I found that for some models, such as deepseek-chat, which does not support tool_call, the program did not convert the input.
My code incorporates a conversion retry after an error is reported to ensure that the entire code does not fail because the user message in the tool_call is not in context.
There are a few issues, including  #1020, that report getting stuck in blank pages and stuck in step1 indefinitely, because our browser-use default installed dependency package does not include google, and the error is not caught properly. To fix this, add google to the dependency package. But the root cause is that users are using an unsupported model.

This pr helps some users solve the problem by adapting the model to solve the error.
程序只对deepseek-reasoner进行了单独的支持，但是我发现对于一些模型，比如deepseek-chat等不支持tool_call的模型，程序没有对输入做转换。
我的代码加入了报错后的转换重试，确保代码不会因为用户消息在tool_call不在context中而导致整个代码失败。
目前有一些issue，包括#1020 这类issue，他们都反馈说会卡在blank页面并无限卡在step1，这是因为我们的browser-use默认安装的依赖包不包括google，错误没有被正常捕获而导致的。要修复这个问题应将google加入到依赖包中。但根本原因是因为用户使用了不受支持的模型。
此pr通过适配模型从而解决报错，进而变相的帮助一部分用户解决了这个问题。


pirate:
please convert the spaces to tabs, we use `ruff` which enforces tabs for indentation. thanks!

pirate:
Related PR: https://github.com/browser-use/browser-use/pull/736

PR #1053, opened 2025-03-18
Fix hardcoded model name check and replace with configurable param

sivaramaaa:
Currently browser-use has a hardcoded model name check for Deepseek like : 

```py
if model_name == 'deepseek-reasoner' or 'deepseek-r1' in model_name
```

This causes an issue if we are using Operouter where model name param can be like `deepseek/deepseek-r1` OR if we are using any other models like LLama etc . 

Screenshot 2025-03-18 at 7 12 38 AM

With the current fix user can easily use any models , Example code : 

```py
from browser_use import Agent, Controller, ActionResult, BrowserConfig, Browser
load_dotenv()

llm = ChatOpenAI(
    model="deepseek/deepseek-r1",
    base_url="https://openrouter.ai/api/v1"
)

    agent = Agent(
        task="Open google.com , Search Deepseek pricing",
        llm=llm,
        use_vision=False,
        max_failures=2,
		max_actions_per_step=1,
        tool_calling_method="raw",
        non_function_calling_models=True
    )
```

### DIFF
--- a/browser_use/agent/message_manager/utils.py
+++ b/browser_use/agent/message_manager/utils.py
@@ -33,11 +33,11 @@ def extract_json_from_model_output(content: str) -> dict:
 		raise ValueError('Could not parse response.')
 
 
-def convert_input_messages(input_messages: list[BaseMessage], model_name: Optional[str]) -> list[BaseMessage]:
+def convert_input_messages(input_messages: list[BaseMessage], model_name: Optional[str], non_function_calling_models: Optional[bool]) -> list[BaseMessage]:
 	"""Convert input messages to a format that is compatible with the planner model"""
 	if model_name is None:
 		return input_messages
-	if model_name == 'deepseek-reasoner' or 'deepseek-r1' in model_name:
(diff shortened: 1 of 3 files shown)


sivaramaaa:
**NOTE:** 

There is still one more hardcoded check in : if tool_calling_method is auto :   https://github.com/browser-use/browser-use/blob/577e15c963f7e3c499aac312ecb8da6e1f4733fe/browser_use/agent/service.py#L296

Also i am not sure if we require a new param `non_function_calling_models` or if we can simply re-use the existing `tool_calling_method` param and add checks bases on that (Any help from maintainer here could help me optimize this code better)

This PR fixes : https://github.com/browser-use/browser-use/issues/1020 
A slightly better fix compared to : https://github.com/browser-use/browser-use/pull/1048

an0th3rhuman:
Nice one. We need to check which model is working better for our needs. We don't have to be all model compatible

Mustafiz04 (approved):

Issue #1056, opened 2025-03-18
If the call to LLM fails due to some reason , throw an error message in the terminal


baluvijay:
### Problem Description

I was trying to run LLM on my local, and it keeps on repeating step1  instead of proceeding further and no helpful error message was thrown.When I went through the code ,I found that the api call to LLM was throwing an error.

### Proposed Solution

It would be great if we throw an error in the terminal if the LLM call is returning an error, be it due to token expiry, or invalid information passed into the LLM ( say invalid model name or something).
Right now the error is caught by an outside function and handled so meaning error messages are not shown

Maybe lets move the try catch block next to the invoke LLM function so that meaningful error message could be thrown.
I would like to work on this if its okay?

### Additional Context

![Image](https://github.com/user-attachments/assets/312211bb-90f7-4dba-a113-f8cb6cd600c1)

When a try catch is introduced you get really helpful error message in the terminal

![Image](https://github.com/user-attachments/assets/4190526e-8c37-4206-b4c1-9c411733318d)

ml5ah:
Implementing #1146. Can you write any suggestions in #1123?

PR #1059, opened 2025-03-18
log meaningful error message if llm doesnt give response

baluvijay:
PR for this issue: https://github.com/browser-use/browser-use/issues/1056
Please let me know if. you need any more changes to be done.
Thinking of adding similar error messages for the other llm invoking functions in the if-else branch too
I just added it onto the function with llm call, if there is a more elegant way ,with less repetitive code please guide me on that

### DIFF
--- a/browser_use/agent/service.py
+++ b/browser_use/agent/service.py
@@ -512,8 +512,7 @@ class Agent(Generic[Context]):
 			try:
 				output = self.llm.invoke(input_messages)
 			except Exception as e:
-				logger.info("error during llm invocation")
-				logger.error(e)
+				logger.error(f"Error during LLM invocation: {e}")
 			# TODO: currently invoke does not return reasoning_content, we should override invoke
 			output.content = self._remove_think_tags(str(output.content))
 			try:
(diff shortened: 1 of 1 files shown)


browser_use/agent/service.py
```diff
@@ -509,7 +509,11 @@ async def get_next_action(self, input_messages: list[BaseMessage]) -> AgentOutpu
 		input_messages = self._convert_input_messages(input_messages)
 
 		if self.tool_calling_method == 'raw':
-			output = self.llm.invoke(input_messages)
+			try:
+				output = self.llm.invoke(input_messages)
+			except Exception as e:
+				logger.info("error during llm invocation")
+				logger.error(e)
```
Simaky, line 516: I believe it might be more appropriate to use a single error log instead of separate info and error logs, as this appears to be an error rather than just informational.
Like that: `logger.error(f"Error during LLM invocation: {e}")`

baluvijay: good point will update the PR like that

MagMueller:
This should stay: max_steps: int = 100

MagMueller:
This doesn't seem correct to me because if there is an error, it would currently just continue and try to get output.content even though there is no output - it should raise the error.

baluvijay:
> This should stay: max_steps: int = 100

fixed it sorry , was trying with fewer steps



> This doesn't seem correct to me because if there is an error, it would currently just continue and try to get output.content even though there is no output - it should raise the error.

it actually doesnt. My api key was invalid and the process kept repeating step1 and not show any error. I think catching it like this will help people who are getting started. This will throw an error in the console if model is invalid/api is facing an error

@MagMueller

pirate:
@baluvijay currently the code is broken because it does not define the variables in the case of exception, execution will continue after the log lines and then crash immediately. You should add a `raise` after each log line.

PR #1060, opened 2025-03-18
Feature - Add custom captcha solver

Simaky:
## ℹ️ PR Responsibility

- Add a new Protocol for a custom captcha solver
- Add an example with the implementation https://github.com/Xewdy444/Playwright-reCAPTCHA

## 🛠 Side effects

- Upgrade the Playwright version to 1.51.0 to fix the issue: https://github.com/Xewdy444/Playwright-reCAPTCHA/issues/151

## 🎥 Screen Recording: Run example
https://github.com/user-attachments/assets/d83d970c-a717-4e4e-a57b-e2f681c492fa

### DIFF
--- a/browser_use/agent/captcha_solver.py
+++ b/browser_use/agent/captcha_solver.py
@@ -4,7 +4,6 @@ from typing import Protocol, TYPE_CHECKING
 if TYPE_CHECKING:
     from browser_use.agent.service import Agent
 
-
 class CaptchaSolverProtocol(Protocol):
     async def solve_captcha(self, context: Agent) -> None:
         ...
--- a/browser_use/agent/captcha_solver.py
+++ b/browser_use/agent/captcha_solver.py
@@ -1,0 +1,10 @@
+from __future__ import annotations
+from typing import Protocol, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from browser_use.agent.service import Agent
+
+
+class CaptchaSolverProtocol(Protocol):
+    async def solve_captcha(self, context: Agent) -> None:
+        ...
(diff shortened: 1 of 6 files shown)


browser_use/agent/captcha_solver.py
```diff
@@ -0,0 +1,10 @@
+from __future__ import annotations
+from typing import Protocol, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from browser_use.agent.service import Agent
+
```
dymytrii-zmarko, line 6:
```suggestion
```

examples/custom_captcha_solver/google_recaptcha_example.py
```diff
@@ -0,0 +1,40 @@
...
+from dotenv import load_dotenv
+from langchain_openai import ChatOpenAI
+
+from browser_use.agent.service import Agent
+from browser_use.browser.browser import Browser, BrowserConfig
+
+from examples.custom_captcha_solver.captcha_solver import ReCaptchaSolver
+
+load_dotenv()
+
```
dymytrii-zmarko, line 11:
```suggestion
```

dymytrii-zmarko (approved):
LGTM!
GREAT JOB MAN, you have saved 100 man-hours by preparing this pr

MagMueller:
I like that! Currently, it checks at every step. I think it's better to do this as a custom action, and when the model thinks it needs to solve a captcha, it simply calls the action.

zaheerahmad33:
Does it solves the Slider Rotational Captcha like Tiktok?

Simaky:
> Does it solves the Slider Rotational Captcha like Tiktok?

It depends on the captcha solver you used.
As an example I used a lib that solve only reCaptcha

pirate:
imo it's better to let the user load an extension like 2captcha into the browser and configure it themselves than trying to check for and solve captchas at the agent-level. Auto-captcha-solving is already working very well with these extensions and it's a difficult cat-and-mouse game to try and keep up with in-house.

PR #1076, opened 2025-03-19
Takeover plawyright browser for single AI actions

jersobh:
I've added the page object to the Browser class, and adjusted the functions to use the existing playwright's browser and context. So, now, it's possible to perform a single action using AI, instead of a fully automated operation.

### DIFF
--- a/browser_use/browser/browser.py
+++ b/browser_use/browser/browser.py
@@ -6,9 +6,10 @@ import asyncio
 import gc
 import logging
 from dataclasses import dataclass, field
+from typing import Optional
 
 from playwright._impl._api_structures import ProxySettings
-from playwright.async_api import Browser as PlaywrightBrowser
+from playwright.async_api import Browser as PlaywrightBrowser, BrowserContext as PlaywrightBrowserContext, Page
 from playwright.async_api import (
 	Playwright,
 	async_playwright,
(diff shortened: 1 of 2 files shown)


jersobh:
Note that the example available in the documentation won't work, as it actually expects a Browser-Use BrowserContext, and not a playwright BrowserContext. It will have errors while trying to call methods that won't exist in the Playwright context:

```python
from playwright.async_api import async_playwright

async with async_playwright() as p:
    browser = await p.chromium.launch()
    page = await browser.new_page()
    
    # Navigate to your desired page
    await page.goto("https://example.com")
Then, create a Browser Use Agent using the context of this page:
from browser_use import Agent

agent = Agent(
    task="Your task description",
    llm=your_llm,
    browser_context=page.context  # Use the context of the specific page
)

# Run the agent
await agent.run()
```

It will throw an error:

 ```
  'Browser' object has no attribute 'get_state'
```

**Comment by MagMueller** [comment]
Awesome!
@gregpr07 - what do you think?

**Comment by pirate** [comment]
I think we should build the remainder of this to get it working, playwright-interop is worth it.

**Comment by prasadpilla** [comment]
> I think we should build the remainder of this to get it working, playwright-interop is worth it.

@pirate, what are the remaining pieces here? I have a production use case for this and don't mind finishing the rest of it.

**Comment by pirate** [comment]
We have refactored the Browser models to make this much easier, please see the new design:

https://github.com/browser-use/browser-use/pull/1699

The new design allows you to do:

```python
from browser_use.browser import BrowserSession

browser_session = BrowserSession(
    browser_context=... normal playwright BrowserContext object here...,
    # and/or
    browser=... normal playwright Browser object here...,
    # and/or
    agent_current_page=... normal playwright Page object here...,
)

agent = Agent(browser_session=browser_session, ...)
```

PR #1078, opened 2025-03-19
Adds support for span and pre to be clickable

### LINKED ISSUE #1074: it is unable to click on span pre or div like elements.
### Problem Description

so  may website use span and pre element as a dynamic button, and if i need to click on it, browser-use can't do as it is not listed in interactive Elements.

### Proposed Solution

1. go the your_venv_path/.venv/lib/python3.11/site-packages/browser_use/dom/buildDomTree.js 
2. search for constant "interactiveElements" (line : 465)
3. add  "span" and "pre" if want to interact (click or hover) with these element or if want to interact with any other custom element you can add them in this set.
4. Done.

### Alternative Solutions

_No response_

### Additional Context

_No response_

vmesel:
Closes #1074 adding the `pre` and `span` tag to clickable elements

### DIFF
--- a/browser_use/dom/buildDomTree.js
+++ b/browser_use/dom/buildDomTree.js
@@ -494,7 +494,7 @@
     const interactiveElements = new Set([
       "a", "button", "details", "embed", "input", "menu", "menuitem",
       "object", "select", "textarea", "canvas", "summary", "dialog",
-      "banner"
+      "banner", "div", "span", "pre"
     ]);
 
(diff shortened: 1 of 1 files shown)


jersobh:
Nice adding :)

raivisdejus:
`label` elements are often used for custom checkbox styling. Example https://www.w3schools.com/howto/howto_css_custom_checkbox.asp

raivisdejus:
In my tests having the any `div` considered as interactive causes side effects where AI can't target specific element as there are too many interactive elements on the page, it tries to click unrelated divs.

pirate:
I have not seen many clickable pre elements, and only a few clickable spans. perhaps this should be part of a second pass that searches harder only when needed to avoid too much noise?

MagMueller:
Thanks @raivisdejus - yes this adds many elements as interactive which are not. -

Issue #993, opened 2025-03-10
Make "greedy focus" an optional behavior instead of the default


righdforsa:
### Bug Description

Running browser-use with the default settings steals my window focus for every action run by the Agent. This damages my workflow, but I want to have the window open to passively monitor, so headless mode doesn't work for me, either.

We should minimize the effect of "bring_to_front" so it doesn't steal focus all the time. 

### Reproduction Steps

1. Create an agent
2. Run any task
3. Once browser has been launched, navigate to any other window
4. Next time the Agent starts a new action, it will bring the browser to front

### Code Sample

```python
# default quickstart agent script causes this behavior
```

### Version

main d5cab0a9e40147e25b21f29581f825886a54a4a6

### LLM Model

Claude 3.5 Sonnet

### Operating System

macOS 13.7.4

### Relevant Log Output

```shell

```

righdforsa:
Here is a patch to do it the way I recommend:

```
From afdb9f0dfc4af084aa8a521b09d5e86672009ad2 Mon Sep 17 00:00:00 2001
From: Christopher Peters <righdforsa@gmail.com>
Date: Mon, 10 Mar 2025 10:51:01 -0700
Subject: [PATCH] make greedy focus an option instead of the default

---
 browser_use/browser/browser.py |  4 ++++
 browser_use/browser/context.py | 13 +++++++++----
 2 files changed, 13 insertions(+), 4 deletions(-)

diff --git a/browser_use/browser/browser.py b/browser_use/browser/browser.py
index 9278ac3..4b56338 100644
--- a/browser_use/browser/browser.py
+++ b/browser_use/browser/browser.py
@@ -32,6 +32,9 @@ class BrowserConfig:
 		disable_security: True
 			Disable browser security features
 
+		greedy_focus: False
+			Whether browser windows should aggressively take focus when interacting
+
 		extra_chromium_args: []
 			Extra arguments to pass to the browser
 
@@ -48,6 +51,7 @@ class BrowserConfig:
 
 	headless: bool = False
 	disable_security: bool = True
+	greedy_focus: bool = False
 	extra_chromium_args: list[str] = field(default_factory=list)
 	chrome_instance_path: str | None = None
 	wss_url: str | None = None
diff --git a/browser_use/browser/context.py b/browser_use/browser/context.py
index 44b7595..0e794cd 100644
--- a/browser_use/browser/context.py
+++ b/browser_use/browser/context.py
[...]

Issue #974, opened 2025-03-08
get_current_page does not return current page when switching tabs


xander-io:
### Bug Description

When using multiple tabs, `browser_ctx.get_current_page()` only returns the last tab. This is a problem if you decide to use `browser_ctx.switch_to_tab`, as I would've expected `get_current_page` to return the page that I switched to.

### Reproduction Steps

1. Create a new Browser context
2. `await browser_ctx.navigate("https://google.com")`
3. `await browser_ctx.create_new_tab("https://youtube.com")`
4. `await browser_ctx.switch_to_tab(0)`
5. `page = await browser_ctx.get_current_page()`
6. `print(page.url)`
Expected https://google.com, result is https://youtube.com

### Code Sample

```python
`   browser = Browser()
    b_ctx = await browser.new_context()
    #playwright_browser = await browser.get_playwright_browser()
    await b_ctx.navigate_to("https://google.com")
    await b_ctx.create_new_tab("https://youtube.com/")
    await b_ctx.switch_to_tab(0)
    page = await b_ctx.get_current_page()
    print(page.url)
`
```

### Version

0.1.40

### LLM Model

Other (specify in description)

### Operating System

Windows 11

### Relevant Log Output

```shell

```

xander-io:
Just to get the conversation going, it looks like the at some point recently the `_get_current_page` code was simplified down to resorting to the last page. Not sure what the reasoning was or if that was the intended behavior. But, the line that seems to be the culprit is: https://github.com/browser-use/browser-use/blob/3076ba0e83f30b45971af58fe2aeff64472da812/browser_use/browser/context.py#L1166 

If this is the new expected behavior, please let me know what function I should be using instead. Otherwise, I'll probably resort to using an older version of browser-use until this is fixed. Thanks!

Yutong-Dai:
I've encountered the same issues. This becomes a problem, if we initailize a browser with some tabs being opened. Due to the line point out by @xander-io, the browser  keeps focusing  on the last tab,  even after action switch tab is taken.
```
return pages[-1] if pages else await session.context.new_page() 
```

agupagu:
Any idea on this issue:

INFO     [agent] 🛠️  Action 1/1: {"switch_tab":{"page_id":4}}
INFO     [controller] 🔄  Switched to tab 4
INFO     [agent] 📍 Step 24
INFO     [agent] ⚠ Eval: Failed - Misalignment in tab interaction leading to incorrect page.

[...]

Issue #660, opened 2025-02-11
Follow up tasks don't work with local browser


ms337:
### Bug Description

The example for running follow up tasks as given in `examples/features/follow_up_tasks.py` does not work for the local browser. 

One gets the error:

```
WARNING  [browser] Page load failed, continuing...
ERROR    [agent] ❌ Result failed 1/3 times:
 Browser.new_context: Target page, context or browser has been closed
```

### Reproduction Steps

Run the reproduction script given below.

The first task runs. The second "new task" that should've been run as a follow up fails.

### Code Sample

```python
import asyncio
from dotenv import load_dotenv
from langchain_openai import ChatOpenAI
from browser_use import Agent, Browser, BrowserConfig, Controller

load_dotenv()

async def main():
    
    # Initialize the model
    llm = ChatOpenAI(
        model='gpt-4o',
        temperature=0.0,
    )
    
    # Get your chrome path
    chrome_path = '/Applications/Google Chrome.app/Contents/MacOS/Google Chrome'
    browser = Browser(
        config=BrowserConfig(
            chrome_instance_path=chrome_path,
        )
    )

    task = 'Go to hn'
    agent = Agent(task=task, llm=llm, browser=browser)
    await agent.run()
    
    new_task = 'Go to the first article on AI'
    agent.add_new_task(new_task)
    await agent.run()

    await browser.close()

if __name__ == '__main__':
    asyncio.run(main())
```

### Version

pip: 0.1.36

### LLM Model

GPT-4o

### Operating System

macOS

### Relevant Log Output

```shell
INFO     [browser_use] BrowserUse logging setup complete with level info
INFO     [root] Anonymized telemetry enabled. See https://docs.browser-use.com/development/telemetry for more information.
INFO     [agent] 🚀 Starting task: Go to hn
INFO     [agent] 📍 Step 1
INFO     [browser] Reusing existing Chrome instance
INFO     [agent] 🤷 Eval: Unknown - No previous goal to evaluate.
INFO     [agent] 🧠 Memory: Hacker News is open in tab 1.
INFO     [agent] 🎯 Next goal: Switch to the Hacker News tab.
INFO     [agent] 🛠️  Action 1/1: {"switch_tab":{"page_id":1}}
INFO     [controller] 🔄  Switched to tab 1
INFO     [agent] 📍 Step 2
INFO     [agent] 👍 Eval: Success - Successfully switched to the Hacker News tab.
[...]

Issue #475, opened 2025-01-30
Target page, context or browser has been closed


lawzlo:
### Bug Description

INFO     [agent] 🚀 Starting task: go to google.com and type 'OpenAI' click search and give me the first url
INFO     [agent] 
📍 Step 1
WARNING  [browser] Page load failed, continuing...
ERROR    [agent] ❌ Result failed 1/3 times:
 Browser.new_context: Target page, context or browser has been closed
INFO     [agent] 
📍 Step 1
WARNING  [browser] Page load failed, continuing...
ERROR    [agent] ❌ Result failed 2/3 times:
 Browser.new_context: Target page, context or browser has been closed
INFO     [agent] 
📍 Step 1
WARNING  [browser] Page load failed, continuing...
ERROR    [agent] ❌ Result failed 3/3 times:
 Browser.new_context: Target page, context or browser has been closed
ERROR    [agent] ❌ Stopping due to 3 consecutive failures

### Reproduction Steps

Follow the instruction step by step.
Installed playwright.
Run Agent.
The chromium opened and closed in 1 second.
Got errors.

### Code Sample

```python
Same as instruction code sample.
```

### Version

git main branch

### LLM Model

GPT-4o

### Operating System

macOS 15.3

### Relevant Log Output

```shell

```

mazahaler:
I had a similar problem: I asked it to find a flight, and after it found one, Chromium immediately closed

nesterone:
I just tried `Starting task: go to google.com and type 'OpenAI' click search and give me the first url`

everything is ok 

```
browser-use                  0.1.23
```

wejarrard:
Did anyone figure this out? Im using version `0.1.34` and having this same error follow the example

vovanmix:
I have this issue when creating multiple agents sequentially, and it fails on the second iteration:

```
async def main():

    controller = Controller(output_model=X)

    for url in urls:
        agent = Agent(
            task="",
            llm=llm,
            browser=browser,
            controller = controller,
        )
        history = await agent.run()
        result = history.final_result()
        print(result)
```

`Browser.new_context: Target page, context or browser has been closed`

pirate:
Can you pull the latest `main` and try again, there have been several improvements to active tab detection in the last few days.

Ovi3:
> Can you pull the latest `main` and try again, there have been several improvements to active tab detection and context-re-use in the last few days.
> 
> If you want to re-use a context for multiple task runs you can use:
> 
[...]

PR #1080, opened 2025-03-20
Improve detection of foreground active tab using document.hasFocus()

pirate:
`(WIP) pending confirmation that this is useful/needed`

My code in this PR fixes the immediate issue of not being able to detect the most recently opened foreground tab properly, but the long-term fix is to refactor the code to better track which tab is "current" for the agent vs "current" for the human watching. Keeping those separate would allow a user to continue using their browser normally without breaking `browser-use`'s interactions with a particular tab.

If it's not a design goal of `browser-use` to allow parallel browser driving by agent & human on separate tabs, let me know and I will change my approach.

Subtasks:

- [x] adds a new helper `context.get_page_visibility(page)` to get tab visibility info
- [x] change `_get_current_page` to use new helper method and filter for focused tab
- [ ] make sure detection of newly opened tabs can distinguish human-opened vs agent-opened
	- create new `get_current_human_page` to detect which tab the human user is focused on (can use my new `document.Focus()` code)
	- create new `get_current_agent_page` to detect which tab the agent is focused on (can use `browser_context.on('page', handler)` and `context.expect_page()` to update an attr on self that keeps a reference to the current tab the agent is driving)

```python
async with context.expect_page() as new_page_info:
	# agent opens new tab 
	await page.get_by_text("Open new tab").click()
	
	# context handler immediately gives us the new tab reference,
	# no need to get_current_page() or iterate through all pages checking visiblity
    page = await new_page_info.value
```

Possibly related issues:

- https://github.com/browser-use/browser-use/issues/974 (should definitely fix this one)
- https://github.com/browser-use/browser-use/issues/993
- https://github.com/browser-use/browser-use/issues/475#issuecomment-2642422124
- https://github.com/browser-use/browser-use/issues/660

### DIFF
--- a/browser_use/browser/context.py
+++ b/browser_use/browser/context.py
@@ -1251,6 +1251,39 @@ class BrowserContext:
 		# Fallback to last page
 		return pages[-1] if pages else await session.context.new_page()
 
+        async def get_page_visibility(page):
+	    	cdp_session = await page.context.new_cdp_session(page)
+	    	target_info = await cdp_session.send("Target.getTargetInfo")
+	    	target_id = target_info["targetInfo"]["targetId"]
+	    
+	    	window_info = await cdp_session.send("Browser.getWindowForTarget", {"targetId": target_id})
+	    	window_id = window_info["windowId"]
+	    	window_bounds = await cdp_session.send("Browser.getWindowBounds", {"windowId": window_id})
+	    	window_state = window_bounds.get("bounds", {}).get("windowState")	 # The state can be "normal", "minimized", "maximized", "fullscreen"
+	    	is_foreground = window_state != "minimized"
+	    
+	    	is_visible = await cdp_session.send("Runtime.evaluate", {
+			"expression": "document.visibilityState === 'visible'",
+			"returnByValue": True
+	    	})
+		is_focused = await cdp_session.send("Runtime.evaluate", {
+			"expression": "document.hasFocus()",
+			"returnByValue": True
+	    	})
+                page_bounds = await cdp_session.send("Page.getLayoutMetrics")  # https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-getLayoutMetrics
+	    	await cdp_session.detach()
+	    
+	    	return {
+			"target_id": target_id,
+	        	"window_id": window_id,
+	        	"window_state": window_state,
+			"window_bounds": window_bounds,
+			"window_is_visible": is_foreground,
+			"page_is_visible": is_visible,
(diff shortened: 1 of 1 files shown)


gregpr07:
Hey, multiple problems with this approach:

1. (`get_page_visibility` should be internal method, first param self, because python obvs)
2. `_get_current_page` is spammed multiple times across the codebase, so it becomes super slow to spam that method on every single call (`get_current_page` -> 19 different places in this file) so we can't calculate it every time, especially using third party cdp connections (latency kills you). I'm not 100% sure what the fix could be for this. Maybe some sort of cache of current active websites (but that's how we had it now and it didn't work well).
3. no types on `page` input and dict output -> something like codeblock below is much nicer

```python
                ...
		return PageVisibility(
			target_id=target_id,
			window_id=window_id,
			window_state=window_state,
			window_bounds=window_bounds,
			window_is_visible=is_foreground,
			page_is_visible=bool(is_visible),
			page_is_focused=bool(is_focused),
			page_bounds=page_bounds,
		)
```

pirate:
> (get_page_visibility should be internal method, first param self, because python obvs)

Whoops embarassing, thats what I get for hastily coding this in the github UI without cursor and my linters

> _get_current_page is spammed multiple times across the codebase

Yes I discovered this as well after making this PR. I see you are using it generically to get the page the agent should be actively focused on, but there are some issues because the human can change the active tab in a headful browser and mess it up.

> third party cdp connections (latency kills you)

right, makes sense

> Maybe some sort of cache of current active websites (but that's how we had it now and it didn't work well).

Ah ok that's sort of what I was suggesting above. What are your thoughts on the snippet I show above with the context manager? Thats the recommended way to do it with playwright according to their docs. If you just keep the reference immediately and pass it around, you dont need the stateful cache of agent pages.


> no types on page input and dict output -> something like codeblock below is much nicer
> ```python
> return PageVisibility(
> 	target_id=target_id,
> 	window_id=window_id,
> 	window_state=window_state,
> 	window_bounds=window_bounds,
> 	window_is_visible=is_foreground,
> 	page_is_visible=bool(is_visible),
> 	page_is_focused=bool(is_focused),
> 	page_bounds=page_bounds,
> )
> ```

Actually tbh I think it's better to remove all those other attrs until you really need them. I was just showing off all the APIs I know haha, you only really need `page_is_focused` (and maybe `page_is_visible`) for this check.

pirate:
I'm going to close this PR for now as I think my fixes in my other PRs help the situation enough on startup.

Issue #1044, opened 2025-03-17
how to change other search ?


lonngxiang:
### Bug Description

The default search engine is Google. How do I change it to another one, like Bing?

### Reproduction Steps

1\

### Code Sample

```python
111
```

### Version

0.12

### LLM Model

DeepSeek Coder

### Operating System

windows11

### Relevant Log Output

```shell

```

Lathezero:
i want woo

dwniu2000:
it is nothing to do with broswer use, it is all about the llm you use to plan the task. the llm itself decide to use google at the first place. You can simply tell llm by  putting some words like "use bing or baidu to..." in your task promote

dwniu2000:
you have to add a new bing search action in your own controller like :
          class Controller_Ext(Controller)
		async def search_bing(params: SearchGoogleAction, browser: BrowserContext):
			page = await browser.get_current_page()
			await page.goto(f'https://www.bing.com/search?q={params.query}')
			await page.wait_for_load_state()
			msg = f'🔍  Searched for "{params.query}" in Bing'
			logger.info(msg)
			return ActionResult(extracted_content=msg, include_in_memory=True)


and pass it to agent like: Agent(controller= your controller, .......)

yonghuname:
I have some modify，you can see my issue

Cfomodz:
To use Bing instead of Google, modify your task instructions to explicitly navigate to Bing: 'go to bing.com and search for [term]' instead of letting the agent choose.

PR #1086, opened 2025-03-20
feat: Added the ability to select the default search engine

umerhd:
## 📌 Description  
Added a default_engine option for users that want to use a different search engine.

e.g. yahoo 

## 🔗 Related Issue  
https://github.com/browser-use/browser-use/issues/1044

## ✨ Type of Change  
- [X] 🚀 New Feature  
- [ ] 🐛 Bug Fix  
- [ ] 🛠 Refactor  
- [x] 📖 Documentation Update  
- [ ] 🔄 Other (please describe)  

## ✅ How Has This Been Tested?  
  
Unit Test

### DIFF
--- a/browser_use/agent/service.py
+++ b/browser_use/agent/service.py
@@ -86,6 +86,7 @@ class Agent(Generic[Context]):
 		browser: Browser | None = None,
 		browser_context: BrowserContext | None = None,
 		controller: Controller[Context] = Controller(),
+		default_search_engine: str = "google",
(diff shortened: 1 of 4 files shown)


MagMueller:
It seems now it's duplicated—the user can set the search engine, but the agent can also decide the search engine. I think it should be only one of the two. I would give the decision to the LLM, then the user can just specify it in the prompt.

umerhd:
@MagMueller Makes sense. I'll close it.

PR #1093, opened 2025-03-21
Ensure viewPortExpansion properly filters elements outside the viewport

mathisarends-viadee:
#### **Description**  
While working with browser interactions, I noticed that the `viewPortExpansion` parameter had no effect on filtering elements in the `SelectorMap`. As a result, elements outside the viewport were still selectable by the LLM, even when `viewport_expansion = 0`.  

This issue was caused by the fact that `isInExpandedViewport` was **defined but never actually used**. The previous implementation contained an inline viewport check, but it did not consistently exclude elements that were outside the visible area on scrollable pages.  

#### **Fix**  
- The existing function `isInExpandedViewport` is now used to filter out elements properly.  
- This ensures that only elements within the expanded viewport are included in the `SelectorMap`.  
- The behavior now correctly respects `viewport_expansion` values, preventing unintended selections.  

#### **Impact**  
- Fixes incorrect element selection on scrollable pages.  
- Ensures that `viewport_expansion` behaves as expected.  
- Reduces unnecessary processing of off-screen elements.  

Let me know if any further refinements are needed! 🚀

### DIFF
--- a/browser_use/dom/buildDomTree.js
+++ b/browser_use/dom/buildDomTree.js
@@ -858,26 +858,9 @@
       return null;
     }
 
-    // Early viewport check - only filter out elements clearly outside viewport
-    if (viewportExpansion !== -1) {
-      const rect = getCachedBoundingRect(node);
-      const style = getCachedComputedStyle(node);
-
-      // Skip viewport check for fixed/sticky elements as they may appear anywhere
-      const isFixedOrSticky = style && (style.position === 'fixed' || style.position === 'sticky');
-
-      // Check if element has actual dimensions
-      const hasSize = node.offsetWidth > 0 || node.offsetHeight > 0;
-
-      if (!rect || (!isFixedOrSticky && !hasSize && (
-        rect.bottom < -viewportExpansion ||
-        rect.top > window.innerHeight + viewportExpansion ||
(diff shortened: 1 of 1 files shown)


pirate:
@MagMueller you last touched this code, is this fix ok? or was your custom code added because `isInExpandedViewport` was insufficient somehow?

pirate:
I believe it was removed for a reason, hang tight @mathisarends-viadee the team is evaluating a lot of improvements in o this area of the code currently and this fix might not be necessary soon.

pirate:
I believe @gregpr07 has already fixed this issue on main now. Can you test `main` and see if it's still happening for you @lln556 @mathisarends-viadee?

lln556:
@pirate Hi, I have tried the latest code in buildDomTree.js. When I set viewportExpansion=0, it actually only detect elements in viewport. But when I set viewportExpansion to -1 or any positive number, it still only detect elements in viewport. It's not clear to me if I'm ignoring some of the settings because I don't know much about the frontend. Please check it again, thanks！

mathisarends-viadee:
I'm experiencing the same issue as @lln556 — isInViewport seems to be undefined.
I'll open a PR shortly and link it here once it's ready.

image

PR #1094, opened 2025-03-21
Added option to customize GIF output path

Anshdeep-Singh:
- Add gif_output_path parameter to AgentSettings (default: 'agent_history.gif')

- Add example script demonstrating custom GIF path usage ('custom_gif_path.py')

This change allows users to specify a custom path for generated GIFs


MagMueller:
Hi - we have that functionality already by setting `generate_gif="abc.gif"` so it can be bool or str:)

PR #1095, opened 2025-03-21
Feature to save webpage as PDF to given path

ml5ah:
This is a simple feature to save a webpage as PDF to a given path.
Brief summary to changes:
    1. service.py: A new SavePDFAction has been introduced. 
    2. views.py: Definition for SavePDFAction params. 

Happy to incorporate any feedback.

### DIFF
--- a/browser_use/controller/service.py
+++ b/browser_use/controller/service.py
@@ -168,14 +168,16 @@ class Controller(Generic[Context]):
 		# Save PDF
 		@self.registry.action(
 			'Save the current page as a PDF file',
-			param_model=SavePDFAction,
 		)
-		async def save_pdf(params: SavePDFAction, browser: BrowserContext):
+		async def save_pdf(browser: BrowserContext):
 			page = await browser.get_current_page()
-			await page.emulate_media('screen')
-			await page.pdf(path=params.file_path, format='A4', print_background=params.print_background)
+			short_url = re.sub(r'^https?://(?:www\.)?|/$', '', page.url)
+			slug = re.sub(r'[^a-zA-Z0-9]+', '-', short_url).strip('-').lower()
+			sanitized_filename = f'{slug}.pdf'
 
-			msg = f"Saving page with URL {page.url} as PDF to {params.file_path}"
+			await page.emulate_media('screen')
+			await page.pdf(path=sanitized_filename, format='A4', print_background=params.print_background)
+			msg = f"Saving page with URL {page.url} as PDF to ./{sanitized_filename}"
 			logger.info(msg)
 			return ActionResult(extracted_content=msg, include_in_memory=True)
 
--- a/browser_use/controller/service.py
+++ b/browser_use/controller/service.py
@@ -174,7 +174,7 @@ class Controller(Generic[Context]):
 		async def save_pdf(params: SavePDFAction, browser: BrowserContext):
 			page = await browser.get_current_page()
 			await page.emulate_media(params.media_type)
-			await page.pdf(path=params.file_name, format='A4')
+			await page.pdf(path=params.file_path, format=params.format, print_background=params.print_background)
 
 			msg = f"Saving page with URL {page.url} as PDF to {params.file_path}"
 			logger.info(msg)
--- a/browser_use/controller/service.py
+++ b/browser_use/controller/service.py
@@ -22,7 +22,6 @@ from browser_use.controller.views import (
 	OpenTabAction,
 	ScrollAction,
 	SearchGoogleAction,
-	SavePDFAction,
 	SendKeysAction,
 	SwitchTabAction,
 )
--- a/browser_use/controller/service.py
+++ b/browser_use/controller/service.py
@@ -173,8 +173,8 @@ class Controller(Generic[Context]):
 		)
 		async def save_pdf(params: SavePDFAction, browser: BrowserContext):
 			page = await browser.get_current_page()
-			await page.emulate_media(params.media_type)
-			await page.pdf(path=params.file_path, format=params.format, print_background=params.print_background)
+			await page.emulate_media('screen')
+			await page.pdf(path=params.file_path, format='A4', print_background=params.print_background)
 
 			msg = f"Saving page with URL {page.url} as PDF to {params.file_path}"
 			logger.info(msg)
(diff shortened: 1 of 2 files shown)


MagMueller:
Awesome! What happens if the file location already exists? Can the agent cause any damage to the system, e.g., by replacing the home folder, etc.?

ml5ah:
Thats a good question. At the moment, there is no specific safeguard against writing to the filesystem except if the python process does not have write permissions for a folder. 
The expected behavior is that, as part of the SavePDFAction, a file path is returned by the model (with an appropriate file name based on page URL, content, etc.). Since it only writes files to systems, I'm thinking its unlikely that it can replace a crucial folder.

How do you suggest we can introduce some guardrails?

pirate:
webdriver-bidi (the new CDP replacement) return screenshots and PDFs as base64 to the caller to avoid the issue of direct filesystem writes from the browser. you can do this in playwright too  by running some https://developer.mozilla.org/en-US/docs/Web/API/Screen_Capture_API JS in the context of the page and returning base64

browser_use/controller/views.py
```diff
@@ -48,6 +48,13 @@ class ExtractPageContentAction(BaseModel):
 	value: str
 
 
+class SavePDFAction(BaseModel):
+	file_path: str
+	media_type: Literal['screen', 'print'] = 'screen'
+	format: Literal['A4', 'A3', 'A5', 'Legal', 'Letter'] = 'A4'
+	print_background: bool = False
```
pirate, line 55: imho this is too much choice for LLMs, this will work for 99% of ppl fixed at `A4` & `screen`. if someone really actually needs it, we could add env vars for these options

ml5ah:
@pirate thanks for the feedback. So is your proposal to add a JS file in browser/ and then have the save_pdf controller action call a JS function? The function would then return a base64 image/pdf and we can write to filesystem separately? 

How exactly would that prevent malicious intent to write to file system though?

pirate:
No concrete proposal to change this PR, just making a note for future that we may want to switch methods as BIDI gets closer to being standardized. 

I think the security risk is acceptable for now as long as the Python code puts output in a dedicated output folder and doesn't allow writes elsewhere.

browser_use/controller/views.py
ml5ah: Sure. I set the media type as 'screen' and format as 'A4'. Kept the print_background parameter since that does change the actual pdf that is stored, and might be useful for customization. @pirate , let me know if the PR looks good now.

ml5ah:
@pirate , makes sense. Please let me know if you are able to add a note somewhere for a future enhancement (or I can write it)

browser_use/controller/views.py
```diff
@@ -48,6 +48,11 @@ class ExtractPageContentAction(BaseModel):
 	value: str
 
 
+class SavePDFAction(BaseModel):
+	file_path: str
+	print_background: bool = False
```
pirate, line 53: I think we can remove params entirely, just save to [page-title-slug].pdf

browser_use/controller/service.py
```diff
@@ -22,6 +22,7 @@
 	OpenTabAction,
 	ScrollAction,
 	SearchGoogleAction,
+	SavePDFAction,
```
pirate, line 25:
```suggestion
```

browser_use/controller/views.py
ml5ah: My only concern is, If the user wants to save to a specific folder, by saying "go to apple.com and save the homepage as a PDF to /Users/abc/xyz", we would lose this capability. What are your thoughts?

browser_use/controller/service.py
```diff
@@ -164,6 +165,20 @@ async def input_text(params: InputTextAction, browser: BrowserContext, has_sensi
...
+		
+		# Save PDF
+		@self.registry.action(
+			'Save the current page as a PDF file',
+			param_model=SavePDFAction,
+		)
+		async def save_pdf(params: SavePDFAction, browser: BrowserContext):
+			page = await browser.get_current_page()
+			await page.emulate_media('screen')
+			await page.pdf(path=params.file_path, format='A4', print_background=params.print_background)
```
pirate, line 177:
```suggestion
		)
		async def save_pdf(browser: BrowserContext):
			page = await browser.get_current_page()
			short_url = re.sub(r'^https?://(?:www\.)?|/$', '', page.url)
			slug = re.sub(r'[^a-zA-Z0-9]+', '-', short_url).strip('-').lower()

			await page.emulate_media('screen')
			await page.pdf(path=f'{slug}.pdf', format='A4', print_background=params.print_background)
```

browser_use/controller/service.py
```diff
@@ -164,6 +165,20 @@ async def input_text(params: InputTextAction, browser: BrowserContext, has_sensi
...
+
+			msg = f"Saving page with URL {page.url} as PDF to {params.file_path}"
```
pirate, line 179:
```suggestion
		)
		async def save_pdf(browser: BrowserContext):
			page = await browser.get_current_page()
			short_url = re.sub(r'^https?://(?:www\.)?|/$', '', page.url)
			slug = re.sub(r'[^a-zA-Z0-9]+', '-', short_url).strip('-').lower()
			sanitized_filename = f'{slug}.pdf'

			await page.emulate_media('screen')
			await page.pdf(path=sanitized_filename, format='A4', print_background=params.print_background)
			msg = f"Saving page with URL {page.url} as PDF to ./{sanitized_filename}"
```

browser_use/controller/views.py
pirate: yes that is intentional. the LLM should not be able to write to arbitrary paths on the filesystem for security reasons, we should definitely restrict it to the current working directory or a dedicated temp directory.

ml5ah: Done

MagMueller:
@ml5ah thanks!
Can you think of a system on how to still get the saved path as a developer when the agent is done to further process it? For example, via ActionResult.

pirate:
@MagMueller that ties into @gregpr07's work on file uploads. lets wait until he merges that and then use a consistent naming, santiization, and temp dir approach for all download/upload handling

PR #1105, opened 2025-03-22
chore: minor refactor

DarkPhoenix2704:
- Added early return in possible scenarios and reduced branching
- Updated some comments

### DIFF
--- a/browser_use/browser/browser.py
+++ b/browser_use/browser/browser.py
@@ -5,15 +5,13 @@ Playwright browser on steroids.
 import asyncio
 import gc
 import logging
+import subprocess
 from dataclasses import dataclass, field
 from typing import Literal
 
+import requests
 from playwright._impl._api_structures import ProxySettings
-from playwright.async_api import Browser as PlaywrightBrowser
-from playwright.async_api import (
-	Playwright,
-	async_playwright,
-)
+from playwright.async_api import Browser as PlaywrightBrowser, Playwright, async_playwright
 
 from browser_use.browser.context import BrowserContext, BrowserContextConfig
 from browser_use.utils import time_execution_async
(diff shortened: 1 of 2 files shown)


pirate:
any functionality changes or is this just style/refactoring?

MagMueller:
What do you mean with `Added early return in possible scenarios and reduced branching`.?

DarkPhoenix2704:
> any functionality changes or is this just style/refactoring?

Just refactoring

> What do you mean with Added early return in possible scenarios and reduced branching.?

Was going through the code and it was difficult to follow the execution flow with unnecessary assignments and if else branching. 

Before

```python
if not self.config._force_keep_browser_alive:
	if self.playwright_browser:
		await self.playwright_browser.close()
		del self.playwright_browser
	if self.playwright:
		await self.playwright.stop()
		del self.playwright

```

After
```python3
if self.config._force_keep_browser_alive:
	return
if self.playwright_browser:
	await self.playwright_browser.close()
	self.playwright_browser = None
if self.playwright:
	await self.playwright.stop()
	self.playwright = None
```

browser_use/browser/browser.py
```diff
@@ -246,24 +224,21 @@ async def _setup_browser(self, playwright: Playwright) -> PlaywrightBrowser:
...
+			if self.playwright:
+				await self.playwright.stop()
+				self.playwright = None
 		except Exception as e:
 			logger.debug(f'Failed to close browser properly: {e}')
 		finally:
-			self.playwright_browser = None
-			self.playwright = None
-
 			gc.collect()
```
DarkPhoenix2704, line 238: I am not sure if the earlier one has a correct behaviour. The playwright_browser and playwright is set to None irrespective of `_force_keep_browser_alive`.

pirate: I believe the existing behavior is correct, it releases the playwright channel memory but does not close the browser, which is what's expected if the user sets `_force_keep_browser_alive`.

DarkPhoenix2704: updated.

MagMueller:
There are so many changes, - could you break it down into smaller PRs?

PR #1108, opened 2025-03-23
Simplify browser setup

SmartManoj:
Removed unnecessary checks.


MagMueller:
We could even put the browser assignment outside?

browser_use/agent/service.py
```diff
@@ -199,14 +199,12 @@ def __init__(
 		# Browser setup
 		self.injected_browser = browser is not None
 		self.injected_browser_context = browser_context is not None
-		self.browser = browser if browser is not None else (None if browser_context else Browser())
```
SmartManoj, line 202: `self.browser = browser or (None if browser_context else Browser())`

Yes, but readability reduces.

Issue #1100, opened 2025-03-21
Adding callback causes agent to break


ellispinsky:
### Bug Description

When I add ` register_new_step_callback=new_step_callback` argument to the agent, the agent does not work and gets stuck in an endless loop. 

### Reproduction Steps

1. Config agent normally
2. Config agent with a simple callback with function signature `def new_step_callback(state: BrowserState, model_output: AgentOutput, steps: int)`

### Code Sample

```python
async def main():
    agent = Agent(
        task="go to the NYTimes and take a screenshot of the front page",
        llm=llm,
        register_new_step_callback=new_step_callback
    )
    result = await agent.run()
```

### Version

 0.1.40

### LLM Model

GPT-4o

### Operating System

Mac os

### Relevant Log Output

```shell

```

taijusanagi:
I'm not sure if this helps, but I ran into the exact same issue earlier. In my case, the callback needed to be asynchronous (async def). Changing it to async fixed it for me!

```
    def new_step_callback(state: BrowserState, model_output: AgentOutput, steps: int):
        print("this does not work")
```

```
    async def new_step_callback(state: BrowserState, model_output: AgentOutput, steps: int):
        print("this works")
```

sthitap0:
I see a problem where if I call a large function inside the callback, the controller doesn't take the action.

PR #1109, opened 2025-03-23
fix: allow callback to be sync functions

DarkPhoenix2704:
Closes https://github.com/browser-use/browser-use/issues/1100

Updates the Agent class to support both synchronous (def) and asynchronous (async def) callbacks for register_new_step_callback and register_done_callback

### DIFF
--- a/browser_use/agent/service.py
+++ b/browser_use/agent/service.py
@@ -1,12 +1,13 @@
 from __future__ import annotations
 
 import asyncio
+import inspect
 import json
 import logging
 import re
 import time
 from pathlib import Path
-from typing import Any, Awaitable, Callable, Dict, Generic, List, Optional, TypeVar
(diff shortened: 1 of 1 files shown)


agn-7:
I didn't know that before, the step callback function had to be async! instead, I was using a sync function for the step callback, and it worked well in version 0.1.37!

agn-7:
@DarkPhoenix2704  did you mean "allow callback to be **async** functions "?

DarkPhoenix2704:
Earlier it used to be support only async functions. Now you can use async and sync

agn-7:
Ahh, I am using version 0.1.37, it uses only sync callback, apparently in the newer versions it became async only, and now with your PR it supports both . . . @DarkPhoenix2704

PR #1116, opened 2025-03-23
Added feature for continuing task execution after reaching maximum steps

Anshdeep-Singh:
### The agent now can continue running tasks after reaching the maximum steps:

   - Added user prompts when approaching the last step, giving users the option to add more steps before the agent tries to finish
   - Users can now specify how many additional steps they want to add
  
Screenshot 2025-03-23 at 3 04 25 AM

Improved resource cleanup after finishing:
   - Enhanced browser resource cleanup to check if resources are already closed before attempting to close them
   - Implemented silent error handling for cleanup operations to prevent unnecessary error messages
   - Added checks for browser.is_connected() and browser_context.is_closed() to avoid redundant close operations

Added test case:
   - Created a test script with a multi-step task to verify the new functionality test_agent_additional_steps.py

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,5 +1,6 @@
 OPENAI_API_KEY=
 ANTHROPIC_API_KEY=
+GEMINI_API_KEY=
 
 # Set to false to disable anonymized telemetry
 ANONYMIZED_TELEMETRY=true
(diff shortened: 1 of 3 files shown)


MagMueller:
`input()` blocks the main thread if you run many agents, and many people don't want human input. I would suggest you implement this function in your own script. If you want to continue, then just increase the number of max steps and continue the agent.

Anshdeep-Singh:
Okay, sorry, I thought this would work as a failsafe for larger tasks, but I missed how this messes things up when using multiple agents.

I think making it an optional parameter in the agent settings would be more flexible, as this won't affect the default agent behavior.

Thanks for the feedback!

PR #1117, opened 2025-03-23
add the html download feature

ai-naymul:
save the html in the current directory with the timestamp to not to overwrite it again...

reference: #886

@pirate @MagMueller


pirate:
Thanks for fixing that @ai-naymul!


> I think this is best moved to the `examples/use-cases` subdirectory so people can use it when they need it. There are many other better scraping tools out there if you really just need to download raw html.

Maybe @MagMueller can chime in about this part, what are your thoughts on keeping this in the main services vs moving it to examples?

pirate:
No we are not giving the LLM unrestricted filesystem access to save to any path on disk. At most, you could suggest a filename, but not the entire path @Animaxx

ai-naymul:
> Are we able to save the file to specified path?

It saves the html in the project directory so you can add the script after saving there move that file to your specific location as well

browser_use/controller/service.py
```diff
@@ -330,6 +330,42 @@ async def extract_content(
 				msg = f'📄  Extracted from page\n: {content}\n'
 				logger.info(msg)
 				return ActionResult(extracted_content=msg)
+		
+		# HTML Download
+		@self.registry.action(
+			'Download the full HTML content of the current page',
```
pirate, line 336:
```suggestion
			'Save the raw HTML content of the current page to a local file',
```

browser_use/controller/service.py
```diff
@@ -330,6 +330,42 @@ async def extract_content(
 				msg = f'📄  Extracted from page\n: {content}\n'
 				logger.info(msg)
 				return ActionResult(extracted_content=msg)
+		
+		# HTML Download
+		@self.registry.action(
+			'Save the raw HTML content of the current page to a local file',
+			param_model=NoParamsAction,
+		)
+		async def download_html(_: NoParamsAction, browser: BrowserContext) -> ActionResult:
```
pirate, line 339:
```suggestion
		async def save_html_to_file(_: NoParamsAction, browser: BrowserContext) -> ActionResult:
```

cclauss:
Please try to avoid merging pull requests that have failing GitHub Actions.

PR #1142, opened 2025-03-25
Update supported-models.mdx

deanyxu:
Update examples with GoogleGenerativeAI and DeepSeek

- For the GoogleGenerativeAI example: there is some redundant code and since there is no need to explicitly pass in api_key for GoogleGenerativeAI, simply loading `GOOGLE_API_KEY` into the env vars would be sufficient (reference: https://python.langchain.com/api_reference/google_genai/chat_models/langchain_google_genai.chat_models.ChatGoogleGenerativeAI.html)
- For the DeepSeek examples: add the missing logic of loading api_key


pirate:
thanks!

PR #1145, opened 2025-03-26
Added missing env variables in .env.example 

ml5ah:
Minor updates to 2 files:

Added env variables to the example file since from the code, it is not obvious that these are supported and what the env variables needed are. 
We can provide all in the example file, and devs can copy to .env and provide values accordingly. 

Added check for empty vars for azure openai example


pirate (approved):

pirate:
nice, thanks

PR #1146, opened 2025-03-26
Improve error logging for LLM API calls and missing env variables

ml5ah:
Draft PR for better exception handling

1. Based on LLM set during Agent initialization, will check for corresponding env variables, if empty --> raise ValueError and exit immediately.
2. If invalid credentials, will raise an LLMException with status 401 and add a debug log informing user of invalid keys.


browser_use/agent/service.py
```diff
@@ -519,7 +540,12 @@ async def get_next_action(self, input_messages: list[BaseMessage]) -> AgentOutpu
 		input_messages = self._convert_input_messages(input_messages)
 
 		if self.tool_calling_method == 'raw':
-			output = self.llm.invoke(input_messages)
+			logger.debug(f"Using {self.tool_calling_method} for {self.chat_model_library}")
+			try:
+				output = self.llm.invoke(input_messages)
+			except Exception as e:
+				logger.error(f'Failed to invoke model: {str(e)}')
+				raise LLMException(401, 'Failed to invoke model')
```
pirate, line 548:
```suggestion
				raise LLMException(401, 'LLM API call failed') from e
```

browser_use/agent/service.py
```diff
@@ -238,6 +245,20 @@ def _set_message_context(self) -> str | None:
 				self.settings.message_context = f'Available actions: {self.available_actions}'
 		return self.settings.message_context
 
+	def _check_env_variables(self) -> bool:
```
pirate, line 248:
```suggestion
def check_env_variables(keys: list[str], any_or_all=all) -> bool:
    """top-level helper function to check env var values"""
    return any_or_all(os.getenv(key).strip() for key in keys)
```

- can be a pure util function so we can import it elsewhere too (e.g. from the examples)
- then at startup we can call and check for the keys needed for that run based on config e.g. `check_env_variables(['OPENAI_API_KEY', 'CHROME_USER_DATA_DIR'])`

ml5ah: Makes sense

browser_use/agent/service.py
ml5ah: Sure

browser_use/agent/service.py
pirate: yes, moving it to `utils.py` is what I had in mind 👍

browser_use/agent/service.py
```diff
@@ -238,6 +245,20 @@ def _set_message_context(self) -> str | None:
...
+		"""Check if all required environment variables are set"""
+		if isinstance(self.llm, ChatOpenAI):
+			if not os.getenv('OPENAI_API_KEY'):
+				logger.error('OPENAI_API_KEY is not set')
+				return False
+
+		if isinstance(self.llm, AzureChatOpenAI):
+			if not os.getenv('AZURE_OPENAI_API_KEY') or not os.getenv('AZURE_OPENAI_ENDPOINT'):
+				logger.error('AZURE_OPENAI_API_KEY or AZURE_OPENAI_ENDPOINT is not set')
+				return False
```
pirate, line 258: Here is the list of supported ones so far btw:

```
OPENAI_API_KEY=
ANTHROPIC_API_KEY=
AZURE_ENDPOINT=
AZURE_OPENAI_API_KEY=
GEMINI_API_KEY=
DEEPSEEK_API_KEY=
```

```
# could do something like:
llm_env_vars = {
    'ChatOpenAI': 'OPENAI_API_KEY',
    'AzureChatOpenAI': 'AZURE_OPENAI_API_KEY',
    ...
}
llm_api_key_var = llm_env_vars[self.llm.__class__.__name__]
if not check_env_variables([llm_api_key_var], all):
    raise ...
```

ml5ah: Yep perfect, following exactly this structure

browser_use/agent/service.py
```diff
@@ -531,12 +545,24 @@ async def get_next_action(self, input_messages: list[BaseMessage]) -> AgentOutpu
...
 			structured_llm = self.llm.with_structured_output(self.AgentOutput, include_raw=True)
-			response: dict[str, Any] = await structured_llm.ainvoke(input_messages)  # type: ignore
-			parsed: AgentOutput | None = response['parsed']
+			try:
+				response: dict[str, Any] = await structured_llm.ainvoke(input_messages)  # type: ignore
+				parsed: AgentOutput | None = response['parsed']
+
+			except Exception as e:
+				logger.error(f'Failed to invoke model: {str(e)}')
+				raise LLMException(401, 'Failed to invoke model')
```
pirate, line 554:
```suggestion
				raise LLMException(401, 'Failed to invoke model') from e
```
always preserve the original error in the traceback with `from e`

browser_use/agent/service.py
```diff
@@ -531,12 +545,24 @@ async def get_next_action(self, input_messages: list[BaseMessage]) -> AgentOutpu
...
+			logger.debug(f"Using {self.tool_calling_method} for {self.chat_model_library}")
 			structured_llm = self.llm.with_structured_output(self.AgentOutput, include_raw=True, method=self.tool_calling_method)
-			response: dict[str, Any] = await structured_llm.ainvoke(input_messages)  # type: ignore
-			parsed: AgentOutput | None = response['parsed']
+			try:
+				response: dict[str, Any] = await structured_llm.ainvoke(input_messages)  # type: ignore
+				parsed: AgentOutput | None = response['parsed']
+			except Exception as e:
+				logger.error(f'Failed to invoke model: {str(e)}')
+				raise LLMException(401, 'Failed to invoke model')
```
pirate, line 564:
```suggestion
				raise LLMException(401, 'LLM API call failed') from e
```

ml5ah:
@pirate , given that many folks are experiencing LLM call failures and aren't able to diagnose as such, let me know if this PR is in a good state to start reviewing, or if there are any other major areas to address.
We can continue optimizing and improving logging.

browser_use/constants.py
```diff
@@ -0,0 +1,7 @@
+REQUIRED_ENV_VARIABLES = {
```
pirate, line 1:
```suggestion
REQUIRED_LLM_API_ENV_VARS = {
```
Can you rename+move this dict to `browser_use/agent/views.py` 🙏 
We will likely add a pydantic settings model for managing these env vars soon so it makes sense to put it near all the other pydantic LLM API config.

pirate (approved):

pirate:
ready to merge after this: https://github.com/browser-use/browser-use/pull/1146#discussion_r2015951549

thanks 👍

browser_use/constants.py
ml5ah: Sure, makes sense. Done.

pirate:
I have also just pushed some more logging and exception handling improvements to `main` as part of my browser launch process improvements. Next priority in this area is fixing the `Ctrl+C` experience: https://github.com/browser-use/browser-use/pull/816 if you want to help @ml5ah that would be awesome :)

PR #1147, opened 2025-03-26
fix error: name 'browser_class' is not defined

vikingden8:
When I first tried to run examples/simple.py，there is an error：Failed to initialize Playwright browser: name 'browser_class' is not defined

This PR fixes the error.

### DIFF
--- a/browser_use/browser/browser.py
+++ b/browser_use/browser/browser.py
@@ -226,6 +226,7 @@ class Browser:
 				'--no-startup-window',
 			],
 		}
+		browser_class = getattr(playwright, self.config.browser_class)
 		browser = await browser_class.launch(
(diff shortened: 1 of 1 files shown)


ml5ah:
Perhaps, there needs to be a guideline to update the version when making bug fixes or adding features. 
Otherwise, the same version (in this case 0.1.40) will have different behaviors based on when the library is installed.

pirate:
ah yes sorry I cherry picked some and left off the commit that fixed that, thanks for catching.

I just went through a huge backlog of stale PRs and I'm merging and fixing actively. If you're looking for a specific version like `0.1.40` you should be using a [release tag](https://github.com/browser-use/browser-use/releases) though, not a moving target like `main`.

Issue #918, opened 2025-03-02
o1-mini not supported


john-the-dev:
### Bug Description

Got the error when using o1-mini via Azure:

`{'error': {'message': "Unsupported value: 'messages[0].role' does not support 'system' with this model.", 'type': 'invalid_request_error', 'param': 'messages[0].role', 'code': 'unsupported_value'}}`

Did some research, it seems it is known issue with o1 models and the fix is replacing system message with user message:

https://community.openai.com/t/o1-models-do-not-support-system-role-in-chat-completion/953880

Anyone know how to get this fixed with browser use?

### Reproduction Steps

1. Run browser use with o1-mini with a task
2. Then you should see error like this:
`{'error': {'message': "Unsupported value: 'messages[0].role' does not support 'system' with this model.", 'type': 'invalid_request_error', 'param': 'messages[0].role', 'code': 'unsupported_value'}}`

### Code Sample

```python
"""
Simple try of the agent.

@dev You need to add AZURE_OPENAI_API_KEY and AZURE_OPENAI_ENDPOINT to your environment variables.
"""

import os
import sys

sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))

import asyncio

from langchain_openai import AzureChatOpenAI

from browser_use import Agent

# Retrieve Azure-specific environment variables
azure_openai_api_key = os.environ.get('AZURE_OPENAI_API_KEY')
azure_openai_endpoint = os.environ.get('AZURE_OPENAI_ENDPOINT')

# Initialize the Azure OpenAI client
llm = AzureChatOpenAI(
    model_name='o1-mini', 
    openai_api_key=azure_openai_api_key,
    azure_endpoint=azure_openai_endpoint,  # Corrected to use azure_endpoint instead of openai_api_base
    deployment_name='o1-mini',  # Use deployment_name for Azure models
    api_version='2024-10-21'  # Explicitly set the API version here
)

agent = Agent(
    task='Go to amazon.com, search for laptop, sort by best rating, and give me the price of the first result',
    llm=llm,
)


async def main():
    await agent.run(max_steps=10)
    input('Press Enter to continue...')


asyncio.run(main())
```

### Version

0.1.37

### LLM Model

Other (specify in description)

### Operating System

macOS 14.5

### Relevant Log Output

```shell

```

MrMarsMic:
Hi @john-the-dev,
I tried today and it worked flawlessly.

[...]

PR #1166, opened 2025-03-27
Support for reasoning models o1, o1-mini, etc for planner LLM

ml5ah:
The reason o1-preview and o1-mini models do not work (silent failures) is that they do not accept System Messages.
Currently, the planner prompt by default wraps the text into a SystemMessage. 

This PR introduces a new flag `is_planner_reasoning` as part of the agent configuration. If set to true, the planner text is wrapped around a HumanMessage, so that agent execution can continue. 

Related issues: #918


browser_use/agent/views.py
```diff
@@ -58,6 +58,7 @@ class AgentSettings(BaseModel):
 	page_extraction_llm: Optional[BaseChatModel] = None
 	planner_llm: Optional[BaseChatModel] = None
 	planner_interval: int = 1  # Run planner every N steps
+	is_planner_reasoning: bool = False, # type: ignore
```
pirate, line 61:
```suggestion
	is_planner_reasoning: bool = False # type: ignore
```
unless you meant it to be a tuple

ml5ah: Thanks! @pirate

ml5ah:
@pirate fixed. The lint check is failing. Is that something you can take care of for now? I see a bunch of errors of my side. Can fix them in main once merged.

pirate:
thanks!

PR #1173, opened 2025-03-27
Improve click by text handling

jersobh:
This PR will improve the text click by adding the element type (optional, eg.: button) and a fallback click try using js evaluate, as we have on the click by index.


jersobh:
@pirate this is a good improvement for yesterday's PR.

pirate (approved):

pirate:
nice, thanks! 🚀

PR #1180, opened 2025-03-28
temp commit on ctrl+c bahavior

ml5ah:
This is a sample method to handle the existing problem of ctrl+c interrupt from a user.
Currently, multiple interrupt commands need to be provided by the user, and graceful exit is not done.

Asking for feedback - We can discuss this approach or other potential approaches here. 

Related issues: #816 
@pirate 
Edit: I just saw your commits as well -- the reason I did not go with sigterm is not well supported on windows
https://anyio.readthedocs.io/en/3.x/signals.html

### DIFF
--- a/examples/browser/real_browser.py
+++ b/examples/browser/real_browser.py
@@ -1,5 +1,6 @@
 import os
 import sys
+import logging
 
 sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 import asyncio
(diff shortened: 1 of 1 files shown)


pirate:
I just added support for Ctrl+C in a few recent PRs https://github.com/browser-use/browser-use/pull/1178 and https://github.com/browser-use/browser-use/pull/1179

If you have access to a windows machine can you help test it?

Issue #917, opened 2025-03-02
Does not recognise clickable button


mikkovaltonen:
### Bug Description

When automation form at https://verkkopalvelu.lahitapiola.fi/e2/autovakuutus/vakuutuslaskuri/ does not recognise button below: 

![Image](https://github.com/user-attachments/assets/ecd4ca1d-81de-4dcb-bb5f-2c868be49829)

### Reproduction Steps

1. Install broser use
2. Run following task: 
3. """Navigate to: https://verkkopalvelu.lahitapiola.fi/e2/autovakuutus/vakuutuslaskuri/

    1. Cookie Settings:
    - Click "Jatka välttämöttömillä" on the cookie consent popup

    2. Basic Information (1 "Täytä perustiedot"):
    - Enter Postal Code: 02620
    - Enter Social Security Number: 240204A590W
    - Find and click the blue "Seuraava" button (Submit button with class="duet-button primary medium") . 

4. You will see that "Seuraava" button is not indicated in scaper abstraction

### Code Sample

```python
from langchain_openai import ChatOpenAI
from browser_use import Agent
import asyncio
from dotenv import load_dotenv

load_dotenv()

async def main():
    # Main LLM for execution
    llm = ChatOpenAI(model='gpt-4o', temperature=0.0)
    
    # Planner LLM for task breakdown
    planner_llm = ChatOpenAI(model='o3-mini')
    
    task = """Navigate to: https://verkkopalvelu.lahitapiola.fi/e2/autovakuutus/vakuutuslaskuri/

    1. Cookie Settings:
    - Click "Jatka välttämöttömillä" on the cookie consent popup

    2. Basic Information (1 "Täytä perustiedot"):
    - Enter Postal Code: 02620
    - Enter Social Security Number: 270304A590W
    - Find and click the blue "Seuraava" button (Submit button with class="duet-button primary medium") . This button is hard to find. It is above "2 Täytä ajoneuvon tiedot". <button class="duet-no-wrap duet-button primary medium sc-duet-button" type="submit" aria-haspopup="false"><span class="duet-button-contents sc-duet-button sc-duet-button-s">Seuraava</span></button>

    3. Vehicle Information (2 "Täytä ajoneuvon tiedot"):
    - Enter Registration Number: KTM-291
    - In dropdown menu, select vehicle type: "Henkilöauto"
    - For financing question, select: "Ei"
    - Click the blue "Seuraava" button to proceed

[...]

PR #857, opened 2025-02-24
Add anti bot detection via patchright

neo773:
This PR adds anti bot detection via ~~`rebrowser`~~ `patchright`

As per their documentation they don't recommend applying patches manually as it's error prone and instead recommend using their drop-in package

This achieves a trust score of 69, previous score was 0

https://github.com/user-attachments/assets/3a781733-1702-4e99-a4d9-a087ed3d3d15

/claim #852
/closes #852

PR #1184, opened 2025-03-28
removed forced shadow root open

PaperBoardOfficial:
Issue : https://github.com/browser-use/browser-use/issues/917

Background: Shadow Dom elements are used to encapsulate frontend logic. There are two types: open and closed. The ones which are open are easily accessible using element.shadowRoot. But, the ones which are a closed will give null. [ref](https://stackoverflow.com/questions/39931284/what-is-the-difference-between-open-and-closed-shadow-dom-encapsulation-mode)

When you want to create shadow dom elements you use the global attachShadow function. for eg,
```
class MyComponent extends HTMLElement {
    constructor() {
        super();
        this.attachShadow({ mode: "open" }); // Calls attachShadow with open mode
        this.shadowRoot.innerHTML = `<button>Click Me</button>`;
    }
}
customElements.define("my-component", MyComponent);
```
The issue is occurring because there is an init_script (which I am removing in this PR) which is being triggered before any webpage loads.  This script contains a code to intercept the call to global attachShadow function and passes a patch to use only open mode.

This is causing issues. This script code is basically taken from this [issue](https://github.com/microsoft/playwright/issues/23047) from playwright. This a flaky. If you check the comments of one of the maintainers of playwright ([comment](https://github.com/microsoft/playwright/issues/23047#issuecomment-1551652078)), he has mentioned that `Simply turning closed shadow roots into open ones will probably affect the behavior of some pages, because DOM APIs will behave differently now`. 
And I verified this, I ran the code without these lines of script and the shadowRoot was intialized successfully (i could see the `# shadow-root (open)` element and upon clicking it revealed the interactive button element it contained. But when this buggy script is applied, it doesnt let the shadow root to initialize (it showed `<!---->` instead of `# shadow-root(open)` and this `<!---->` was not clickable, hence the shadow root was not intialized properly). If this shadow element doesnt get initialized, the dom tree will not have the interactive button and the buildDomTree.js will not be able to highlight the interactive button.

Possible fixes:
Level 1: Remove the buggy script. This PR does this. I think this is one of the easiest solutions. Note that I tried other usual ways like using if statement like this but it didn't work:
```
 if (options.mode === "closed") {
            console.log("Forcing shadow root to open for:", this);
            return originalAttachShadow.call(this, { ...options, mode: "open" });
        } else {
            console.log("Shadow root mode is already open for:", this);
            return originalAttachShadow.call(this, options);
        }
```

Level 2: If you see this [comment](https://github.com/microsoft/playwright/issues/23047#issuecomment-2393077737) in the playwright issue, one person has recommended to use chrome's api called [chrome.dom.openOrClosedShadowRoot](https://developer.chrome.com/docs/extensions/reference/api/dom). But this would require loading a dummy extension because chrome api is only avaible in extension context. And to load an extension, one would require persistent context and user data dir. Also, it would require different handling for firefox and webkit. I am not even sure if this approach will work, haven't tested it.

Level 3: Move to some other library than playwright which is more apt for anti detection. Here is one more [comment](https://github.com/microsoft/playwright/issues/33725#issuecomment-2497195075) by the maintainer of playwright where he states: `Helping to circumvent anitbot mechanisms is out of scope for Playwright, which is focused on testing.`
So we can use [Patchright](https://github.com/Kaliiiiiiiiii-Vinyzu/patchright/issues/28#issuecomment-2737129193). This is an improved patched version of playwright which is more undetectable and recently they have handled closed shadow roots as well. But it is somewhat buggy too (check their github there are a lot of thing to be done). There could be other libraries which we can explore.

### DIFF
--- a/browser_use/browser/context.py
+++ b/browser_use/browser/context.py
@@ -399,12 +399,6 @@ class BrowserContext:
                     Promise.resolve({ state: Notification.permission }) :
                     originalQuery(parameters)
             );
-            (function () {
-                const originalAttachShadow = Element.prototype.attachShadow;
-                Element.prototype.attachShadow = function attachShadow(options) {
-                    return originalAttachShadow.call(this, { ...options, mode: "open" });
-                };
-            })();
             """
 		)
 


pirate:
The issue of closed shadow roots is well known and plagues all the automation and archiving projects. `browser-use` is currently ok breaking page JS in a few edge cases if it means  faster and simpler buildDOMTree for the other 99% of pages. That may change in the future.

---

> Level 2: If you see this https://github.com/microsoft/playwright/issues/23047#issuecomment-2393077737 in the playwright issue, one person has recommended to use chrome's api called [chrome.dom.openOrClosedShadowRoot](https://developer.chrome.com/docs/extensions/reference/api/dom). But this would require loading a dummy extension because chrome api is only avaible in extension context. And to load an extension, one would require persistent context and user data dir. Also, it would require different handling for firefox and webkit. I am not even sure if this approach will work, haven't tested it.

Yup, this is what we're already working on, I'm grafting an idea from ArchiveBox. Bridge extension + other extensions preloaded for access to chrome runtime APIs + ability to drive extensions actions & context menus directly. Might also do copy-on-write/overlayfs or memory-snappsotted firecracker VMs for the `user_data_dir` so multiple chrome procs can effectively share one and boot instantly without shipping gigabyte zip files around.

---

patchwright is appealing but unfortunately it has this issue:

> XPath locators are not supported in Closed Shadow Roots (CSR). This isnt a limitation of Patchright, but of Chromium itself. Furthermore, some functionalities like Element Counts might be off because Evaluation of CSR-Locators happens in an Isolated World.

`browser-use` depends somewhat heavily on xpath selectors in `buildDOMTree.js`, we'd have to reimplement a simple version of it somehow.

https://github.com/Kaliiiiiiiiii-Vinyzu/patchright/issues/28#issuecomment-2737129193

pirate:
This is now a part of #857, so I'm going to close this PR as patchright added support for xpath selectors in shadow roots!

onlybiju:
Browser-use is still not able to detect elements under Shadow-DOM, Please help or guide me how it can be fixed. Thank you!

PR #1191, opened 2025-03-28
clean: Improve result paring in run_browser_task

welli7ngton:
- Adds detailed analysis of duration and tokens used, and enhances result return with an error message in case of a return with a different type of AgentHistoryList


pirate:
this is great, but perhaps you could move it to a dedicated demo under `examples/functionality/step_history_summary.py` so it doesn't get lose?

Issue #1185, opened 2025-03-28
Agent run does not work on a Windows machine


ml5ah:
### Bug Description

There seems to be a bug introduced in the latest main for windows OS.
When executing agent.run(), there is a NotImplementedError raised relating to the Agent._verify_llm_connection() coroutine. 

### Reproduction Steps

1. Install browser-use (main)
2. Add OPENAI_API_KEY to .env
3. Update browser binary path in `examples\browser\real_browser.py`
4. Run python real_browser.py

### Code Sample

```python
Same as `real_browser.py`
```

### Version

git main branch

### LLM Model

GPT-4o

### Operating System

Windows

### Relevant Log Output

```shell
Traceback (most recent call last):
File "C:\Users\Desktop\browser-use\examples\browser\real_browser.py", line 37, in <module>
asyncio.run(main())
File "\.pyenv-win\versions\3.11.1\Lib\asyncio\runners.py", line 190, in run
return runner.run(main)
File "\.pyenv-win\versions\3.11.1\Lib\asyncio\runners.py", line 118, in run
return self._loop.run_until_complete(task)
...

File "C:\Users\Desktop\browser-use\examples\browser\real_browser.py", line 30, in main
await agent.run()
File "C:\Users\Desktop\browser-use\browser_use\utils.py", line 264, in wrapper
result = await func(*args, **kwargs)
File "C:\Users\Desktop\browser-use\browser_use\agent\service.py", line 694, in run
signal_handler.register()
File "C:\Users\Desktop\browser-use\browser_use\utils.py", line 74, in register
self.original_sigint_handler = self.loop.add_signal_handler(signal.SIGINT, lambda: self.sigint_handler())
File "\.pyenv-win\versions\3.11.1\Lib\asyncio\events.py", line 574, in add_signal_handler
raise NotImplementedError

NotImplementedError
sys:1: RuntimeWarning: coroutine 'Agent._verify_llm_connection' was never awaited
```

pirate:
ah that's annoying, apparently windows doesn't implement SIGINT signals... I'll fix it

welli7ngton:
@pirate I was working on this right now. Do you have any updates, or would you prefer I dive deeper into this issue?

ml5ah:
Since, ctrl+c is not a core functionality here, (agree that pressing ctrl+c multiple times is not ideal), but this is definitely not worth making browser-use incompatible on windows. 

I gave it a quick shot here: #1180.

[...]

PR #1192, opened 2025-03-28
Add windows support to ctrl+C handling

pirate:
Fixes https://github.com/browser-use/browser-use/issues/1185

### DIFF
--- a/browser_use/agent/service.py
+++ b/browser_use/agent/service.py
@@ -6,6 +6,7 @@ import inspect
 import json
 import logging
 import os
+import platform
 import re
 import time
 from pathlib import Path
--- a/browser_use/utils.py
+++ b/browser_use/utils.py
@@ -1,6 +1,7 @@
 import asyncio
 import logging
 import os
+import platform
 import signal
 import time
 from functools import wraps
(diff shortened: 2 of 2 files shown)


ml5ah (commented):
I see that you're accessing self.is_windows in the SignalHandler class, but I don't see this property defined in __init__. Did it get missed?

pirate:
whoops I had it stashed forgot to push the commit, thanks @ml5ah

ml5ah:
No worries, thanks for taking care of this @pirate! 
This moves ahead and I don't see the previous error! 

However, I'm not sure what changed, when running `real_browser.py`, it throws an error saying Playwright browser is not installed, and asks me to run playwright install. 

For the real_browser use-case, this should not be a requirement, correct? This was working fine until very recently.

pirate:
Can you share the debug log from startup with `real_browser.py`, I can fix the missing playwright error. It should not need the playwright browser installed when running with a user-provided `browser_binary_path`.

ml5ah:
[debug_log.log](https://github.com/user-attachments/files/19515877/debug_log.log)

Issue #871, opened 2025-02-26
Some form elements not being detected for highlighting


BaiqingL:
### Bug Description

Browser use is unable to detect the checkbox element
Here is the relevant div
```
<div class="checkbox">
                          <input type="checkbox" name="agree" value="1" id="checkbox1"><label for="checkbox1">
                            I agree with <a href="rules">Terms and conditions</a></label>
                        </div>
```

Here is a screenshot of the relevant areas:

<img width="1840" alt="Image" src="https://github.com/user-attachments/assets/8d86755f-4448-4000-8086-274becd849a9" />

### Reproduction Steps

Use the latest browser use version, ask the agent to create an account on https://availfinance.online/home/auth/signup.php

The checkbox is not detectable

### Code Sample

```python
agent.run("go to https://availfinance.online/home/auth/signup.php, and try to agree to the terms and conditions")
```

### Version

git main branch

### LLM Model

Claude 3.5 Sonnet

### Operating System

macOS

### Relevant Log Output

```shell

```

ai-naymul:
@BaiqingL Probably by just telling the model in the task that by click the checkbox agree to the terms and conditions... A specific instruction to model can save you a lot of time....

BaiqingL:
> @BaiqingL Probably by just telling the model in the task that by click the checkbox agree to the terms and conditions... A specific instruction to model can save you a lot of time....

Please read the bug report

ai-naymul:
> > [@BaiqingL](https://github.com/BaiqingL) Probably by just telling the model in the task that by click the checkbox agree to the terms and conditions... A specific instruction to model can save you a lot of time....
> 
> Please read the bug report

Oops Sorry I didn't noticed... Let me see

ai-naymul:
I can see that browser-use is not able to recognize that as a clickable element even if you move the cursor to the checkbox you can see cursor is not recognizing it as a clickable element

<img width="1282" alt="Image" src="https://github.com/user-attachments/assets/4179176e-7a38-4d34-bc6c-8a4f0ba06ec2" />

MrMarsMic:
+1
I'm experiencing the same issue in some sites

raivisdejus:
I used custom functions to accept terms. 

```python
from langchain_openai import ChatOpenAI
from browser_use import (
    Agent,
    Browser,
    BrowserConfig,
    BrowserContextConfig,
    Controller,
    ActionResult
)
from browser_use.browser.context import BrowserContext
import asyncio
[...]

PR #869, opened 2025-02-25
Swapnil/rerun

SwapnilSonker:
/claim #763 
i have created another domCache file from where i am caching the paths (XPaths)

PR #1195, opened 2025-03-28
[ Interactive Elements ] Improve detection of interactive labels ( checkbox/radio ) 

Rahul-Sharma-1729:
## Background

- In web applications, `<label>` elements are often used to enhance accessibility and usability by moving focus to or toggling elements like checkboxes and radio buttons.  
- A `<label>` with a non-empty `for` attribute is considered interactive since it associates the label with another form control.  
- Some users have reported issues with detecting checkboxes due to improper label handling. This PR improves the detection mechanism and addresses #871.

## Implementation

This PR introduces the `isInteractiveLabel` function to improve label detection by:  
- Ensuring the element is a `<label>` tag.  
- Verifying that the `for` attribute is non-empty.  
- Checking if the `for` attribute references a valid interactive element (`<input>`, `<textarea>`, `<select>`, or `<button>`).  

By implementing this function, we enhance the ability to detect interactive labels and improve overall user interaction.

@pirate @MagMueller

### DIFF
--- a/browser_use/dom/buildDomTree.js
+++ b/browser_use/dom/buildDomTree.js
@@ -479,6 +479,27 @@
       return true;
     }
 
+    // special handling for element which is an interactive label (generally toggles radio/checkbox )
+    function isInteractiveLabel(element) {
+      if (!(element instanceof HTMLLabelElement)) return false;
+
+      const forAttr = element.getAttribute("for");
+      if (!forAttr || forAttr.trim() === "") return false;
+
+      const targetElement = document.getElementById(forAttr);
+      if (!targetElement) return false;
+
+      return (
+        targetElement instanceof HTMLInputElement ||
+        targetElement instanceof HTMLTextAreaElement ||
+        targetElement instanceof HTMLSelectElement ||
+        targetElement instanceof HTMLButtonElement
+      );
+    }
+
+    let isInteractiveLabelElement = isInteractiveLabel(element);
+    if (isInteractiveLabelElement) return true;
+
     // Special handling for cookie banner elements
     const isCookieBannerElement =
       (typeof element.closest === 'function') && (


pirate:
Awesome, LGTM. Thanks!

Rahul-Sharma-1729:
Hey @pirate , I had missed a case for handling implicit labels, I have pushed some changes, please check, thanks! 

Example :
`<label class="d-flex">
      <input type="checkbox" class="mr-2 js-collab-checkbox" name="collab_privs" checked="">
      Allow edits by maintainers
</label>`

browser_use/dom/buildDomTree.js
```diff
@@ -479,6 +479,30 @@
...
+      const targetElement = document.getElementById(forAttr);
+      if (!targetElement) return false;
+
+      return (
+        targetElement instanceof HTMLInputElement ||
+        targetElement instanceof HTMLTextAreaElement ||
+        targetElement instanceof HTMLSelectElement ||
+        targetElement instanceof HTMLButtonElement
+      );
+      }
```
pirate, line 497:
```suggestion
      const forAttr = element.getAttribute("for");
      if (forAttr && forAttr.trim() !== "") {
	      const targetElement = document.getElementById(forAttr);
	      if (!targetElement) return false;
	
	      return (
	        targetElement instanceof HTMLInputElement ||
	        targetElement instanceof HTMLTextAreaElement ||
	        targetElement instanceof HTMLSelectElement ||
	        targetElement instanceof HTMLButtonElement
	      );
      }
```

pirate:
My only concern with this is that it will mark labels as interactive even if the underlying input is already also marked interactive. I don't want to add duplicate highlights if they are not needed, can you exclude the labels who's `for=...` elem is already going to be marked highlighted?

Rahul-Sharma-1729:
> My only concern with this is that it will mark labels as interactive even if the underlying input is already also marked interactive. I don't want to add duplicate highlights if they are not needed, can you exclude the labels who's `for=...` elem is already going to be marked highlighted?

Well that's a valid concern, I'll add that check to avoid the redundancies.

Rahul-Sharma-1729:
@pirate since calling the isInteractiveElement function on the for=... element would be expensive, I believe a caching based solution would be appropriate for this, keeping a cache of interactive elements which are visible and are the topElement, what do you think?

pirate:
I would rather not add a cache yet as we may change the behavior of `buildDOMTree.js` soon as part of our cross-origin-iframe and closed shadow DOM support, and it will be extra work to refactor the cache as well.

Some people have already tried similar approaches and it is quite complex: https://github.com/browser-use/browser-use/pull/869

What if instead of caching you run a separate pass at the end just for the labels, and you look at the already collected list of interactive elements to see if the `for=...` target has already been found interactive.

deveshruttala-edu:

mertunsall:
What's the status of this PR? Seems very useful!!

browser_use/dom/buildDomTree.js
```diff
@@ -479,6 +479,30 @@
...
+      if (forAttr && forAttr.trim() !== "") {
+	      const targetElement = document.getElementById(forAttr);
+	      if (!targetElement) return false;
+	
+	      return (
+	        targetElement instanceof HTMLInputElement ||
+	        targetElement instanceof HTMLTextAreaElement ||
+	        targetElement instanceof HTMLSelectElement ||
+	        targetElement instanceof HTMLButtonElement
+	      );
```
pirate, line 732: what if we change this to:

```suggestion
          const target_is_interactive = isInteractiveElement(targetElement)
	      return !target_is_interactive
```

that way we dont highlight the labels if they point to an input that's already highlighted. if you test that and confirm it works on https://browser-use.github.io/stress-tests/forms-comparison.html then I'm down to merge this

PR #1203, opened 2025-03-29
Combine the index and text click in a single function with text as fallback

jersobh:
This PR will combine both functions, leavnig the text as a fallback option. I've also fixed a small issue regarding the env. variable validation (when or if the env. variable is None)

### DIFF
--- a/browser_use/controller/service.py
+++ b/browser_use/controller/service.py
@@ -129,42 +129,88 @@ class Controller(Generic[Context]):
 				raise Exception(err_msg)
 
 		# Element Interaction Actions
-		@self.registry.action('Click element by index', param_model=ClickElementAction)
-		async def click_element_by_index(params: ClickElementAction, browser: BrowserContext):
+		@self.registry.action('Click on element', param_model=ClickElementAction)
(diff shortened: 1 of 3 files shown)


pirate:
I think it's better to have them as separate actions, I want the LLM to almost always click the normal way and only manually choose the CSS selector, text, or Xpath methods as a fallback if it fails the normal way.

jersobh:
But that's the catch.  It still has the css selectors and xpath in a separate way, and it works really well. But when you do something like "Click the accecpt button" it might be confusing if the user expects to click on the button with text "accept", or a element with an index that has the accept prop. So, by doing like in my PR, it will try the index first, then use the text mode as a fallback.

pirate:
@jersobh are you on [Discord](https://discord.gg/pYHCRw2m)? we are discusing click method refactoring and trying to figure out the best way to integrate these 3 related PRs:

- https://github.com/browser-use/browser-use/pull/726
- https://github.com/browser-use/browser-use/pull/1203
- https://github.com/browser-use/browser-use/pull/845

PR #864, opened 2025-02-25
Planning and Cache

PaperBoardOfficial:
Issue link:
https://github.com/browser-use/browser-use/issues/778
https://github.com/browser-use/browser-use/issues/763

/claim #778 
/claim #763

PR #1205, opened 2025-03-29
Add support for procedural memory for running agents for longer tasks

deshraj:
# Add Procedural Memory Support Using Mem0

This PR adds support for procedural memory in browser-use using [Mem0](https://mem0.ai), which helps agents run longer tasks more reliably while using fewer tokens.

## What's Changed

- Added a Memory service that generates procedural memories at configurable intervals
- Integrated memory generation into the agent's step cycle
- Added memory configuration options to the Agent constructor
- Implemented conditions to generate memory only when enabled and at specified intervals

## Configuration Options

Memory can be configured through these parameters when initializing an Agent:

```python
# Memory settings
enable_memory: bool = True             # Enable/disable memory feature
memory_interval: int = 10              # Generate memory every N steps
memory_config: Optional[dict] = None   # Additional Mem0-specific configuration
```

## Benefits

- **Token Efficiency**: By summarizing past steps into procedural memory, we reduce the token count in the agent's context
- **Improved Reliability**: Agents can better remember previous actions and their outcomes over longer tasks
- **Configurable Usage**: Memory generation is fully configurable and can be disabled if not needed

## Results

Ran for a long task for visiting 50 links and summarizing the content for all the links. Here are the results:

image


PaperBoardOfficial:
@deshraj Its great to see you here! I am big fan of your and Taranjeet's Mem0.
I had also tried to add some memory logic before in a PR using qdrant (both docker and cloud version). But, I guess the repo owners of browser use want to have this library lean. Could we use faiss file based storage instead of chromadb.
Also, I would like your opinion on this (because you guys are expert in memory): currently this is the flow of history messages sent to llm:
1. system prompt
2. example task
3. actual/current task
4. all the previous history (which basically consists of list of previous actions and their results. for eg, action: click_element: index 14, result: clicked on element with index 14 or error: element with index 14 not interactive.
5. current webpage state. it consists of pruned html structure (which contains the interactive elements with index numbers) and a base64 string of image
currently when I run a task like "Go to amazon.com and search for laptop and sort by customer review and return price of first result.", it works perfectly. But I want that if I run the same task again, I should leverage the memory and save my tokens. Or if I run the same task but instead I replace laptop with iphone, again I should save tokens.
What's your take on this?

pirate:
@PaperBoardOfficial I am just looking at https://github.com/browser-use/browser-use/pull/864/files and this PR now, give me some time to dig into both of these and discuss the architecture more with the team. I know @MagMueller and @gregpr07 have both tried various cache and planner optimization approaches but the gains are not always straightforward to measure and it takes time to run good evals.

MagMueller:
@PaperBoardOfficial, this type of memory is only for one task, allowing us to perform 100+ steps. The memory you are referring to is across tasks.

pirate:
@deshraj ok after a bunch of internal discussion we are ready to try this, thanks for contributing!

deshraj:
Awesome, thank you folks. Happy to brainstorm together how to build the memory across tasks as follow up steps.

pirate:
One small nit that maybe you can help with, this merge added a few extra error log lines that would be nice to silence if they aren't critical:

```python
DEBUG    [faiss.loader] Environment variable FAISS_OPT_LEVEL is not set, so let's pick the instruction set according to the current CPU
INFO     [faiss.loader] Loading faiss.
<frozen importlib._bootstrap>:241: DeprecationWarning: builtin type SwigPyPacked has no __module__ attribute
<frozen importlib._bootstrap>:241: DeprecationWarning: builtin type SwigPyObject has no __module__ attribute
<frozen importlib._bootstrap>:241: DeprecationWarning: builtin type swigvarlink has no __module__ attribute
INFO     [faiss.loader] Successfully loaded faiss.
INFO     [faiss] Failed to load GPU Faiss: name 'GpuIndexIVFFlat' is not defined. Will not load constructor refs for GPU indexes.
INFO     [mem0.vector_stores.faiss] Loaded FAISS index from /tmp/faiss/mem0.faiss with 0 vectors
```

PR #1211, opened 2025-03-29
pytest: uv run playwright install --with-deps

cclauss:


.github/workflows/ci.yml
```diff
@@ -1,16 +1,21 @@
...
       - uses: actions/checkout@v4
       - uses: astral-sh/setup-uv@v5
-      - run: uv run ruff format
       - run: uv run pre-commit run --all-files
+  pytest:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: astral-sh/setup-uv@v5
+      - run: uv run playwright install --with-deps
```
pirate, line 18:
```suggestion
      - run: uv run playwright install --with-deps chromium
```

cclauss:
I have never had much luck getting playwright to do its thing.  This is about as far as I can take this one.

pirate:
No worries @cclauss playwright in CI/CD is fairly complex, we'll dig into this more later.

For now we are using a separate YC startup that's helping us run massive integration test evals every week, so pytest in CI/CD is not a high priority.

PR #1216, opened 2025-03-29
Fix typos discovered by codespell

cclauss:
https://pypi.org/project/codespell

% `codespell --ignore-words-list=bu --skip="*.json" --write-changes`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,3 +23,10 @@ repos:
       - id: detect-private-key
       - id: mixed-line-ending
       - id: fix-byte-order-marker
+
+  - repo: https://github.com/codespell-project/codespell
+    rev: v2.4.1
+    hooks:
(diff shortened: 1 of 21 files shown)


Mustafiz04 (approved):

pirate:
Can you also add codespell as a dev dependency: `uv add --dev codespell` and commit the change.

cclauss:
I have done `uv add --dev codespell`.

I usually do `uv tool install codespell" so the tool is installed globally on my machine and can be used on any project.

I do not `uv add --dev codespell` because it is not really a dev dependency but is instead a pre-commit dependency.  When using codespell on a particular project, I use `pre-commit run -a` or `pre-commit run codespell -a`.  Storing the dependency both in pyproject.toml for uv and in .pre-commit-config.yaml can lead to out of sync dependencies.

Issue #1217, opened 2025-03-29
Log which model is being used when starting the agent


gregpr07:
### Problem Description

Sometimes people don't understand which model is _actually_ being used for main, planner, and extraction LLM.

### Proposed Solution

Just log it

pirate:
This is already done, it shows which keys are used on startup now in the debug log.

PR #1219, opened 2025-03-29
Log active LLM models on agent startup

federicovilla55:
## Changes

Enhance logging by explicitly reporting which LLM models are being used when the agent starts.
These loggings were included in the `service.py` file, into the constructor of the `Agent` class (since logging should be done when models are in use at startup).

## Example log

```
INFO [agent] Starting an agent with main_model=gpt-4o, planner_model=gpt-3.5-turbo, extraction_model=None
```

Fixes: #1217

### DIFF
--- a/browser_use/agent/service.py
+++ b/browser_use/agent/service.py
@@ -189,6 +189,11 @@ class Agent(Generic[Context]):
 
 		# Model setup
 		self._set_model_names()
+		logger.info(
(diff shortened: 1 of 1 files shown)


pirate:
Hey @federicovilla55 thanks for contributing this so fast! If you can sign that CLA I can merge it :)

federicovilla55:
@pirate signed it!

Issue #1207, opened 2025-03-29
Repeatedly stuck trying to switch tabs


german94:
### Bug Description

Hello!

First of all thanks for implementing this great tool, I really like it. I'm just having some issue with my use case. Basically, I'm trying to enter an OTP in a login flow, so I need to inject this value dynamically. My workaround is to make the browser navigate to a page that will provide this OTP. The problem comes up when it tries to go to the original tab and enter this value. Basically the task looks like this:

```
    1. Go to 'https://some_page'. and login
    2. Click on the send OTP button
    4. Wait 10 seconds
    5. Go to 'https://some_other_page' and get the number from the body, remember it
    6. Go to the original page
    7. Enter the number as the OTP
```

Everything goes well until step 5 (included). But then it starts switching between the two tabs. I have tried several things, like for example asking to close all tabs but the original one or things like that, but I didn't have luck. Do you know what can be happening or have any suggestion regarding the prompt? From the logs I see things like:

`Eval: Failed - Currently on the wrong tab, need to switch to login tab to enter the OTP.` and 
`Eval: Unknown - Repeatedly stuck trying to switch tabs.`

### Reproduction Steps

N/A

### Code Sample

```python
N/A
```

### Version

0.1.40

### LLM Model

GPT-4

### Operating System

MacOS

### Relevant Log Output

```shell
Eval: Failed - Currently on the wrong tab, need to switch to login tab to enter the OTP.
Eval: Unknown - Repeatedly stuck trying to switch tabs.
```

devarsh-mavani-19:
Getting similar issue

pirate:
Interesting, if you are able to set up langsmith or LiteLLM or a similar LLM tracing tool and show us the full prompt the LLM is getting for the eval that would be super helpful:

In `.env`:
```bash
LANGSMITH_TRACING=true
LANGSMITH_ENDPOINT="https://api.smith.langchain.com"
LANGSMITH_API_KEY=lsv2_pt_your_api_key_here
LANGSMITH_PROJECT="browser-use"
```

We will also test ourselves and try to reproduce this.

PaperBoardOfficial:
It has nothing to do with prompt. Its a code bug, I am working on fixing it
.

PaperBoardOfficial:
@devarsh-mavani-19 The code you mentioned is basically handling those cases when we click a link and new tab is opened (for eg, click any link in https://www.bing.com/search?q=hello).
[...]

PR #1221, opened 2025-03-30
added active_tab state management

PaperBoardOfficial:
Issue link: https://github.com/browser-use/browser-use/issues/1207
### Root cause:
In this code, the `await browser.switch_to_tab(params.page_id)` perfectly switches to the desired tab but the next line `page = await browser.get_current_page()` is switching to the last tab because it internally calls `return non_extension_pages[-1]`:
```
@self.registry.action('Switch tab', param_model=SwitchTabAction)
		async def switch_tab(params: SwitchTabAction, browser: BrowserContext):
			await browser.switch_to_tab(params.page_id)
			# Wait for tab to be ready
			await page.wait_for_load_state()
			msg = f'🔄  Switched to tab {params.page_id}'
			logger.info(msg)
			return ActionResult(extracted_content=msg, include_in_memory=True)
```

### Fix:
I have fixed it by adding active_tab class member field. I have added it in all those methods that are responsible for opening/closing a tab.

[logs](https://drive.google.com/file/d/1ZRsLOqK2Z50LwKmRBiHy2Zw75NH6OGg3/view?usp=sharing)
[video](https://youtu.be/YWejc0Tccio)

### DIFF
--- a/browser_use/browser/context.py
+++ b/browser_use/browser/context.py
@@ -203,6 +203,7 @@ class BrowserSession:
 				}
 			})()
 			"""
+		self.active_tab = None
 		self.context = context
 		self.cached_state = cached_state
 		self.context.on('page', lambda page: page.add_init_script(init_script))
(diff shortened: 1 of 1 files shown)


pirate:
There actually was some code already doing this approach a few months ago and it got removed because of some bugs. I think it's ok to try again as this is obviously a better approach than scanning all tabs every time.

PR #1222, opened 2025-03-30
Fix: Disable signal handlers to prevent main thread error

kitchan98:


pirate:
what error were you seeing? can you share the error even if you dont want to fix it yourself, we want to make sure other people dont hit the error

Issue #1223, opened 2025-03-30
Browser_binary_path variable not getting initliazed


PaperBoardOfficial:
### Bug Description

I took a recent pull of browser use and I am unable to set the browser_binary_path.
Try running this:
```
browser_config = BrowserConfig(
		browser_binary_path='/usr/bin/google-chrome',
	)
	print(browser_config.browser_binary_path)
```
It will print `None`.

### Reproduction Steps

Described above

### Code Sample

```python
Described above
```

### Version

latest

### LLM Model

GPT-4o

### Operating System

Ubuntu

### Relevant Log Output

```shell

```

ml5ah:
Thanks @PaperBoardOfficial . I've been trying to see why it was using playwright chromium even when I specific the browser_binary_path. I guess, this related to (and fixed) #1209 as well. Can you confirm?

PaperBoardOfficial:
It seems like it is related, but it is not fixed yet. that's why I raised this PR.

pirate:
Fixed in https://github.com/browser-use/browser-use/pull/1232

PR #1224, opened 2025-03-30
fixed browser_binary_path

PaperBoardOfficial:
Issue link: https://github.com/browser-use/browser-use/issues/1223


pirate:
Nah the issue is actually before it ever gets to the model, adding getters to the config model wont fix it. See my fix in https://github.com/browser-use/browser-use/pull/1232

Issue #1225, opened 2025-03-30
Local Browser instance deleting my logged in session/states


PaperBoardOfficial:
### Bug Description

I took the latest pull of browser use. Earlier I was able to use my local browser and even use my already logged in state. But now when I try to use local browser, It deletes all my cookies and logged in sessions. I have to login again. It even logs me out of gmail, github.

### Reproduction Steps

Try the code sample below using your local browser. It will not go to the already logged in gmail account. Instead it will take you to the google sign in page. Also, when you are done with this, you will realize that you are even logged out from github now.

### Code Sample
```
import asyncio
import os

from dotenv import load_dotenv
from langchain_google_genai import ChatGoogleGenerativeAI
from pydantic import SecretStr

from browser_use import Agent, BrowserConfig
from browser_use.browser.browser import Browser
from browser_use.browser.context import BrowserContextConfig

load_dotenv()
api_key = os.getenv('GEMINI_API_KEY')
if not api_key:
	raise ValueError('GEMINI_API_KEY is not set')

llm = ChatGoogleGenerativeAI(model='gemini-2.0-flash-exp', api_key=SecretStr(api_key))

browser = Browser(
	config=BrowserConfig(
		browser_binary_path='/usr/bin/google-chrome',
	)
)

task="""Go to 'https://mail.google.com/mail/u/0/#inbox'"""
async def run_search():
	agent = Agent(
		task=task,
		llm=llm,
		max_actions_per_step=1,
		browser=browser,
	)

	await agent.run(max_steps=25)


if __name__ == '__main__':
	asyncio.run(run_search())

```

### Version

latest

### LLM Model

GPT-4o

### Operating System

Ubuntu

### Relevant Log Output

```shell

```

PR #1226, opened 2025-03-30
fixed logged in state

PaperBoardOfficial:
Issue link: https://github.com/browser-use/browser-use/issues/1225

@pirate I saw that you refactored the code to use chrome_args from the chrome.py constants file, but the two flags I have commented out in this PR, are causing issues. I know you are progressing towards auth.json. But, meanwhile that is achieved, lets comment these flags. Because if these are enabled, one will not be able to even run the 'Google Docs' demo in the readme.


pirate:
fair enough, thanks

PR #1227, opened 2025-03-30
refactor: rename _input_text_element_node to input_text_element_node

mannaandpoem:
Changed the variable name from _input_text_element_node to input_text_element_node to make it accessible outside its original scope since it was being used externally.


pirate:
What do you mean by `being used externally`? It's underscore prefixed to discourage use by library users, not to prevent use between different files in the same codebase internally.

PR #1229, opened 2025-03-30
update example tool call

Alezander9:
this changes the example tool use given to the agent in the _init_message function in browser-use/agent/message_manager/service.py to remove typos and encourage the agent to write out some reasoning before making tool calls.


pirate:
Can you provide some examples of pages where this improves the behavior?

We need empirical evidence of anything that changes the LLM prompts, as it's hard to predict what actually helps and what hurts by vibes alone.

Mustafiz04 (approved):

PR #1230, opened 2025-03-30
add anthropic RateLimitError to list of defined rate limit errors

Alezander9:
add anthropic RateLimitError to list of defined rate limit errors


pirate:
awesome thanks!

PR #1238, opened 2025-03-31
fix: use correct config when creating browser context

lynnzc:
### Issue
The browser context creation was failing with an error: `'BrowserConfig' object has no attribute 'user_agent'`.  
This occurred because we were trying to initialize a BrowserContextConfig with attributes from a BrowserConfig object, which doesn't have all the required attributes.  

### Fix
Use the existing new_context_config attribute from BrowserConfig instead of attempting to convert the entire BrowserConfig object to a BrowserContextConfig.


pirate:
We are merging Browser and BrowserContext eventually (and their corresponding configs) so this should be a lot simpler soon.

PR #1256, opened 2025-04-01
Fix custom-functions.mdx confusion

oswy-cpu:
.get_current_page() is an async function which have to be awaited.

### DIFF
--- a/docs/customize/custom-functions.mdx
+++ b/docs/customize/custom-functions.mdx
@@ -55,7 +55,7 @@ from browser_use import Browser, Controller, ActionResult
 controller = Controller()
 @controller.action('Open website')
 async def open_website(url: str, browser: Browser):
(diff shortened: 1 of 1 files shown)


pirate:
ah thank you!

docs/customize/custom-functions.mdx
```diff
@@ -55,7 +55,7 @@ from browser_use import Browser, Controller, ActionResult
 controller = Controller()
 @controller.action('Open website')
 async def open_website(url: str, browser: Browser):
-    page = browser.get_current_page()
+    page = await browser.get_current_page()
```
pirate, line 58: actually I think this should maybe be:
```suggestion
async def open_website(url: str, browser: BrowserContext):
    page = await browser.get_current_page()
```

Issue #1237, opened 2025-03-31
Support for gemma instruction-tuned (IT) models


KamilU01:
### Problem Description

When using browser-use with google/gemma-3-27b-it model, I encounter the following error:

`Error code: 400 - {'object': 'error', 'message': 'Conversation roles must alternate user/assistant/user/assistant/...', 'type': 'BadRequestError', 'param': None, 'code': 400}`

Based on https://ai.google.dev/gemma/docs/core/prompt-structure#system-instructions instruction-tuned models are designed to work with only two roles: user and model amd system role is not supported.

### Proposed Solution

After debugging, I noticed that it is enough to apply similar conditions as for deepseek models and the operations perform correctly

I modified:
`agent/message_manager/utils.py`

```
def convert_input_messages(
    input_messages: list[BaseMessage], model_name: Optional[str]
) -> list[BaseMessage]:
    """Convert input messages to a format that is compatible with the planner model"""
    if model_name is None:
        return input_messages
    if (
        model_name == "deepseek-reasoner"
        or "deepseek-r1" in model_name
        or "gemma" in model_name # I added this condition
        and "-it" in model_name
    ):
        converted_input_messages = _convert_messages_for_non_function_calling_models(
            input_messages
        )
        merged_input_messages = _merge_successive_messages(
            converted_input_messages, HumanMessage
        )
        merged_input_messages = _merge_successive_messages(
            merged_input_messages, AIMessage
        )
        return merged_input_messages
    return input_messages
```

I am also not sure if, in such a case, a change in `agent/service.py` will provide more convincing results:
`agent/service.py`
```
def _set_tool_calling_method(self) -> Optional[ToolCallingMethod]:
        tool_calling_method = self.settings.tool_calling_method
        if tool_calling_method == "auto":
            if (
                "deepseek-reasoner" in self.model_name
                or "deepseek-r1" in self.model_name
                or "gemma" in self.model_name
                and "-it" in self.model_name
            ):
                return "raw"
            elif self.chat_model_library == "ChatGoogleGenerativeAI":
                return None
            elif self.chat_model_library == "ChatOpenAI":
                return "function_calling"
            elif self.chat_model_library == "AzureChatOpenAI":
                return "function_calling"
            else:
                return None
        else:
            return tool_calling_method
```

pirate:
Thanks!

Want to open a PR with your suggested fix? It will make it easier for other people to help test it too.

deveshruttala-edu:
[...]

PR #1257, opened 2025-04-01
Added support for gemma instruction-tuned models

KamilU01:
**Feature Request**: https://github.com/browser-use/browser-use/issues/1237

**Problem Description**
When using browser-use with google/gemma-3-27b-it model, I encounter the following error:
`
Error code: 400 - {'object': 'error', 'message': 'Conversation roles must alternate user/assistant/user/assistant/...', 'type': 'BadRequestError', 'param': None, 'code': 400}`

Based on https://ai.google.dev/gemma/docs/core/prompt-structure#system-instructions instruction-tuned models are designed to work with only two roles: user and model amd system role is not supported.

This PR contains the changes I used to make it work with the Gemma model

### DIFF
--- a/browser_use/agent/message_manager/utils.py
+++ b/browser_use/agent/message_manager/utils.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import json
 import logging
 import os
+import re
 from typing import Any, Optional, Type
 
 from langchain_core.messages import (
@@ -14,6 +15,19 @@ from langchain_core.messages import (
 )
 
 logger = logging.getLogger(__name__)
+
+MODELS_WITHOUT_TOOL_SUPPORT_PATTERNS = [
+	"deepseek-reasoner",
+	"deepseek-r1",
+	".*gemma.*-it",
+]
+
+
+def is_model_without_tool_support(model_name: str) -> bool:
+	return any(
+		re.match(pattern, model_name)
+		for pattern in MODELS_WITHOUT_TOOL_SUPPORT_PATTERNS
+	)
 
 
 def extract_json_from_model_output(content: str) -> dict:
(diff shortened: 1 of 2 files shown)


pirate:
Thanks for contributing! 

Can you actually move this list of models that we have to check for to a `MODELS_WITHOUT_TOOL_SUPPORT = [...]` at the top of the file? I think it's better than just hardocding the values in inline in multiple places now that the list is getting longer.

KamilU01:
Absolutely. I also created a function` is_model_without_tool_support` that will check if the model name meets these conditions

browser_use/agent/message_manager/utils.py
```diff
@@ -15,6 +15,29 @@
...
+List of model names that do not support tool/function calling capabilities.
+Each entry can be either:
+- A string representing the exact model name
+- A list of strings where all parts must be present in the model name
+"""
+MODELS_WITHOUT_TOOL_SUPPORT = [
+	"deepseek-reasoner",
+	["deepseek-r1"],
+	["gemma", "-it"],
+]
```
pirate, line 28: sorry to nit but I think this would be clearer/simpler as a list of regexes.

```suggestion
MODELS_WITHOUT_TOOL_SUPPORT = [
	r"deepseek-reasoner",
	r"deepseek-r1",
	r"gemma.*-it",
]
```

KamilU01: You're right. This is a much better sollution. I've pushed a fix

pirate (approved):

pirate:
ok, can you sign the CLA? then I'm good to merge it

danielblnc:
Having the same issue, any chance you could take a look at the CLA? @KamilU01

KamilU01:
> ok, can you sign the CLA? then I'm good to merge it

Sorry, I accidentally pushed commit with the wrong email. It should be correct now. The code hasn't changed

PR #1259, opened 2025-04-01
Fix: Remove unnecessary await from _verified_api_keys

BurnyCoder:
This PR fixes an issue where the code was unnecessarily awaiting the `_verified_api_keys` property in the Agent's `run` method. 

The property is a boolean attribute and doesn't need to be awaited since it's not a coroutine or awaitable object.

Fix:
- Remove the `await` keyword from `assert await self.llm._verified_api_keys`


pirate:
Thanks!

Issue #1267, opened 2025-04-02
Bug: deepseek example not working, logging cannot access local variable 'response' where it is not associated with a value


zsjie:
### Bug Description

deepseek example is not working, logging cannot access local variable 'response' where it is not associated with a value

<img width="1034" alt="Image" src="https://github.com/user-attachments/assets/acbbb90e-64f6-44e2-bf77-8e154556f22a" />

### Reproduction Steps

1. git clone the repo to local: `git@github.com:browser-use/browser-use.git`
2. set up local environment following [docs/development/local-setup.mdx](https://github.com/browser-use/browser-use/blob/main/docs/development/local-setup.mdx)
3. run the deepseek example

```sh
uv run examples/models/deepseek-r1.py
```

### Code Sample

```python
uv run examples/models/deepseek-r1.py
```

### Version

git main branch

### LLM Model

DeepSeek Coder

### Operating System

macOS 15.3.2

### Relevant Log Output

```shell
INFO     [browser_use] BrowserUse logging setup complete with level info
INFO     [telemetry] Anonymized telemetry enabled. See https://docs.browser-use.com/development/telemetry for more information.
INFO     [agent] 🧠 Starting an agent with main_model=deepseek-reasoner, planner_model=None, extraction_model=deepseek-reasoner
INFO     [faiss.loader] Loading faiss.
<frozen importlib._bootstrap>:241: DeprecationWarning: builtin type SwigPyPacked has no __module__ attribute
<frozen importlib._bootstrap>:241: DeprecationWarning: builtin type SwigPyObject has no __module__ attribute
<frozen importlib._bootstrap>:241: DeprecationWarning: builtin type swigvarlink has no __module__ attribute
INFO     [faiss.loader] Successfully loaded faiss.
INFO     [faiss] Failed to load GPU Faiss: name 'GpuIndexIVFFlat' is not defined. Will not load constructor refs for GPU indexes.
INFO     [mem0.vector_stores.faiss] Loaded FAISS index from /tmp/faiss/mem0.faiss with 1 vectors
INFO     [agent] 🚀 Starting task: go to amazon.com, search for laptop, sort by best rating, and give me the price of the first result
INFO     [agent] 📍 Step 1
ERROR    [agent] ❌ Result failed 1/2 times:
 cannot access local variable 'response' where it is not associated with a value
INFO     [agent] 📍 Step 1
ERROR    [agent] ❌ Result failed 2/2 times:
[...]

PR #1269, opened 2025-04-02
Fix: resolve undefined response variable in Deepseek model raw tool calling

zsjie:
bug description can be found at https://github.com/browser-use/browser-use/issues/1267


CloneOfAlex (approved):

pirate:
thanks!

PR #1279, opened 2025-04-03
Fix Azure OpenAI example by disabling memory feature

mathisarends-viadee:
This PR addresses an issue where the Azure OpenAI example fails to run after recent memory handling changes introduced in PR #1205. 

The simple fix adds `enable_memory=False` to the Agent initialization, which should be sufficient for demo purposes, allowing the example to run without requiring OpenAI API credentials for memory features.

image


pirate:
Thanks, didn't realize the azure API isn't totally cross compatible.

PR #1281, opened 2025-04-04
Add extended system prompt capability for planner

mathisarends-viadee:
This PR introduces the ability to extend the planner's system prompt with additional instructions, similar to how the agent's system prompt can be customized using the existing `override_system_message` and `extend_system_message` parameters.

## Changes
* Added `extend_planner_system_message` parameter to `AgentSettings` class and constructor
* Refactored the `PlannerPrompt` class to accept and incorporate extended prompt instructions
* Updated the `_run_planner` method to pass the extended system message to the planner

## Benefits
* **Consistent with existing customization patterns**: Follows the same pattern as the agent's `extend_system_message` parameter
* **Fine-grained planning control**: Developers can now guide the planner's behavior with additional instructions without replacing the entire system prompt


pirate (approved):

pirate:
Great idea! Could also add a short section to the docs under `customize/system-prompt.mdx` explaining this new feature?

mathisarends-viadee:
Thanks for the feedback :) I just added the documentation for it. Let me know what you think ^^

docs/customize/system-prompt.mdx
```diff
@@ -64,3 +53,25 @@ agent = Agent(
...
+# Create agent with extended planner system prompt
+llm = ChatOpenAI(model='gpt-4o')
+planner_llm = ChatOpenAI(model='gpt-4o-mini')
+
+agent = Agent(
+  task="Your task here",
+  llm=llm,
+  planner_llm=planner_llm,
+  extend_planner_system_message=extend_planner_system_message
+)
```
pirate, line 76:
```suggestion
agent = Agent(
	task="Your task here",
	llm=llm,
	planner_llm=planner_llm,
	extend_planner_system_message=extend_planner_system_message
)
```

browser_use/agent/prompts.py
```diff
@@ -138,28 +138,51 @@ def get_user_message(self, use_vision: bool = True) -> HumanMessage:
...
+		planner_prompt_text = """
+You are a planning agent that helps break down tasks into smaller steps and reason about the current state.
 Your role is to:
 1. Analyze the current state and history
 2. Evaluate progress towards the ultimate goal
 3. Identify potential challenges or roadblocks
 4. Suggest the next high-level steps to take
 
+Available actions:
+{available_actions}
```
pirate, line 166: Adding the full actions list to the planner prompt is a bigger change than the rest, can you PR this separately so we can test the impact?

pirate (approved):
I'm down to merge it all except the part that adds the full actions list (at least in this PR), that has a wider potential impact than just adding extended prompt text.

mathisarends-viadee:
I've removed the available_actions parameter as requested. That said, I'm not entirely sure why the class still keeps this variable if it's no longer being used anywhere.

Also, the inheritance from SystemPrompt seems a bit unnecessary — we’re only overriding get_system_message and not using any of the base class functionality.
Should I remove that as well, or am I missing something here?

pirate (approved):

pirate:
The inheritance is technically unecessary, but helpful because the editor hints help us keep the APIs consistent between the two similar types.

PR #1293, opened 2025-04-05
Add timeout param to go_to_url

mattyhempstead:
We are trying to work with a website that has a very slow initial load time (>30 seconds). We are currently providing a `go_to_url` action in the `initial_actions` param of the `Agent` class, which has a default timeout of 30 seconds and so fails each time. Unfortunately this action does not support a timeout param, despite the `page.goto` method supporting it.

I've made an attempt to fix this (it works for our local instance) and figured I would contribute those changes upstream. Feel free to reject the PR and/or implement it another way.

:)

### DIFF
--- a/browser_use/controller/service.py
+++ b/browser_use/controller/service.py
@@ -98,7 +98,7 @@ class Controller(Generic[Context]):
 		@self.registry.action('Navigate to URL in the current tab', param_model=GoToUrlAction)
 		async def go_to_url(params: GoToUrlAction, browser: BrowserContext):
 			page = await browser.get_current_page()
-			await page.goto(params.url)
+			await page.goto(params.url, timeout=params.timeout)
 			await page.wait_for_load_state()
 			msg = f'🔗  Navigated to {params.url}'
 			logger.info(msg)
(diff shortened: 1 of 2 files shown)


chrisppa:
was this fixed?

browser_use/controller/service.py
```diff
@@ -98,7 +98,7 @@ async def search_google(params: SearchGoogleAction, browser: BrowserContext):
 		@self.registry.action('Navigate to URL in the current tab', param_model=GoToUrlAction)
 		async def go_to_url(params: GoToUrlAction, browser: BrowserContext):
 			page = await browser.get_current_page()
-			await page.goto(params.url)
+			await page.goto(params.url, timeout=params.timeout)
```
pirate, line 101:
```suggestion
			await page.goto(params.url, timeout=params.timeout or 30.0)
```

pirate:
We don't want to expose more parameters/complexity to the LLM on such a critical action. Instead we already offer several options to tune page load times, which can even be modified mid-run using hook functions if you need to change them on a page-by-page basis:

- https://docs.browser-use.com/customize/browser-settings#page-load-settings
	- `minimum_wait_page_load_time`
	- `wait_for_network_idle_page_load_time`
	- `maximum_wait_page_load_time`
- https://docs.browser-use.com/customize/hooks

PR #1303, opened 2025-04-07
fix: fix BrowserContext initialization in new_context method

lynnzc:
### Problem  
When `config=None` is passed to `Browser.new_context()`, it passes the BrowserConfig object directly to BrowserContext rather than allowing `BrowserContext.__init__` to handle the config creation logic properly.
```  
def __init__(
  self,
  browser: 'Browser',
  config: BrowserContextConfig | None = None,
  state: Optional[BrowserContextState] = None,
):
  self.context_id = str(uuid.uuid4())
  
  self.config = config or BrowserContextConfig(**(browser.config.model_dump() if browser.config else {}))
```  

resulting in errors like:
```  
'BrowserConfig' object has no attribute 'user_agent'
```  

image  

### Testing
```python  
import os
import sys
import asyncio

from browser_use.browser.browser import Browser, BrowserConfig, BrowserContext
from browser_use import Agent

load_dotenv()

llm = ChatOpenAI(
	model='gpt-4o-mini',
)
task = 'Find the founders of browser-use and draft them a short personalized message'

async def main():
  browser = Browser(
    config=BrowserConfig(
      headless=False,
    )
  )
  async with await browser.new_context() as context:
    agent = Agent(
      task=task, 
      llm=llm, 
      browser_context=context,
      generate_gif=False,
      enable_memory=False,
    )
    await agent.run()

  await browser.close()

if __name__ == '__main__':
	asyncio.run(main())

```

### DIFF
--- a/browser_use/browser/browser.py
+++ b/browser_use/browser/browser.py
@@ -133,7 +133,7 @@ class Browser:
 
 	async def new_context(self, config: BrowserContextConfig | None = None) -> BrowserContext:
 		"""Create a browser context"""
-		return BrowserContext(config=config or self.config, browser=self)
+		return BrowserContext(config=config, browser=self)
 
 	async def get_playwright_browser(self) -> PlaywrightBrowser:
 		"""Get a browser context"""


alevern:
I think the fix should rather be
```python
return BrowserContext(config=config or self.config.new_context_config, browser=self)
```

lynnzc:
This PR can now be closed as its changes have been merged into the main branch via #1379.
But I think we should pass `None` then rely on the original `init` method to handle it. 

@pirate @azhengzz

PR #1313, opened 2025-04-08
Use ainvoke for async LLM call in extract_content

Haskely:
https://github.com/browser-use/browser-use/blob/eb6ad20bd981ec83ad7aa7383f47eafb65b5342b/browser_use/controller/service.py#L328

Replaced synchronous invoke with await ainvoke in the extract_content method to support proper async behavior under concurrent load.

This improves performance in high-concurrency situations and avoids blocking the event loop.


gregpr07:
Oh wow, how did we miss this haha - thank you :)

Haskely:
By the way, I found that PR #1310, by adding the ruff ASYNC rule, more aggressively detects all possible issues with using blocking sync methods in async contexts. This even requires modifications for all local file I/O operations. However, ruff still ignores the issue of "extract_content" addressed in this PR.

PR #1327, opened 2025-04-09
fix _verify_llm_connection

hongjingzhou:
1. use sync version llm invoke since it is used in __init__
2. remove unreachable return


pirate:
it's async for a reason, it runs in parallel with initial setup in order to not block the rest of the code

PR #1333, opened 2025-04-10
[Mem0] Fix issue related to memory not working with other llms expect OpenAI LLMs

deshraj:
Users have reported previously that memory is not working in browser-use if you are using any other LLM provider except OpenAI LLMs. This PR fixes this issue. 

Tested it with other LLM providers such as anthropic and Azure OpenAI and works fine.


deshraj:
@pirate @MagMueller would appreciate your review 🙏

pirate (approved):

PR #1337, opened 2025-04-10
Remove unused browser-use.png image file

GibzB:


pirate:
we use this file in a bunch of places

PR #1342, opened 2025-04-10
Fix prompt variable syntax: replace {{max_actions}} with {max_actions}

koreng-nimble:
The double braces prevented the value from being interpolated properly. Updated to single braces so `format` correctly replaces it with the real value.


gregpr07:
Damn, you are very right haha - thank you for this :):)

PR #1347, opened 2025-04-11
Fix web_voyager_agent.py example

soeffing:
## Summary by mrge
Fixed the web_voyager_agent.py example to resolve runtime errors related to the Memory/Mem0 addition. Added support for both Azure and OpenAI providers based on available environment variables.

**Bug Fixes**
- Fixed browser window size configuration by using the proper BrowserContextWindowSize type instead of a dictionary.
- Resolved Memory initialization errors by adding an enable_memory=False parameter.

**Refactors**
- Improved LLM initialization to support both OpenAI and Azure OpenAI based on available environment variables.
- Replaced hard-coded environment variable validation with more flexible provider detection.
- Simplified browser context configuration with proper type usage.

Running this example led to errors, mainly caused by the new "Memory/Mem0" addition (see below).

Also, made it possible to run the example with Azure or OpenAI, depending on the env variables that are set

Screenshot 2025-04-10 at 8 35 34 PM


examples/use-cases/web_voyager_agent.py
```diff
@@ -56,13 +58,13 @@
...
 """
 
-
 async def main():
 	agent = Agent(
 		task=TASK,
 		llm=llm,
 		browser=browser,
 		validate_output=True,
+		enable_memory=False,
```
deshraj, line 67: I think you are getting the Mem0 error since using an older version of Mem0 probably. Please use "mem0ai==0.1.88" or later.

browser_use/browser/context.py
```diff
@@ -87,7 +87,7 @@ class BrowserContextConfig(BaseModel):
 	    wait_between_actions: 1.0
 	        Time to wait between multiple per step actions
 
-	    browser_window_size: {'width': 1280, 'height': 1100}
+	    browser_window_size: BrowserContextWindowSize
```
pirate, line 90:
```suggestion
	    browser_window_size: BrowserContextWindowSize(width=1280, height=1100)
```

pirate:
thanks!

PR #1352, opened 2025-04-11
Assign playwright if successful start prior to browser initialization

### LINKED ISSUE #1351: Initialization Edge Case Leads to Traceback
### Bug Description

When attempting to initialize the browser/browser context, there is a potential failure case that causes the playwright_browser in the underlying session to not get cleaned up properly, causing a traceback to occur:

Image

### Reproduction Steps

1. Install browser-use
2. Have the playwright browser initialization succeed but the browser initialization fail
3. Exit application

I was able to make this happen via wss and having the remote side close down it's socket prior to completing initialization

### Code Sample

```python
None - I'll put up a PR that fixes this
```

### Version

main branch

### LLM Model

Gemini 1.5 Pro

### Operating System

Windows 11

### Relevant Log Output

```shell

```

zacka026:
## Summary by mrge
Fixed a bug in browser initialization by storing the playwright reference before browser setup. This change ensures the playwright instance is properly assigned even if browser initialization fails.

**Bug Fixes**
- Moved the playwright assignment before browser setup to prevent reference loss during errors.
- Reordered initialization sequence to ensure proper cleanup if browser setup fails.

Fixes #1351

### DIFF
--- a/browser_use/browser/browser.py
+++ b/browser_use/browser/browser.py
@@ -146,9 +146,9 @@ class Browser:
 	async def _init(self):
 		"""Initialize the browser session"""
 		playwright = await async_playwright().start()
-		browser = await self._setup_browser(playwright)
-
 		self.playwright = playwright
+
+		browser = await self._setup_browser(playwright)
 		self.playwright_browser = browser
 
 		return self.playwright_browser


pirate:
thanks!

PR #1354, opened 2025-04-11
feat: add agent setting to generate playwright locators

rideron89:
## Summary by mrge
Added Playwright locator generation for interacted elements with language selection options. This feature helps users create reliable Playwright test scripts by automatically generating locators in JavaScript or Python.

**New Features**
- Added `playwright_locator_language` agent setting that accepts `"javascript"` or `"python"` values
- Implemented automatic generation of Playwright locators for elements with XPaths
- Added locator strings to state history for each interacted element
- Updated documentation with usage instructions and requirements

**Refactors**
- Restructured element interaction tracking to support locator generation
- Improved error handling for locator generation failures

# Preface

Hello! We had a project requirement to generate Playwright locators which required changes to browser-use, so I figured I'd create a PR here to see if it made sense to include. I will say, that the, as far as I can tell, full implementation of this feature is outside the scope of browser-use, because `playwright` must be made accessible in the browser, which, by default, it is not, and seems outside the scope of browser-use.

One method of injecting `playwright` is to launch a browser with CDP enabled, run some Playwright code, "pause" the page (which opens a codegen window), and connect to that browser with browser-use over CDP. You can read more here: https://playwright.dev/python/docs/codegen#record-using-custom-setup

# Summary

This change adds a new agent setting, `playwright_locator_language` which can be set to enable Playwright locator codegen for all interacted elements. It defaults to `None`, but accepts values of `"javascript"` and `"python"`, which control the language the codegen outputs.

The generated locator string will exist in the state history as part of the agent history item.

### Example

Element with XPath `html/body/dialog/form/button` could generate a Playwright locator string of `get_by_role("button", name="Login")`.

### DIFF
--- a/browser_use/agent/service.py
+++ b/browser_use/agent/service.py
@@ -37,6 +37,7 @@ from browser_use.agent.views import (
 	AgentSettings,
 	AgentState,
 	AgentStepInfo,
+	PlaywrightLocatorLanguage,
 	StepMetadata,
 	ToolCallingMethod,
 )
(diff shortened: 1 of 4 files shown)


deveshruttala-edu:

PR #1357, opened 2025-04-12
Moved username and password to env instead of hardcoding in the file

vasu-pocketfm:
## Summary by mrge
Improved security by moving hardcoded credentials to environment variables. This change follows security best practices by storing sensitive information in the environment rather than in code.

**Refactors**
- Moved username and password from hardcoded values to environment variables
- Added code to retrieve credentials from MIGROS_USERNAME and MIGROS_PASSWORD env vars
- Updated the task string to use f-string formatting for credential insertion
- Added comments to improve code readability

**Security Improvements**
- Removed plaintext credentials from the codebase
- Implemented proper credential management using dotenv

Moved username and password to env instead of hardcoding in the file

### DIFF
--- a/examples/use-cases/shopping.py
+++ b/examples/use-cases/shopping.py
@@ -1,13 +1,18 @@
 from dotenv import load_dotenv
 from langchain_openai import ChatOpenAI
-
 from browser_use import Agent, Browser
-
-load_dotenv()
-
+import os
 import asyncio
 
-task = """
+# Load environment variables from .env file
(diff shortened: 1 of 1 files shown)


vasu-pocketfm (commented):
Nit pick changes

pirate:
dont use env vars for passwords, this is just an example but for real use cases you should be getting them from a secrets manager and passing them via: https://docs.browser-use.com/customize/sensitive-data

PR #1359, opened 2025-04-13
Fix: Properly await LLM connection verification before agent run

RogelioRichmanAstronaut:
## Summary by mrge
Fixed a critical issue where the agent would fail at startup due to unverified LLM connections. The asynchronous connection verification is now properly awaited before the agent begins operation, preventing premature failures.

**Bug Fixes**
- Fixed the LLM connection verification by properly awaiting it at the beginning of the agent's run method.
- Removed the non-awaited verification call from the initialization method that was causing a RuntimeWarning.

**Refactors**
- Moved connection verification to ensure it completes before any LLM-dependent operations begin.
- Replaced an assertion check with the actual verification process for more robust error handling.

## Description

This PR addresses an issue where the agent could fail immediately upon starting if the LLM connection verification hadn't completed. The `_verify_llm_connection` asynchronous function was called but not awaited before the main agent loop began in the `run` method.

This resulted in a `RuntimeWarning: coroutine 'Agent._verify_llm_connection' was never awaited` and could cause the agent to hit its `max_failures` limit without performing any actual steps, as the LLM calls would fail.

## Changes Made

 - Removed the non-awaited call to `_verify_llm_connection` from the `__init__` method of the `Agent` class.
 - Added `await self._verify_llm_connection(self.llm)` at the beginning of the `run` method in `browser_use/agent/service.py`.

## Impact

This change ensures the agent robustly verifies its connection to the LLM service *before* attempting any operations that require it, preventing premature failures and improving startup reliability.


ncrmro:
I think this is also missing the `sentence-transformers` package. When I try to install from your commit directly.

RogelioRichmanAstronaut:
> I think this is also missing the `sentence-transformers` package. When I try to install from your commit directly.

Hi ncrmro,

Thanks so much for testing the commit and raising the point about `sentence-transformers`! It's valuable feedback.

You've highlighted an interesting challenge regarding the `sentence-transformers `dependency. It appears likely required by the default `mem0ai` component.

This particular PR is narrowly focused on fixing the `RuntimeWarning: coroutine '_verify_llm_connection' was never awaited `by ensuring the LLM connection verification is properly awaited.

Addressing the `sentence-transformers` dependency presents a great opportunity for improvement in a separate context. We could investigate whether it should be an explicit dependency in `pyproject.toml`or clarify how `mem0ai` should handle its installation consistently across different environments. Tackling this as a distinct issue or PR would allow us to give it the focused attention needed.

For now, keeping this PR dedicated solely to the `await` fix helps ensure we can integrate this specific improvement smoothly.

Thanks again for bringing this up – it’s definitely a point worth exploring further! Let me know your thoughts.

pirate:
This misunderstands how the code works, the original design is correct but noisy, if you are seeing an error about an unawaited coroutine it's because setup is failing for some other reason.

RogelioRichmanAstronaut:
@pirate Thanks for the clarification Nick

pirate:
Improved it in this PR https://github.com/browser-use/browser-use/pull/1398

Issue #648, opened 2025-02-10
Final result gif not support Chinese chars


baiyyee:
### Bug Description

I have task like "goole 搜索 pytest, 进入 pytest 官方网站, 下载 pdf 格式帮助文档到本地", after its done, the final result gif not support Chinese chars, can't show Chinese words correct

<img width="917" alt="Image" src="https://github.com/user-attachments/assets/28d5d17b-7396-41e3-a7e4-b67582e9b6f6" />

### Reproduction Steps

N/A

### Code Sample

```python
import asyncio
import os

from dotenv import load_dotenv
from langchain_openai import ChatOpenAI
from pydantic import SecretStr

from browser_use import Agent

# dotenv
load_dotenv()

api_key = os.getenv("DEEPSEEK_API_KEY", "")
if not api_key:
    raise ValueError("DEEPSEEK_API_KEY is not set")


async def run_search():
    agent = Agent(
        task=("goole 搜索 pytest, 进入 pytest 官方网站, 下载 pdf 格式帮助文档到本地"),
        llm=ChatOpenAI(
	    base_url='https://api.deepseek.com/v1',
	    model='deepseek-reasoner',
            api_key=SecretStr(api_key),
        ),
        use_vision=False,
        max_failures=2,
        max_actions_per_step=1,
    )

    await agent.run()


if __name__ == "__main__":
    asyncio.run(run_search())
```

### Version

0.1.36

### LLM Model

Other (specify in description)

### Operating System

deepseek-r1

### Relevant Log Output

```shell
INFO     [browser_use] BrowserUse logging setup complete with level info
INFO     [root] Anonymized telemetry enabled. See https://docs.browser-use.com/development/telemetry for more information.
INFO     [agent] 🚀 Starting task: goole 搜索 pytest, 进入 pytest 官方网站, 下载 pdf 格式帮助文档到本地
INFO     [agent] 📍 Step 1
ERROR    [agent] ❌ Result failed 1/2 times:
 browser_use.agent.views.AgentOutput() argument after ** must be a mapping, not list
INFO     [agent] 📍 Step 1
INFO     [agent] 🤷 Eval: Unknown - No previous actions taken yet
INFO     [agent] 🧠 Memory: Task initiated. Need to: 1) Google search 'pytest' 2) Find official site 3) Locate PDF docs
INFO     [agent] 🎯 Next goal: Perform initial Google search for 'pytest'
[...]

PR #1368, opened 2025-04-14
Fix non-alphabet character rendering issues in gif picture

asasasasasbc:
## Summary by mrge
Added ArialUni font to the rendering options for GIF images to properly display non-alphabetic Unicode characters. This fix addresses issue #648 by supporting Chinese, Japanese, Korean, and other international character sets.

**Bug Fixes**
- Added ArialUni font to the font preference list for GIF rendering
- Prioritized ArialUni before Arial to better handle non-alphabetic characters

**Dependencies**
- Requires ArialUni font, which comes with Microsoft Office
- Font can be downloaded from GitHub if Office is not installed

Added ArialUni font rendering option, which is an Arial Unicode font that comes with MS Office and can be used to render most non-alphabetic Unicode characters, including Chinese, Japanese, Korean, etc. This addresses issue #648.

Dependencies:

If Microsoft Office is installed on the system, this font is automatically usable. If not, this font can be easily downloaded from: https://github.com/kaienfr/Font/blob/master/font/ARIALUNI.TTF

### DIFF
--- a/browser_use/agent/gif.py
+++ b/browser_use/agent/gif.py
@@ -49,7 +49,8 @@ def create_history_gif(
 	# Try to load nicer fonts
 	try:
 		# Try different font options in order of preference
-		font_options = ['Helvetica', 'Arial', 'DejaVuSans', 'Verdana']
+		# ArialUni is a font that comes with Office and can render most non-alphabet characters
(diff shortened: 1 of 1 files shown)


pirate:
thanks!

Issue #853, opened 2025-02-24
Failed to locate element: Locator.element_handle: Error: strict mode violation: locator(...) resolved to 2 elements


ameyapathare:
### Bug Description

I am trying to get browser use to navigate a page with multiple action buttons on it, but it cannot click the button because the locator locates multiple elements.

I believe the error is being triggered in [browser/context.py](https://github.com/browser-use/browser-use/blob/main/browser_use/browser/context.py#L989).  

Possible solution could be to modify the logic in [this code block](https://github.com/browser-use/browser-use/blob/738bad1af5f79d1e52ed0c187e5817c620873529/browser_use/browser/context.py#L973-L987) to always click the first element if multiple are returned.

### Reproduction Steps

This example is from Salesforce.com: https://playful-otter-smx8dr-dev-ed.trailblaze.lightning.force.com/lightning/setup/ObjectManager/Account/FieldsAndRelationships/new.  I give it the instructions:
"""
go to https://playful-otter-smx8dr-dev-ed.trailblaze.lightning.force.com/lightning/setup/ObjectManager/Account/FieldsAndRelationships/new
click the Formula radio button
click the first 'Next' button
"""

It correctly identifies the first 'Next' button to click:
![Image](https://github.com/user-attachments/assets/655d07ed-20d6-42c7-ab5e-65a4cac0075d)

But when it goes to click it, it runs into the error.

I tried creating a toy example with two identical buttons [here](https://v0-react-button-redirect.vercel.app/), but browser-use was able to navigate this page effectively, so I've only been able to recreate within Salesforce.com.  

I had to use web-ui to leverage the 'use same browser' functionality since I needed to log into Salesforce before launching browser-use on it.

### Code Sample

```python
Run prompt: """go to https://playful-otter-smx8dr-dev-ed.trailblaze.lightning.force.com/lightning/setup/ObjectManager/Account/FieldsAndRelationships/new
click the Formula radio button
click the first 'Next' button"""

Unfortunately, I don't have a reproducible example that doesn't require a login.
```

### Version

0.1.29

### LLM Model

GPT-4o

### Operating System

macOS 14.6.1

### Relevant Log Output

```shell
INFO     [agent] 🚀 Starting task: 
        go to https://playful-otter-smx8dr-dev-ed.trailblaze.lightning.force.com/lightning/setup/ObjectManager/Account/FieldsAndRelationships/new
        click the Formula radio button
        click the first 'Next' button
        
[...]

PR #744, opened 2025-02-17
【bug fix】xpath caused error: strict mode conflict: locator

timeash:
【bug fix】
ERROR    [browser] Failed to locate element: Locator.element_handle: Error: strict mode violation: locator("html > body > div > div > div > section > aside > div > ul > li.el-menu-item[role=\"menuitem\"]") resolved to 7 elements

【description】
I tried to find elements on the page through a locator, but the locator matched multiple elements (7). The xpath's judgment logic causes the first css_selector of the element group to be'li' instead of'li: nth-of-type(1)', resulting in Failed to locate element

PR #1369, opened 2025-04-14
Fix: Ensure first element is correctly indexed in sibling list

shivamkhatri:
## Summary by mrge
Fixed the element position calculation in DOM tree building to correctly index the first element among siblings. This ensures accurate XPath generation by properly handling elements regardless of their position in the sibling list.

**Bug Fixes**
- Fixed indexing of the first element in sibling lists that was previously skipped or incorrectly calculated
- Added new `getElementPosition` function that correctly determines element position among siblings with the same tag name
- Improved handling of edge cases where an element is the only one of its type in the parent
- https://github.com/browser-use/browser-use/issues/853
- Better implementation than Open PR https://github.com/browser-use/browser-use/pull/744

**Refactors**
- Replaced previous sibling-traversal logic with a cleaner Array-based approach
- Changed position calculation to use direct children of the parent element, avoiding non-element nodes
- Implemented 1-based indexing for consistent XPath generation

#### **Overview:**
This update introduces a more reliable and efficient method for determining the position of an element among its siblings, specifically focusing on elements with the same tag name. The solution now properly accounts for the position of the **first element** and ensures correct indexing whether it's the first, middle, or last element of its type in the parent.

#### **Key Changes:**
- **Element Position Calculation:**
  - The `getElementPosition` function has been enhanced to correctly calculate the position of an element among its **direct siblings** with the same tag name.
  - The function now:
    - Filters siblings by tag name.
    - Returns `0` if the element is the only one of its tag type (first in the list).
    - Returns the **1-based index** of the element if multiple siblings share the same tag.
    - Ensures that the **first element** is also properly indexed (no longer skipped).

- **Parent Element Handling:**
  - Uses the parent’s direct children (via `parentElement.children`) to ensure the correct position within the sibling list.
  - This avoids counting non-element nodes (such as text nodes or comments), ensuring more accurate calculations.
  
- **Robustness:**
  - Returns `null` for cases where no siblings exist (i.e., the element has no siblings in the DOM).
  - Handles edge cases where the element might be the only one of its type in the parent.
  - Accounts for the first element, ensuring the position calculation begins correctly.

#### **Why This Change?**
- **Previous XPath Issue:**
  The previous approach to generating the XPath would resolve to multiple elements because it didn't correctly account for the position of the **first element** in the sibling list. This caused an issue where the XPath would resolve to more than one matching element, leading to errors like:
  ```
  Failed to locate element: Locator.element_handle: Error: strict mode violation: locator("html > body > div:nth-of-type(2) > div > div > main > div > div:nth-of-type(2) > div > div:nth-of-type(2) > div > div > div:nth-of-type(2) > div > div > div > div > div > div:nth-of-type(2) > div > div > div > div:nth-of-type(2) > div > div > div > div:nth-of-type(2) > div > button.Dniwja_spectrum-ActionButton.Dniwja_spectrum-BaseButton.Dniwja_i18nFontFamily.Dniwja_spectrum-FocusRing.Dniwja_spectrum-FocusRing-ring[type=\"button\"][aria-label=\"Add guideline\"][data-testid=\"content-popover-actionBtn\"]") resolved to 9 elements.
  ```

Earlier, the generated XPath looked like this:
```
html/body/div[2]/div/div/main/div/div[2]/div/div[2]/div/div/div[2]/div/div/div[2]/div/div/div/div/div/div[2]/div/div/div/div[2]/div/div/div/div[2]/div/button
```

After this change the XPath looks like below:

```
html/body/div[2]/div/div/main/div/div[2]/div/div[2]/div/div/div[2]/div/div/div[2]/div/div[1]/div/div/div/div[2]/div/div[1]/div/div[2]/div/div[1]/div[1]/div[2]/div/button
```

- With this change, the **correct 1-based position** of the first element is now properly considered, ensuring that the XPath or element locator resolves to a single, accurate element.


shivamkhatri:
@gregpr07 @pirate could you please review this PR?

shivamkhatri:
@pirate , @gregpr07 , could you please check it? Thanks

pirate:
thanks!

Issue #1377, opened 2025-04-15
multiple_agents_same_browser.py example fails to run with error "AttributeError: 'BrowserConfig' object has no attribute 'user_agent'"


azhengzz:
### Bug Description

`browser-use/examples/features/multiple_agents_same_browser.py`

When running the multiple_agents_same_browser.py example script, it throws the following

error:
```py
AttributeError: 'BrowserConfig' object has no attribute 'user_agent'
```

### Reproduction Steps

1. Execute the example script multiple_agents_same_browser.py

### Code Sample

```python
import os
import sys

from langchain_openai import ChatOpenAI

sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))

import asyncio

from browser_use import Agent, Browser


# Video: https://preview.screen.studio/share/8Elaq9sm
async def main():
	# Persist the browser state across agents

	browser = Browser()
	async with await browser.new_context() as context:
		model = ChatOpenAI(model='gpt-4o')
		current_agent = None

		async def get_input():
			return await asyncio.get_event_loop().run_in_executor(
				None, lambda: input('Enter task (p: pause current agent, r: resume, b: break): ')
			)

		while True:
			task = await get_input()

			if task.lower() == 'p':
				# Pause the current agent if one exists
				if current_agent:
					current_agent.pause()
				continue
			elif task.lower() == 'r':
				# Resume the current agent if one exists
				if current_agent:
					current_agent.resume()
				continue
			elif task.lower() == 'b':
				# Break the current agent's execution if one exists
				if current_agent:
					current_agent.stop()
					current_agent = None
				continue

			# If there's a current agent running, pause it before starting new one
			if current_agent:
				current_agent.pause()

			# Create and run new agent with the task
			current_agent = Agent(
				task=task,
				llm=model,
				browser_context=context,
			)

			# Run the agent asynchronously without blocking
			asyncio.create_task(current_agent.run())


asyncio.run(main())

# Now aad the cheapest to the cart
```

### Version

main

### LLM Model

GPT-4o

### Operating System

Windows11

### Relevant Log Output

```shell
DEBUG    [browser] 🌎  Initializing new browser
DEBUG    [browser] 🌎  Initializing new browser context with id: 4d6f8133-ba5b-43c0-926b-5eccd7f99255
DEBUG    [browser_use] --init (browser) Execution time: 3.40 seconds
Traceback (most recent call last):
[...]

PR #1379, opened 2025-04-15
Fix: incorrect config type in BrowserContext instantiation

azhengzz:
## Summary by mrge
Fixed a type mismatch in the BrowserContext instantiation process. This change ensures the correct configuration type is passed when creating a new browser context.

**Bug Fixes**
- Updated the `new_context` method to properly convert the browser's config to a BrowserContextConfig object.
- Resolved issue #1377 where incorrect config type was causing instantiation problems.

The problem description refers to this PR https://github.com/browser-use/browser-use/issues/1377


pirate:
Thanks!

PR #1389, opened 2025-04-16
feat: new actions: hover and input number

konrad-czarnota-ds:
## Summary by mrge
Added hover and number input capabilities to the browser automation framework. These new actions allow for more complex user interactions like hovering over elements and entering numeric values in forms.

**New Features**
- Added hover functionality to interact with hover-triggered UI elements
- Created number input action for entering numeric values in form fields
- Built action handlers for both features in the controller service
- Added corresponding action models with proper parameter definitions

**Examples**
- Added example script for hovering over Wikipedia links to show page previews
- Created example for numeric input using a USPS zip code lookup form


examples/features/hover_over_element.py
```diff
@@ -0,0 +1,31 @@
...
+from browser_use import Agent
+from browser_use.browser.browser import Browser
+
+load_dotenv()
+if not os.getenv('OPENAI_API_KEY'):
+	raise ValueError('OPENAI_API_KEY is not set. Please add it to your environment variables.')
+
+
+async def run_hover():
+	browser = Browser()
```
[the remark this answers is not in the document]
konrad-czarnota-ds: browser creation deleted

mathisarends-viadee:
This is a great idea! I'm currently working on a tool for evaluating the usability of websites and use browser-use for the automation part. The hover action would be super helpful for interacting with tooltips and other dynamic elements. Looking forward to seeing this merged!

browser_use/browser/context.py
```diff
@@ -1395,6 +1395,29 @@ async def _input_text_element_node(self, element_node: DOMElementNode, text: str
...
+			if element_handle is None:
+				raise BrowserError(f'Element: {repr(element_node)} not found')
+
+			try:
+				await element_handle.wait_for_element_state('stable', timeout=1000)
+				await element_handle.scroll_into_view_if_needed(timeout=1000)
+			except Exception:
+				pass
+
+			await element_handle.evaluate(f'el => el.value = {number}')
```
pirate, line 1415: you should simulate keystrokes, not set the value in JS

as-is this wont work for most 2fa fields that have separate input fields but allow you to type the full number into the first one

Also how is this function any different from `input_text()` currently?

konrad-czarnota-ds: Thanks for the feedback — you're absolutely right. I’ve updated the implementation to simulate keystrokes instead of setting the value directly. In our specific case, input_text() didn't work directly, but typing is definitely a more general solution here.

deveshruttala-edu:

deveshruttala-edu:

Issue #1365, opened 2025-04-14
Error using DeepSeek LLM with Browser-Use Agent: Deserialization Failure


dukeyc138:
### Bug Description

Hi team,
I'm trying to use the browser-use Agent with a DeepSeek LLM (deepseek-chat model via langchain-deepseek), but I'm consistently encountering an error during the agent's run.

### Reproduction Steps

The DEEPSEEK_API_KEY has been verified and works correctly when calling the DeepSeek API directly (e.g., using the deepseek-ai library or curl). Multiple valid keys were tested.
The error persists regardless of the specific task given to the agent.
Attempts to use the ChatOpenAI class while pointing the openai_api_base to https://api.deepseek.com/v1 and using the DeepSeek key also resulted in the same deserialization error.
This appears to be a known issue, potentially related to message format incompatibility between browser-use / langchain-core message structures and the DeepSeek API expectations. Similar problems seem to be reported in:
#1362 1
#1355 2
#1317 3
Any guidance on resolving this incompatibility or potential workarounds specifically for using DeepSeek with the browser-use Agent would be greatly appreciated. Happy to provide more details if needed.
Thanks!

### Code Sample

```python
import asyncio
from dotenv import load_dotenv
from browser_use import Agent
from langchain_deepseek import ChatDeepSeek
import os

# Load environment variables (optional if hardcoding key)
load_dotenv()

# Get API key
DEEPSEEK_API_KEY = os.getenv("DEEPSEEK_API_KEY")
# Or hardcode: DEEPSEEK_API_KEY = "sk-..." 
if not DEEPSEEK_API_KEY:
    raise ValueError("Please set the DEEPSEEK_API_KEY")

async def main():
    print("Initializing DeepSeek LLM...")
    deepseek_llm = ChatDeepSeek(
        api_key=DEEPSEEK_API_KEY,
        model="deepseek-chat", # Tested with deepseek-chat
        temperature=0.7,
        max_tokens=4000
    )
    print("LLM Initialized.")

    print("Initializing Browser-Use Agent...")
    # Note: 'headless' argument removed as it caused TypeError in recent versions
    agent = Agent(
        task="Search for the latest research on quantum computing and summarize it.",
        llm=deepseek_llm
    )
    print("Agent Initialized.")
    
    print("Running agent...")
    try:
        # The error occurs during this step
        result = await agent.run() 
        print("\\n===== Task Result =====")
        print(result)
        print("=====================")
    except Exception as e:
[...]

Issue #1362, opened 2025-04-13
DeepSeek Example not work.


eliojing:
### Bug Description

The execution failed while attempting the deepseek example.
Already set the deepseek api-key in the .env

### Reproduction Steps

Run as the document, use the deepseek example.

### Code Sample

```python
import asyncio
import os

from dotenv import load_dotenv
from langchain_deepseek import ChatDeepSeek
from pydantic import SecretStr

from browser_use import Agent

# dotenv
load_dotenv()

api_key = os.getenv('DEEPSEEK_API_KEY', '')
print(api_key)
if not api_key:
        raise ValueError('DEEPSEEK_API_KEY is not set')


async def run_search():
        agent = Agent(
                task=(
                        '1. Go to https://www.baidu.com'
                        "2. Search for 'browser use' in the search bar"
                        '3. Click on first result'
                        '4. Return the first comment'
                ),
                llm=ChatDeepSeek(
                        base_url='https://api.deepseek.com/v1',
                        model='deepseek-chat',
                        api_key=SecretStr(api_key),
                ),
                use_vision=False,
                max_failures=10
        )

        await agent.run()


if __name__ == '__main__':
        asyncio.run(run_search())
```

### Version

pip Version: 0.1.40

### LLM Model

DeepSeek Coder

### Operating System

macOS 13.4 

### Relevant Log Output

```shell
INFO     [agent] 🚀 Starting task: 1. Go to https://www.baidu.com2. Search for 'browser use' in the search bar3. Click on first result4. Return the first comment
INFO     [agent] 📍 Step 1
INFO     [agent] 📍 Step 1
..
..
INFO     [agent] 📍 Step 1
INFO     [agent] Last step finishing up
INFO     [agent] ❌ Failed to complete task in maximum steps
```

neonwatty:
Your example worked for me.

Does the basic "hello world" example from deepseek work on your end?  

```python
import os
from dotenv import load_dotenv
load_dotenv()

from openai import OpenAI
api_key = os.getenv("DEEPSEEK_API_KEY")

client = OpenAI(api_key=api_key, base_url="https://api.deepseek.com")

response = client.chat.completions.create(
    model="deepseek-chat",
    messages=[
        {"role": "system", "content": "You are a helpful assistant"},
        {"role": "user", "content": "Hello"},
    ],
    stream=False
)

print(response.choices[0].message.content)
```

Issue #1355, opened 2025-04-11
Trying the example for deepseek do not work


ndinev:
### Bug Description

- Windows 11, python 3.11.9, playwright-1.51.0
- Install browser use as per instruction
- Configure env variables as per instruction
- get examples/models/deepseek.py and run it

Result: Chromium is started but never navigate to any site
In console I see many attempts to execute Step1 that eventually end with error

INFO     [agent] 📍 Step 1
INFO     [agent] 📍 Step 1
INFO     [agent] 📍 Step 1
INFO     [agent] 📍 Step 1
INFO     [agent] 📍 Step 1
INFO     [agent] 📍 Step 1
INFO     [agent] Last step finishing up
INFO     [agent] ❌ Failed to complete task in maximum steps

### Reproduction Steps

- Install browser use as per instruction
- Configure env variables as per instruction
- get examples/models/deepseek.py and run it

### Code Sample

```python
examples/models/deepseek.py
```

### Version

browser-use         0.1.40

### LLM Model

DeepSeek Coder

### Operating System

Windows 11

### Relevant Log Output

```shell
INFO     [agent] 🚀 Starting task: 1. Go to https://www.reddit.com/r/LocalLLaMA 2. Search for 'browser use' in the search bar3. Click on first result4. Return the first comment
INFO     [agent] 📍 Step 1
INFO     [agent] 📍 Step 1
.
.
INFO     [agent] 📍 Step 1
INFO     [agent] 📍 Step 1
INFO     [agent] 📍 Step 1
INFO     [agent] 📍 Step 1
INFO     [agent] 📍 Step 1
INFO     [agent] 📍 Step 1
INFO     [agent] Last step finishing up
INFO     [agent] ❌ Failed to complete task in maximum steps
```

yha9806:
hello have you fix it?

GhouI:
Can you show us the proper code sample?

pokitpeng:
same question

Issue #1317, opened 2025-04-08
DeepseekV3 not work！


kukugpt:
### Bug Description

<img width="738" alt="Image" src="https://github.com/user-attachments/assets/88d353e7-6d6a-4009-b9b2-491406896c2f" />
When I configure deepseekv3, it doesn't work properly

### Reproduction Steps

1. configure the deepseek model
2. python example/ui/gradio_demo.py

### Code Sample

```python
async def run_browser_task(
	task: str,
	api_key: str,
	model: str = 'gpt-4o',
	headless: bool = True,
) -> str:
	if not api_key.strip():
		return 'Please provide an API key'

	os.environ['OPENAI_API_KEY'] = api_key


	try:
		agent = Agent(
			task=task,
			llm=ChatOpenAI(
				base_url="https://api.deepseek.com",
				api_key='sk-#####',
				model='deepseek-chat')
		)
		result = await agent.run()
		#  TODO: The result cloud be parsed better
		return result
	except Exception as e:
		return f'Error: {str(e)}'
```

### Version

latest

### LLM Model

Other (specify in description)

### Operating System

MAC os

### Relevant Log Output

```shell

```

imecodehub:
deepseek-chat 不支持视觉模型，所以需要加上use_vision=False，此处更改 		agent = Agent(
			task=task,
			llm=ChatOpenAI(
				base_url="https://api.deepseek.com",
				api_key='sk-#####',
				model='deepseek-chat')
              use_vision=False,
		)

kukugpt:
I configure ‘use_vision=False’, but same wrong is presented.

kukugpt:
thx, I have sloved my problem

PR #1393, opened 2025-04-16
Autoconfigure `use_vision=False` for models that dont support it

neonwatty:
## Summary by mrge
Fixed vision setting for DeepSeek models to prevent common errors. The code now dynamically sets `use_vision=False` by default for DeepSeek models while maintaining `True` for other models.

**Bug Fixes**
- Changed `use_vision` parameter to accept `None` to enable dynamic configuration
- Added logic to automatically set `use_vision=False` for DeepSeek models
- Moved model setup before agent settings initialization to ensure proper configuration

**Refactors**
- Reordered initialization sequence to set model names before configuring agent settings
- Improved code organization by grouping related initialization steps

There are a number of open issues related to a common error using deepseek models (https://github.com/browser-use/browser-use/issues/1365, https://github.com/browser-use/browser-use/issues/1362, https://github.com/browser-use/browser-use/issues/1355, https://github.com/browser-use/browser-use/issues/1317) that appear to be related to the universal default setting of `use_vision` to `True`.

To avoid future confusion I propose here moving model setup before that`use_vision` can be set dynamically based on `self.model_name` - and in particular so that it is set to `False` by default for deepseek models.


pirate:
Please fix the lint errors then it's good to merge, thanks!

run `ruff check --fix; ruff format` manually and commit the results  
or  
run `pre-commit install --install-hooks` to auto-format automatically on every commit

neonwatty:
All set.

pirate:
Your fix had a circular dependency issue so I just rewrote it quickly here: https://github.com/browser-use/browser-use/pull/1399

Issue #1442, opened 2025-04-21
AttributeError: 'AgentSettings' object has no attribute 'enable_memory'


Anirudha619:
### Bug Description

While initializing the agent, the code raises the following error:

AttributeError: 'AgentSettings' object has no attribute 'enable_memory'

### Reproduction Steps

1. install browser use
2. `python examples/models/gemini.py`

### Code Sample

```python
python examples/models/gemini.py
```

### Version

commit = ed6540b67e6241f0cea6602c5e2f5966870d5b86

### LLM Model

Other (specify in description)

### Operating System

gemini-2.0-flash-exp

### Relevant Log Output

![Image](https://github.com/user-attachments/assets/ec63d3e2-1d82-48e8-8ced-6013e4489d5b)

PR #1443, opened 2025-04-21
Fixs #1442 : 'AgentSettings' object has no attribute 'enable_memory'

Anirudha619:
Fixes https://github.com/browser-use/browser-use/issues/1442
Fix: Corrected reference to enable_memory attribute in agent initialization.

Replaced self.settings.enable_memory with self.enable_memory to fix the error: 'AgentSettings' object has no attribute 'enable_memory'.

---

## Summary by mrge
Fixed an error in agent initialization by using self.enable_memory instead of self.settings.enable_memory. This prevents attribute errors when starting the agent.


pirate:
Thanks!

PR #1456, opened 2025-04-22
Fix RuntimeError: no running event loop

Anirudha619:
This PR fixes the test_vision.py file that was failing with "RuntimeError: no running event loop" during test collection.

Changes:

- Added explicit event_loop fixture to properly manage asyncio resources
- Moved Agent initialization inside the test function to prevent asyncio tasks from being created outside a running event loop
- Added proper cleanup with try/finally to ensure browser resources are closed
    

---

## Summary by mrge
Fixed "RuntimeError: no running event loop" in test_vision.py by adding an event loop fixture and moving Agent initialization inside the test function.

- **Bug Fixes**
  - Added an event_loop fixture to manage asyncio resources.
  - Moved Agent creation inside the test to avoid creating asyncio tasks outside a running event loop.
  - Ensured browser resources are closed after the test.

### DIFF
--- a/examples/custom-functions/file_upload.py
+++ b/examples/custom-functions/file_upload.py
@@ -70,7 +70,7 @@ async def read_file(path: str, available_file_paths: list[str]):
 	if path not in available_file_paths:
 		return ActionResult(error=f'File path {path} is not available')
 
-	async with (await anyio.open_file(path, 'r')) as f:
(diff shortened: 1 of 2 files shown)


pirate:
can you run the linter with `pre-commit run --all-files`, then it's good to merge, thanks!

Issue #1397, opened 2025-04-17
browser_window_size not working


edwardysun:
### Bug Description

After trying out 0.1.41rc3, I'm noticing that the `browser_window_size` no longer changes the browser's window size. I verified the numbers are still being passed in correctly.

### Reproduction Steps

see code sample below

### Code Sample

```python
window_size = BrowserContextWindowSize(width=800, height=400)
config = BrowserContextConfig(browser_window_size=window_size)

browser_context = BrowserContext(
    browser=browser,
    config=config
)

agent = Agent(
    task=prompt,
    llm=actual_llm,
    browser_context=browser_context,
)
```

### Version

0.41.1rc3

### LLM Model

Gemini 1.5 Pro

### Operating System

Mac

### Relevant Log Output

```shell

```

edwardysun:
Note: this worked in 0.1.40

pirate:
can you try the new 0.1.41, I believe this should be fixed now

edwardysun:
@pirate just tried it with 0.1.41 but still not seeing the screen size changes

kpmc-anu:
same here as well ..

fantasyisking:
![Image](https://github.com/user-attachments/assets/e9530288-d995-4e7f-a8cc-4a81de3370f7)

edwardysun:
@pirate would appreciate a review on the PR above, thank you!

pirate:
PR #1472 is merged, please help test!

sandeepsalwan1:
Hi @kpmc-anu,

Thanks for your feedback! Let me know if any other issues and I will work to resolve it within the day. I've submitted a PR to fix this issue. 

The problem occurs because the browser's window size and viewport settings aren't synchronized properly. When you set browser_window_size but leave the default viewport settings, the mismatch causes elements to appear at the bottom of the page.

## Temporary fix for your issue:

```python
# Solution: Set both viewport and window size to match
config = BrowserContextConfig(
    viewport=BrowserContextViewport(width=1440, height=900),  # Match your desired dimensions
    browser_window_size=BrowserContextWindowSize(width=1440, height=900)
)
```

Or alternatively:

```python
# Alternative: Disable viewport entirely and just use window size
config = BrowserContextConfig(
    no_viewport=True,  # Let the window size control everything
    browser_window_size=BrowserContextWindowSize(width=1440, height=900)
)
```

My PR properly respects the user-provided no_viewport setting when browser_window_size is specified.

[...]

PR #1472, opened 2025-04-24
Fix browser_window_size not changing window size in non-headless mode

sandeepsalwan1:
## Description
This PR fixes issue #1397 where the `browser_window_size` setting in `BrowserContextConfig` was not changing the browser's window size in non-headless mode.

## Changes
- Updated the `_setup_builtin_browser` method in `browser_use/browser/browser.py` to use the configured window size from `BrowserContextConfig` when available
- Enhanced the `_resize_window` method in `browser_use/browser/context.py` to set both the viewport size and the actual window size using CDP
- Added an example script to demonstrate the fix

## Testing
- Added an example script `examples/browserDemo.py` that demonstrates the fix
- Tested in both headless and non-headless modes
- Verified that the window size is correctly set to the configured size

### Screenshots
image
image
    

---

## Summary by mrge
Fixed an issue where the browser_window_size setting did not change the window size in non-headless mode.

- **Bug Fixes**
  - Ensured the configured window size is applied to both the viewport and the actual browser window in non-headless mode.
  - Added an example script to show the fix in action.

### DIFF
--- a/browser_use/browser/browser.py
+++ b/browser_use/browser/browser.py
@@ -321,7 +321,7 @@ class Browser:
 			args=args[self.config.browser_class],
 			proxy=self.config.proxy.model_dump() if self.config.proxy else None,
 			handle_sigterm=False,
-			handle_sigint=False
+			handle_sigint=False,
 		)
 		return browser
 
--- a/browser_use/browser/browser.py
+++ b/browser_use/browser/browser.py
@@ -27,7 +27,7 @@ from browser_use.browser.chrome import (
 	CHROME_HEADLESS_ARGS,
 )
 from browser_use.browser.context import BrowserContext, BrowserContextConfig
-from browser_use.browser.utils.screen_resolution import get_window_adjustments, get_screen_resolution
+from browser_use.browser.utils.screen_resolution import get_screen_resolution, get_window_adjustments
 from browser_use.utils import time_execution_async
 
 logger = logging.getLogger(__name__)
@@ -299,22 +299,13 @@ class Browser:
 			],
 		}
 
-		# Add viewport size to launch options for non-headless mode
-		viewport = None
-		if (
-			not self.config.headless
-			and hasattr(self.config, 'new_context_config')
-			and hasattr(self.config.new_context_config, 'browser_window_size')
-		):
-			viewport = self.config.new_context_config.browser_window_size.model_dump()
-
 		browser = await browser_class.launch(
 			headless=self.config.headless,
 			channel='chrome',
 			args=args[self.config.browser_class],
 			proxy=self.config.proxy.model_dump() if self.config.proxy else None,
 			handle_sigterm=False,
-			handle_sigint=False,
+			handle_sigint=False
 		)
 		return browser
 
--- a/browser_use/browser/browser.py
+++ b/browser_use/browser/browser.py
@@ -27,7 +27,7 @@ from browser_use.browser.chrome import (
 	CHROME_HEADLESS_ARGS,
 )
 from browser_use.browser.context import BrowserContext, BrowserContextConfig
-from browser_use.browser.utils.screen_resolution import get_screen_resolution, get_window_adjustments
+from browser_use.browser.utils.screen_resolution import get_window_adjustments, get_screen_resolution
 from browser_use.utils import time_execution_async
 
 logger = logging.getLogger(__name__)
@@ -251,7 +251,15 @@ class Browser:
 		"""Sets up and returns a Playwright Browser instance with anti-detection measures."""
 		assert self.config.browser_binary_path is None, 'browser_binary_path should be None if trying to use the builtin browsers'
 
-		if self.config.headless:
+		# Use the configured window size from new_context_config if available
+		if (
+			not self.config.headless
+			and hasattr(self.config, 'new_context_config')
+			and hasattr(self.config.new_context_config, 'browser_window_size')
+		):
+			screen_size = self.config.new_context_config.browser_window_size.model_dump()
+			offset_x, offset_y = get_window_adjustments()
+		elif self.config.headless:
 			screen_size = {'width': 1920, 'height': 1080}
 			offset_x, offset_y = 0, 0
 		else:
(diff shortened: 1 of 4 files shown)


browser_use/browser/context.py
```diff
@@ -43,6 +43,31 @@
...
+# Constants
+# Default height adjustment for browser chrome (title bar, tabs, etc.)
+# This can vary by platform, browser, and display settings
+BROWSER_CHROME_HEIGHT_DEFAULT = 85
+BROWSER_CHROME_HEIGHT_WINDOWS = 85
+BROWSER_CHROME_HEIGHT_MACOS = 80
+BROWSER_CHROME_HEIGHT_LINUX = 90
+
+def get_browser_chrome_height() -> int:
+    """Determine appropriate browser chrome height based on platform"""
```
pirate, line 55:
```suggestion
def get_browser_menubar_height() -> int:
    """Determine how much browser window height used up by the navbar, titlebar, and window frame based on platform"""
```

browser_use/browser/context.py
```diff
@@ -43,6 +43,31 @@
 
 logger = logging.getLogger(__name__)
 
+# Constants
+# Default height adjustment for browser chrome (title bar, tabs, etc.)
+# This can vary by platform, browser, and display settings
+BROWSER_CHROME_HEIGHT_DEFAULT = 85
+BROWSER_CHROME_HEIGHT_WINDOWS = 85
+BROWSER_CHROME_HEIGHT_MACOS = 80
+BROWSER_CHROME_HEIGHT_LINUX = 90
```
pirate, line 52:
```suggestion
BROWSER_MENUBAR_HEIGHT_DEFAULT = 85
BROWSER_MENUBAR_HEIGHT_WINDOWS = 85
BROWSER_MENUBAR_HEIGHT_MACOS = 80
BROWSER_MENUBAR_HEIGHT_LINUX = 90
```

browser_use/browser/context.py
```diff
@@ -442,14 +474,22 @@ async def _create_context(self, browser: PlaywrightBrowser):
...
+				await self._set_viewport_size_for_page(context.pages[0])
 		else:
 			kwargs = {}
-			if self.browser.config.headless:
-				kwargs['viewport'] = self.config.browser_window_size.model_dump()
-				kwargs['no_viewport'] = False
+			# Set viewport for both headless and non-headless modes
+			kwargs['viewport'] = self.config.browser_window_size.model_dump()
+			# Important: set no_viewport to False to ensure the viewport size is applied
+			kwargs['no_viewport'] = False
```
pirate, line 491: viewport size should not be set in headful mode, it should default to the window size

browser_use/browser/browser.py
```diff
@@ -313,7 +321,7 @@ async def _setup_builtin_browser(self, playwright: Playwright) -> PlaywrightBrow
 			args=args[self.config.browser_class],
 			proxy=self.config.proxy.model_dump() if self.config.proxy else None,
 			handle_sigterm=False,
-			handle_sigint=False,
+			handle_sigint=False
```
pirate, line 324: leave the trailing comma please

browser_use/browser/context.py
```diff
@@ -43,6 +43,31 @@
...
+        return BROWSER_CHROME_HEIGHT_MACOS
+    elif system == 'windows':
+        return BROWSER_CHROME_HEIGHT_WINDOWS
+    elif system == 'linux':
+        return BROWSER_CHROME_HEIGHT_LINUX
+    else:
+        return BROWSER_CHROME_HEIGHT_DEFAULT
+
+# Get the appropriate chrome height for current platform
+BROWSER_CHROME_HEIGHT = get_browser_chrome_height()
```
pirate, line 69:
```suggestion
BROWSER_NAVBAR_HEIGHT = {
	'windows': 85,
	'macos': 80,
	'linux': 90,
}.get(platform.system().lower(), 85)
```

browser_use/browser/context.py
```diff
@@ -442,14 +474,22 @@ async def _create_context(self, browser: PlaywrightBrowser):
 		"""Creates a new browser context with anti-detection measures and loads cookies if available."""
 		if self.browser.config.cdp_url and len(browser.contexts) > 0 and not self.config.force_new_context:
 			context = browser.contexts[0]
+			# For existing contexts, we need to set the viewport size manually
+			if context.pages and not self.browser.config.headless:
+				await self._set_viewport_size_for_page(context.pages[0])
```
pirate, line 479:
```suggestion
				for page in context.pages:
					try:
						await self._set_viewport_size_for_page(page)
					except Exception as e:
						logger.debug(f'Failed to set viewport size for page {page}: {e}')
```

browser_use/browser/context.py
```diff
@@ -442,14 +474,22 @@ async def _create_context(self, browser: PlaywrightBrowser):
...
 		elif self.browser.config.browser_binary_path and len(browser.contexts) > 0 and not self.config.force_new_context:
 			# Connect to existing Chrome instance instead of creating new one
 			context = browser.contexts[0]
+			# For existing contexts, we need to set the viewport size manually
+			if context.pages and not self.browser.config.headless:
+				await self._set_viewport_size_for_page(context.pages[0])
```
pirate, line 485:
```suggestion
			for page in context.pages:
				await self._set_viewport_size_for_page(page)
```

pirate:
argh this is getting a bit complicated for my liking but I guess it's unavoidable with the combinations of different options possible.

in an ideal world we would:

- dont change behavior based on headless=True or not, just always set window and viewport size
- remove all the code that handles setting initial window and viewport size on launch
- move everything into `get_current_page()` or some other function guaranteed to be called during any step, just check if the viewport and window size are correct, and if not, change them there using the same code for headless/non-headless

sandeepsalwan1:
> argh this is getting a bit complicated for my liking but I guess it's unavoidable with the combinations of different options possible.
> 
> in an ideal world we would:
> 
> * dont change behavior based on headless=True or not, just always set window and viewport size
> * remove all the code that handles setting initial window and viewport size on launch
> * move everything into `get_current_page()` or some other function guaranteed to be called during any step, just check if the viewport and window size are correct, and if not, change them there using the same code for headless/non-headless

Thank you for your feedback! I will make sure to use this info in future prs.

pirate:
changed this behavior here: https://github.com/browser-use/browser-use/pull/1557

PR #1474, opened 2025-04-25
feat(browser): option to control window focusing

andrewgazelka:
Add new configuration option that determines whether browser windows/tabs are automatically brought to front when interacting with them. Default is set to true for backward compatibility with existing behavior.

This gives users more control over browser behavior, particularly useful in when running browser automation tasks in the background where focus stealing might be undesirable.

Should be tested more before merging.
    

---

## Summary by mrge
Added a new setting to control whether browser windows or tabs are brought to the front when used. The default is true, but users can now turn this off to prevent focus stealing during automation.

### DIFF
--- a/browser_use/browser/dolphin_service.py
+++ b/browser_use/browser/dolphin_service.py
@@ -98,7 +98,6 @@ class DolphinBrowser(Browser):
 
 		# Set the current page to the selected tab
 		self.page = self._pages[page_id]
-		await self.page.bring_to_front()  # Bring tab to the front
 		await self.wait_for_page_load()
 
 	async def get_tabs_info(self) -> list[TabInfo]:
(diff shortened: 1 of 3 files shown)


browser_use/browser/dolphin_service.py
```diff
@@ -98,7 +98,6 @@ async def switch_to_tab(self, page_id: int) -> None:
 
 		# Set the current page to the selected tab
 		self.page = self._pages[page_id]
-		await self.page.bring_to_front()  # Bring tab to the front
```
andrewgazelka, line 101: unless I am missing something this is already called by `BrowserContext` 

https://github.com/browser-use/browser-use/blob/2948fc4889e38d0eb9f2a2c0e68cc3c9cab7e430/browser_use/browser/context.py#L404

pirate:
Wont work, foregrounding is actually mostly a result of the screenshotting code, not the `bring_to_front` calls. We're working on improving this already, a full solution for being able to run headful tasks in a backgrounded browser is going to be much more involved than just removing those calls :+1:

PR #1480, opened 2025-04-26
fix status code when checking if chrome browser is already running

prenit-wankhede:
This PR fixes connecting to already running chrome instance or when using user provided chrome binary path.

We check if browser is already running using below code

```
async with httpx.AsyncClient() as client:
				response = await client.get('http://localhost:9222/json/version', timeout=2)
				if response.status == 200:
					logger.info('🔌  Reusing existing browser found running on http://localhost:9222')
					browser_class = getattr(playwright, self.config.browser_class)

```
`response` object is an instance of `Response` class in `httpx` library. `response` object has `status_code` property and not `status`. This PR fixes this issue.
    

---

## Summary by mrge
Fixed a bug where checking if Chrome is already running used the wrong status code property, which could cause connection issues.

- **Bug Fixes**
  - Replaced response.status with response.status_code when verifying the running Chrome instance.

### DIFF
--- a/browser_use/browser/browser.py
+++ b/browser_use/browser/browser.py
@@ -190,7 +190,7 @@ class Browser:
 			# Check if browser is already running
 			async with httpx.AsyncClient() as client:
 				response = await client.get('http://localhost:9222/json/version', timeout=2)
-				if response.status == 200:
+				if response.status_code == 200:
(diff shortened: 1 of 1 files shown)


pirate:
Thanks!

PR #1483, opened 2025-04-26
Add handling of interactive elements inside html slot tags

enrique-specht:
This PR adds the functionality that interactive elements inside a html slot tag are taken into account to be marked as interactive and not ignored. Similar to the handling of iframes and shadow DOMs.


pirate:
Can you provide some example URLs and screenrecordings showing this change in action?

Issue #1476, opened 2025-04-25
Once the page loads -- not able to take any action


akshathsk:
### Bug Description

I am trying with multiple websites and it loads the page but once it loads the "action" is empty and it is unable to follow the prompt.

INFO     [agent] 👍 Eval: Success - I have successfully filled in the username and password fields with valid user data.
INFO     [agent] 🧠 Memory: Current URL is 'https://parabank.parasoft.com/parabank/index.htm'. I have filled in the username as 'testuser' and the password as 'password123'. Current step is 24/30.
INFO     [agent] 🎯 Next goal: Submit the registration form.
INFO     [agent] 🛠️  Action 1/1: {}
INFO     [agent] 📍 Step 24
INFO     [agent] 👍 Eval: Success - I have successfully filled in the username and password fields with valid user data and am ready to submit the registration form.
INFO     [agent] 🧠 Memory: Current URL is 'https://parabank.parasoft.com/parabank/index.htm'. I have filled in the username as 'testuser' and the password as 'password123'. Current step is 25/30.
INFO     [agent] 🎯 Next goal: Submit the registration form.
INFO     [agent] 🛠️  Action 1/1: {}
INFO     [agent] 📍 Step 25
INFO     [agent] 👍 Eval: Success - I have successfully filled in the username and password fields with valid user data and am ready to submit the registration form.
INFO     [agent] 🧠 Memory: Current URL is 'https://parabank.parasoft.com/parabank/index.htm'. I have filled in the username as 'testuser' and the password as 'password123'. Current step is 26/30.
INFO     [agent] 🎯 Next goal: Submit the registration form.
INFO     [agent] 🛠️  Action 1/1: {}
INFO     [agent] 📍 Step 26
INFO     [agent] 👍 Eval: Success - I have successfully filled in the username and password fields with valid user data and am ready to submit the registration form.
INFO     [agent] 🧠 Memory: Current URL is 'https://parabank.parasoft.com/parabank/index.htm'. I have filled in the username as 'testuser' and the password as 'password123'. Current step is 27/30.
INFO     [agent] 🎯 Next goal: Submit the registration form.
INFO     [agent] 🛠️  Action 1/1: {}

### Reproduction Steps

Prompt - Go to https://parabank.parasoft.com/parabank/index.htm Click on register, fill the form, then login

### Code Sample

```python
N/A
```

### Version

main branch

### LLM Model

GPT-4o

### Operating System

Azure cloud - k8s

### Relevant Log Output

[...]

PR #1485, opened 2025-04-26
 Fix: "Once the page loads -- not able to take any action" (#1476)

Shyanil:
### Summary
This PR fixes the issue where, after loading, **no action was taken** if the model returned an empty or invalid response.  
We now **retry once with a clarification prompt**, and if the model still does not respond correctly, we **safely insert a default "no-op" action** to prevent the system from hanging.

---

### Code Explanation

```python
model_output = await self.get_next_action(input_messages)

# Handle empty action
if not model_output.action or all(action.model_dump() == {} for action in model_output.action):
    logger.warning("Model returned empty action. Retrying once with clarification prompt...")

    # Add a clarification message
    clarification_message = HumanMessage(
        content="You forgot to return an action. Please respond only with a valid JSON action according to the expected format."
    )
    retry_messages = input_messages + [clarification_message]

    # Retry getting next action with clarification
    model_output = await self.get_next_action(retry_messages)

    if not model_output.action or all(action.model_dump() == {} for action in model_output.action):
        logger.warning("Model still returned empty after retry. Inserting safe noop action.")
        
        # Insert a safe "do nothing" action to prevent crash
        model_output.action = [self.ActionModel(done={"success": False, "text": "No action returned, safe exit."})]
```

---

### How it works:
-  **First try:** Call the model normally.
- **If empty:** Retry once, politely asking the model to **return a valid JSON action**.
-  **If still empty:** Insert a **safe fallback action** (`success: False, safe exit`) to **gracefully handle** the failure without breaking the system.

---

### Why this is needed:
Without this fix, when the model returns no actions, the app **hangs** and **users cannot continue**.  
This approach ensures **reliability** and **improves the user experience** even if the model behaves unexpectedly.

---

    

---

## Summary by mrge
Fixed an issue where the app would hang if the model returned an empty or invalid action by retrying once with a clarification prompt and then inserting a safe fallback action if needed.

- **Bug Fixes**
  - Retries once with a clarification prompt if the model returns no action.
  - Inserts a default "no-op" action to prevent the system from hanging if the retry also fails.

### DIFF
--- a/browser_use/agent/service.py
+++ b/browser_use/agent/service.py
@@ -62,7 +62,9 @@ from browser_use.utils import check_env_variables, time_execution_async, time_ex
 load_dotenv()
 logger = logging.getLogger(__name__)
 
-SKIP_LLM_API_KEY_VERIFICATION = os.environ.get('SKIP_LLM_API_KEY_VERIFICATION', 'false').lower()[0] in 'ty1'
+SKIP_LLM_API_KEY_VERIFICATION = os.environ.get(
+    'SKIP_LLM_API_KEY_VERIFICATION',
+     'false').lower()[0] in 'ty1'
 
 
 def log_response(response: AgentOutput) -> None:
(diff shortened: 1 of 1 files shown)


pirate:
This seems like a good improvement. Can you make sure to sign the CLA and run the pre-commit checks?

`pre-commit run --all-files`

Shyanil:
@pirate  Sure i will do that soon

deveshruttala-edu:

Issue #221, opened 2025-01-11
[Feature Request] human-in-the-loop 


MagMueller:
# Add human-in-the-loop 

## Description
Add capability to pause agent execution and allow human intervention. This would enable hybrid automation where humans can assist when needed.

### 1. Interactive Pause & Resume:
   - Add pause functionality in agent execution

### 2. New prompt during pause:
   - Human should be able to give a new task / redifine the task during a pause

### 3. Human verification mode:
   - Before the agent executes the action a human has to verify with e.g. `enter` that it wants to continue with that action

### 4. Interactive Browser Control:
   - Allow human to take control of browser during pauses

aflansburg:
I believe `humanlayer` has some support for Langchain as a tool - might be worth a peek.

I'm on holiday, but would be interested in looking-at/working-on this.

[Github project](https://www.humanlayer.dev/docs/frameworks/langchain)

[Relevant Docs](https://www.humanlayer.dev/docs/frameworks/langchain)

Cfomodz:
I think this is a great idea. I know when running crew I make sure to *always* include an option for any agent to ask me for help if it doesn't know what else to do, or does, but needs more info (or thinks it needs more info).

You could even have intervention be a part of the 3 (or whatever number given) failed attempts error out sequence where, if enabled, human intervention is triggered.

Then from there go into (m)anual mode to edit the current task, (r)eturn to kill that step and ask for a new plan, basically, (s)pecify a new model for that step or for n future steps or until the task is complete, or (a)bort the current task [continuing with the next].

Also is there some reason this isn't already possible using [this](https://docs.browser-use.com/customize/custom-functions#basic-function-registration)?

galatian44to7:
This is a great idea!!!

xitex:
sorry if i misunderstood the concept but it seems this can be done by custom function mentioned in docs, right?
like this:
`from browser_use.controller.service import Controller

# Initialize the controller
controller = Controller()

@controller.action('Ask user for information')
def ask_human(question: str, display_question: bool) -> str:
    return input(f'\n{question}\nInput: ')`

I'm new on github and in python but if I can be useful, I will be glad to realize small tasks.
Thank you guy's project is amazing!

xitex:
> I think this is a great idea. I know when running crew I make sure to _always_ include an option for any agent to ask me for help if it doesn't know what else to do, or does, but needs more info (or thinks it needs more info).
> 
[...]

PR #1486, opened 2025-04-26
Added an action to request human help

minamaged095:
This feature is to allow the agent to request human help when needed.

Example of usage in a task: 
"""
  If you are stuck request human help, if you have too many failures instead of stopping request human help 
  after getting human assistance continue the task
"""

This feature is requested here as well by several memebers: https://github.com/browser-use/browser-use/issues/221 
    

---

## Summary by mrge
Added a new action that lets the agent request human help by showing a popup to the user and waiting for their response.

- **New Features**
  - Displays a popup with a message and instructions when human help is needed.
  - Waits for the user to mark the task as completed or cancelled before continuing.


browser_use/controller/service.py
```diff
@@ -77,6 +77,90 @@ async def done(params: ExtendedOutputModel):
...
+		                  bg="green", fg="white", padx=10, pady=5).pack(side=tk.LEFT, padx=10)
+		        tk.Button(button_frame, text="Cancel", command=on_cancel,
+		                  bg="red", fg="white", padx=10, pady=5).pack(side=tk.LEFT, padx=10)
+		        
+		        # Start the main loop
+		        popup.mainloop()
+		    
+		    # Launch popup in a separate thread
+		    popup_thread = threading.Thread(target=show_popup)
+		    popup_thread.daemon = True
```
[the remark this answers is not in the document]
minamaged095: It is meant to remain open for the user to see

browser_use/controller/service.py
```diff
@@ -77,6 +77,90 @@ async def done(params: ExtendedOutputModel):
...
+		    popup_thread.start()
+		    
+		    # Wait for user to respond
+		    while not user_responded:
```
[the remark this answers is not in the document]
minamaged095: changed and tested

browser_use/controller/service.py
```diff
@@ -77,6 +77,90 @@ async def done(params: ExtendedOutputModel):
...
+		        message (str): The main message explaining what help is needed
+		        instructions (str, optional): Additional instructions for the user
+		    
+		    Returns:
+		        str: The response from the user (either "completed" or "cancelled")
+		    """
+		    global user_responded, user_response
+		    # message = "dfd"
+		    # instructions = "asdwsdws"
+		    # Reset state
```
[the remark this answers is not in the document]
minamaged095: I checked and these variables are not used elsewhere

browser_use/controller/service.py
```diff
@@ -77,6 +77,90 @@ async def done(params: ExtendedOutputModel):
...
+		    user_responded = False
+		    user_response = None
+		    
+		    # Function to create and show the popup in a separate thread
+		    def show_popup():
+		        # Create the main window
+		        popup = tk.Tk()
```
nnWhisperer, line 104: why do you use tk? When I searched for tk or tkinter, there isn't any in this repo, am I mistaken?
there is no import tk as statement either, hence, have you checked that this pull request works without errors?
To display the popup, or any GUI in general, it must be using the existing/imported libraries and integrated with existing code structure.

minamaged095: I was using a different repo/fork that was unorganized as I did not expect to work as much on this and want to contribute
, It had one branch with many features so it made sense to create a new repo and start clean, the code was fully tested on the old repo, 

Thanks for pointing out that I need to test on the current repo as well, I tested it with different prompts after adding the necessary imports and it is fully functioning, Kindly find the screenshot below taken after adding the necessary imports (all on the new repo and branch of course). 

image


the agent task was:
"go to https://narmertech.com and clcik on book a call, book a call for 3rd or april 2022
If you are stuck request human help if you have too many failures instead of stopping request human help 
after getting human assistance continue" 

which of course is expected to not find the date and ask for human help

minamaged095: image

minamaged095:
Since tkinter is better installed on the machine and not with pip, I switched to easygui and below is a screenshot after testing for the same task

image

pirate:
For a few reasons I'm not going to merge this right now:

- separation of concerns: we already have many other dedicated front ends that run on top of browser use. I don't want to add a gui within the library itself, the UI part would be better suited in GitHub.com/browser-use/web-ui
- it provides an easy escape for the agent to give up and stop trying workarounds, we want the user to more explicitly opt in to offering help during tasks rather than have it as an action available by default
- we want the agent to be able to do everything itself. Whatever you think you need human assistance for, open an issue for that and we'll work on addding a feature to directly fill that need.

pirate:
I'm ok to merge this if you remove the GUI part entirely and make it just trigger a Python callback instead + put it behind an agent config flag so that the "give up and ask the human" option isn't available by default. 

Take a look at the on_step_start and on_step_end callbacks to see how we've handled pausing and resuming so far.

minamaged095:
@pirate Thanks for taking teh time to review.

It makes sense, if the UI is to be used then the library shouldn't open pop-ups from its side. I can add the allow_yield_to_human and will look into the callbacks you mentioned.

minamaged095:
> I'm ok to merge this if you remove the GUI part entirely and make it just trigger a Python callback instead + put it behind an agent config flag like allow_yield_to_human so that the "give up and ask the human" option isn't available by default.
> 
> Take a look at the on_step_start and on_step_end callbacks to see how we've handled pausing and resuming so far.

@pirate  Thanks for taking the time to review.

It makes sense, if the UI is to be used then the library shouldn't open pop-ups from its side. I can add the allow_yield_to_human and will look into the callbacks you mentioned.

minamaged095:
@pirate 

I removed GUI and now the user needs to use the terminal, also used an agent config flag "allow_yield_to_human" so that the "give up and ask the human" option isn't available by default. 

example usage: 

agent = Agent(task=task, llm=llm, allow_yield_to_human=True)

if allow_yield_to_human was not specified, then the agent can't use "request_human_help"

minamaged095:
Hi @pirate 

After a fix the workflow is running successfully on the fork main branch, I think the lint will probably finish successfully this time

browser_use/agent/service.py
```diff
@@ -99,9 +99,10 @@ def __init__(
 		task: str,
 		llm: BaseChatModel,
 		# Optional parameters
+		allow_yield_to_human: bool = False,
 		browser: Browser | None = None,
 		browser_context: BrowserContext | None = None,
-		controller: Controller[Context] = Controller(),
+		controller: Controller[Context] = None,  # Make this None by default
```
pirate, line 105:
```suggestion
		controller: Controller[Context] | None = None,  # Make this None by default
```

browser_use/agent/service.py
```diff
@@ -164,7 +165,17 @@ def __init__(
 		# Core components
 		self.task = task
 		self.llm = llm
-		self.controller = controller
+
+		self.allow_yield_to_human = allow_yield_to_human
```
pirate, line 169: this should probably be stored in `AgentSettings`, not just a loose attr on Agent

browser_use/agent/service.py
```diff
@@ -191,6 +202,9 @@ def __init__(
 			extend_planner_system_message=extend_planner_system_message,
 		)
 
+		# to allow human in loop
+		self.controller.agent = self  # Give the controller a reference to the agent
```
pirate, line 206: this is a very bad idea, we don't want to introduce circular references between the child and parent class

browser_use/controller/service.py
```diff
@@ -77,6 +79,30 @@ async def done(params: ExtendedOutputModel):
 			async def done(params: DoneAction):
 				return ActionResult(is_done=True, success=params.success, extracted_content=params.text)
 
+		@self.registry.action('Request human help')
```
pirate, line 82:
```suggestion
		@self.registry.action('Request human help', agent_filter=lambda agent: agent.settings.allow_yield_to_human)
```

browser_use/controller/service.py
```diff
@@ -48,6 +48,8 @@ def __init__(
 
 		"""Register all default browser actions"""
 
+		# Initialize these attributes (they'll be set by Agent later)
+		self.allow_yield_to_human = False
```
pirate, line 52:
```suggestion
```

browser_use/controller/service.py
```diff
@@ -77,6 +79,30 @@ async def done(params: ExtendedOutputModel):
...
 				return ActionResult(is_done=True, success=params.success, extracted_content=params.text)
 
+		@self.registry.action('Request human help')
+		async def request_human_help(message: str, instructions: str):
+			"""
+			Yield to the embedder over a simple Python callback.
+			"""
+			# Access the agent through the reference
+			if not hasattr(self, 'agent') or not self.agent.allow_yield_to_human:
+				return 'Human help is not enabled.'
```
pirate, line 89:
```suggestion
```

pirate:
I think as it stands this design doesn't provide enough benefit over just adding an example in our documentation showing how users can do this using existing mechanisms:

```python
controller = Controller()

@controller.action('Ask user for information or assistance')
def ask_for_human_assistance(prompt: str) -> str:
	print(f'\n✋ The AI Agent is requesting assistance: {prompt}')
	print('\t(press [Enter] when ready to continue)')
	human_response = input(' > ')
	
	return ActionResult(output=human_response.strip(), save_in_memory=True)
```

^ this style has several benefits:

- it's easier for library users to customize how they want to prompt the user (e.g. they could replace it with a call to twillio to SMS the user and wait for a reply!)
- it doesn't require any extra config options
- it doesn't require adding a circular dependency between the action registry and `Agent` class (I really want to avoid this, even if it's dependency-injected it's not great)

If we want to build it into the library we could do:

`browser_use/controller/service.py`:
```diff
class Controller:

+    def ask_for_human_assistance(prompt: str) -> str:
+		print(f'\n✋ The AI Agent is requesting assistance: {prompt}')
+		print('\t(press [Enter] when ready to continue)')
+		human_response = input(' > ')		
+		return ActionResult(output=human_response.strip(), save_in_memory=True)

+	 if os.getenv('ALLOW_REQUESTING_HUMAN_ASSISTANCE', 'false').lower()[0] in 'ty1':
+		 controller.action('Ask user for information or assistance')(ask_for_human_assistance)
```

pirate:
Clsoing this for now, but if you want to open a PR to add something in `examples/` I'd be happy to review it.

PR #1482, opened 2025-04-26
Alllow providing chrome_remote_debugging_port in browser config. Fix launching user provided chrome browser.

prenit-wankhede:
This PR add a feature to provide custom chrome remote debugging port in `BrowserConfig` class. 

This allows running multiple agents having chrome browsers with same browser_binary_path but running on different ports.

Also, this makes it possible to launch new user provided chrome browser without having to close already opened chrome instances, by providing non-default chrome debugging port.

This PR also fixed launching chrome using user provided chrome binary path. 

Earlier, chrome with user provided chrome binary path was launched with this code below:
```
self._chrome_subprocess = psutil.Process(
	await asyncio.create_subprocess_shell(
		chrome_launch_cmd,
		stdout=subprocess.DEVNULL,
		stderr=subprocess.DEVNULL,
		shell=False,
	).pid
) 
``` 

which was raising exceptions. Also `playwright` library itself launches its browsers using `asyncio.create_subprocess_exec`  as shown in the code below from the library:
```

    async def connect(self) -> None:
        self._stopped_future: asyncio.Future = asyncio.Future()
        try:
            env = get_driver_env()
            if getattr(sys, "frozen", False) or globals().get("__compiled__"):
                env.setdefault("PLAYWRIGHT_BROWSERS_PATH", "0")
            startupinfo = None
            if sys.platform == "win32":
                startupinfo = subprocess.STARTUPINFO()
                startupinfo.dwFlags |= subprocess.STARTF_USESHOWWINDOW
                startupinfo.wShowWindow = subprocess.SW_HIDE
            executable_path, entrypoint_path = compute_driver_executable()
            self._proc = await asyncio.create_subprocess_exec(
                executable_path,
                entrypoint_path,
                "run-driver",
                stdin=asyncio.subprocess.PIPE,
                stdout=asyncio.subprocess.PIPE,
                stderr=_get_stderr_fileno(),
                limit=32768,
                env=env,
                startupinfo=startupinfo,
            ) 
``` 

So it is natural to use the same approach in this library as well. 
    

---

## Summary by mrge
Added support for setting a custom Chrome remote debugging port in browser config, allowing multiple Chrome instances and fixing issues with launching user-provided Chrome browsers.

- **New Features**
  - Added `chrome_remote_debugging_port` option to `BrowserConfig`.
  - Allows running multiple Chrome browsers on different ports.
  - Supports launching user-provided Chrome without closing existing instances.

- **Bug Fixes**
[...]

PR #1488, opened 2025-04-26
Fix: Correct subprocess creation using asyncio.create_subprocess_exec

PawiX25:
When using `browser_binary_path` to connect to a user-provided browser instance, the code incorrectly attempted to access the `.pid` attribute of the coroutine object returned by `asyncio.create_subprocess_shell` before it was awaited. This caused a `RuntimeWarning` and an `AttributeError: 'coroutine' object has no attribute 'pid'`.

A subsequent attempt to fix this by changing to `shell=True` and joining the command arguments into a string resulted in an `ECONNREFUSED` error, likely because the browser process failed to start correctly with the debugging port enabled.

## Solution

This PR fixes the subprocess creation by:

1.  Using `asyncio.create_subprocess_exec` instead of `asyncio.create_subprocess_shell`. `create_subprocess_exec` is the appropriate function for launching an executable directly with a list of arguments (equivalent to `shell=False`).
2.  Ensuring that `await` is correctly used before accessing the `.pid` attribute of the returned process object.

This resolves the `RuntimeWarning`, `AttributeError`, and `ECONNREFUSED` errors, allowing the agent to correctly launch and connect to a user-provided browser instance.
    

---

## Summary by mrge
Fixed subprocess creation for user-provided browser instances by using asyncio.create_subprocess_exec and awaiting the process before accessing its pid. This prevents errors and allows the agent to launch and connect to the browser correctly.

### DIFF
--- a/browser_use/browser/browser.py
+++ b/browser_use/browser/browser.py
@@ -213,14 +213,13 @@ class Browser:
 				*self.config.extra_browser_args,
 			},
 		]
-		self._chrome_subprocess = psutil.Process(
-			await asyncio.create_subprocess_shell(
-				chrome_launch_cmd,
-				stdout=subprocess.DEVNULL,
-				stderr=subprocess.DEVNULL,
-				shell=False,
-			).pid
+		process = await asyncio.create_subprocess_exec(
+			chrome_launch_cmd[0],
+			*chrome_launch_cmd[1:],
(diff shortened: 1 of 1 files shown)


pirate:
This is fixes as part of: https://github.com/browser-use/browser-use/pull/1482 which is getting merged first.

Good catch though! Great minds think alike :wink:

Issue #1433, opened 2025-04-21
The agent recognizes the file tree as a common element (and unable to click internal elements)


prt0117:
### Bug Description

The previous version 0.1.40 recognized clickable elements inside the file tree. Screenshot 0.1.40:
![Image](https://github.com/user-attachments/assets/827cb873-2522-444b-9a7e-203fce0cde2d)

Files are downloadable with success rate of ~60%.

Version 0.1.41 recognizes the file tree as a common element. Screenshot 0.1.41:
![Image](https://github.com/user-attachments/assets/51f556ff-b6e7-4ae0-8a3b-a5a4096da31c)

And unable to download files. Also reports that files were downloaded.

### Reproduction Steps

1. Install Browser Use 0.1.41.
2. Run agent (see code sample).
3. The actual result: none of the files in file tree were downloaded
4. The required result: all files in file tree have been downloaded.

### Code Sample

```python
import sys
from langchain_openai import ChatOpenAI
from browser_use import Agent
from browser_use import BrowserConfig
from browser_use import Browser
from browser_use import BrowserContextConfig
from dotenv import load_dotenv
from pprint import pprint
import os
load_dotenv()

import asyncio

llm = ChatOpenAI(model="gpt-4o")

async def main():
    # Basic configuration
    config = BrowserConfig(
        headless=False,
        new_context_config=BrowserContextConfig(
            save_downloads_path=os.path.join(os.path.expanduser('~'), 'browser-use-test-rc', 'downloads')
        )
    )

    browser = Browser(config=config)

    agent = Agent(
        browser=browser,
        # sensitive_data=sensitive_data,
        task="""
Find and download all files (pdf, images):
a) First open "Document" section.
b) Then download files listed in the left sidebar, on each file: click on file, then file's page will open, then click download button, then proceed to next file in the left sidebar.
""",
        llm=llm,
    )
    result = await agent.run()
    print(result)

if __name__ == "__main__":
    asyncio.run(main())
```

### Version

0.1.41

### LLM Model

GPT-4o

### Operating System

Ubuntu 22.04

### Relevant Log Output

```shell
INFO     [agent] 🎯 Next goal: Open the 'Document' section to begin downloading files if available.
INFO     [agent] 🛠️  Action 1/1: {"click_element_by_index":{"index":16}}
INFO     [controller] 🖱️  Clicked button with index 16: 
INFO     [agent] 📍 Step 28
[...]

PR #1502, opened 2025-04-28
fix: add cursor:pointer handling in buildDomTree and update test URLs to handle expander icons

satya-nutella:
## Summary by mrge
Elements with cursor:pointer are now treated as distinct in buildDomTree, improving detection of clickable icons. Test URLs were updated to include pages with expander icons.

Fixes https://github.com/browser-use/browser-use/issues/1433


pirate:
did you already sign the CLA?

PR #1533, opened 2025-05-01
Fix Gemini 500 Error Handling: Prevent Indefinite Agent.run() Hanging

### LINKED ISSUE #1497: LLM API error and then Agent.run() hangs
### Bug Description

I randomly get this following error and then the agent run will never finish.

WARNING:langchain_google_genai.chat_models:Retrying langchain_google_genai.chat_models._achat_with_retry.<locals>._achat_with_retry in 2.0 seconds as it raised InternalServerError: 500 An internal error has occurred. Please retry or report in https://developers.generativeai.google/guide/troubleshooting.
ERROR:browser_use.agent.service:Failed to invoke model: 500 An internal error has occurred. Please retry or report in https://developers.generativeai.google/guide/troubleshooting
ERROR:browser_use.agent.service:❌ Result failed 1/3 times:
 Error 401: LLM API call failed
INFO:browser_use.agent.service:📍 Step 20
WARNING:browser_use.browser.context:⚠️  No browser window available, opening a new window
WARNING:browser_use.browser.context:⚠️  Page load failed, continuing...
WARNING:browser_use.browser.context:⚠️  No browser window available, opening a new window

### Reproduction Steps

call Agent.run with gemini 2.5 pro as LLM

### Code Sample

```python
agent = Agent(
            task=task_template.format(task=task),
            llm=llm,
            controller=controller,
            browser_context=browser_context,
            browser=browser
        )
history = await agent.run(on_step_start=hook, max_steps=50)
                result = history.final_result()
```

### Version

0.1.41

### LLM Model

### Operating System

macos 15.4

### Relevant Log Output

```shell

```

pyoneerC:
this pr fixes an issue where agent.run() would hang forever after getting a 500 error from google’s gemini api.

fixes #1497 

the problem
when gemini returns a 500 internal server error, the agent doesn’t know it’s unrecoverable. it just logs the error, keeps incrementing the failure counter, and gets stuck in an endless loop trying to recover the browser window. basically, it never gives up, even when it should.

what changed
added a retriable flag to llmexception so we can tell which errors are worth retrying and which aren’t

gemini’s “500 an internal error has occurred” is now flagged as non-retriable

if the agent hits a non-retriable error, we immediately max out the failure counter so it shuts down gracefully

added clear messaging so it’s obvious why the agent stopped

stops wasting resources trying to recover from something that’s not fixable

this improves the user experience without changing how retriable errors are handled.

this is a focused fix that only affects error handling, normal agent behavior stays the same.
    

---

## Summary by mrge
Fixed an issue where the agent would hang forever after a Gemini API 500 error by marking these errors as non-retriable and shutting down the agent gracefully.

- **Bug Fixes**
  - Added a retriable flag to LLMException to distinguish between retriable and non-retriable errors.
  - Gemini 500 errors are now treated as non-retriable, causing the agent to stop instead of looping.
  - Improved error messages to make shutdown reasons clear.

### DIFF
--- a/browser_use/agent/service.py
+++ b/browser_use/agent/service.py
@@ -501,7 +501,6 @@ class Agent(Generic[Context]):
 						model_output.action = [action_instance]
 
 				# Check again for paused/stopped state after getting model output
-				# This is needed in case Ctrl+C was pressed during the get_next_action call
 				await self._raise_if_stopped_or_paused()
 
 				self.state.n_steps += 1
@@ -595,6 +594,13 @@ class Agent(Generic[Context]):
 		if 'Browser closed' in error_msg:
 			logger.error('❌  Browser is closed or disconnected, unable to proceed')
 			return [ActionResult(error='Browser closed or disconnected, unable to proceed', include_in_memory=False)]
+			
+		# Check for non-retriable LLM exceptions (like Gemini 500 errors)
+		if isinstance(error, LLMException) and not getattr(error, 'retriable', True):
+			logger.error(f'{prefix}Non-retriable LLM error: {error_msg}')
+			# Force consecutive failures to max to terminate the agent
+			self.state.consecutive_failures = self.settings.max_failures
+			return [ActionResult(error=f'Non-retriable LLM error: {error_msg}', include_in_memory=True)]
 
 		if isinstance(error, (ValidationError, ValueError)):
 			logger.error(f'{prefix}{error_msg}')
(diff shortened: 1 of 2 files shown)


pyoneerC:
https://github.com/browser-use/browser-use/issues/1497

pirate:
What is the underlying cause of the 500 error?

Do we know for a fact that *all* 500 errors from google are non-retriable? 500 is pretty generic I feel like it could cover a wide array of things.

pyoneerC:
good point. 500 is broad, but this targets gemini exact "500 an internal error has occurred" message. in testing, it never recovers, even after retries. looks like a hard backend fail. if they give better signals later, we can adjust. open to better ways of flagging it if you’ve got one.

pirate:
I think I would prefer to make the agent not crash from any 500 rather than catch this error specifically.

e.g. after N number of failed LLM API calls just exit the agent regarless of what the error is

deveshruttala-edu:

PR #1537, opened 2025-05-02
enable include_in_memory by default, standardize temperature=0.0

Alezander9:
## Summary by mrge
Set include_in_memory to true by default for tool actions and updated all example scripts to use temperature=0.0 for LLMs. 

- **Defaults**
  - Tool outputs are now included in memory by default.
  - All LLM examples now explicitly set temperature=0.0.


MagMueller:
Did you set everywhere where we did not set `include_in_memory` as parameter (because we dont want it) now to `include_in_memory=False`?

pirate (approved):

PR #1542, opened 2025-05-02
Added Docker and .dockerignore file for containerizing "browser-use"

Ameya-2003:
## Summary

This PR introduces a Dockerfile and corresponding .dockerignore file to the `browser-use` project. These additions aim to provide a standardized, secure, and efficient containerized environment suitable for both development and deployment.

---

## Dockerfile Highlights

- **Base Image**: Uses `python:3.11.3-slim` for a lightweight, consistent runtime.
- **Security Best Practices**:
  - Runs the application as a non-root user (`appuser`) to enhance container security.
  - Removes unnecessary packages after installation to reduce the image's attack surface.
- **Optimized Build Layers**:
  - Groups `RUN` commands to minimize layers and image size.
- **Health Check**:
  - Includes a placeholder `HEALTHCHECK` instruction for production monitoring (can be customized as needed).

### Build & Run

```bash
docker build -t browser-use .
docker run -it --rm browser-use
```

---

## .dockerignore Highlights

The `.dockerignore` file excludes unnecessary files from the Docker build context, significantly improving build speed and reducing final image size.

- **Ignored items include**:
      - Git metadata (`.git`, `.gitignore`)
      - Python caches (`__pycache__,` `*.pyc`, etc.)
      - IDE/workspace files (`.vscode/`, `.idea/`)
      - Virtual environments (`venv/`,` env/`)
      - System files (`.DS_Store`, `Thumbs.db`)
      - Docker build artifacts themselves (`Dockerfile`,` .dockerignore`)


pirate:
cool! can you sign the CLA and fix the lint errors? thanks for working on this!

Ameya-2003:
@cla-assistant check

Ameya-2003:
Hi maintainers 👋,

I've signed the CLA using the correct GitHub account and verified that my Git email matches my GitHub email. All checks have passed, but the CLA check seems to be stuck in a pending state (“Waiting for status to be reported”).

I've also tried re-signing the CLA and triggering a new commit, but the check remains unchanged.

Attaching a screenshot here for reference.

Could you please help by re-triggering the CLA check or overriding it if appropriate? Thank you!

CLA check pending screenshot

Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+
+# Change ownership to the non-root user
+RUN chown -R appuser:appgroup /app
+
+# Switch to non-root user
+USER appuser
+
+# Install Python dependencies
+RUN pip install --upgrade pip \
+    && pip install .
```
pirate, line 49: shouldnt we install chrome here too?

```suggestion
    && pip install . \
    && patchright install chrome
```

Ameya-2003: No, installing Chrome isn't necessary. The `playwright` Docker image already comes with the required browsers (Chromium, Firefox, WebKit) pre-installed via `playwright install --with-deps`. This covers all major use cases for browser automation and testing.

If there's a specific dependency on Google Chrome (not just Chromium), we can consider adding it, but for now, Chromium via Playwright should be sufficient and more lightweight.

Sounds good?

pirate: but this isn't based on the playwright docker image, it's just `FROM python:3.11.3-slim`

Ameya-2003: You're right — since the base image is `python:3.11.3-slim`, it doesn't include any browsers or Playwright dependencies by default. I mistakenly assumed a Playwright base image earlier.

To handle this properly, we should explicitly install the necessary dependencies using:

`RUN playwright install --with-deps `

This will install Chromium, Firefox, and WebKit along with required system packages — covering most use cases without needing to install full Chrome separately.

Let me know if you'd prefer I push a patch for this.

pirate: You can just accept my comment above, don't install webkit and firefox, just patchright chrome.

Issue #1527, opened 2025-04-30
Regression from 0.1.40 to 0.1.41 on recognition of contenteditable elements


edwardysun:
### Bug Description

I noticed that the older version (0.1.40) correctly identifies `contenteditable` elements as interact-able but this regressed in 0.1.41 (including what's on `main` right now). Was this an intentional change or can we add it back?

### Reproduction Steps

Pages with an element like this:
```
<span contenteditable="true">
```
Element not recognized

### Code Sample

```python
^
```

### Version

0.1.41

### LLM Model

Other (specify in description)

### Operating System

macOS

### Relevant Log Output

```shell

```

pirate:
I've created a dedicated test for us to evaluate the agent and help fix this and keep it fixed:

https://browser-use.github.io/stress-tests/challenge.html#contenteditable-div

PR #1548, opened 2025-05-02
Ensure contenteditable fields are interactable

edwardysun:
Resolves https://github.com/browser-use/browser-use/issues/1527
    

## Summary by mrge
Made contenteditable fields fully interactable by updating checks in the DOM tree builder. This ensures editable elements are recognized as interactive.

- **Bug Fixes**
  - Added logic to detect both contenteditable="true" and isContentEditable properties.


pirate:
thanks!

PR #1559, opened 2025-05-03
Add ALPHA browser-use CLI

pirate:
image

    

---

## Summary by mrge
Added an interactive CLI for browser-use that lets users run web automation tasks with LLMs from the terminal. User settings, API keys, and command history are saved to a config file for easy reuse.

- **New Features**
  - Supports model selection (OpenAI, Anthropic, Google) via config or CLI flags.
  - Remembers command history and settings across sessions.
  - Provides a guided prompt for entering and running web tasks interactively.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,16 +75,14 @@ For other settings, models, and more, check out the [documentation 📕](https:/
 
 ### Test with UI
 
-You can test [browser-use with a UI repository](https://github.com/browser-use/web-ui)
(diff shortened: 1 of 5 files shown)


pirate:
v2 is with textual and rich, but it's much more complicated and not working great yet because it's hard to get the output from the agent to not go straight to stdout/stderr 
image

pyoneerC:
looks awesome

pirate:
image

Issue #1565, opened 2025-05-04
viewport_expansion=-1 not working


tonystark262:
### Bug Description

Running `browser_use/dom/tests/extraction_test.py` using `viewport_expansion=-1,` only outputs visible screen dom.

https://github.com/browser-use/browser-use/issues/1516
https://github.com/browser-use/browser-use/pull/1552

### Reproduction Steps

Run `browser_use/dom/tests/extraction_test.py` using `viewport_expansion=-1

### Code Sample

```python
import asyncio
import os

import anyio
from langchain_openai import ChatOpenAI

from browser_use.agent.prompts import AgentMessagePrompt
from browser_use.browser.browser import Browser, BrowserConfig
from browser_use.browser.context import BrowserContext, BrowserContextConfig
from browser_use.dom.service import DomService


def count_string_tokens(string: str, model: str) -> tuple[int, float]:
	"""Count the number of tokens in a string using a specified model."""

	def get_price_per_token(model: str) -> float:
		"""Get the price per token for a specified model.

		@todo: move to utils, use a package or sth
		"""
		prices = {
			'gpt-4o': 2.5 / 1e6,
			'gpt-4o-mini': 0.15 / 1e6,
		}
		return prices[model]

	llm = ChatOpenAI(model=model)
	token_count = llm.get_num_tokens(string)
	price = token_count * get_price_per_token(model)
	return token_count, price


TIMEOUT = 60

DEFAULT_INCLUDE_ATTRIBUTES = [
	'id',
	'title',
	'type',
	'name',
	'role',
	'aria-label',
	'placeholder',
	'value',
	'alt',
	'aria-expanded',
	'data-date-format',
]


async def test_focus_vs_all_elements():
	config = BrowserContextConfig(
		# cookies_file='cookies3.json',
		disable_security=True,
		wait_for_network_idle_page_load_time=1,
		viewport_expansion=-1,
	)

	browser = Browser(
		config=BrowserConfig(
			# browser_binary_path='/Applications/Google Chrome.app/Contents/MacOS/Google Chrome',
		)
	)
	context = BrowserContext(browser=browser, config=config)

	websites = [
		# 'https://demos.telerik.com/kendo-react-ui/treeview/overview/basic/func?theme=default-ocean-blue-a11y',
		'https://www.ycombinator.com/companies',
		# 'https://kayak.com/flights',
		# 'https://en.wikipedia.org/wiki/Humanist_Party_of_Ontario',
[...]

PR #1571, opened 2025-05-04
Additional fixes for viewportExpansion=-1 mode to force-include all elements

pirate:
Fixes https://github.com/browser-use/browser-use/issues/1565

Need help testing it.
    

---

## Summary by mrge
Improved support for viewportExpansion = -1 mode so all elements and text nodes are force-included, even if outside the viewport.

- **Bug Fixes**
  - Ensured all elements and text nodes are treated as visible and topmost when viewportExpansion is -1.
  - Fixed highlight indexing so all interactive elements are included in this mode.

### DIFF
--- a/browser_use/dom/buildDomTree.js
+++ b/browser_use/dom/buildDomTree.js
@@ -523,6 +523,26 @@
    */
   function isTextNodeVisible(textNode) {
     try {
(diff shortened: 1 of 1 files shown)


tonystark262:
seems to be working for few cases i tried

PR #1572, opened 2025-05-04
Fix: Exclude disabled LWC components from interactive elements list

### LINKED ISSUE #1469: disabled elements are considered in the interactive elements list.
### Bug Description

Websites built using LWC framework will have buttons and input fields which are generally greyed out or disabled by using this addditional attribute disabled=true. Such elements are getting considered while constructing interactive elements list and sometimes LLM gives action list which forces it do interact with such elements resulting in failures or agent to be stuck in the loop.

### Reproduction Steps

Run any prompt which forces it to interact with LWC based websites or applications.

### Code Sample

```python
need small additional check for disabled attribute in this function: function isInteractiveElement
```

### Version

latest main branch

### LLM Model

GPT-4o

### Operating System

macOS

### Relevant Log Output

```shell

```

pyoneerC:
fixes a bug where disabled elements on lwc sites were being treated as interactive. for example, buttons with disabled="true" were still being clicked, which caused loops or failed interactions.

what changed:
added a check for the disabled attribute inside isInteractiveElement. this applies to regular form elements, elements styled to look interactive, and those using aria roles.

how to test:
run the agent on a site built with lwc. it should now ignore anything marked as disabled.

fixes #1469

### DIFF
--- a/browser_use/dom/buildDomTree.js
+++ b/browser_use/dom/buildDomTree.js
@@ -753,6 +753,11 @@
       if (element.inert) {
         return false;
       }
+      
+      // Check specifically for disabled="true" attribute which is common in LWC framework
+      if (element.getAttribute('disabled') === 'true') {
+        return false;
+      }
 
       return true;
     }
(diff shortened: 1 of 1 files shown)


pirate:
`element.getAttribute('disabled') === 'true'` should just be `element.getAttribute('disabled')`, boolean html attributes are truthy if present and only falsy if not present.

image

deveshruttala-edu:

PR #1576, opened 2025-05-05
Add chess example using browser-use

sandeepsalwan1:
1/30
    

---

## Summary by mrge
Added a new example script that uses browser-use to automate playing a chess game on lichess.org.

- **New Features**
  - Example script logs in to lichess.org, starts a game against the computer, and plays moves using browser automation.


sandeepsalwan1:
Hi @pirate  feel free to let me know if I should add these in pull requests because I plan to do a lot more of these.

PR #1579, opened 2025-05-05
Fix Dockerfile to match compatibility with patchright chrome

Ameya-2003:
## Updated Dockerfile based on [maintainer's](https://github.com/pirate) feedback:

- Removed redundant `pip install patchright`.
- Removed unnecessary system dependencies.

Take a look at reference to get a complete picture.
    

---

## Summary by mrge
Cleaned up the Dockerfile by removing extra dependencies and redundant pip install steps to match patchright chrome compatibility.

- **Refactors**
  - Removed unnecessary system packages.
  - Simplified user creation and file ownership steps.


Dockerfile
```diff
@@ -48,9 +33,9 @@ USER appuser
 RUN pip install --upgrade pip \
```
pirate, line 33: this should be removed too, trust the image to provide `pip` otherwise we have two copies in the layers

Dockerfile
```diff
@@ -48,9 +33,9 @@ USER appuser
...
     && pip install .
 
-# Define health check (example: adjust as needed)
+# Healthcheck (optional)
 HEALTHCHECK --interval=30s --timeout=10s --start-period=5s --retries=3 \
   CMD curl -f http://localhost:8000/health || exit 1
```
pirate, line 39: this should be removed we dont have anything that responds to this healthcheck

Dockerfile
```diff
@@ -48,9 +33,9 @@ USER appuser
...
 
-# Define default command
+# Default command
 CMD ["python", "-m", "browser_use"]
```
pirate, line 41:
```suggestion
CMD ["browser-use"]
```

Dockerfile
```diff
@@ -1,44 +1,29 @@
-# Use a specific Python 3.11 slim image as the base
 FROM python:3.11.3-slim
 
-# Set environment variables
 ENV PYTHONDONTWRITEBYTECODE=1 \
     PYTHONUNBUFFERED=1
 
-# Install system dependencies
+# Install minimal system dependencies required for Chrome
 RUN apt-get update && apt-get install -y --no-install-recommends \
```
pirate, line 7:
```suggestion
RUN apt-get update -qq && apt-get install -y --no-install-recommends \
```

quieter docker build

Ameya-2003:
I've applied the changes to the Dockerfile. It should be all set now, but let me know if there's anything else you'd like me to update or add.

Dockerfile
```diff
@@ -1,56 +1,40 @@
...
-HEALTHCHECK --interval=30s --timeout=10s --start-period=5s --retries=3 \
-  CMD curl -f http://localhost:8000/health || exit 1
+# Install Python dependencies from built package
+RUN python3 -m pip install --upgrade pip --quiet \
+    && python3 -m pip install dist/*.tar.gz --quiet
 
-# Define default command
-CMD ["python", "-m", "browser_use"]
+# Default command
+CMD ["python3", "-m", "browser_use"]
```
pirate, line 40:
```suggestion
CMD ["browser-use"]
```

Dockerfile
```diff
@@ -1,56 +1,40 @@
...
-    && pip install .
+# Build Python package
+RUN python3 setup.py sdist
 
-# Define health check (example: adjust as needed)
-HEALTHCHECK --interval=30s --timeout=10s --start-period=5s --retries=3 \
-  CMD curl -f http://localhost:8000/health || exit 1
+# Install Python dependencies from built package
+RUN python3 -m pip install --upgrade pip --quiet \
+    && python3 -m pip install dist/*.tar.gz --quiet
```
pirate, line 37:
```suggestion
RUN pip install .
```

Dockerfile
```diff
@@ -1,56 +1,40 @@
...
 RUN chown -R appuser:appgroup /app
 
 # Switch to non-root user
 USER appuser
 
-# Install Python dependencies
-RUN pip install --upgrade pip \
-    && pip install .
+# Build Python package
+RUN python3 setup.py sdist
```
pirate, line 33:
```suggestion
```

Dockerfile
```diff
@@ -1,56 +1,40 @@
...
+FROM python:3.11-slim
 
-# Set environment variables
 ENV PYTHONDONTWRITEBYTECODE=1 \
     PYTHONUNBUFFERED=1
 
-# Install system dependencies
-RUN apt-get update && apt-get install -y --no-install-recommends \
+# Install minimal system dependencies required for Chrome
+RUN apt-get update -qq && apt-get install -y --no-install-recommends -qq \
```
pirate, line 7:
```suggestion
RUN apt-get update -qq && apt-get install -y --no-install-recommends \
```

pirate:
be careful you introduced a bunch of broken changes in the last round of edits, also you're now missing the patchright install.

The order should be (each in a separate layer):

1. create non-root user
2. install base apt packages
3. copy `pyproject.toml` only into `/app`
4. install patchright *only* based on version in `pyproject.toml`
5. install chromium: `patchright install --with-deps --no-shell chromium`
6. copy the rest of the `.` codebase into `/app`
7. install codebase as pip package: `pip install .`
8. switch to non-root user: `USER appgroup`

For now this should be enough ^ but eventually we should add labels, per-layer caching, and a build summary to speed up repeated builds. See here how to do a proper chromium-based image: https://github.com/ArchiveBox/ArchiveBox/blob/dev/Dockerfile

Ameya-2003:
Thank you very much for your detailed step-by-step guidance. I’ve followed your instructions closely. I hope it meets your expectations — please let me know if any further adjustments are needed.

Dockerfile
```diff
@@ -1,56 +1,38 @@
...
 
-# Copy project files
-COPY . /app/
+# 5. Copy only dependency manifest
+COPY pyproject.toml /app/
 
-# Change ownership to the non-root user
-RUN chown -R appuser:appgroup /app
+# 6. Install patchright (version from pyproject.toml)
+RUN python3 -m pip install --upgrade pip --quiet \
```
pirate, line 22: don't upgrade pip, the container already provides this you're just introducing duplication again

Dockerfile
```diff
@@ -1,56 +1,38 @@
...
+      curl git \
+ && rm -rf /var/lib/apt/lists/*
+
+# 4. Set working directory
 WORKDIR /app
 
-# Copy project files
-COPY . /app/
+# 5. Copy only dependency manifest
+COPY pyproject.toml /app/
```
pirate, line 19: this is not used anywhere currently, you need to also extract the playwright version somehow or tell pip to use it somehow

Dockerfile
```diff
@@ -1,56 +1,38 @@
...
 
-# Change ownership to the non-root user
-RUN chown -R appuser:appgroup /app
+# 6. Install patchright (version from pyproject.toml)
+RUN python3 -m pip install --upgrade pip --quiet \
+ && python3 -m pip install patchright --quiet
```
pirate, line 23:
```suggestion
 && pip install patchright==$PLAYWRIGHT_VERSION
```

Dockerfile
```diff
@@ -1,56 +1,38 @@
...
+RUN patchright install --with-deps --no-shell chromium
+
+# 8. Copy the rest of the codebase
+COPY . /app
 
-# Install Python dependencies
-RUN pip install --upgrade pip \
-    && pip install .
+# 9. Install the application package
+RUN python3 -m pip install . --quiet
```
pirate, line 32:
```suggestion
RUN pip install -e .
```

Dockerfile
pirate, line 23:
```suggestion
RUN pip install --upgrade playwright==$PLAYWRIGHT_VERSION
```

Dockerfile
```diff
@@ -1,56 +1,38 @@
...
-# Change ownership to the non-root user
-RUN chown -R appuser:appgroup /app
+# 6. Install patchright (version from pyproject.toml)
+RUN python3 -m pip install --upgrade pip --quiet \
+ && python3 -m pip install patchright --quiet
 
-# Switch to non-root user
-USER appuser
+# 7. Install Chromium via patchright
+RUN patchright install --with-deps --no-shell chromium
```
pirate, line 26:
```suggestion
RUN playwright install --with-deps --no-shell chromium
```

Dockerfile
pirate, line 23:
```suggestion
RUN pip install playwright==$PLAYWRIGHT_VERSION
```

Dockerfile
```diff
@@ -1,56 +1,37 @@
...
 
-# Copy project files
-COPY . /app/
+# 5. Copy only dependency manifest
+COPY pyproject.toml /app/
 
-# Change ownership to the non-root user
-RUN chown -R appuser:appgroup /app
+# 6. Install patchright (version from pyproject.toml)
+RUN pip install playwright==$PLAYWRIGHT_VERSION
```
pirate, line 22:
```suggestion
# 6. Install playwright (version from pyproject.toml)
RUN PLAYWRIGHT_VERSION=$(grep -oP '(p....right>=\K[0-9.]+' pyproject.toml)\
	&& pip install playwright==$PLAYWRIGHT_VERSION
```

Issue #1312, opened 2025-04-07
(Low-impact) Remove `botocore` dependency


tarasyarema:
### Bug Description

Remove/bump the `botocore` dep, or make it optional (as only used in example) as it mess up upstream projects using s3 libraries.

See https://github.com/tarasyarema/browser-use/commit/8fd0b8bd1cc99492d323ca9f183692c17e2db097 (I removed the example completely, but can be moved to a notebook like or something)

### Reproduction Steps

1. Install browser-use
2. Try installing latest version of s3fs (`uv add s3fs==2025.3.1`)

### Code Sample

```python
uv add s3fs==2025.3.1
```

### Version

0.1.41rc2

### LLM Model

Other (specify in description)

### Operating System

macos

### Relevant Log Output

```shell

```

PR #1586, opened 2025-05-05
chore: Bedrock optional dependencies

tarasyarema:
Moved the Bedrock Claude example's botocore dependency to an optional "examples" extra, making it easier to install only when needed.

- **Dependencies**
  - Removed botocore from main dependencies.
  - Added a "examples" optional extra for botocore.
  - Updated example instructions to mention the new install method.

Fixes https://github.com/browser-use/browser-use/issues/1312

### DIFF
--- a/examples/models/bedrock_claude.py
+++ b/examples/models/bedrock_claude.py
@@ -1,6 +1,8 @@
 """
 Automated news analysis and sentiment scoring using Bedrock.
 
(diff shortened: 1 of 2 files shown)


tarasyarema:
@pirate late follow-up on the issue

pirate:
thanks!

PR #1601, opened 2025-05-07
PDF

ProphetDiceBot:
Pdf functionality 
    

---

## Summary by mrge
Added a new function to download PDFs from a given URL using a browser context and save them to a local downloads folder.


ProphetDiceBot (commented):
#

PR #1633, opened 2025-05-10
switch back from patchright to playwright until issues are resolved

pirate:
- drop in cloud stats
- encoding headers issues possibly related to initScripts, or CDP NetworkService anti-leak interfering with legitimate library usage?
- persistent CAPTCHAS still
- PDF script injection issues: https://browser-use.slack.com/archives/C08K7B19DKM/p1747040347629579?thread_ts=1747040324.134799&cid=C08K7B19DKM
    

---

## Summary by mrge
Reverted all code and documentation from Patchright back to Playwright due to ongoing issues with Patchright.

- **Dependencies**
  - Replaced Patchright with Playwright in all imports and requirements.


evgeny-kim:
> persistent CAPTCHAS still
This is likely if we use Chromium, for me it triggers CF captcha. Patchright suggests using Chrome.

pirate:
I asked for clarification about that @evgeny-kim https://github.com/Kaliiiiiiiiii-Vinyzu/patchright-python/issues/56

The `chromium` vs `chrome` playwright channels actually differ significantly in the headless modes they provide, so switching them has lots of other side effects.

PR #1636, opened 2025-05-11
fix(context): add _is_visible method for better element visibility checks

satya-nutella:
- Introduced a new private method `_is_visible` to determine if an element is visible based on its hidden state and bounding box dimensions.
- Updated element selection logic to utilize the new visibility check, ensuring only visible elements are processed.
- Added .idea to .gitignore to exclude IDE-specific files from version control.
    

---

## Summary by mrge
Added a new _is_visible method to improve element visibility checks, ensuring only visible elements are processed. Also updated .gitignore to exclude .idea files.

- **Bug Fixes**
  - Fixed cases where hidden elements were incorrectly treated as visible by checking both hidden state and bounding box size.


pirate (approved):

LarsenCundric:
Please include this in all cases where we try to scroll into view.

PR #1651, opened 2025-05-12
Add Integration Tests for Browser Window Dimensions

Anirudha619:
This PR introduces integration tests to verify the correct application of custom window dimensions in both headless and headful browser contexts. The tests ensure that viewport and window sizes are accurately configured and validated, enhancing the reliability of the browser functionality.
Additionally, this PR identifies an issue where the window_height and window_width variables in the context do not match the actual Playwright browser context. The assertions at lines 240 and 241 currently fail, indicating a discrepancy that needs to be addressed.
    

---

## Summary by mrge
Added integration tests to check that custom browser window and viewport dimensions are set correctly in both headless and headful modes. Found a mismatch between configured and actual window size values in some cases.

### DIFF
--- a/browser_use/browser/context.py
+++ b/browser_use/browser/context.py
@@ -219,9 +219,6 @@ class BrowserContext:
 		state: BrowserContextState | None = None,
 	):
 		self.context_id = str(uuid.uuid4())
-		if not browser.config.headless and config.no_viewport:
-			# keep the window_width and window_height to default values
-			pass  # TODO
 
 		self.config = config or BrowserContextConfig(**(browser.config.model_dump() if browser.config else {}))
 		self.browser = browser
--- a/browser_use/browser/context.py
+++ b/browser_use/browser/context.py
@@ -349,6 +349,13 @@ class BrowserContext:
 				await current_page.goto('about:blank')
 				logger.debug('🆕  Created new page: %s', current_page.url)
 
+			if self.config.no_viewport and not self.browser.config.headless:
+				# Sync the config window dimensions with the actual browser viewport size
+				actual_size = await current_page.evaluate("""() => ({ width: window.innerWidth, height: window.innerHeight })""")
+
+				self.config.window_width = actual_size['width']
+				self.config.window_height = actual_size['height']
+
 			# Get target ID for the active page
 			if self.browser.config.cdp_url:
 				targets = await self._get_cdp_targets()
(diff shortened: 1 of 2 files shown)


browser_use/browser/context.py
```diff
@@ -219,6 +219,9 @@ def __init__(
 		state: BrowserContextState | None = None,
 	):
 		self.context_id = str(uuid.uuid4())
+		if not browser.config.headless and config.no_viewport:
+			# keep the window_width and window_height to default values
+			pass  # TODO
```
Anirudha619, line 224: when no_viewport == True and headless == False , then the context state doesn't match with actual browser viewport

```python
assert context.config.window_height == default_viewport_size['height']
assert context.config.window_width == default_viewport_size['width']
```

Anirudha619: below line of code sets the window.outerWidth,  window.outerHeight of the browser

https://github.com/browser-use/browser-use/blob/942935f0be767562ad08abfa3ffb91168c60d36c/browser_use/browser/browser.py#L322

tests/test_browser_context.py
```diff
@@ -0,0 +1,360 @@
...
+			)
+
+			assert viewport_size['width'] != custom_width, (
+				f'Viewport width ({viewport_size["width"]}px) should not be equal to {custom_width}px.'
+			)
+			assert viewport_size['height'] != custom_height, (
+				f'Viewport height ({viewport_size["height"]}px) should not be equal to {custom_height}px.'
+			)
+
+			assert context.config.window_height == default_viewport_size['height']
```
Anirudha619, line 240: Because of all the above reasons, currently the test fails at this line

browser_use/browser/context.py
pirate: imo window_size and window_height should refer to the inner size of the viewport, and headful browsers should never use viewport

browser_use/browser/context.py
```diff
@@ -346,6 +346,13 @@ async def _initialize_session(self):
 				await current_page.goto('about:blank')
 				logger.debug('🆕  Created new page: %s', current_page.url)
 
+			if self.config.no_viewport and not self.browser.config.headless:
+				# Sync the config window dimensions with the actual browser viewport size
+				actual_size = await current_page.evaluate("""() => ({ width: window.innerWidth, height: window.innerHeight })""")
+
+				self.config.window_width = actual_size['width']
+				self.config.window_height = actual_size['height']
```
Anirudha619, line 354: @pirate 
added above if condition that syncs the context config window dimensions states with the actual browser viewport size.

deveshruttala-edu:

deveshruttala-edu:

Issue #1594, opened 2025-05-06
When I call the drop-down menu of the input component, I want to select one of the fields (such as Beijing School), but it can never be selected.


hanxiao55:
### Bug Description

When I call the drop-down menu of the input component, I want to select one of the fields (such as Beijing School), but it can never be selected.
This is an internal address and is temporarily inaccessible.
Version: 0.1.45

![Image](https://github.com/user-attachments/assets/77de13ad-35ee-400e-a0bd-fc78e5d72d8e)

![Image](https://github.com/user-attachments/assets/9b738a89-98db-40e6-b919-9b97e8daff65)

### Reproduction Steps

INFO [agent] 📍 Step 1
INFO [agent] 🤷 Eval: Unknown - Starting the task
INFO [agent] 🧠 Memory: Step 1/3: Navigating to target URL https://qerp.test.xdf.cn/nis/index
INFO [agent] 🎯 Next goal: Check if page loaded and setup popup appears
INFO [agent] 🛠️ Action 1/1: {"go_to_url":{"url":"https://qerp.test.xdf.cn/nis/index"}}
INFO [controller] 🔗 Navigated to https://qerp.test.xdf.cn/nis/index
INFO [agent] 📍 Step 2
INFO [agent] ⚠ Eval: Failed - Navigation to target URL redirected to a login page instead of showing setup popup. The setup popup appears after login
INFO [agent] 🧠 Memory: Step 1 completed (0/3 sub-steps). Need to login first using credentials: account in [0], password in [1], test environment security code in [2]
INFO [agent] 🎯 Next goal: Perform login to access target page
INFO [agent] 🛠️ Action 1/4: {"input_text":{"index":0,"text":"account"}}
INFO [agent] 🛠️ Action 2/4: {"input_text":{"index":1,"text":"password"}}
INFO [agent] 🛠️ Action 3/4: {"input_text":{"index":2,"text":"test environment security code"}}
INFO [agent] 🛠️ Action 4/4: {"click_element":{"index":3}}
INFO [controller] ⌨️Input sensitive data into index 0
INFO [controller] ⌨️ Input sensitive data into index 1
INFO [controller] ⌨️ Input sensitive data into index 2
INFO [controller] 🖱️ Clicked button with index 3:
INFO [agent] 📍 Step 3
INFO [agent] 👍 Eval: Success - Login completed and setup popup appears
[...]

PR #1660, opened 2025-05-13
Fix: detect nested interactive elements using heuristics to avoid skipping actionable items

s00d:
### Description

When an interactive element is nested inside another interactive parent (e.g. `.btn-menu-item` inside a `<button>`), it often gets ignored and not highlighted. This happens because the parent is already marked as interactive, and the child is skipped by default.

To address this, a new helper function `isHeuristicallyInteractive()` was introduced. It adds a secondary detection path for elements that are:

* potentially visible,
* have interactive attributes or semantic classes,
* are placed inside known interactive containers (like menus or lists),
* and have visible children.

This logic helps surface deeply nested actionable elements that would otherwise be missed.

---

### Example HTML (repro case):

```html
<button class="sidebar-button">
  <span class="sidebar-button-label">Open Menu</span>

  <div class="btn-menu">
    <div class="btn-menu-item">
      <span class="icon">⚙️</span>
      <span class="label">Settings</span>
    </div>

    <div class="btn-menu-item">
      <span class="icon">📁</span>
      <span class="label">Archived Chats</span>
    </div>

    <div class="btn-menu-item">
      <span class="icon">🔒</span>
      <span class="label">Privacy & Security</span>
    </div>

    <div class="btn-menu-item">
      <span class="icon">👤</span>
      <span class="label">My Profile</span>
    </div>
  </div>
</button>
```

In this case, the `.btn-menu-item` elements are clearly actionable, but were previously skipped.

image

---

### Changes

* Extracted heuristic logic to `isHeuristicallyInteractive(element)`
* Used it inside `isElementDistinctInteraction()` to treat such elements as separate interactions
* Improved readability and comments
* Added safeguards to avoid false positives (e.g., decorative spans)

---

### Summary

Fixes an issue where nested interactive items inside buttons or containers were skipped during DOM scanning. Now they are detected using heuristic signals.

✅ Closes: [#1594](https://github.com/browser-use/browser-use/issues/1594)

### DIFF
--- a/browser_use/dom/buildDomTree.js
+++ b/browser_use/dom/buildDomTree.js
@@ -1037,8 +1037,8 @@
    * Heuristically determines if an element should be considered as independently interactive,
    * even if it's nested inside another interactive container.
    *
-   * This function is useful for identifying nested actionable elements (e.g., buttons inside menus),
-   * especially when they do not meet the full criteria of interactive elements but still appear clickable.
+   * This function helps detect deeply nested actionable elements (e.g., menu items within a button)
+   * that may not be picked up by strict interactivity checks.
    */
   function isHeuristicallyInteractive(element) {
     if (!element || element.nodeType !== Node.ELEMENT_NODE) return false;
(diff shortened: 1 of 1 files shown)


Alezander9:
On it!

Alezander9:
The model performance seems to be the same

pirate:
awesome, thanks @Alezander9 and thank you @s00d for the contribution!

PR #1664, opened 2025-05-13
add searchBing action (to skip CAPTCHA in google)

yonghuname:
### Problem Description
Suggest add an action to use searchBing instead of searchGoogle for stuck in CAPTCHA .Espcially for Chinese user，In most time, they was stuck in google CAPTCHA anti-robot test ，due to use vpn in same ip. I suggest add more action to use bing to skip it，and improve its applicability range.

### Proposed Solution
I add a action in
browse_use/agent/PlaywrightScriptGenerator.py
browse_use/controller/service.py
browse_use/controller/view.py
and I have tried to modify the
browser-use/browser_use/agent/system_prompt.md
to indicate "new search (e.g., using search_bing action with a query )"and "if encounter with CAPTCHA, skip it."

of course，the system_prompt.md could keep former one
I suggest add "search in bing" in task description prompt. It will work also, llm will do aciton "saerchBing" first .
    

---

## Summary by mrge
Added a new searchBing action to let users search with Bing instead of Google, helping avoid CAPTCHA issues, especially for users in China.

- **New Features**
  - Added searchBing action support in the backend and API.
  - Updated prompts to suggest using Bing when Google CAPTCHA blocks searches.

### DIFF
--- a/browser_use/agent/playwright_script_generator.py
+++ b/browser_use/agent/playwright_script_generator.py
@@ -416,7 +416,6 @@ class PlaywrightScriptGenerator:
 				]
 			)
 		else:
-	
 			script_lines.append(f'            # Skipping search_bing ({step_info_str}): missing or invalid query')
 		return script_lines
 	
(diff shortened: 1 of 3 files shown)


deveshruttala-edu:

mertunsall:
I think we are trying to keep the action space as clean as possible: one principle for this is that we do not want to add an action unless it
1. would be called extremely often if it exists
2. cannot be done using other actions

PR #1667, opened 2025-05-13
fix user_data_dir not expanded in chrome_launch_args

khesui:
Fixed tilde expansion in browser profile path handling. 
Previously, when creating a user data directory with a tilde (\~) in the path, the code correctly expanded the tilde when creating the directory but didn't save the expanded path. Thus later on in `chrome_launch_args` Chrome receiving a path with a literal "~" character in the --user-data-dir argument, causing profiles to be created in unexpected locations such as `/Users/<username>/browser_use_workspace/~/.config/browseruse/profiles/default`. 
The fix properly stores the expanded path by assigning the result of expanduser() back to the user_data_dir variable, ensuring Chrome receives the fully resolved home directory path.
    

---

## Summary by mrge
Fixed an issue where the user data directory path with a tilde (~) was not fully expanded, causing Chrome to use the wrong profile location.

- **Bug Fixes**
  - Ensured the expanded home directory path is used when launching Chrome.

### DIFF
--- a/browser_use/browser/browser.py
+++ b/browser_use/browser/browser.py
@@ -225,7 +225,8 @@ class Browser:
 			try:
 				# ~/.config/browseruse/profiles/default
 				user_data_dir = Path('~/.config') / 'browseruse' / 'profiles' / 'default'
-				user_data_dir.expanduser().mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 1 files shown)


pirate:
Thanks!

Issue #1628, opened 2025-05-09
Browser.close() method closes external HTTPX clients used by parent processes


Saik0s:
### Bug Description

The [`Browser.close()` method](https://github.com/browser-use/browser-use/blob/main/browser_use/browser/browser.py#L373) currently closes **all active HTTPX client connections**, not just its own.
This breaks external processes that rely on persistent HTTPX connections, such as an agent calling the browser via tools.

**Specific Example:**

* [DeepResearch agent](https://github.com/browser-use/web-ui/blob/main/src/agent/deep_research/deep_research_agent.py) has an LLM which uses HTTPX AsyncClient under the hood.
* DeepResearch agent invokes a search tool that internally creates a browser.
* When the browser completes the search, it triggers its `close()` method.
* This action forcefully closes **every HTTPX connection**, including the external connection still actively used by the DeepResearch agent’s LLM API call.
* As a result, the DeepResearch agent immediately fails after one browser invocation.

**Expected Behavior:**
Browser should close **only HTTPX client instances it directly owns and manages**.

**Actual Behavior:**
Browser aggressively collects and closes **all** HTTPX clients in the Python process, unintentionally breaking external components.

**Suggested Fix:**
Modify the cleanup logic so the browser explicitly tracks and closes **only HTTPX instances it created**, rather than globally closing all HTTPX clients via garbage collection.

### Reproduction Steps

1. Start deep research from https://github.com/browser-use/web-ui
2. Wait for the first search task's browser gets closed
3. Deep research agent stops working

### Related Code

[`Browser.close()` method](https://github.com/browser-use/browser-use/blob/main/browser_use/browser/browser.py#L373)
[Where it is called in deep research code](https://github.com/browser-use/web-ui/blob/main/src/agent/deep_research/deep_research_agent.py#L184)

### Version

0.1.45

### LLM Model

Other (specify in description)

### Operating System

macos

### Relevant Log Output

```shell
2025-05-09 18:35:14,212 - openai._base_client - DEBUG - Encountered Exception
Traceback (most recent call last):
  File "/Users/igortarasenko/Projects/mcp-browser-use/.venv/lib/python3.11/site-packages/openai/_base_client.py", line 1484, in request
    response = await self._client.send(
               ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/igortarasenko/Projects/mcp-browser-use/.venv/lib/python3.11/site-packages/httpx/_client.py", line 1616, in send
[...]

PR #1688, opened 2025-05-15
Fix Httpx issue closing all clients

sandeepsalwan1:
Resolves Issue #1628. The httpx clients are typically closed automatically when they go out of scope or when their owners close them. External libraries like OpenAI's SDK or Langchain manage their own HTTPX clients. Global cleanup can be dangerous - it's better for each component to manage its own resources. Removing this method shouldn't cause memory leaks since responsible libraries handle their own cleanup.

Test showing before and after change.  Screenshot:
image

    

---

## Summary by mrge
Removed global cleanup of httpx clients from the Browser class to prevent closing clients owned by external libraries.

- **Bug Fixes**
  - Browser.close() no longer closes httpx clients it did not create.
  - Added a test to confirm external httpx clients remain open after closing the browser.

### DIFF
--- a/browser_use/browser/browser.py
+++ b/browser_use/browser/browser.py
@@ -422,6 +422,6 @@ class Browser:
 		except Exception as e:
 			logger.debug(f'Failed to cleanup browser in destructor: {e}')
 
-	# no-op because httpx clients are by their creators, not by the Browser class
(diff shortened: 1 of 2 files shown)


pirate:
Thank you, I was going to remove this too but you beat me to it!

PR #1692, opened 2025-05-16
Code cleanup for cleanup_httpx_clients function call

ajuijas:
cleanup_httpx_clients function removed from the codebase. 
    

---

## Summary by mrge
Removed the unused cleanup_httpx_clients function and its call from browser.py to simplify the code.


pirate:
thanks!

PR #1698, opened 2025-05-17
add cookies support

itamartal:
## Summary by mrge
Added support for collecting, clearing, and adding cookies in the browser controller.

- **New Features**
  - Collect cookies from the current page.
  - Clear cookies by name, domain, or path.
  - Add cookies with full attribute support.


pirate:
we don't want to provide this as an action to the LLM, it's not the kind of action LLMs are good at (too many parameters that have to be exactly character-for-character correct), instead we want to handle it up-front at the start of the run using https://github.com/thewh1teagle/rookie

PR #1703, opened 2025-05-18
fixed scroll when whole page lives inside a nested container

Guido1Alessandro1Trevisan:
### **Problem**
Pages with nested overflow containers (e.g., Airtable forms) ignored window.scrollBy, leaving bots “stuck.”
This PR lets the agent scroll inside the element users actually see, improving reliability across modern web apps.

**Data entry** is an important use case for browser use

### **Before:** (Doesn't scroll down or up)
agent_history_before

### **After:** (Scroll down and up)
agent_history_after

### **Fixed how?:**
added a container-aware scrolling helper that works the same way on every site. Instead of blindly calling window.scrollBy, the helper runs a tiny in-page script that (1) starts at the element in focus, (2) walks up the DOM until it finds the first ancestor whose overflow-y is auto | scroll | overlay and whose content actually overflows, and (3) invokes scrollBy on that element (or on window as a fallback). Because the logic is driven purely by runtime CSS/DOM properties, it doesn’t rely on hard-coded selectors or site-specific quirks so the solution is website-agnostic and handles complex layouts like Airtable or any app with nested scroll regions.

    

---

## Summary by cubic
Scrolling now works inside nested containers, not just the main window, so bots can scroll reliably in apps like Airtable forms.

- **Bug Fixes**
  - Added a helper that finds the correct scrollable container and scrolls it, instead of always using window.scrollBy.
  - Improved fallback logic to handle sites with complex layouts.

### DIFF
--- a/browser_use/browser/context.py
+++ b/browser_use/browser/context.py
@@ -1910,12 +1910,12 @@ class BrowserContext:
 		pixels_above = scroll_y
 		pixels_below = total_height - (scroll_y + viewport_height)
 		return pixels_above, pixels_below
-	
+
 	async def _scroll_container(self, pixels: int) -> None:
 		"""Scroll the element that truly owns vertical scroll.Starts at the focused node ➜ climbs to the first big, scroll-enabled ancestor otherwise picks the first scrollable element or the root, then calls `element.scrollBy` (or `window.scrollBy` for the root) by the supplied pixel value."""
-		
+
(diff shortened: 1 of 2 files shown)


pirate:
awsome! we've gotten a few different PRs attempting this but I like this approach

there are some situations where it doesn't work (side-by-side nested text scolling frames that arent forms / dont grab focus), but I think it's an improvement over the current status quo

PR #1706, opened 2025-05-18
fix: detect interactive elements with click event listener 

limingzhong61:
fix: detect interactive elements with click event listener : **adding 'click' mouse event check and fixing element mouse event check**

## chrome debug
The element of "不购买行李额" is not hightlighted  in the picture，and **it is a div element with class "baggage-chooser".**
image

When we use `window.getEventListenersForNode`,it **return array of listener object** and div element with class "baggage-chooser" has the **click event.**
img_v3_02me_187b388f-1388-4234-b835-cc7b9a7c62ag

## before fix
before fix: the element of "不购买行李额" is not hightlighted  in the picture
image

## after fix
after fix: the element of "不购买行李额" is **hightlighted with number 8**  in the picture
image

    

---

## Summary by cubic
Fixed detection of interactive elements by checking for click event listeners, so elements with click handlers are now correctly highlighted.

### DIFF
--- a/browser_use/dom/buildDomTree.js
+++ b/browser_use/dom/buildDomTree.js
@@ -830,8 +830,10 @@
         const listeners = getEventListeners(element);
         const mouseEvents = ['click', 'mousedown', 'mouseup', 'dblclick'];
         for (const eventType of mouseEvents) {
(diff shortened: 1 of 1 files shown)


pirate:
nice, thanks!

PR #1708, opened 2025-05-18
Route Playwright Imports through a browser-use specific class

yasithdev:
Just routing all playwright imports through a `driver.py` class to make it easy to swap out playwright with other drivers. The motivation is to have own types and classes for browser driving, and have playwright as the default implementation of these classes.

This PR by itself doesn't add much value, but my followup idea is this.
1. create Protocol classes (from typing import Protocol) in `driver.py` to match the playwright spec.
2. create implementations of these protocol classes for playwright.
3. once done, create other implementations of these protocol classes (e.g., for extension-based driving).

I didn't start on the protocol classes yet, as there are pending API changes (e.g., #1699)

    

---

## Summary by cubic
Routed all Playwright imports through a new driver.py class to make it easier to swap Playwright with other browser drivers in the future. This sets up a single place to manage browser driver types and imports.


yasithdev:
@pirate I made a full abstraction layer (changes included). there's a bunch of failing tests after the change, so I'll keep this on Draft until these are fixed. but you can test it out. now playwright implementation is fully abstracted and is in

```
from browser_use.drivers.playwright import PlaywrightBrowser
```

pirate:
Please see the latest design in: https://github.com/browser-use/browser-use/pull/1699

I added support for choosing `playwright`/`patchright` at runtime, you can see an example in: `examples/browser/stealth.py`.

yasithdev:
Thanks @pirate! I'll resolve the merge conflicts, make sure the tests pass, and then promote it from draft status.

pirate:
there are a bunch of good bugfixes in this PR, but as for the driver/accessor class, do we still need it now that `BrowserSession(...)` can take `playwright`, `browser`, and `browser_context` objects at runtime?

yasithdev:
@pirate I'm targeting a different usage pattern from this PR. I want to build an extension that purely runs from chrome/edge. I want to run browser-use as a server (e.g., wrapped in a Flask/FastAPI server) that can accept user actions, and then orchestrate the task purely through sockets + extension. In this case, I cannot use playwright/patchwright, and have to rely on what's possible natively through chrome. Once this PR is in, it opens up a range of use cases that can't otherwise be done from playwright. e.g., hosting browser-use on a public URL, and communicating with for automating user tasks from a non-debugging-mode browser context.

I can't think of a way to do this without this PR in. But if there is a way, I'd love to know!

pirate:
ah ok yeah we're building that next! both for the cloud and for the library we're working on designing the event schema for several layers of the internals right now.

yasithdev:
Yes, absolutely! There's few naggy ones I fixed, like the gpt4 check breaking the use of gpt4o (had to add a - to workaround it). I'll raise a separate PR with those improvements.

PR #1709, opened 2025-05-19
fix:using f-string to refer outer variable, fix glob WARNING log in allowed_domains

limingzhong61:
fix: **using f-string** to refer outer variable to fix glob WARNING   log  in allowed_domains

when I use glob "*.xx.com" in allowed_domains,the log show as below:
```
⚠️ Allowing agent to visit {domain} based on allowed_domains=['{glob}', ...]. Set allowed_domains=['{domain}', ...] explicitly to avoid the security risks of glob patterns!
```
so,I add f-string  to refer outer variable `domain` and `glob`
```
WARNING  [browser] ⚠️ Allowing agent to visit www.xx.com based on allowed_domains=['*.xx.com', ...]. Set allowed_domains=['www.xx.com', ...] explicitly to avoid the security risks of glob patterns!
```
    

---

## Summary by cubic
Fixed log message to correctly show the actual domain and glob values by using an f-string.


pirate:
thanks!

Issue #1715, opened 2025-05-19
Auto-detect tool calling method at startup


pirate:
We should enhance `Agent._set_tool_calling_method()` to auto-detect the best supported tool-calling method at startup by just trying all of them, instead of manually guessing it based on hardcoded model name `if` statements.

- `Agent._set_tool_calling_method()`
  - if `Agent.settings.tool_calling_method == 'auto'`:
    - 1. try `Agent.settings.tool_calling_method` `=` `function_calling` | `json` | `tools` | `raw` in order of best -> worst
    - 2. set `Agent.settings.tool_calling_method` accordingly

For speed reasons and to avoid 2 wasted LLM calls on every startup, we may want to merge this with `Agent._verify_llm_connection()`.

This should help clear up a ton of basic support tickets that we get from weird provider + model issues around certain tool calling methods: https://github.com/browser-use/browser-use/issues?q=sort%3Aupdated-desc%20is%3Aissue%20tool%20calling

MagMueller:
Mh - Some models only work with a specific tool_calling_method, which is why I hard-coded that. So e.g. for ChatGoogleGenerativeAI like gemini it needs to be None.

GeorgePPP:
I believe there are some bugs related to some models that do not support Langchain ToolMessage but returns a tool call format.

deveshruttala-edu:
Mock comment: Hyperbrowser can help with this issue by providing AI-native browsing support.

deveshruttala-edu:
Mock comment: Hyperbrowser can help with this issue by providing AI-native browsing support.

RMartires:
hey looking at the merged PR should this be a open issue be closed ?

MagellaX:
> hey looking at the merged PR should this be a open issue be closed ?

yeah, had the same question, is this an issue now?

PR #1718, opened 2025-05-20
Auto-detect tool calling method and verify llm api at startup

Anirudha619:
https://github.com/browser-use/browser-use/issues/1715

- Enhanced `_set_tool_calling_method()` to auto-detect the best supported tool-calling method by testing each method in order of preference
- Integrated with `_verify_llm_connection()` to avoid extra LLM calls
    

---

## Summary by cubic
The agent now auto-detects the best tool calling method for the LLM by testing each option at startup and verifies the LLM API connection in the same step.

- **Refactors**
  - Combined tool calling method detection and LLM API verification to avoid extra API calls.
  - Removed redundant environment variable checks and simplified connection logic.


browser_use/agent/service.py
```diff
@@ -396,27 +394,78 @@ def _setup_action_models(self) -> None:
...
-				else:
-					return 'function_calling'
+	def _test_tool_calling_method(self, method: str) -> bool:
+		"""Test if a specific tool calling method works with the current LLM."""
+		try:
+			# Create a simple test message
+			test_message = HumanMessage(content='Hello, world!')
+
+			if method == 'raw':
+				# For raw mode, we just check if we can invoke the model
```
pirate, line 404: the test for `raw` should probably ask the model to respond with some JSON manually in the same way we do in the actual system prompt in `raw` mode.

I think for the same reason we check if the model knows the capital of France, basic sanity checks are very useful to weed out low power models that struggle to follow directions + return JSON

browser_use/agent/service.py
```diff
@@ -396,27 +394,78 @@ def _setup_action_models(self) -> None:
...
+				return True
 			else:
-				return None
-		else:
-			return tool_calling_method
+				# For other methods, try to use structured output
+				structured_llm = self.llm.with_structured_output(self.ActionModel, include_raw=True, method=method)
+				# Try a simple invocation to test the method
+				response = structured_llm.invoke([test_message])
+				# TODO: add validation of response
```
pirate, line 415: can you try and keep the capital of France question + answer check here as part of the tool call prompt & validation logic? we've found it very useful in practice to set a minimum intelligence bar for models, it really reduces the support load from people trying to run very low-power or broken models

pirate:
this looks great so far nice work! 🚀

PR #1742, opened 2025-05-21
fix(controller): wait on page change

Thytu:
This introduces a mechanism to wait for page loads after an action causes a URL change. 

By ensuring the page is fully loaded before proceeding, it prevents the agent from attempting to scroll, interact with elements, or potentially failing on pages that are still in the process of loading.

Changes :
* Adding a `_wait_on_url_change` parameter to the `Controller` (defaulting to True).
* In `act`, if `_wait_on_url_change` is `True` and the URL changes, the controller now waits for the page and its frames to complete loading before proceeding.

Note : I was (and still) hesitating with putting this parameter inside `AgentSettings` and provide it as arg to `controller.act`. Let me know what you prefer.

### DIFF
--- a/browser_use/controller/service.py
+++ b/browser_use/controller/service.py
@@ -43,8 +43,10 @@ class Controller(Generic[Context]):
 		self,
 		exclude_actions: list[str] = [],
 		output_model: type[BaseModel] | None = None,
+		wait_on_url_change: bool = True,
 	):
 		self.registry = Registry[Context](exclude_actions)
+		self._wait_on_url_change = wait_on_url_change
 
 		"""Register all default browser actions"""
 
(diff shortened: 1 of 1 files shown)


browser_use/controller/service.py
```diff
@@ -868,6 +872,11 @@ async def act(
 					context=context,
 				)
 
+				if self._wait_on_url_change and initial_url != (new_url := (await browser_session.get_current_page()).url):
+					initial_url = new_url
+					logger.info(f'🕒 URL changed, waiting for page load state.')
```
Thytu, line 877:
```suggestion
					logger.info('🕒 URL changed, waiting for page load state.')
```

pirate:
I think we don't even need the config flag, we can just always wait for load after actions as long as we can guarantee:

- no wait should happen in the best case when the URL was changed by JS `history.pushState()` and there is no loading to wait for
- in the worst case where loading is triggered but it never finishes, it should cap it at the `browser_profile.maximum_wait_page_load_time`

pirate:
Related:

- https://github.com/browser-use/browser-use/pull/1831

deveshruttala-edu:

PR #1759, opened 2025-05-23
Improve file upload detection

evgeny-kim:
File input search fails when the file input itself is not visible and the label referencing it is not its sibling. For example:
```
<input id="input-file-id" type="file">
<div class="fancy">
    <label for="input-file-id">Drop your files here</label>
<div>
```

This PR fixes the issue.

If a label's `for` attribute points to a file input, we now correctly find that input. I added `"label"` to the `interactiveElements` set so its `for` attribute is preserved.

I also combined the file upload search logic from `is_file_uploader` and `get_file_upload_element` in one place.
    

---

## Summary by cubic
Improved file upload detection so labels linked to hidden file inputs are now correctly found, even if the label is not a sibling of the input.

- **Bug Fixes**
  - Fixed cases where file upload elements were missed if the label and input were not siblings.
  - Combined and simplified file upload search logic for better reliability.

### DIFF
--- a/browser_use/browser/session.py
+++ b/browser_use/browser/session.py
@@ -1932,10 +1932,10 @@ class BrowserSession(BaseModel):
 		- Check if it's a label pointing to a file input
 		- Recursively search children for file inputs
 		- Check siblings for file inputs
-		
+
 		Args:
 			index: The index of the candidate element (could be a file input, label, or parent element)
-			
+
 		Returns:
 			The DOM element for the file input if found, None otherwise
 		"""
@@ -1943,12 +1943,12 @@ class BrowserSession(BaseModel):
 			selector_map = await self.get_selector_map()
 			if index not in selector_map:
 				return None
-				
+
 			candidate_element = selector_map[index]
-			
+
 			def is_file_input(node: DOMElementNode) -> bool:
 				return isinstance(node, DOMElementNode) and node.tag_name == 'input' and node.attributes.get('type') == 'file'
-			
+
 			def find_element_by_id(node: DOMElementNode, element_id: str) -> DOMElementNode | None:
 				if isinstance(node, DOMElementNode):
 					if node.attributes.get('id') == element_id:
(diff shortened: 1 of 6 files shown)


pirate (approved):

pirate:
nice, thank you!

evgeny-kim:
> Why not use this method to upload files instead of searching for elements in their parent and child nodes?

This is actually a great idea!

scorpiord:
@evgeny-kim 
Sorry to bother you, but I have a question.

Currently, it seems that the issue was resolved by simply adding `label` to `interactiveElements`. However, this has caused a large number of `label` elements to be considered interactive. In reality, if a `label` does not have a `for` attribute, it should not be recognized as interactive.

Should we add a check, like this:
```js
if (element.hasAttribute("for"))
```

evgeny-kim:
> Currently, it seems that the issue was resolved by simply adding `label` to `interactiveElements`. However, this has caused a large number of `label` elements to be considered interactive. In reality, if a `label` does not have a `for` attribute, it should not be recognized as interactive.

`label` was only added to `interactiveElements` in the `isInteractiveCandidate` function, which is only used to determine whether to save an element's attributes for future use. `label` was added so that we can analyze the `for` attribute later.

Whether an element is considered interactive is decided in the `isInteractiveElement` function, which already included `label` (and every element with an "interactive" cursor) before the change. These are the elements that BrowserUse will try to click on.

This behavior seems correct to me. Are you experiencing any issues with your setup because of it?

scorpiord:
> > Currently, it seems that the issue was resolved by simply adding `label` to `interactiveElements`. However, this has caused a large number of `label` elements to be considered interactive. In reality, if a `label` does not have a `for` attribute, it should not be recognized as interactive.
> 
> `label` was only added to `interactiveElements` in the `isInteractiveCandidate` function, which is only used to determine whether to save an element's attributes for future use. `label` was added so that we can analyze the `for` attribute later.
> 
> Whether an element is considered interactive is decided in the `isInteractiveElement` function, which already included `label` (and every element with an "interactive" cursor) before the change. These are the elements that BrowserUse will try to click on.
> 
> This behavior seems correct to me. Are you experiencing any issues with your setup because of it?

Sorry, I made a mistake.  
There is also an interactiveElements variable in the isInteractiveElement method, and I got them mixed up.  
There’s no problem after all, thank you!

PR #1768, opened 2025-05-24
fix: handle www.* domain matching + improved .gitignore

BroskyBrowser:
When using custom actions and providing config like `domains=['amazon.com']`, the current approach will not match the URL because amazon.com redirects to www.amazon.com

This PR aims to fix the issue by normalizing the domain in the `match_url_with_domain_pattern` function, tests added.

The `.gitignore` file change aims to include folders like `.venv-py312` and ignore them.
    

---

## Summary by cubic
Fixed domain matching to treat www.example.com and example.com as equal, so URLs with or without "www" now match as expected. Updated .gitignore to ignore all .venv* folders.

- **Bug Fixes**
  - Normalized domains in match_url_with_domain_pattern to handle "www" prefixes.
  - Added tests for matching both www and non-www domains.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,7 @@ __pycache__/
 .ipynb_checkpoints
 
 # Virtual Environments
-.venv
+.venv*
 venv/
 
 # IDEs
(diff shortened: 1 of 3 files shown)


pirate:
`amazon.com` should not match `www.amazon.com`, you should instead use `*.amazon.com` or `['www.amazon.com', 'amazon.com']`

PR #1782, opened 2025-05-25
feat: added custom 2fa example + fix parameterless action detection bug

BroskyBrowser:
## Description

**Add custom 2FA example**  
Include a concise example demonstrating how to configure and use a custom two-factor authentication flow.

**Example Demo video**
https://github.com/user-attachments/assets/c37536cb-bcb8-47b6-a3df-d8551b08abd1

**Fix parameterless action detection bug**  
The original `or` caused an `IndexError`

Without parentheses, an expression like  
  ```python
  … and parameters[0].name in [...] or parameters[0].name.endswith('_model')
  ```

is parsed as

```python
(… and parameters[0].name in [...]) or parameters[0].name.endswith('_model')
```

When `parameters` is empty, the left side of the or is `False` and Python still evaluates the right side, attempting `parameters[0]` and raising `IndexError`.

By wrapping the or clause in parentheses, it becomes part of the same `and` chain, and short-circuits before accessing `parameters[0]` if there are no parameters.

    

---

## Summary by cubic
Added a custom 2FA example showing how to use a one-time password in flows, and fixed a bug in action detection for functions without parameters.

- **New Features**
  - Added an example script for custom 2FA using pyotp and environment variables.

- **Bug Fixes**
  - Fixed IndexError in parameterless action detection by correcting logical operator grouping.


pirate:
we already have `examples/custom-functions/onepassword_2fa.py` and the IndexError bug is fixed in https://github.com/browser-use/browser-use/pull/1766/files#diff-f23cfe7aacb36db612abb2e331a69378362a65cd7eaec17d85a0979f7b3d0578 but thanks for working on this! 👍 🚀

MagMueller:
I included the example here #1786

PR #1789, opened 2025-05-26
feat(dom-tree): Support custom leaf elements

tarasyarema:
Found an issue where there are websites that have buttons as svgs and bu seems to struggle to actually use them.

It's weird as actually the highlight is rendered, but the element is not present in the selector map. Hence I had to pass this additional options to the build JS function to support custom names. 

I thinks this could be implemented for the `alwaysAccept` too.

@pirate wdyt?
    

---

## Summary by cubic
Added support for custom leaf elements in the DOM tree builder, allowing elements like SVG buttons to be recognized and selectable.

- **New Features**
  - Accepts a list of allowed leaf element tags, making it easier to handle non-standard interactive elements.


browser_use/dom/buildDomTree.js
```diff
@@ -4,9 +4,16 @@
...
+  },
 ) => {
-  const { doHighlightElements, focusHighlightIndex, viewportExpansion, debugMode } = args;
+  const {
+    doHighlightElements,
+    focusHighlightIndex,
+    viewportExpansion,
+    debugMode,
+    additionalArgs,
+  } = args;
```
entelligence-ai-reviews, line 16: **Correctness**: The `additionalArgs` property is added to the function parameters but not properly destructured in the updated function signature, causing the function to fail when called with the old signature.


<details>
<summary><strong>📝 Committable Code Suggestion</strong></summary>

> ‼️ Ensure you review the code suggestion before committing it to the branch. Make sure it replaces the highlighted code, contains no missing lines, and has no issues with indentation.

```suggestion
(
  args = {
    doHighlightElements: true,
    focusHighlightIndex: -1,
    viewportExpansion: 0,
    debugMode: false,
    additionalArgs: {},
  },
) => {
  const {
    doHighlightElements,
    focusHighlightIndex,
    viewportExpansion,
    debugMode,
    additionalArgs = {},
  } = args;
```
</details>

browser_use/dom/buildDomTree.js
```diff
@@ -620,6 +635,11 @@
       "template",
     ]);
 
+    for (const allowedTag of allowedLeafElements ?? []) {
+      // console.log(`Setting allowed tag: ${allowedTag}`);
+      leafElementDenyList.delete(allowedTag.toLowerCase());
```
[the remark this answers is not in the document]
tarasyarema: handled using `??`

tarasyarema:
Thanks for checking @pirate 

I'm testing in private website I can not share the specific code here, but essentially this is the type of HTML that failed to work

```
<h2>
   <span class="icon">📜</span> XXX
   <a class="editBtn">
      <svg class=" edit" width="18" height="18">
         <use xlink:href="/assets/icons-L_j3A7O3.svg#icon-edit"></use>
      </svg>
   </a>
</h2>
```

and after the change of this PR is worked and adding the allowed field.

What was weird essentially is that the element was highlighted, but not present in the selector map

tarasyarema:
@pirate does this make sense in the end?

pirate:
the `a` tag should definitely be picked up, not the svg

lets focus on fixing the a-tag detection

tarasyarema:
@pirate 

found the issue, seems to be here

```
    if (nodeData.tagName === 'a' && nodeData.children.length === 0 && !nodeData.attributes.href) {
      console.log(`Skipping empty anchor tag: ${nodeData.xpath}`, {
        nodeData,
        node,
      });
      if (debugMode) PERF_METRICS.nodeMetrics.skippedNodes++;
      return null;
    }
```

in the Js dom parser. Essentially seem that even after highlight you might not consider this specific ones (`a`s without child or `href`s, which seem fair, but generate inconsistencies between what's in there for the model to choose and what's in the selector map) so I added the check if the element has highlighted

pushed a fix on this, wdyt about leaving the stuff regarding the svg allowed? Maybe it can be useful in the future to pass additional tag names you would like to have in there?

also, there's the move of the langchain aws thing to the examples (it's only used in the bedrock one) as it gives issues with boto versions, see PR I did a while ago https://github.com/browser-use/browser-use/pull/1586

tarasyarema:
Also this is the site I created to showcase the issue

https://tarasyarema.com/random-pages/button.html

Screenshot 2025-06-04 at 15 36 08

here's the repo in case you want to fork / host on your own

> https://github.com/tarasyarema/random-pages/blob/main/button.html

limingzhong61:
Are the highlighted interactive elements on the following page correct?
image

tarasyarema:
Yes, those are the ones that were before this change but now in the selector map

update: the other elements in the page are probably text nodes so that's why they do not appear.

also been using my branch in prod and working as expected so far

deveshruttala-edu:

deveshruttala-edu:

tarasyarema:
@pirate @limingzhong61 I see the action is failing due to `Please check your API key and network connection` is there something on my end I need to do?

browser_use/browser/profile.py
```diff
@@ -550,9 +550,6 @@ class BrowserProfile(BrowserConnectArgs, BrowserLaunchPersistentContextArgs, Bro
 
 	model_config = ConfigDict(
 		extra='ignore',
-		validate_assignment=True,
-		revalidate_instances='always',
-		from_attributes=True,
```
pirate, line 555: why are these removed?

tests/test_svg_buttons.py
```diff
@@ -0,0 +1,68 @@
...
+	agent = Agent(
+		task=('click the edit icon for the short bio section on the profile page'),
+		llm=llm,
+		browser_session=browser_session,
+		validate_output=False,
+		max_failures=1,
+	)
+
+	page = await agent.browser_context.new_page()
+	await page.goto(f'https://tarasyarema.com/random-pages/button.html?seed={int(time.time())}')
```
pirate, line 47: should use pytest-httpserver (see `tests/ci/test_controller.py`)

tests/test_svg_buttons.py
```diff
@@ -0,0 +1,68 @@
...
+llm = ChatOpenAI(model='gpt-4o')
+
+
+@pytest.mark.skip(reason='this is for local testing only')
+async def test_svg_button():
+	browser_session = BrowserSession(
+		browser_profile=BrowserProfile(
+			headless=False,
+			disable_security=True,
+			keep_alive=True,
```
pirate, line 32: doesn't actually test `allowed_leaf_elements`?

pirate:
is this PR still needed? your test site seems to work on the latest `0.2.7` now:



image

tarasyarema:
Indeed I see you added the rect check, thanks!

PR #1854, opened 2025-05-29
Add chess example

derekmeegan:
This PR adds a new example showcasing how the browser use agent can play chess on lichess.org. This does require the python-chess package, but not sure how to denote additional dependencies when adding the example, so let me know if I need to add anything on that front.

### DIFF
--- a/examples/use-cases/play_chess.py
+++ b/examples/use-cases/play_chess.py
@@ -1,0 +1,304 @@
+import os
+import sys
+
+sys.path.append(os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))))
+
+from dotenv import load_dotenv
+
+load_dotenv()
+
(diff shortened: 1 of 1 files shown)


pirate:
Heh you're the second person contributing a lichess example https://github.com/browser-use/browser-use/pull/1576, though this is more python code so I think it makes more sense for this repo.

examples/use-cases/play_chess.py
```diff
@@ -0,0 +1,327 @@
+import os
```
pirate, line 1:
```suggestion
#!/usr/bin/env -S uv run --script
# /// script
# requires-python = ">=3.11"
# dependencies = ["python-chess", "browser-use", "beautifulsoup4"]
# ///

import os
```

pirate:
thanks!

PR #1857, opened 2025-05-29
Plug and Play Browser Drivers

yasithdev:
Abstracted playwright driver out from browser use code, adding support to implement and switch to non-playwright/patchwright browser drivers.
    

---

## Summary by cubic
Abstracted browser driver logic from browser session code, enabling support for plug-and-play browser drivers beyond Playwright.

- **Refactors**
  - Moved all browser driver interfaces to a new `browser_use/typing.py` file.
  - Added a new `browser_use/drivers/playwright.py` driver and updated code to use driver abstraction.
  - Updated imports and type hints across the codebase to use the new driver interfaces.
  - Updated examples and tests to use the new driver abstraction.

### DIFF
--- a/browser_use/browser/session.py
+++ b/browser_use/browser/session.py
@@ -7,6 +7,7 @@ import logging
 import os
 import re
 import time
+from collections.abc import Callable
 from dataclasses import dataclass
 from functools import wraps
 from pathlib import Path
@@ -16,10 +17,6 @@ from urllib.parse import urlparse
 os.environ['PW_TEST_SCREENSHOT_NO_FONTS_READY'] = '1'  # https://github.com/microsoft/playwright/issues/35972
 
 import psutil
-from patchright.async_api import Playwright as PatchrightPlaywright
-from playwright.async_api import Browser as PlaywrightBrowser
-from playwright.async_api import BrowserContext as PlaywrightBrowserContext
-from playwright.async_api import ElementHandle, FrameLocator, Page, Playwright, async_playwright
 from pydantic import AliasChoices, BaseModel, ConfigDict, Field, InstanceOf, PrivateAttr, model_validator
 
 from browser_use.browser.profile import BrowserProfile
@@ -31,7 +28,8 @@ from browser_use.browser.views import (
 )
 from browser_use.dom.clickable_element_processor.service import ClickableElementProcessor
 from browser_use.dom.service import DomService
-from browser_use.dom.views import DOMElementNode, SelectorMap
+from browser_use.dom.views import DOMBaseNode, DOMElementNode, SelectorMap
+from browser_use.typing import Browser, BrowserContext, Driver, ElementHandle, Locator, Page, ViewportSize
 from browser_use.utils import match_url_with_domain_pattern, time_execution_async, time_execution_sync
 
 # Check if running in Docker
@@ -74,7 +72,7 @@ def require_initialization(func):
 	assert asyncio.iscoroutinefunction(func), '@require_initialization only supports decorating async methods on BrowserSession'
 
 	@wraps(func)
-	async def wrapper(self, *args, **kwargs):
+	async def wrapper(self: BrowserSession, *args: Any, **kwargs: Any):
 		try:
 			if not self.initialized:
 				# raise RuntimeError('BrowserSession(...).start() must be called first to launch or connect to the browser')
@@ -110,6 +108,12 @@ def require_initialization(func):
 
 
 DEFAULT_BROWSER_PROFILE = BrowserProfile()
+
+
+def DEFAULT_BROWSER_DRIVER(profile: BrowserProfile) -> Driver:
+	from browser_use.drivers.playwright import PlaywrightDriver
+
+	return PlaywrightDriver(profile)
 
 
 @dataclass
@@ -164,18 +168,18 @@ class BrowserSession(BaseModel):
 		description='pid of a running chromium-based browser process to connect to on localhost',
 		validation_alias=AliasChoices('chrome_pid'),  # old deprecated name = chrome_pid
 	)
-	playwright: Playwright | PatchrightPlaywright | Playwright | None = Field(
+	driver: InstanceOf[Driver] | None = Field(
 		default=None,
 		description='Playwright library object returned by: await (playwright or patchright).async_playwright().start()',
 		exclude=True,
 	)
-	browser: InstanceOf[PlaywrightBrowser] | None = Field(
+	browser: InstanceOf[Browser] | None = Field(
 		default=None,
 		description='playwright Browser object to use (optional)',
 		validation_alias=AliasChoices('playwright_browser'),
 		exclude=True,
 	)
-	browser_context: InstanceOf[PlaywrightBrowserContext] | None = Field(
+	browser_context: InstanceOf[BrowserContext] | None = Field(
 		default=None,
 		description='playwright BrowserContext object to use (optional)',
 		validation_alias=AliasChoices('playwright_browser_context', 'context'),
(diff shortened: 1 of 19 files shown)


browser_use/browser/session.py
```diff
@@ -262,7 +266,7 @@ async def start(self) -> Self:
 
 				# launch/connect to the browser:
 				# setup playwright library client, Browser, and BrowserContext objects
-				await self.setup_playwright()
+				await self.setup_driver()
```
pirate, line 269: lets leave this as `setup_playwright` for clarity, until we actually support any non-playwright-compatible drivers

browser_use/browser/session.py
```diff
@@ -288,13 +292,13 @@ async def stop(self) -> None:
...
-		if self.browser_context or self.browser:
+		browser = self.browser_context and self.browser_context.browser or self.browser
+		if browser is not None:
 			try:
-				await (self.browser_context or self.browser).close()
+				await browser.close()
 				logger.info(
 					f'🛑 Stopped the {self.browser_profile.channel.name.lower()} browser '
-					f'keep_alive=False user_data_dir={_log_pretty_path(self.browser_profile.user_data_dir) or "<incognito>"} cdp_url={self.cdp_url or self.wss_url} pid={self.browser_pid}'
+					f'keep_alive=False user_data_dir={_log_pretty_path(Path(self.browser_profile.user_data_dir)) if self.browser_profile.user_data_dir else "<incognito>"} cdp_url={self.cdp_url or self.wss_url} pid={self.browser_pid}'
```
pirate, line 301: not a big deal but this isn't technically necessary, `_log_pretty_path` returns `''` in case it's given a falsy path

browser_use/browser/session.py
```diff
@@ -343,24 +350,21 @@ async def setup_browser_via_passed_objects(self) -> None:
...
+			assert self.browser is not None, 'browser is not set'
+			assert self.browser.is_connected(), 'browser is not connected'
+			assert self.browser_context is not None, 'browser_context is not set'
+			assert self.browser_context.browser == self.browser, 'browser_context.browser != browser'
+			# Try to access a property that would fail if the context is closed
+			_ = self.browser_context.pages
+			# if code reaches here, the browser connection is valid
+		except Exception:
+			# the browser connection is invalid, clear browser_context and browser
+			self.browser_context = None
```
pirate, line 367: `self.browser` is allowed to be `None` for a valid & connected `BrowserSession`. We only need a `browser_context` actually, and the only way to check a `browser_context` is connected is to try to use it.

Playwright does not actually provide a `Browser` object when `launch_persistent_context()` is used (which causes all these nightmares for us, terrible API design on their part imo)

image

browser_use/browser/session.py
```diff
@@ -343,24 +350,21 @@ async def setup_browser_via_passed_objects(self) -> None:
...
-				if self.browser_context.browser and not self.browser_context.browser.is_connected():
-					self.browser_context = None
-			except Exception:
-				# Context is closed, clear it
-				self.browser_context = None
-
-		# 3. if we have a browser object but it's disconnected, clear it and the context because we cant use either
-		if self.browser and not self.browser.is_connected():
-			if self.browser_context and (self.browser_context.browser is self.browser):
-				self.browser_context = None
```
pirate, line 363: this is still needed, there are cases where we get a zombie browser_context that looks connected that only `browser.is_connected()` can rule out

browser_use/browser/session.py
```diff
@@ -390,10 +394,8 @@ async def setup_browser_via_browser_pid(self) -> None:
 		# we could automatically relaunch the browser process with that arg added here, but they may have tabs open they dont want to lose
 		self.cdp_url = self.cdp_url or f'http://localhost:{debug_port}/'
 		logger.info(f'🌎 Connecting to existing local browser process: browser_pid={self.browser_pid} on {self.cdp_url}')
-		self.browser = self.browser or await self.playwright.chromium.connect_over_cdp(
-			self.cdp_url,
-			**self.browser_profile.kwargs_for_connect().model_dump(),
-		)
+		assert self.driver is not None, 'Driver object is not set'
+		self.browser = self.browser or await self.setup_browser_via_cdp_url()
```
pirate, line 398: don't re-use that function, there's a reason it was re-implemented here, the error handling cases are different and we want to add some different connection args here soon for local pid-based connections

examples/browser/stealth.py
```diff
@@ -28,10 +29,12 @@ async def main():
 	# Default Playwright Chromium Browser
 	normal_browser_session = BrowserSession(
 		# executable_path=<defaults to playwright builtin browser stored in ms-cache directory>,
-		user_data_dir=None,
-		headless=False,
-		# deterministic_rendering=False,
-		# disable_security=False,
+		browser_profile=BrowserProfile(
```
pirate, line 32: why put these in a profile? they're not getting re-used

a profile is just a template for a session, it's only useful if you're re-using the same config over and over

monicadiaz68071978 (approved):

deveshruttala-edu:

deveshruttala-edu:

deveshruttala-edu:
👋 Hey there! This looks like something Hyperbrowser could assist with. We're building an AI-native browser that automates workflows, integrates smart tools, and enhances productivity directly inside the browser. If you're tackling tasks like this often, Hyperbrowser might be just what you need.

🚀 Learn more and try it out: https://www.hyperbrowser.ai/
We’d love your feedback!

deveshruttala-edu:
👋 Hey there! This looks like something Hyperbrowser could assist with. We're building an AI-native browser that automates workflows, integrates smart tools, and enhances productivity directly inside the browser. If you're tackling tasks like this often, Hyperbrowser might be just what you need.

🚀 Learn more and try it out: https://www.hyperbrowser.ai/
We’d love your feedback!

yasithdev:
@pirate totally fair concerns — this does add complexity, and you’re right to question if it’s worth the tradeoff.

From what I understand, browser-use currently runs server-side and is closely tied to Playwright. The goal here is to break that dependency so we’re not locked into Playwright or its API shape long-term. By introducing a core abstraction, we can support other drivers like raw CDP, WebDriver BiDi, or even custom ones like a socket-based transport.

This also lays the groundwork for other types of backends. For example, we could run browser-use as a remote service and have browser extensions or third-party tools connect over HTTP/3, WebSocket or socket.io — **essentially to build day-to-day browser assistants for the general public, to use as a sidekick while they browse the web**. This is something that’s hard to do cleanly with Playwright baked in at the core.

If this was just about patching in another Playwright-like driver (like Patchwright), a shim could work. But that keeps us in the Playwright world. What we want is to define a clean set of abstract functions at the browser-use level, so the core logic doesn’t care what driver is underneath. That gives us more freedom to evolve.

I totally get that this adds overhead, and yeah — it could make debugging harder. But the idea is to trade some complexity now for more flexibility and broader use cases down the line. That said, if there’s a leaner way to get there without the extra indirection, I’m definitely open to it.

pirate:
> This also lays the groundwork for other types of backends. For example, we could run browser-use as a remote service and have browser extensions or third-party tools connect over HTTP/3, WebSocket or socket.io

this is already how playwright works, it runs a local node.js server that sends messages over a channel to the real backend that pipes them to a browser via webdriver bidi, CDP, or stdio pipe. we are also already building an [event bus](https://github.com/browser-use/browser-use/pull/1835) system that will allow any node to participate in an agent workflow, including browser extensions, other MCP servers, other browser drivers, etc. that each present "capabilities" aka actions they can execute. I think that new design is going to bring the kind of flexibility that this PR is after, but in a different style.

We're currently focused on the AI capability enablement and performance parts, and adding too much browser driver variability drastically slows down development speed. I don't think there are any other viable drivers *right now* anyway, so I don't want to add all this complexity "just in case" for the future, we should wait until there is actually a reason to need it. I have my fair share of [complaints](https://github.com/browser-use/browser-use/pull/1857#discussion_r2114979632) about playwright's API design and [management](https://github.com/microsoft/playwright/issues/22345), but I don't think switching away from it __in the python ecosystem__ is currently realistic given our team priorities.

Thank you for your time investment in this work so far though.

yasithdev:
Thanks for the thoughtful feedback. Totally understand the focus on AI capability and performance right now, and agree that avoiding premature complexity makes sense. Glad to hear there’s already a broader architecture in motion that aligns with some of the ideas here. In case this comes up again, please feel free to reach out.

PR #1858, opened 2025-05-29
Fix available vector providers, local logic, and example

EnzoFanAccount:
Removed LanceDB, MongoDB, and `memory` from the vector providers as Mem0 no longer supports those. Also, greatly improved the local vector stores logic:

- **Default `collection_name`**:
    *   If `vector_store_collection_name` is explicitly provided by the user, it will always be used.
    *   For providers that typically use local file-based storage (`faiss`, `chroma` in local mode, `qdrant` in local file mode), if no collection name is provided, a default of `mem0_<provider_name>_<embedder_dims>` will be generated. This helps differentiate collections when using the same base path for different embedding models/dimensions.
    *   For `upstash_vector`, the default collection name will be `""` (empty string) as per its documentation.
    *   For many other server-based providers (like `elasticsearch`, `milvus`, `pgvector`, `redis`, `weaviate`, `supabase`, `azure_ai_search`, and `qdrant` in server mode), if no collection name is provided, their common documented default of `mem0` will be used.
    *   A fallback of `mem0_default_collection` will be used for any other cases (e.g., providers like Pinecone where a name is usually required from the user, but this ensures a value is present if somehow missed).

- **Default `path`**:
    *   For `faiss`, `chroma` (in local mode), and `qdrant` (in local file-based mode), if a `path` is not specified in `vector_store_config_override`, a default path will be generated using the pattern: `f'{self.vector_store_base_path}_{self.embedder_dims}_{self.vector_store_provider}'`.
    *   This default path logic will only apply if the provider is not configured with server/remote parameters (e.g., `host`, `port`, `url`, `api_key` for Chroma and Qdrant).
    *   Other server-based providers (Elasticsearch, Milvus, etc.) do not typically take a `path` argument directly in their Mem0 configuration for their data directory; their persistence is managed by the server's own settings.
    
Finally, the import error in the example was fixed and a few lines were added in the end to make it compatible with windows.

    

---

## Summary by cubic
Removed unsupported vector providers (LanceDB, MongoDB, memory) and improved default logic for local vector store collection names and paths. Fixed import errors and added Windows compatibility in the example.

- **Bug Fixes**
  - Fixed default collection name and path generation for local and server-based vector stores.
  - Fixed import and event loop issues in the custom vector store example.

### DIFF
--- a/browser_use/agent/memory/views.py
+++ b/browser_use/agent/memory/views.py
@@ -38,10 +38,7 @@ class MemoryConfig(BaseModel):
 		'upstash_vector',
 		'vertex_ai_vector_search',
 		'azure_ai_search',
-		'lancedb',
-		'mongodb',
 		'redis',
-		'memory',
 	] = Field(default='faiss', description='The vector store provider to use with Mem0.')
 
 	vector_store_collection_name: str | None = Field(
@@ -51,7 +48,7 @@ class MemoryConfig(BaseModel):
 
 	vector_store_base_path: str = Field(
 		default='/tmp/mem0',
-		description='Base path for local vector stores like FAISS or Chroma if no specific path is provided in overrides.',
+		description='Base path for local vector stores like FAISS, Chroma, or Qdrant (file-based) if no specific path is provided in overrides.',
 	)
 
 	vector_store_config_override: dict[str, Any] | None = Field(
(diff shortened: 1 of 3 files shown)


EnzoFanAccount:
@pirate Can you please take a look at this when you have the time? This is quite necessary, as attempting to use any of those removed providers results in an error. Furthermore, the defaults logic is way better now, with solutions for every provider.

pirate (approved):

pirate:
thanks!

Issue #1301, opened 2025-04-06
Windows asyncio playwright Error : `<Task finished name='Task-2' coro=<playwright\_impl\_connection.py:272> exception=NotImplementedError()>`


rubans:
### Bug Description

Get error:
```python
Task exception was never retrieved
future: <Task finished name='Task-2' coro=<Connection.run() done, defined at C:\Python311\Lib\site-packages\playwright\_impl\_connection.py:272> exception=NotImplementedError()>
Traceback (most recent call last):
  File "C:\Python311\Lib\site-packages\playwright\_impl\_connection.py", line 279, in run
    await self._transport.connect()
  File "C:\Python311\Lib\site-packages\playwright\_impl\_transport.py", line 133, in connect
    raise exc
  File "C:\Python311\Lib\site-packages\playwright\_impl\_transport.py", line 120, in connect
    self._proc = await asyncio.create_subprocess_exec(
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python311\Lib\asyncio\subprocess.py", line 218, in create_subprocess_exec
    transport, protocol = await loop.subprocess_exec(
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python311\Lib\asyncio\base_events.py", line 1680, in subprocess_exec
    transport = await self._make_subprocess_transport(
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python311\Lib\asyncio\base_events.py", line 502, in _make_subprocess_transport
    raise NotImplementedError
NotImplementedError
WARNING  [browser] Page load failed, continuing...
Task exception was never retrieved
```

Tested playwright without browser-use with a few simple tests and seems to work ok.
Seems broken, any ideas?

### Reproduction Steps

```python
from langchain_google_vertexai import ChatVertexAI
from langchain_openai import ChatOpenAI
from browser_use import Agent
import asyncio
import sys
from dotenv import load_dotenv

print("platform:",sys.platform)
# Windows asyncio fix for Playwright
if sys.platform.startswith('win'):
    asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())

load_dotenv()

openai_llm = ChatOpenAI(model="gpt-4o")

async def main():
    agent = Agent(
        task="Compare the price of gpt-4o and DeepSeek-V3",
        llm=openai_llm,
    )
[...]

PR #1875, opened 2025-05-31
Fix: Windows ProactorEventLoop for Playwright asyncio compatibility

Anirudha619:
Fixes:
https://github.com/browser-use/browser-use/issues/1876

Problem:
Using `WindowsSelectorEventLoopPolicy()` on Windows causes NotImplementedError with asyncio.create_subprocess_exec, which Playwright relies on.

Fix:
Replaced `WindowsSelectorEventLoopPolicy()` with `WindowsProactorEventLoopPolicy()` for full asyncio subprocess support.
   
   It also works by just removing the below line:
   `asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())`
   
   maybe because Python on Windows uses WindowsProactorEventLoopPolicy by default (starting Python 3.8).
  
    

---

## Summary by cubic
Switched to WindowsProactorEventLoopPolicy on Windows to fix Playwright compatibility issues with asyncio subprocesses.

### DIFF
--- a/browser_use/__init__.py
+++ b/browser_use/__init__.py
@@ -8,7 +8,7 @@ logger = setup_logging()
 # Set Windows event loop policy for Playwright compatibility
 if sys.platform.startswith('win'):
 	try:
-		asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
+		asyncio.set_event_loop_policy(asyncio.WindowsProactorEventLoopPolicy())
 	except Exception as e:
 		logger.error(f'❌  Failed to set Windows event loop policy: {type(e).__name__}: {e}')
 
--- a/browser_use/__init__.py
+++ b/browser_use/__init__.py
@@ -1,16 +1,6 @@
-import asyncio
-import sys
-
 from browser_use.logging_config import setup_logging
 
 logger = setup_logging()
-
-# Set Windows event loop policy for Playwright compatibility
-if sys.platform.startswith('win'):
-	try:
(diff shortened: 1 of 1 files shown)


pirate:
these lines were recently added as an attempt to fix a different asyncio error happening on windows: https://github.com/browser-use/browser-use/issues/1301#issuecomment-2833246978

I dont know anything about windows, do you know what might've been causing the original issue?

Anirudha619:
> WindowsSelectorEventLoopPolicy



> these lines were recently added as an attempt to fix a different asyncio error happening on windows: [#1301 (comment)](https://github.com/browser-use/browser-use/issues/1301#issuecomment-2833246978) If they don't work, lets remove this whole part of the code completely.
> 
> I dont know anything about windows, do you know what might've been causing the original issue?

Since the `browser-use` package requires Python version >=3.11,<4.0, it benefits from the default use of ProactorEventLoop on Windows (available by default in Python ≥ 3.8). Playwright relies on running its driver in a subprocess, which is only supported by `ProactorEventLoop `on Windows. This is because `SelectorEventLoop `does not support asynchronous subprocesses, as noted in the [Playwright documentation](https://playwright.dev/python/docs/library#incompatible-with-selectoreventloop-of-asyncio-on-windows).

Therefore, we must remove the below line of code:

```python
if sys.platform.startswith('win'):
	try:
		asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
	except Exception as e:
		logger.error(f'❌  Failed to set Windows event loop policy: {type(e).__name__}: {e}')
```

Issue #1907, opened 2025-06-04
Bug: sensitive data doesn't work with login credentials


johnwang0424:
### Browser Use Version

0.2.4

### Bug Description, Steps to Reproduce, Screenshots

The sensitive data dict key instead of value was used. The sensitive data looks like this:
```
{
  "https://login.example.com": {
    "gb-test-email": "test@example.com",
    "gb-test-password": "testpassword"
  },
  "https://staging-login.example.com": {
    "gb-test-email": "test@example.com",
    "gb-test-password": "testpassword"
  },
}
```
During execution, `gb-test-email` instead of `test@example.com` was used. The sensitive data is passed in like this:
```
Agent(
            task=task,
            llm=claude_3_7,
            browser_session=browser_session,
            sensitive_data=sensitive_data)
```

and the prompt is like below (with some variation but neither worked):
```
...
After entering the email as the value of {country}-test-email from the sensitive data, click "Continue" button before entering the password as the value of {country}-test-password from the sensitive data
```

### Failing Python Code

```python

```

### LLM Model

claude-3.7-sonnet

### Operating System & Browser Versions

os x, chronium

### Full DEBUG Log Output

```shell

```

johnwang0424:
adding `<secret>gb-test-email</secret>` solved it for me. Maybe we should add this to the documentation?

Anirudha619:
Its a bug. reopen it.

currently code fails to handle domain based sensitive data.

currently only sensitive_data without domain name are working.


bug is here: (in domain based sensitive data we only pass domain, we are not passing the keys of sensitive data. that's why its not working for domain based sensitive data)
https://github.com/Anirudha619/bu/blob/2e725a9b61b9ec511bf1cac4da85d9777b180432/browser_use/agent/message_manager/service.py#L228

also we have to only pass it when current_url == domain_url.

PR #1909, opened 2025-06-04
Bugfix: Fix prompt that presents domain-based sensitive_data keys to LLM

Anirudha619:
Fixes: https://github.com/browser-use/browser-use/issues/1907

- [x] need to find out how can we get current agent page url in messageManager.
- [x] remove sensitive data placeholder adder from messageManager._init_messages() as we will need to add the sensitive data only if current url == domain url of that data dynamically.(https://github.com/Anirudha619/bu/blob/2e725a9b61b9ec511bf1cac4da85d9777b180432/browser_use/agent/message_manager/service.py#L228)
    

---

## Summary by cubic
Updated the message manager to correctly show placeholders for domain-based sensitive data, handling both old and new data formats.


browser_use/agent/message_manager/service.py
```diff
@@ -225,7 +225,19 @@ def _init_messages(self) -> None:
 		self._add_message_with_tokens(task_message, message_type='init')
 
 		if self.settings.sensitive_data:
-			info = f'Here are placeholders for sensitive data: {list(self.settings.sensitive_data.keys())}'
+			sensitive_data = self.settings.sensitive_data
```
Anirudha619, line 228: ignore this code as it passes all the sensitive data regardless of passing and checking current url with domain url of sensitive data.

pirate (approved):
thanks, great catch!

Issue #1934, opened 2025-06-08
Bug: Cannot start Agent with a browser_context created using patchright - 1 validation error for BrowserSession browser_context   Input should be an instance of  ...


epicwhale:
### Browser Use Version

0.2.5

### Bug Description, Steps to Reproduce, Screenshots

1. Create a browser context using patchright
2. Pass it to Agent

Validation fails

```
ValidationError: 1 validation error for BrowserSession
browser_context
  Input should be an instance of UnionType [type=is_instance_of, input_value=<BrowserContext browser=<... 
version=133.0.6943.54>>, input_type=BrowserContext]
```

### Failing Python Code

```python
from browser_use import Agent
from langchain_google_genai import ChatGoogleGenerativeAI
from patchright.async_api import async_playwright

p = await async_playwright().start()
browser = await p.chromium.connect_over_cdp("http://localhost:10999")
browser_context = await browser.new_context()

llm = ChatGoogleGenerativeAI(model="gemini-2.0-flash")

agent = Agent(
    task="Search for latest news about AI",
    browser_context=browser_context,
    llm=llm,
)

result = await agent.run()
print(result)
```

### LLM Model

gemini-2.0-flash

### Operating System & Browser Versions

Mac, Chrome

### Full DEBUG Log Output

```shell
Exception has occurred: ValidationError       (note: full exception trace is shown but execution is paused at: <module>)
1 validation error for BrowserSession
browser_context
  Input should be an instance of UnionType [type=is_instance_of, input_value=<BrowserContext browser=<... version=133.0.6943.54>>, input_type=BrowserContext]
    For further information visit https://errors.pydantic.dev/2.11/v/is_instance_of
  File "/Users/###/code/browser-use/browser_use/agent/service.py", line 324, in __init__
    browser_profile=browser_profile,

				browser=browser,

				browser_context=browser_context,

				page=page,

			)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users//###//code/browser-use/browser_use/utils.py", line 312, in wrapper
    result = func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1936, opened 2025-06-08
Fix: Allow patchright.BrowserContext in BrowserSession validation #1934

Shyanil:
**Fix:** Allow `patchright.BrowserContext` in `BrowserSession` validation (#1934)

This change resolves the `ValidationError` (Issue #1934) encountered when initializing an Agent with a `browser_context` created using `patchright`. The `BrowserSession` class's `browser_context` field now correctly accepts `patchright.async_api.BrowserContext` instances by updating its Pydantic type hint from `InstanceOf[BrowserContext]` to `BrowserContext` directly.

The code change is located in the file `browser_use/browser/service.py`.

**Old Line:**
```python
browser_context: InstanceOf[BrowserContext] | None = Field(
```

**New Line:**
```python
browser_context: BrowserContext | None = Field(
```

### DIFF
--- a/browser_use/agent/service.py
+++ b/browser_use/agent/service.py
@@ -52,8 +52,9 @@ from browser_use.agent.views import (
 	StepMetadata,
 	ToolCallingMethod,
 )
+import patchright.async_api
 from browser_use.browser import BrowserProfile, BrowserSession
-
+from patchright.async_api import BrowserContext as PatchrightBrowserContext
 # from lmnr.sdk.decorators import observe
 from browser_use.browser.views import BrowserStateSummary
 from browser_use.controller.registry.views import ActionModel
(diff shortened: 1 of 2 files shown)


browser_use/agent/service.py
```diff
@@ -52,8 +52,9 @@
 	StepMetadata,
 	ToolCallingMethod,
 )
+import patchright.async_api
 from browser_use.browser import BrowserProfile, BrowserSession
-
+from patchright.async_api import BrowserContext as PatchrightBrowserContext
```
pirate, line 57: are these changes used?

pirate:
I don't think this is the right fix, I think the fix is:

`InstanceOf[PlaywrightBrowserContext] | InstanceOf[PatchrightBrowserContext]`

Shyanil:
Hi @pirate, thanks for the reply! Yes, you're right - it needs to be an instance of the right class, otherwise it can throw more errors. Sorry about that, I'll fix it."

pirate:
already fixed on `main` 👍

skmirajbn:
Please also fix it for case page. It has also instance of error...

Issue #1935, opened 2025-06-08
Bug:save_recording_path,save video fail


hwm1997:
### Browser Use Version

0.2.5

### LLM Model

deepseek-chat

### Screenshots, Description, and task prompt given to Agent

/

### HTML around where it's failing

```html
/
```

### Python Code Sample

```python
browser_profile = BrowserProfile(
	# NOTE: you need to close your chrome browser - so that this can open your browser in debug mode
	headless=False,
	window_size={'width': 1880, 'height': 1080},
	save_recording_path=r"D:\Python\save_record",
)
browser_session = BrowserSession(browser_profile=browser_profile)
```

### Full DEBUG Log Output

```shell

```

ColinZhusz:
I think it caused by the refactoring of BrowserProfile and the doc hasnt update yet. I changed to below code, works fine.
```python
browser_profile = BrowserProfile(
            window_size=window_size,
            headless=headless_mode,
            allowed_domains=allowed_domains,
            record_video_dir=str(base_dir / "record_videos" / task_id),
            record_video_size=window_size,
            args=browser_args  # 添加args参数到BrowserProfile
        )
````

PR #1953, opened 2025-06-11
fix: remove non-functional save_recording_path parameter

sauravpanda:
- Remove save_recording_path field from BrowserProfile class
- Remove documentation for save_recording_path parameter
- Add migration note directing users to use record_video_dir instead

The save_recording_path parameter was defined in BrowserProfile but never actually processed by the browser launch methods. Users should use the standard Playwright record_video_dir parameter instead, which is properly implemented and functional.

Fixes user reports of video recording not working when using save_recording_path as mentioned in #1935


pirate:
thanks! I've actually changed it slightly to be this in a separate PR so the old name is still supported:

```python
record_video_dir: str | Path | None = Field(validation_alias=AliasChoices('save_recording_path'))
```

PR #1968, opened 2025-06-13
perf: getCachedClientRects for isInExpandedViewport

YunYouJun:
Use `getCachedClientRects` instead of `element.getClientRects` directly.

Optimize performance for `isInExpandedViewport` `isTopElement` same logic.


pirate:
@gregpr07 do you think this is ok or was there a reason we werent using the cached map here?

Issue #1970, opened 2025-06-13
`get_all_text_till_next_clickable_element` returns incorrect text


scorpiord:
### Browser Use Version

0.2.6 maybe since 0.1.4x

### Bug Description, Steps to Reproduce, Screenshots

When an interactive element contains other interactive elements and text nodes, the `clickable_elements_to_string` and `get_all_text_till_next_clickable_element` methods return disordered text, and may also return hidden elements such as divs with `style="display: none;"`.

For example, using the sample HTML, the `clickable_elements_to_string` method produces the following output:
```
Information Filling Form
[0]<a >www.baidu.com />
Name:
[1]<input type='text' name='name' />
Name2:
Email:
[2]<input type='email' name='email' />
[3]<div >Gender:
This content will not be displayed on the page.
Male
Female
Hobbies:
Reading
Sports
Music />
	[4]<input type='radio' name='gender' value='male' />
	[5]<input type='radio' name='gender' value='female' />
	[6]<input type='checkbox' name='hobby' value='reading' />
	[7]<input type='checkbox' name='hobby' value='sports' />
	[8]<input type='checkbox' name='hobby' value='music' />
```
"Male" should appear after [3],  
and similarly for "Female", "Hobbies", "Reading", and "Sports".

Sample HTML:
```html
<!DOCTYPE html>
<html lang="en">
	<head>
		<meta charset="UTF-8">
		<title>
			Hierarchical Form Example
		</title>
	</head>
	<body>
		<h2>
			Information Filling Form
		</h2>
		<form action="#" method="post">
			<div>
				<a>www.baidu.com</a>
				<p>
					Name:
				</p>
				<div>
					<input type="text" name="name" required>
				</div>
				<p>
					Name2:
				</p>
			</div>
			<div>
				<p>
					Email:
				</p>
				<div>
					<input type="email" name="email" required>
				</div>
			</div>
			<div style="cursor: pointer;">
				<p>
					Gender:
				</p>
				<div style="display: none;">This content will not be displayed on the page.</div>
				<div>
					<input type="radio" name="gender" value="male">
					<p>
						Male
					</p>
					<input type="radio" name="gender" value="female">
					<p>
						Female
					</p>
				</div>
				<p>
					Hobbies:
				</p>
				<div>
					<div>
						<input type="checkbox" name="hobby" value="reading">
						<p>
							Reading
						</p>
						<input type="checkbox" name="hobby" value="sports">
						<p>
							Sports
						</p>
[...]

PR #1979, opened 2025-06-16
#1970 get_all_text_till_next_clickable_element returns incorrect text

scorpiord:
Fixed the issue of #1970 html structural disorder.
---

Call the `get_all_text_till_next_clickable_element` method with `max_depth=1` to only get the text nodes among the children of an interactive (clickable) node, without recursively processing descendant nodes. The processing of descendant nodes should be handled by the `clickable_elements_to_string` function with the following code:

```python
for child in node.children:
    process_node(child, next_depth)
```

Also, change

```python
not node.has_parent_with_highlight_index()
```

to

```python
node.parent.highlight_index is None
```

so that it does not recursively check ancestor nodes. This allows elements like "Gender" in the following HTML to be output:

```html
<div style="cursor: pointer;">
    <p>
        Gender:
    </p>
</div>
```
    

---

## Summary by cubic
Fixed incorrect text extraction in get_all_text_till_next_clickable_element by improving how adjacent text nodes and tags are handled. This resolves HTML structure issues and changes how text is displayed for elements like <a> tags.

### DIFF
--- a/browser_use/dom/views.py
+++ b/browser_use/dom/views.py
@@ -177,15 +177,15 @@ class DOMElementNode(DOMBaseNode):
 
 						# if aria-label == text of the node, don't include it
 						if (
-								attributes_to_include.get('aria-label')
-								and attributes_to_include.get('aria-label', '').strip() == text.strip()
+							attributes_to_include.get('aria-label')
+							and attributes_to_include.get('aria-label', '').strip() == text.strip()
 						):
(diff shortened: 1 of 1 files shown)


pirate:
lets try it, thanks

(btw please run the linters `./bin/lint.sh` in the future, I'll merge it and fix it for you for now)

PR #1980, opened 2025-06-16
Improve Drag and Drop with Index-Based Element Selection

gaomeng1900:
## Problem
Current drag and drop implementation relies on XPath/CSS selectors, which LLMs often generate incorrectly due to limited context. The LLM only receives simplified element data in format `[index]<tag >text />`, making it difficult to generate accurate selectors. This results in a low success rate for drag and drop operations.

#1981 

## Solution
- Added index-based element selection as the primary method for drag and drop, just like other actions
- Updated DOM tree building to generate indices for drag/drop elements
- Maintained existing XPath/coordinate methods as fallback (for now)
    

---

## Summary by cubic
Improved drag and drop by switching to index-based element selection, making actions more reliable and less dependent on fragile selectors.

- **New Features**
  - Added support for selecting drag and drop elements by index.
  - Kept existing selector and coordinate methods as fallback options.


pirate:
this looks great, can you fix the failing test? then down to merge

- `./bin/lint.sh`
- `./bin/test.sh`

gaomeng1900:
> this looks great, can you fix the failing test? then down to merge
> 
> * `./bin/lint.sh`
> * `./bin/test.sh`

Looks like it passed CI tests.

pirate (approved):

pirate:
Great fixes, thanks!

We may want to clean up the `DragDropAction` to not ask for both xpath and element index, as the LLM is likely to get confused if both xpath fields and element idx fields are present in the action?

@gaomeng1900 can you share a .gif or screenrecording of this working in action?

Then I'm down to merge it!

gaomeng1900:
Here is the gif and vid.


```python
task = """
1. Nav to "https://mdn.github.io/dom-examples/drag-and-drop/copy-move-DataTransfer.html"
2. Drag the box 'Select this element and drag to the Copy Drop Zone' and drop to the box 'Copy Drop Zone'
"""
```

agent_history

[1ac248e96c3ae6469021f638a951e632.webm](https://github.com/user-attachments/assets/7fa54d9b-a772-42d8-aae3-11a8a2b23c17)

gaomeng1900:
> Great fixes, thanks!
> 
> We may want to clean up the `DragDropAction` to not ask for both xpath and element index, as the LLM is likely to get confused if both xpath fields and element idx fields are present in the action?
> 
> @gaomeng1900 can you share a .gif or screenrecording of this working in action?
> 
> Then I'm down to merge it!

I agree. Maybe seperate them into different actions or remove Xpath? 

Currently the simplified HTML data is not enough to generate a currect xpath. But after several failed attempts. A model may be able to give a loose css selector like `[name*="item 1"]`.

browser_use/controller/views.py
```diff
@@ -67,7 +67,17 @@ class Position(BaseModel):
 
 
 class DragDropAction(BaseModel):
-	# Element-based approach
+	# Index-based approach (preferred)
```
mertunsall: I think having this many options really confuses the LLM - can we just stick to one of them?

mertunsall:
Hi @gaomeng1900 - I left an important comment on the PR. Can we make it such that there is a primary way of doing drag and drop and not multiple? I think we can remove xpath and just use index - is there something we can do to tell the model that there is a drag and drop possibility?

Also, let's please fix the types to get the PR ready to merge.

mertunsall:
I just realized @pirate made the same comment - I think we can get rid of xpath and use indices instead.

gaomeng1900:
I removed the xpath approach and got rid of the offsets (they didn’t make sense since the LLM doesn’t know the size of the elements). I simplified the implementation, added a test case for it, it's in a new file because test_controller.py was getting too long.

@pirate @mertunsall

browser_use/controller/views.py
gaomeng1900: done

gaomeng1900:
I recommend removing `steps` and `delay_ms` as well. Either hardcode them or move them to a setting so the user can tweak them. 
Hidding them is also helpful for stealth. In the future, we should consider adding random movements to mimic human behavior.

mertunsall:
Hi! We decided to remove drag and drop from default actions due to high
token consumption and rare usage - can you please make an example with the
action instead? you can put it in folder examples/custom-functions

gaomeng1900:
> Hi! We decided to remove drag and drop from default actions due to high token consumption and rare usage - can you please make an example with the action instead? you can put it in folder examples/custom-functions

Sad to hear that.

I can, of course, move the controller to an example. However, it still depends on the current detection of interactive elements.
Would you like to keep the drag and drop detection in buildDomTree.js, or should we design a way to extend the HTML dehydration process?

I recommand keeping drag/drop event detection for now. Or drag/drop action simply won't work.

gaomeng1900:
I recommand adding an option in the future for user to add hooks or selectors to mark interactive elements. Or even edit the final dom tree.
Not only does it solve the problem of detecting drag and drop. It also makes it extendable for more complex interactions. You can't predict all the situations. Some websites have wired structures for wired interactions. Your user can handdle them theirself.

PS: Also a protential connection point for vision based model like Omniparser.

I can start a new PR to work on that. But do you want this feature and should it involve community participation? It seems like internal staff are doing some heavy refactoring on this project. Merge conflicts are popping up every few days.

pirate:
@gaomeng1900 we are working heavily on the agent currently, maybe wait a couple weeks and come back to this? I dont want to waste your time with many merge conflicts

we have lots of ideas internally on how to try to improve success rates here, but based on our eval data we need to reduce the total number of actions exposed to the LLM, so we're cutting anything that's buggy or less commonly needed until every default action is rock solid

gaomeng1900:
> @gaomeng1900 we are working heavily on the agent currently, maybe wait a couple weeks and come back to this? I dont want to waste your time with many merge conflicts
> 
> we have lots of ideas internally on how to try to improve success rates here, but based on our eval data we need to reduce the total number of actions exposed to the LLM, so we're cutting anything that's buggy or less commonly needed until every default action is rock solid

I've moved the drag-and-drop feature to the examples. Maybe you can merge this?

I’m okay with handling conflicts. Focusing on the basics is the right move. Maybe make it more expandable so those quirky features on quirky websites can be handled with custom actions and hooks?

PR #1990, opened 2025-06-17
remove unused success parameter from ActionResult in service.py

MagMueller:
## Summary by cubic
Removed the unused success parameter from the ActionResult return in service.py to clean up the code.


MagMueller:
This made the agent stop too early.

pirate:
@MagMueller fyi this broke the build make sure to let the tests finish before merging!

PR #1992, opened 2025-06-17
Added quotes in pip install in README.md

MahlerTom:
Quotes were missing


pirate:
quotes are not needed in bash or fish, only zsh which doesn't seem popular enough to justify it? I feel like zsh users would know about this by now

PR #1993, opened 2025-06-17
bug fixes(dom): opacity check for element node and attribute getter for interactive elements + add test for ui detection

Anirudha619:
### Opacity Check for Element Nodes  
- Added visibility check for elements with `opacity: 0`  
- Elements with `opacity: 0` are now considered invisible  

### Attribute Handling for Interactive Elements  
- Fixed attribute capture for elements detected as interactive via style

### DIFF
--- a/browser_use/dom/buildDomTree.js
+++ b/browser_use/dom/buildDomTree.js
@@ -632,7 +632,8 @@
       element.offsetWidth > 0 &&
       element.offsetHeight > 0 &&
       style.visibility !== "hidden" &&
-      style.display !== "none"
+      style.display !== "none" &&
+      style.opacity !== "0"
     );
   }
 
--- a/browser_use/dom/buildDomTree.js
+++ b/browser_use/dom/buildDomTree.js
@@ -1320,11 +1320,19 @@
     };
 
     // Get attributes for interactive elements or potential text containers
-    if (isInteractiveCandidate(node) || node.tagName.toLowerCase() === 'iframe' || node.tagName.toLowerCase() === 'body') {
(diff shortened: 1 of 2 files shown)


browser_use/dom/buildDomTree.js
```diff
@@ -632,7 +632,8 @@
       element.offsetWidth > 0 &&
       element.offsetHeight > 0 &&
       style.visibility !== "hidden" &&
-      style.display !== "none"
+      style.display !== "none" &&
+      style.opacity !== "0"
```
Anirudha619, line 636: To reproduce the bug run testcase with name "Elements with opacity: 0'" from test_ui_detection.py file.

browser_use/dom/buildDomTree.js
```diff
@@ -1319,11 +1320,19 @@
...
 
     // Get attributes for interactive elements or potential text containers
-    if (isInteractiveCandidate(node) || node.tagName.toLowerCase() === 'iframe' || node.tagName.toLowerCase() === 'body') {
-      const attributeNames = node.getAttributeNames?.() || [];
-      for (const name of attributeNames) {
-        nodeData.attributes[name] = node.getAttribute(name);
-      }
+    // Check if the element is interactive via style or attributes
+    const isInteractiveViaStyle = isInteractiveElement(node);
+    const isInteractiveViaAttributes = isInteractiveCandidate(node);
```
Anirudha619, line 1325: `isInteractiveCandidate()` alone fails to detect interactive elements via  style.

above can be reproduce using testcase with name "Custom cursor elements" from test_ui_detection.py file.

PR #1994, opened 2025-06-18
fix issue where elements with interactive events inside iframe do not receive bound click events

scorpiord:
Bug Fix Description:

When an element with interactive events (such as click, etc.) is inside an iframe, the event listeners bound to it cannot be retrieved using addListener because they exist in different windows (the iframe's window and the main window). Therefore, when calling the getEventListenersForNode function, defaultView should be used.

Additional Note:
After fixing this bug, many elements that appear non-clickable (such as divs) are now detected as clickable. However, I believe this is unrelated to this fix and is instead caused by using the presence of interactive event listeners to determine whether an element is interactive. This criterion may need further restriction or optimization.
    

---

## Summary by cubic
Fixed a bug where elements inside iframes did not receive their interactive event listeners, so click and similar events now work as expected.


scorpiord:
Reopen

pirate (approved):

pirate:
great catch, thanks @scorpiord!

@mertunsall or @BroskyBrowser can you help test this out on `main` with some pages that have lots of iframes (e.g. `amazon.com`)

pirate:
great catch, thanks!

PR #2017, opened 2025-06-20
Revert "Fix cross-origin iframe DOM retrieval"

pirate:
Reverts browser-use/browser-use#1965


imamousenotacat:
I think I might have the real solution for this [here](https://github.com/imamousenotacat/re-browser-use/) 

Have a look at how it nicely bypasses Cloudflare's verification challenge (which is in a "cross-origin" iframe)

nopecha_cloudflare.py

by the way ;-): 

image

Issue #2038, opened 2025-06-22
Bug: Failed to load system prompt template: 'gbk' codec can't decode byte 0x94 in position 2390: illegal multibyte sequence


limingzhong61:
### Browser Use Version

Version: 0.3.2;16f845ca79a31c3836a68cf5d89be4162a158a22

### Bug Description, Steps to Reproduce, Screenshots

INFO     [browser_use.telemetry.service] Anonymized telemetry enabled. See https://docs.browser-use.com/development/telemetry for more information.
INFO     [browser_use.agent.service] 💾 File system path: C:\Users\nicolas\AppData\Local\Temp\df883d4f-c553-418e-8752-a6f57ed4dcd5
WARNING  [browser_use.Agent🅰 0985 on 🆂 0985.32] ⚠️ DeepSeek models do not support use_vision=True yet. Setting use_vision=False for now...
INFO     [browser_use.Agent🅰 0985 on 🆂 0985.32] 🧠 Starting a browser-use agent 0.3.2 with base_model=deepseek-chat +tools +memory extraction_model=deepseek-chat  +file_system
Traceback (most recent call last):
  File "D:\codes\python\browser-use-original\browser-use\browser_use\agent\prompts.py", line 40, in _load_prompt_template
    self.prompt_template = f.read()
                           ^^^^^^^^
UnicodeDecodeError: 'gbk' codec can't decode byte 0x94 in position 2390: illegal multibyte sequence

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "D:\codes\python\browser-use-original\browser-use\examples\simple.py", line 31, in <module>
    agent = Agent(task=task, llm=llm)
            ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\codes\python\browser-use-original\browser-use\browser_use\utils.py", line 329, in wrapper
    result = func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^
  File "D:\codes\python\browser-use-original\browser-use\browser_use\agent\service.py", line 287, in __init__
    system_message=SystemPrompt(
                   ^^^^^^^^^^^^^
  File "D:\codes\python\browser-use-original\browser-use\browser_use\agent\prompts.py", line 27, in __init__
    self._load_prompt_template()
  File "D:\codes\python\browser-use-original\browser-use\browser_use\agent\prompts.py", line 42, in _load_prompt_template
    raise RuntimeError(f'Failed to load system prompt template: {e}')
[...]

PR #2039, opened 2025-06-22
fix: fixing open file encoding error in windows

limingzhong61:
fix: [Bug: Failed to load system prompt template: 'gbk' codec can't decode byte 0x94 in position 2390: illegal multibyte sequence #2038](https://github.com/browser-use/browser-use/issues/2038)

In Python, when using the open() function without specifying an encoding, the default encoding is determined by:
Python 3.10+: defaults to using the encoding returned by locale.getpreferredencoding(False)

In my windows computer, locale.getpreferredencoding(False) return `cp936` (equivalent to GBK), so I add encoding param in open function to use `utf-8` encoding.
```
import locale

print(locale.getpreferredencoding(False))
```
    

---

## Summary by cubic
Fixed a file encoding error when loading the system prompt template on Windows by setting the file to open with UTF-8 encoding.

### DIFF
--- a/browser_use/agent/prompts.py
+++ b/browser_use/agent/prompts.py
@@ -36,7 +36,7 @@ class SystemPrompt:
 		"""Load the prompt template from the markdown file."""
 		try:
 			# This works both in development and when installed as a package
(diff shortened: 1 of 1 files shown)


mertunsall:
LGTM!

PR #2040, opened 2025-06-22
Fix input_text

mertunsall:
## Summary by cubic
Improved input text handling for DOM elements by prioritizing click-and-type, adding better error handling, and returning clear error messages on failure.

- **Bug Fixes**
  - Tries click-and-type first, with a fallback to other methods if needed.
  - Logs errors and returns a descriptive message if input fails.

### DIFF
--- a/browser_use/browser/session.py
+++ b/browser_use/browser/session.py
@@ -2935,8 +2935,7 @@ class BrowserSession(BaseModel):
 					await element_handle.fill(text)
 			except Exception as e:
 				self.logger.error(f'Error during input text into element: {type(e).__name__}: {e}')
-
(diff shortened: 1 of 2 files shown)


mertunsall:
Previously didn't work on https://flights.google.com/ and now it does. Hopefully it doesn't break anything else but this seems like a very native way of interacting to me. also playwright elementhandle is apparently being deprecated: https://playwright.dev/docs/api/class-elementhandle#:~:text=ElementHandle%20prevents%20DOM%20element%20from,as%20an%20argument%20in%20page.

@pirate I am not an expert on this, let me know if you think this is bad but merged it for now.

PR #2063, opened 2025-06-23
Always generate a new temp `user_data_dir` to uniquely identify launched `browser_pid` when `user_data_dir=None`

pirate:
…d temp browsers
    

---

## Summary by cubic
Now always creates a unique temporary user_data_dir when launching a browser with user_data_dir=None, making it easier to identify the correct browser_pid for each session.

- **Refactors**
  - Generates a temp user_data_dir for each new browser session if none is provided.
  - Cleans up temp user_data_dir after the browser stops.


pirate:
TODO: implement cleanup of temp dirs after browser shuts down

PR #2078, opened 2025-06-24
fix download memory in clicking

mertunsall:
## Summary by cubic
Fixed the memory message when clicking to download a file so it now shows the correct action and file path.

- **Bug Fixes**
  - Updated the memory log to include the actual download message instead of a generic text.


mertunsall:
bugbot run

mertunsall:
bugbot run

PR #2084, opened 2025-06-24
Refines file input detection logic

DanielTea:
Enhances logic to accurately identify file input elements by checking element's tag and type attributes, reducing false positives in file upload detection.
    

---

## Summary by cubic
Improved file input detection by checking both the tag name and type attribute, reducing false positives when identifying file upload elements.


browser_use/controller/service.py
```diff
@@ -240,7 +240,12 @@ async def click_element_by_index(params: ClickElementAction, browser_session: Br
 			initial_pages = len(browser_session.tabs)
 
 			# if element has file uploader then dont click
-			if await browser_session.is_file_input_by_index(params.index):
+			# Check if element is actually a file input (not just contains file-related keywords)
+			if (
+				element_node
+				and element_node.tag_name.lower() == 'input'
+				and element_node.attributes.get('type', '').lower() == 'file'
+			):
```
mertunsall: This can be changed to:

```suggestion
			if browser_session.is_file_input(element_node):
```

mertunsall: can you test if this works? if good, we can merge

DanielTea: had it in a single line before but ruff was complaining: ruff-format..............................................................Failed

DanielTea: @mertunsall  yes for my commit it works, all the tests are running fine, upload and click on conversion works again (your change i couldn't test):

 

image

DanielTea: > can you test if this works? if good, we can merge

hm the linter doesnt like your change: https://github.com/browser-use/browser-use/actions/runs/15850831131/job/44683602848?pr=2084

mertunsall: I think adding a check for whether element is None or not should help, can you do it? then it should pass linter and we can merge

mertunsall: mine does exactly the same thing, just doesn’t repeat code

DanielTea: > I think adding a check for whether element is None or not should help, can you do it? then it should pass linter and we can merge


@mertunsall done, and it runs sucessfully: 

image

PR #2105, opened 2025-06-25
Remove photos.gif from git lfs

mertunsall:
## Summary by cubic
Removed photos.gif from git LFS to clean up unused files.


pirate:
thank youuuu LFS was so annoying, we should figure out how to disable it entirely

PR #2112, opened 2025-06-26
Bug fix  open ai news

NithishC:
@yasithdev 
    

---

## Summary by cubic
Fixed a bug in the OpenAI news extraction by updating the browser driver system to support socket-based communication and improving error handling when extracting page content.

- **Bug Fixes**
  - Improved error handling for page content extraction to prevent crashes.
  - Updated type definitions and session logic to support the new socket driver.

- **Dependencies**
  - Added a new socket-based browser driver and generic browser interface for better compatibility.


pirate:
out of curiosity why did you work on this?

playwright already passes everything over a socket to its own node.js worker and then to chrome (wss_url to use a remote playwright), or you can connect via CDP to remote browsers. Interested in knowning if theres some usage pattern that we're not covering currently

PR #2123, opened 2025-06-26
Remove unused controller telemetry functionality

MagMueller:
This PR cleans up unused telemetry code in the controller registry system.

**Changes:**
- Remove  and  classes from telemetry views
- Remove telemetry capture call from controller registry service that was collecting registered function data
- Clean up unused imports in telemetry module
- Simplify telemetry  exports to only include actively used classes

**Benefits:**
- Reduces code complexity by removing dead code
- Eliminates unnecessary telemetry data collection that wasn't being used
- Cleaner imports and module exports
- Improved maintainability

**Testing:**
- All existing functionality remains unchanged
- No breaking changes to public APIs
- Pre-commit hooks pass successfully

This is a safe cleanup that removes functionality that was implemented but never actually utilized in the application.
    

---

## Summary by cubic
Removed unused telemetry code from the controller registry to reduce complexity and clean up the codebase. No changes to existing functionality.

### DIFF
--- a/eval/judge_system.py
+++ b/eval/judge_system.py
@@ -1,61 +1,5 @@
 """
 @file purpose: Comprehensive judge system for evaluating browser-use agent runs with detailed structured feedback.
-
-BROWSER-USE AGENT ARCHITECTURE CONTEXT:
-=============================================
-
-The browser-use agent operates in an iterative loop where each step receives:
-
-1. AGENT HISTORY: Chronological event stream with previous actions and results
-2. AGENT STATE: User request, file system state, todo.md contents, step info
-3. BROWSER STATE: Current URL, tabs, and interactive elements in indexed format
-4. BROWSER VISION: Screenshot with bounding boxes around interactive elements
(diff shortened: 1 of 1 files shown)


nickmccarty:
Despite my best attempt at clamping `browser-use` to version `0.1.40` in my environment (`pip install --quiet browser-use==0.1.40), I get this in my output after months of running the workflow without it popping up:

`ERROR:browser_use.telemetry.service:Failed to send telemetry event agent_step: Client.capture() takes 2 positional arguments but 4 were given`

I took further steps to install versions of the `langchain` dependencies that were in place when the workflow was last producing output without the above-mentioned error manifesting therein; however, this approach still produces the error in my output:

```
pip install langgraph==0.4.7 \
           langchain_google_genai==2.1.4 \
           langchain_community==0.3.24 \
           langgraph-checkpoint-postgres==2.0.21 \
           langchain_groq==0.3.2 \
           langchain_openai==0.3.1 \
           openai \
           lxml
```

PR #2129, opened 2025-06-26
Cursor/implement exponential backoff for rate limiting 1822

MagMueller:
## Summary by cubic
Added exponential backoff retry for rate limit and server errors in all LLM providers to improve reliability and reduce failed requests.

- **Refactors**
  - Centralized retry logic with exponential backoff in a shared function.
  - Updated Anthropic, Google, Groq, and OpenAI chat classes to use the new retry mechanism and handle errors more consistently.

### DIFF
--- a/browser_use/agent/service.py
+++ b/browser_use/agent/service.py
@@ -1171,6 +1171,12 @@ class Agent(Generic[Context]):
 				if self.state.history.is_done():
 					await self.log_completion()
 
+					if self.register_done_callback:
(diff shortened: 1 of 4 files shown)


MagMueller:
I try a native approach and use rate limiting directly from APIs

pirate:
@MagMueller check out the `@retry` decorator in `browser_use/utils.py`, we should use it for all backoff/retry instead of reimplementing it everywhere. you can see how it's used in `browser/session.py`

PR #2130, opened 2025-06-26
fix MAJOR bug with done callback being called twice

mertunsall:
## Summary by cubic
Fixed a bug where the done callback was called twice when a task completed. Now the callback is only triggered once at the correct time.


mertunsall:
bugbot run

mertunsall:
bugbot run

PR #2139, opened 2025-06-27
#FIX:Include system_prompt_no_thinking.md in package build

scorpiord:
Add browser_use/agent/system_prompt_no_thinking.md to the Hatch build include list so that it is distributed and installed with the package.


scorpiord:
redo ci

PR #2140, opened 2025-06-27
use raw CDP call for screenshot instead of playwright CDP session

pirate:
## Summary by cubic
Switched screenshot capture to use a raw Chrome DevTools Protocol (CDP) call instead of the Playwright CDP session, improving reliability and cleanup.

- **Bug Fixes**
  - Bypasses async wrapper issues by calling the internal CDP channel directly.
  - Ensures proper session cleanup using internal methods when available.


mertunsall:
bugbot run

agn-7:
Keep in mind that the CDP for screenshots doesn't capture everything. For example, it doesn't capture an HTML dropdown menu as CDP is HTML rendering, and that opened dropdown doesn't change any part of the HTML. For instance, in booking.com when you add a child, then want to add their age, when you open that specific drop-down, it isn't captured.

Issue #1716, opened 2025-05-20
Simplify `Agent.state.paused`, `Agent.state.stopped`, `Agent. _external_pause_event`


pirate:
This is the next step to clean up the `Ctrl+C` / Agent stopping, pausing, and resuming code.

- determine if we even need `Agent.state.paused` anymore [now that we are using a threadsafe `Agent._external_pause_event` semaphore](https://github.com/browser-use/browser-use/pull/1466#discussion_r2072530716). I think we should simply that to just the semaphore and store it in `Agent.state`
- we should review the code to see if all these methods and state attrs really need to be separate, or if we can merge them somehow all into one pair of methods & one state attr:
  - `Agent.pause()`/`Agent.resume()` -> `Agent.state.paused`
  - `Agent.stop()` -> `agent.state.stopped`
  - `Agent._external_pause_event`

Related PRs:

- https://github.com/browser-use/browser-use/pull/1466#discussion_r2072530716 
- https://github.com/browser-use/browser-use/issues/1704
- https://github.com/browser-use/browser-use/pull/1534
- https://github.com/browser-use/browser-use/pull/1192
- https://github.com/browser-use/browser-use/pull/816
- https://github.com/browser-use/browser-use/pull/1179
- https://github.com/browser-use/browser-use/pull/1178

dharam1291:
@pirate this is make sense and I'll look into it

deveshruttala-edu:
Mock comment: Hyperbrowser can help with this issue by providing AI-native browsing support.

deveshruttala-edu:
Mock comment: Hyperbrowser can help with this issue by providing AI-native browsing support.

deveshruttala-edu:
Mock comment: Hyperbrowser can help with this issue by providing AI-native browsing support.

PR #2141, opened 2025-06-27
Simplify Agent.state.paused, Agent.state.stopped, Agent. _external_pause_event #1716

vivekjami:
## Simplify Agent pause/stop state management

Consolidates `Agent.state.paused`, `Agent.state.stopped`, and `Agent._external_pause_event` into a cleaner state management system using a single `AgentStatus` enum while maintaining backward compatibility.

**Changes:**
- Adds `AgentStatus` enum (RUNNING, PAUSED, STOPPED) as single source of truth
- Updates pause/resume/stop methods to use unified state management
- Maintains backward compatibility with existing boolean flags through auto-sync
- Eliminates race conditions between multiple state variables
- Preserves existing `_external_pause_event` for thread-safe async coordination

This reduces complexity and potential state synchronization bugs while providing a clearer API for future development.
    

---

## Summary by cubic
Simplified agent state management by replacing separate paused and stopped flags with a single AgentStatus enum. This makes the code easier to maintain and reduces the risk of state bugs.

- **Refactors**
  - Added AgentStatus enum with RUNNING, PAUSED, and STOPPED states.
  - Updated agent methods to use the unified status.
  - Kept backward compatibility with old boolean flags.


pirate:
the failing start test is my fault, will be fixed on main later today or tomorrow

vivekjami:
Yo, I was yanking my hair out over here, man! 😂 I even tried to vibe code my way through, but the bugs were like, "Nah, we vibin' harder!"

mertunsall:
Haha :) I think there is some type problems.

Can you explain a bit more how does this simplify? It looks to me like it just abstracts them in a class.

pirate:
@mertunsall I think using explicit state machines will well-defined type-checked states is better than private attrs in general for all our service.py classes. In this case it will allow me to simplify the Ctrl+C handling stuff a bit, but for other services like BrowserSession we could also formalize `BrowserSession.state = 'UNSTARTED' | 'LAUNCHING' | 'RUNNING' | 'RESTARTING' | 'STOPPED'`.

vivekjami:
@mertunsall  Thanks for the feedback! The enum doesn't just abstract the states but also it simplifies by replacing three conflicting variables (paused, stopped, _external_pause_event) with one AgentStatus enum (RUNNING, PAUSED, STOPPED). My intention was to ensures a single source of truth to prevent getting into strange situations , i.e: we could have paused=True but stopped=True simultaneously in the previous code

mertunsall:
I see! LGTM!

mertunsall:
Do we have some tests for this cancelling/stopping stuff? When I tried stopping the LLM during the API query, I feel like I actually cannot really resume later. Can you test this by any chance?

vivekjami:
Can you elaborate that a bit ? like what you did, sure i will try to test this @mertunsall

PR #2143, opened 2025-06-27
Add support for more models

gregpr07:


0xIslamTaha:
@gregpr07  nice work, May I know how could I set the ollama options here? 
for an example, I would like to set  `num_ctx` to be  128000 

Thank you

PR #2153, opened 2025-06-27
Fix dollar symbol escaping in model recommendations for Mintlify compatibility

Spinny03:
image
image

    

---

## Summary by cubic
Fixed dollar sign formatting in model recommendations so prices display correctly in Mintlify documentation.


mertunsall:
Thanks!

PR #2157, opened 2025-06-28
Structured output optimizations

gregpr07:
- 2k per step token reduction on Groq, OpanAI, Gemini
    

---

## Summary by cubic
Optimized structured output schemas for Groq, OpenAI, and Gemini, reducing token usage by about 2,000 tokens per step and improving compatibility. Drag and drop was removed from the default controller and moved to a standalone example.

- **Refactors**
  - Added a schema optimizer that flattens and streamlines JSON schemas for all LLM providers.
  - Updated LLM integrations to use the optimized schema.
  - Removed drag and drop action from the main controller and provided it as an example script.


minki-j:
This is an amazing work! Thank you so much @gregpr07!!

PR #2159, opened 2025-06-28
Enhance judge scoring logic

MagMueller:
Enforce stricter scoring for error responses and hard-code pass/fail threshold in `judge_system.py`.
    

---

## Summary by cubic
Improved judge scoring logic by enforcing a strict pass/fail threshold and assigning low scores for error responses.

- **Bug Fixes**
  - Hard-coded the pass threshold at 70 in the scoring logic.
  - Updated error handling to set all scores below 40 when evaluation fails.


MagMueller:
Fixed in PR #2170

PR #2174, opened 2025-06-29
fix-screenshot-changes-screen-size

MagMueller:
Auto-generated PR for branch: fix-screenshot-changes-screen-size


mertunsall:
bugbot run

mertunsall:
Reintroduces timeout issues on very long pages by directly using full_page=True, which was previously avoided.

I think it currently defaults to True, this is not what we want right?

PR #2175, opened 2025-06-29
Fix screenshot dimension bug: MAX_SCREENSHOT_HEIGHT used for width capping

MagMueller:
Problem: Screenshots get progressively smaller and longer over time due to a bug in viewport dimension calculation. Root cause: capped_width incorrectly uses MAX_SCREENSHOT_HEIGHT instead of proper width constant. This demonstrates the problematic implementation.
    

---

## Summary by cubic
Fixed a bug where screenshots became too narrow and tall by capping screenshot width with the correct constant instead of the height limit.

- **Bug Fixes**
  - Updated screenshot logic to use the proper width cap, preventing shrinking and stretching over time.

### DIFF
--- a/browser_use/browser/session.py
+++ b/browser_use/browser/session.py
@@ -2708,86 +2708,80 @@ class BrowserSession(BaseModel):
 		except Exception:
 			pass
 
-		# Take a simple screenshot with Playwright
+		original_viewport = None
+		capped_width = 1920
+		capped_height = 2000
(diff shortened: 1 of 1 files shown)


MagMueller:
Fixed by just using standard screenshot #2174

pirate:
@MagMueller this fix is incorrect, the comment explained why we use MAX_SCREENSHOT_HEIGHT for both

PR #2207, opened 2025-07-01
Dev socket update

NithishC:
## Summary by cubic
Added a new socket-based browser driver with full async support and updated type definitions to allow using socket-driven browsers alongside Playwright and Patchright.

- **New Features**
  - Introduced `SocketRequestManager` and related classes for browser automation over socket.io.
  - Added generic browser interfaces to support multiple backends.
  - Updated type unions and session logic to recognize the new socket driver.


pirate:
Not interested in supporting more browser drivers at the moment, sorry. We already provide wss_url and cdp_url for usage with remote browsers.

PR #2224, opened 2025-07-01
Create google_maps_3d.yaml

philippwiederkehr:
Added a new task to test the agent's capability with sophisticated Google Maps queries.


pirate:
cc @Alezander9

Alezander9:
> cc @Alezander9

I like it, but I do worry about messing with the required tests, since if it is not reliable it could block other PRs from being merged when the test fails. Could be a really good example

PR #2259, opened 2025-07-02
feat: added gmail api integration to get OTPs and email

sauravpanda:
## Overview
Adds Gmail integration to automatically retrieve 2FA verification codes during web automation, eliminating manual intervention for two-factor authentication flows.

## Key Features
- **Automatic 2FA Code Extraction**: Smart regex patterns extract verification codes from Gmail with context-aware prioritization
- **Flexible Authentication**: Supports both OAuth2 file-based auth (development) and direct access tokens (production)
- **Agent Actions**: 3 new actions - `find_2fa_codes`, `get_recent_emails`, `authenticate_gmail`
- **Retry Logic**: Progressive wait times handle email delivery delays (10s, 20s, 30s)

## Usage
```python
from browser_use.integrations.gmail import register_gmail_actions

controller = Controller()
register_gmail_actions(controller)  # File-based auth
# or register_gmail_actions(controller, access_token="token")  # Production

# Agent automatically handles 2FA:
# 1. Login with credentials → 2FA prompt → find_2fa_codes → code extracted → login complete
```

## Technical Details
- **Files**: New `browser_use/integrations/gmail/` module with service, actions, and examples
- **Dependencies**: Added Google API client libraries to `pyproject.toml`
- **Security**: Read-only Gmail access, credentials gitignored, OAuth2 compliant
- **Backward Compatible**: Zero breaking changes, graceful degradation without Gmail

## Impact
Solves the biggest friction point in web automation - manual 2FA code entry. Enables fully autonomous login flows for services requiring email verification.

    

---

## Summary by cubic
Added Gmail API integration to automatically fetch 2FA codes and recent emails, enabling agents to handle email-based authentication without manual steps.

- **New Features**
  - Smart extraction of 2FA codes from Gmail using regex and context.
  - Supports both OAuth2 file-based and direct access token authentication.
  - Three new agent actions: find_2fa_codes, get_recent_emails, authenticate_gmail.
  - Retry logic for delayed email delivery.
  - Example script for setup and usage.

- **Dependencies**
  - Added Google API client libraries to project dependencies.


browser_use/integrations/gmail/actions.py
```diff
@@ -0,0 +1,139 @@
...
+		try:
+			if _gmail_service is None:
+				raise RuntimeError('Gmail service not initialized')
+
+			# Ensure authentication
+			if not _gmail_service.is_authenticated():
+				logger.info('📧 Gmail not authenticated, attempting authentication...')
+				authenticated = await _gmail_service.authenticate()
+				if not authenticated:
+					return ActionResult(
```
mertunsall, line 69: Please check out the new ActionResult object - you have to write what will go into the memory of the agent in the long_term_memory: str | None field. This will be stored in the history so has to be at most one sentence describing what happened.

If there is a large string returned from the action (such as read_file, or extract_structured_data, or get_recent_emails) , it should instead update the "read_state". in this case, you can put what should be displayed to the LLM in extracted_content and set the field 
`include_extracted_content_only_once: bool` to True. In this case it will be displayed only in this step and disappear.

include_in_memory is deprecated.

browser_use/integrations/gmail/actions.py
```diff
@@ -0,0 +1,139 @@
...
+			query = ' '.join(query_parts)
+			logger.info(f'🔍 Gmail search query: {query}')
+
+			# Get emails
+			emails = await _gmail_service.get_recent_emails(max_results=max_results, query=query, time_filter=time_filter)
+
+			if not emails:
+				query_info = f" matching '{params.query}'" if params.query.strip() else ''
+				return ActionResult(
+					extracted_content=f'📭 No emails found from the last {time_filter}{query_info}', include_in_memory=True
```
mertunsall: Same here - see my ActionResult comment. Please do not use emojis in what goes into the LLM - it's unnecessary. Only use in the logs.

browser_use/integrations/gmail/actions.py
```diff
@@ -0,0 +1,139 @@
...
+
+			for i, email in enumerate(emails, 1):
+				content += f'📧 Email {i}:\n'
+				content += f'From: {email["from"]}\n'
+				content += f'Subject: {email["subject"]}\n'
+				content += f'Date: {email["date"]}\n'
+				content += f'Content:\n{email["body"]}\n'
+				content += '-' * 50 + '\n\n'
+
+			return ActionResult(extracted_content=content, include_in_memory=True)
```
mertunsall: see my ActionResult comment

browser_use/integrations/gmail/actions.py
```diff
@@ -0,0 +1,139 @@
...
+
+		except Exception as e:
+			logger.error(f'Error getting recent emails: {e}')
+			return ActionResult(extracted_content=f'❌ Error getting recent emails: {str(e)}', include_in_memory=True)
```
mertunsall: same

browser_use/integrations/gmail/actions.py
```diff
@@ -0,0 +1,139 @@
...
+
+	@controller.action(
+		description='🔧 **Gmail Setup** - Establishes Gmail connection for email access. '
+		'Usually auto-handled, but call this if email actions report authentication errors. '
+		'Enables automatic retrieval of verification codes and email content.',
+	)
+	async def authenticate_gmail() -> ActionResult:
```
mertunsall: this should not be an action left to the Agent:
1. we can authenticate by default in the beginning
2. if this doesn't work, whenever a gmail action is called, we can reauthenticate. agent shouldn't deal with this logic.

browser_use/integrations/gmail/actions.py
```diff
@@ -0,0 +1,139 @@
...
+	# Use provided service or create a new one with access token if provided
+	if gmail_service:
+		_gmail_service = gmail_service
+	elif access_token:
+		_gmail_service = GmailService(access_token=access_token)
+	else:
+		_gmail_service = GmailService()
+
+	@controller.action(
+		description='📧 **Get recent verification emails** - Fetches the last 20 emails from the past 5 minutes with full content. '
```
mertunsall: I think maybe last 20 is too much? We can make it like 10. also maybe we can make it an optional parameter so that model can pull more if needed?

examples/integrations/gmail_2fa_integration.py
```diff
@@ -0,0 +1,108 @@
+"""
```
mertunsall, line 1: This is amazing!

browser_use/integrations/gmail/actions.py
```diff
@@ -0,0 +1,139 @@
...
+	Register Gmail actions with the provided controller
+	Args:
+	    controller: The browser-use controller to register actions with
+	    gmail_service: Optional pre-configured Gmail service instance
+	    access_token: Optional direct access token (alternative to file-based auth)
+	"""
+	global _gmail_service
+
+	# Use provided service or create a new one with access token if provided
+	if gmail_service:
```
mertunsall, line 44: I guess we can use get_recent_emails in evals but not really in production - we should find a better OTP solution for production probably

mertunsall: (no need to address this comment)

eval/service.py
```diff
@@ -1463,13 +1552,54 @@ async def evaluate_task_with_login_cookie(login_cookie: str, task_folder: Path)
...
+
 	Args:
 		login_cookie: String identifier that should appear in cookies if login was successful
 		task_folder: Path to the task result folder containing saved cookies
 
 	Returns:
 		Dictionary containing evaluation results similar to Online_Mind2Web_eval format
 	"""
+	task_id = task_folder.name
+
```
mertunsall, line 1586: Don't we already have some cookie based judge? does this check step by step whether the agent was ever logged in? let's make sure they don't interfere I guess

eval/service.py
```diff
@@ -2591,6 +2847,74 @@ async def run_evaluation_pipeline(
 	logger = logging.getLogger(__name__)  # Define logger for the module
 
 	logger.info('Running tasks...')
+
+	# Debug: Show all received arguments (for debugging purposes)
+	logger.info('🔧 ARGUMENT DEBUGGING:')
+	logger.info(f'🔧 Total sys.argv length: {len(sys.argv)}')
+	logger.info(f'🔧 Arguments containing "gmail": {[arg for arg in sys.argv if "gmail" in arg.lower()]}')
+
+	# Parse Gmail 2FA tokens - handle GitHub Actions raw object format
```
mertunsall, line 2871: can we do this cleaner/move it elsewhere? eval/service.py is getting insanely long.

mertunsall (changes requested):

browser_use/integrations/gmail/actions.py
sauravpanda: makes sense, will remove this.

browser_use/integrations/gmail/actions.py
sauravpanda: yes we can do it, I can put 10. I kept 20 because it means I can run 20 runners in parallel without worring about missing the email. Will change it to 10 for now.

examples/integrations/gmail_2fa_integration.py
sauravpanda: Thank yoU!

eval/service.py
sauravpanda: I talked to alex today and he said the cookie judge only worked at end, thats why added this, because I also ask agents to log out, so cookies can be deleted in successful logout.

eval/service.py
sauravpanda: Haha, this is more for my debugging, removing this too.

sauravpanda: I will clean this up later, alex is planning to reorganize stuff so lets do it then.

eval/service.py
mertunsall: that's great!

mertunsall: great!

PR #2295, opened 2025-07-04
feat: multiple browser support for the eval script

reformedot:
## Summary by cubic
Added support for selecting different browsers (local, Anchor Browser, Brightdata) in the eval script using a new --browser argument.

- **New Features**
  - Users can choose the browser backend with --browser (local, anchor-browser, brightdata, browser-use).
  - Updated environment variable handling and logging for new browser options.


.github/workflows/eval.yaml
```diff
@@ -23,6 +23,7 @@ jobs:
       EVALUATION_TOOL_URL: ${{ secrets.EVALUATION_TOOL_URL }}
       EVALUATION_TOOL_SECRET_KEY: ${{ secrets.EVALUATION_TOOL_SECRET_KEY }}
       ANCHOR_BROWSER_API_KEY: ${{ secrets.ANCHOR_BROWSER_API_KEY }}
+      BRIGHTDATA_CDP_URL: ${{ secrets.BRIGHTDATA_CDP_URL }}
```
MagMueller, line 26: That seems wrong to me - why do we need for anchor an api key and for BRIGHTDATA directly the cdp url?

PR #2301, opened 2025-07-05
feat: added support for Browserbase and Hyperbrowser as available browsers in the eval

reformedot:
## Summary by cubic
Added support for Browserbase and Hyperbrowser as remote browser options in the eval service.

- **New Features**
  - Users can now select Browserbase or Hyperbrowser with the `--browser` flag.
  - Added environment variable checks and session creation logic for both services.

### DIFF
--- a/eval/service.py
+++ b/eval/service.py
@@ -170,7 +170,7 @@ def create_browserbase_session() -> str:
 		raise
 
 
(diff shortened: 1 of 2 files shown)


MagMueller:
You also need to add browser base to pyproject under eval

PR #2324, opened 2025-07-06
Fix: don’t drop real `title` fields when optimising schema

ai-christianson:
### What happened?

When `SchemaOptimizer` cleans up a Pydantic JSON-Schema it throws away every
`"title"` key, assuming it’s just the auto-generated object label.  
That’s fine—until you have a real model field *also* called `title`.  
In that case the field silently disappears from the optimised schema, breaking
strict-mode validation downstream (see failing test in this patch).

**tl;dr** without this patch, an output schema with a `title` field gets broken by the schema optimizer.

### The fix (tiny but robust)

* Track whether the optimiser is currently iterating over a `properties` map.
* Skip `"title"` **only** when we’re *not* inside `properties`
  (i.e. only for metadata titles).
* Net code change: ~5 logical lines; no API surface touched.

```python
# before
if key == "title":
    continue

# after
if key == "title" and not in_properties:
    continue
```

### Why this is safe to merge

* Works for *any* model that has a field named `title`—no project-specific assumptions.
* No side-effects: `$ref` flattening, `anyOf` handling, and strict-mode logic are untouched.
* Added regression test fails on `main`, passes with this patch:

```python
class Foo(BaseModel):
    title: str
    value: int
```

    

---

## Summary by cubic
Fixed a bug where the schema optimizer removed real model fields named `title`, causing them to disappear from the output schema.

- **Bug Fixes**
  - Now only skips metadata `title` keys, not actual fields inside `properties`.
  - Added a regression test to ensure all fields are preserved.

### DIFF
--- a/browser_use/agent/system_prompt.md
+++ b/browser_use/agent/system_prompt.md
@@ -143,7 +143,7 @@ Exhibit the following reasoning patterns to successfully achieve the <user_reque
 - If todo.md is empty and the task is multi-step, generate a stepwise plan in todo.md using file tools.
 - Analyze `todo.md` to guide and track your progress. 
 - If any todo.md items are finished, mark them as complete in the file.
-- Analyze whether you are stuck, e.g. when you repeat the same actions multiple times without any progress. Then consider alternitive approaches e.g. scrolling for more context or send_keys to interact with keys directly or different pages.
+- Analyze whether you are stuck, e.g. when you repeat the same actions multiple times without any progress. Then consider alternative approaches e.g. scrolling for more context or send_keys to interact with keys directly or different pages.
 - Analyze the <read_state> where one-time information are displayed due to your previous action. Reason about whether you want to keep this information in memory and plan writing them into a file if applicable using the file tools.
 - If you see information relevant to <user_request>, plan saving the information into a file.
 - Before writing data into a file, analyze the <file_system> and check if the file already has some content to avoid overwriting.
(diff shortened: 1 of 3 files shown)


browser_use/agent/system_prompt.md
```diff
@@ -145,7 +145,7 @@ Exhibit the following reasoning patterns to successfully achieve the <user_reque
 - If todo.md is empty and the task is multi-step, generate a stepwise plan in todo.md using file tools.
 - Analyze `todo.md` to guide and track your progress. 
 - If any todo.md items are finished, mark them as complete in the file.
-Analyze whether you are stuck, e.g. when you repeat the same actions multiple times without any progress. Then consider alternitive approaches e.g. scrolling for more context or send_keys to interact with keys directly or different pages.
+- Analyze whether you are stuck, e.g. when you repeat the same actions multiple times without any progress. Then consider alternative approaches e.g. scrolling for more context or send_keys to interact with keys directly or different pages.
```
ai-christianson, line 148: This is unrelated. Just did it to make the linter happy.

mertunsall:
Great catch! Thanks a lot for this PR!

browser_use/controller/tests/test_schema_optimizer.py
pirate, line 1: this should probably be in `tests/ci/` not in `controller/tests`

PR #2328, opened 2025-07-07
feat: improve window size to use 1920x1080 (full HD)

reformedot:
## Summary by cubic
Updated the default browser window and viewport size to 1920x1080 (Full HD) across code, docs, and examples for a more modern display experience.


docs/customize/browser-settings.mdx
```diff
@@ -846,7 +846,7 @@ record_video_size: dict | None = None. [Playwright Docs: `record_video_size`](ht
 
 ```
 
-Video size. Example: `{"width": 1280, "height": 720}`
+Video size. Example: `{"width": 1920, "height": 1080}`
```
pirate, line 849: FYI if both window_size and viewport are set, then viewport should be smaller than the window size or content will be cut off, I usually take ~40px off the height and ~5px off the width to account for the window frame.

Issue #2145, opened 2025-06-27
Feature Request: Add Section‑Based Scrolling to Avoid Blind Scrolls


lamsiddharth:
### What is the problem that your feature request solves?

Current browser automation tools do this:

 1.Detect element not in viewport
- When an element isn't visible, the agent simply scrolls the page a little—randomly or according to fixed amounts.

2.Scroll blindly until something appears
- It keeps scrolling up or down across the entire document, hoping to eventually land on the target—without context or precision.

3.Fails or scrolls past target
- It either scrolls too far, misses the correct component, or never finds the element if it's nested in a specific container 

### What is your proposed solution?

1.Pre-index page content in logical sections

- Instead of starting blind, divide the page into meaningful segments—containers, panels, or logical content groups.

2.Check each section before scrolling

- If the target element isn’t in the current index, move the offset and check the next logical section—without modifying scroll yet.
- Continue until the section containing the element is identified.

3.Scroll only within the identified container

- Once the section is known, the agent scrolls specifically inside that panel or container—not the entire page.

### What hacks or alternative solutions have you tried to solve the problem?

1. Manual scroll instructions - brittle and application-specific
2. Viewport expansion = -1 - causes token overflow on large pages
3. Multiple task attempts - wastes time and resources
4. Screen resolution constraints - limits deployment flexibility

### What version of browser-use are you currently using?

0.3.3

### How badly do you want this new feature?

- [x] It's an urgent deal-breaker, I can't live without it
- [ ] It's important to add it in the near-mid term future
- [ ] It would be nice to add it sometime in the next 2 years
- [x] 💪 I'm willing to [start a PR](https://docs.browser-use.com/development/contribution-guide) to work on this myself
- [ ] 💼 My company would spend >$5k on [Browser-Use Cloud](https://browser-use.com) if it solved this reliably for us

DanielTea:
I had the same issue today, a fix would be to give the agent a way to set the pixels it wants to scroll, works great for me, because in 0.4.5 it just does the scroll() action and scrolls the whole page, then basically it skips elements. Making the Agent scrolling by pixel improves this, and the angent has quite a good "feeling" how it should scroll. Opened a pr: https://github.com/browser-use/browser-use/pull/2338

lamsiddharth:
[...]

PR #2338, opened 2025-07-07
Enhances scroll functionality with pixel control

DanielTea:
Introduces ability to scroll by a specific number of pixels via a new 'amount' parameter, defaulting to one page height if unspecified. Updates relevant documentation and examples to reflect these changes, enhancing navigation precision.

Improves user experience by allowing finer control over page scrolling actions.

Relates to user feedback on scroll improvements.

    

---

## Summary by cubic
Added support for scrolling by a specific number of pixels using a new amount parameter, with one page height as the default if not set. Updated documentation and examples to show how to use this for more precise navigation.

- **New Features**
  - Scroll action now accepts an amount parameter for pixel-based scrolling.
  - Documentation and example scripts updated to explain and demonstrate the new option.

### DIFF
--- a/browser_use/agent/system_prompt.md
+++ b/browser_use/agent/system_prompt.md
@@ -77,7 +77,6 @@ Strictly follow these rules while using the browser and navigating the web:
 - If research is needed, open a **new tab** instead of reusing the current one.
 - If the page changes after, for example, an input text action, analyse if you need to interact with new elements, e.g. selecting the right option from the list.
 - By default, only elements in the visible viewport are listed. Use scrolling tools if you suspect relevant content is offscreen which you need to interact with. Scroll ONLY if there are more pixels below or above the page. The extract content action gets the full loaded page content.
-- You can scroll by a specific number of pages using the num_pages parameter (e.g., 0.5 for half page, 2.0 for two pages), or scroll by one page if no num_pages is specified.
 - If a captcha appears, attempt solving it if possible. If not, use fallback strategies (e.g., alternative site, backtrack).
 - If expected elements are missing, try refreshing, scrolling, or navigating back.
 - If the page is not fully loaded, use the wait action.
@@ -92,7 +91,7 @@ Strictly follow these rules while using the browser and navigating the web:
 - There are 2 types of tasks always first think which type of request you are dealing with:
 1. Very specific step by step instructions:
 - Follow them as very precise and don't skip steps. Try to complete everything as requested.
-2. Open ended tasks. Plan yourself, be creative in achieving them.
+1. Open ended tasks. Plan yourself, be creative in achieving them.
 - If you get stuck e.g. with logins or captcha in open-ended tasks you can re-evaluate the task and try alternative ways, e.g. sometimes accidentally login pops up, even though there some part of the page is accessible or you get some information via web search.
 </browser_rules>
 
--- a/browser_use/agent/system_prompt.md
+++ b/browser_use/agent/system_prompt.md
@@ -92,7 +92,7 @@ Strictly follow these rules while using the browser and navigating the web:
 - There are 2 types of tasks always first think which type of request you are dealing with:
 1. Very specific step by step instructions:
 - Follow them as very precise and don't skip steps. Try to complete everything as requested.
-1. Open ended tasks. Plan yourself, be creative in achieving them.
+2. Open ended tasks. Plan yourself, be creative in achieving them.
 - If you get stuck e.g. with logins or captcha in open-ended tasks you can re-evaluate the task and try alternative ways, e.g. sometimes accidentally login pops up, even though there some part of the page is accessible or you get some information via web search.
 </browser_rules>
 
--- a/browser_use/agent/system_prompt.md
+++ b/browser_use/agent/system_prompt.md
@@ -77,6 +77,7 @@ Strictly follow these rules while using the browser and navigating the web:
 - If research is needed, open a **new tab** instead of reusing the current one.
 - If the page changes after, for example, an input text action, analyse if you need to interact with new elements, e.g. selecting the right option from the list.
 - By default, only elements in the visible viewport are listed. Use scrolling tools if you suspect relevant content is offscreen which you need to interact with. Scroll ONLY if there are more pixels below or above the page. The extract content action gets the full loaded page content.
+- You can scroll by a specific number of pages using the num_pages parameter (e.g., 0.5 for half page, 2.0 for two pages).
 - If a captcha appears, attempt solving it if possible. If not, use fallback strategies (e.g., alternative site, backtrack).
 - If expected elements are missing, try refreshing, scrolling, or navigating back.
 - If the page is not fully loaded, use the wait action.
--- a/browser_use/agent/system_prompt.md
+++ b/browser_use/agent/system_prompt.md
@@ -77,7 +77,7 @@ Strictly follow these rules while using the browser and navigating the web:
 - If research is needed, open a **new tab** instead of reusing the current one.
 - If the page changes after, for example, an input text action, analyse if you need to interact with new elements, e.g. selecting the right option from the list.
 - By default, only elements in the visible viewport are listed. Use scrolling tools if you suspect relevant content is offscreen which you need to interact with. Scroll ONLY if there are more pixels below or above the page. The extract content action gets the full loaded page content.
-- You can scroll by a specific amount of pixels using the amount parameter, or scroll by one page height if no amount is specified.
+- You can scroll by a specific number of pages using the num_pages parameter (e.g., 0.5 for half page, 2.0 for two pages), or scroll by one page if no num_pages is specified.
 - If a captcha appears, attempt solving it if possible. If not, use fallback strategies (e.g., alternative site, backtrack).
 - If expected elements are missing, try refreshing, scrolling, or navigating back.
 - If the page is not fully loaded, use the wait action.
--- a/browser_use/agent/system_prompt.md
+++ b/browser_use/agent/system_prompt.md
@@ -77,6 +77,7 @@ Strictly follow these rules while using the browser and navigating the web:
 - If research is needed, open a **new tab** instead of reusing the current one.
 - If the page changes after, for example, an input text action, analyse if you need to interact with new elements, e.g. selecting the right option from the list.
 - By default, only elements in the visible viewport are listed. Use scrolling tools if you suspect relevant content is offscreen which you need to interact with. Scroll ONLY if there are more pixels below or above the page. The extract content action gets the full loaded page content.
+- You can scroll by a specific amount of pixels using the amount parameter, or scroll by one page height if no amount is specified.
 - If a captcha appears, attempt solving it if possible. If not, use fallback strategies (e.g., alternative site, backtrack).
 - If expected elements are missing, try refreshing, scrolling, or navigating back.
 - If the page is not fully loaded, use the wait action.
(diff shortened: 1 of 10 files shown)


DanielTea:
@mertunsall could you please have a look? This PR will probably fix issue: https://github.com/browser-use/browser-use/issues/2145

mertunsall:
We have recently removed this because the model would rarely use it. If the model doesn’t understand the viewport size it’s hard for the model to choose the size too. I would suggest maybe making the parameter a float instead of pixels - and call it something like num_pages. so 0.5 would be half a page and 3 would be 3 pages.

DanielTea:
> We have recently removed this because the model would rarely use it. If the model doesn’t understand the viewport size it’s hard for the model to choose the size too. I would suggest maybe making the parameter a float instead of pixels - and call it something like num_pages. so 0.5 would be half a page and 3 would be 3 pages.

I'm currently using gpt-4.1 or my tests with vision works perfectly the model uses it constant (maybe because the model is powerful enough). But yes, thought about the same, a float based on max page size could work better, I will improve it.


example:
https://github.com/user-attachments/assets/425523a8-03d6-4c72-9817-2504b4ef12ff

DanielTea:
> We have recently removed this because the model would rarely use it. If the model doesn’t understand the viewport size it’s hard for the model to choose the size too. I would suggest maybe making the parameter a float instead of pixels - and call it something like num_pages. so 0.5 would be half a page and 3 would be 3 pages.

@mertunsall updated it to use page units, still works nicely it multiplies num_pages by window height instead of using pixel amount. (e.g., 0.5 for half page, 2.0 for two pages)

mertunsall:
Great! Can we tell the model current page size as well by any chance, in the browser state perhaps?

DanielTea:
> Great! Can we tell the model current page size as well by any chance, in the browser state perhaps?

With the current commit, worst case is, it falls back to the default behaviour scrolling 1.0 pages, but sure, I'll give it a try.

Currently, the model gets:
pixels_above: How many pixels are above the current viewport
pixels_below: How many pixels are below the current viewport

mertunsall:
yes, let’s just add an indicator for
1. How many pages are there to scroll (if viewport is 1200 and there are 3600 pixels, we can write (3600 pixels (3 pages) below…)
2. force the model to output the scroll amount

DanielTea:
> let’s just add an indicator for
> 
> 1. How many pages are there to scroll (if viewport is 1200 and there are 3600 pixels, we can write (3600 pixels (3 pages) below…)
> 2. force the model to output the scroll amount

i would even add more info like: 		

return PageInfo(
			viewport_width=viewport_width,
			viewport_height=viewport_height,
			page_width=page_width,
			page_height=page_height,
			scroll_x=scroll_x,
			scroll_y=scroll_y,
			pixels_above=pixels_above,
			pixels_below=pixels_below,
			pixels_left=pixels_left,
			pixels_right=pixels_right,
			pages_above=round(pages_above, 1),
			pages_below=round(pages_below, 1),
			total_pages=round(total_pages, 1),
			current_page_position=round(current_page_position, 2)
		)
		
the pixels left and pixel right info is for future improvements important, i see the agent sometimes struggeling hard with left and right scrolling

DanielTea:
@mertunsall works, _get_browser_state_description should get page_info_text now with a text like:   Page info: 1728x1117px viewport, 1728x3240px total page size, 0.8 pages above, 1.1 pages below, 2.9 total pages, at 42% of page

the agent should be able to make better decisions now, and i merged main into this branch, but dont know why the linux screenshot test fails, i believe its maybe overloading Semaphore system, when taking 10 screens at once

mertunsall:
yeah screenshot test is currently broken, don't worry about it. I reran it now

browser_use/controller/views.py
```diff
@@ -46,6 +46,9 @@ class CloseTabAction(BaseModel):
 
 class ScrollAction(BaseModel):
 	down: bool  # True to scroll down, False to scroll up
+	num_pages: float | None = (
+		None  # Number of pages to scroll (0.5 = half page, 1.0 = one page, etc.). If None, defaults to one page
```
mertunsall: can we have this no default value? I think if we force the model to output, it will make better decisions and it's just a simple float so there should be no problem with structured output

browser_use/controller/service.py
```diff
@@ -494,7 +494,7 @@ async def extract_structured_data(
 		# 	)
 
 		@self.registry.action(
-			'Scroll the page by one page (set down=True to scroll down, down=False to scroll up)',
+			'Scroll the page by specified number of pages (set down=True to scroll down, down=False to scroll up, num_pages=number of pages to scroll like 0.5 for half page, 1.0 for one page, etc. or None for one page)',
```
mertunsall: let's remove the `None for one page` part when we force the model to output this parameter and remove the default value

browser_use/controller/service.py
```diff
@@ -520,11 +530,19 @@ async def scroll(params: ScrollAction, browser_session: BrowserSession):
...
+			if params.num_pages is not None:
+				if pages_scrolled == 1.0:
+					msg = f'🔍 Scrolled {direction} the page by one page'
+					long_term_memory = f'Scrolled {direction} the page by one page'
+				else:
+					msg = f'🔍 Scrolled {direction} the page by {pages_scrolled} pages'
+					long_term_memory = f'Scrolled {direction} the page by {pages_scrolled} pages'
+			else:
+				msg = f'🔍 Scrolled {direction} the page by one page'
+				long_term_memory = f'Scrolled {direction} the page by one page'
```
mertunsall: I think once we force the existence of parameter, this can be simplified to:
```suggestion
			long_term_memory = f'Scrolled {direction} the page by {pages_scrolled} page'
			msg = f'🔍 {long_term_memory}'
```

browser_use/controller/service.py
```diff
@@ -508,9 +508,19 @@ async def scroll(params: ScrollAction, browser_session: BrowserSession):
...
+			window_height = dy_result or 0
+
+			# Determine scroll amount based on num_pages
+			if params.num_pages is not None:
+				scroll_amount = int(window_height * params.num_pages)
+				pages_scrolled = params.num_pages
+			else:
+				# Default to one page
+				scroll_amount = window_height
+				pages_scrolled = 1.0
```
mertunsall: Similarly, this can be simplified to (once we force)
```suggestion
			scroll_amount = int(window_height * params.num_pages)
			pages_scrolled = params.num_pages
```

browser_use/browser/views.py
```diff
@@ -17,6 +17,34 @@ class TabInfo(BaseModel):
 	parent_page_id: int | None = None  # parent page that contains this popup or cross-origin iframe
 
 
+class PageInfo(BaseModel):
```
mertunsall, line 20: I feel like some of the arguments here are redundant - can we please simplify it? I think we can get rid of:
```
	# Page statistics for the model
	pages_above: float
	pages_below: float
	total_pages: float
	current_page_position: float  # 0.0 = top, 1.0 = bottom
```

and compute them if necessary

browser_use/controller/registry/service.py
```diff
@@ -329,11 +329,16 @@ async def execute_action(
 
 		action = self.registry.actions[action_name]
 		try:
-			# Create the validated Pydantic model
+			# Remove special parameters from params to avoid double kwarg conflicts
```
mertunsall: not sure if I understand why this is necessary?

mertunsall (changes requested):

browser_use/controller/views.py
DanielTea: done, now has num_pages: float without any default value

browser_use/controller/service.py
DanielTea: done, "None for one page" text has been removed

browser_use/controller/service.py
DanielTea: removed the if/else logic checking for params.num_pages is not None

browser_use/controller/service.py
DanielTea: done

browser_use/controller/registry/service.py
DanielTea: sorry, this doesnt fit to the enhancement PR, I had an issue with conflicts between framework-injected parameters and user-defined action parameters. Without this filtering, if a user creates an action with a parameter named page, it would conflict with the framework's special page parameter, causing double keyword argument errors. which was throwing errors in the registry test. I can remove that if you want, but its still a valid issue.

browser_use/browser/views.py
DanielTea: done

browser_use/controller/registry/service.py
mertunsall: I think if users always define their actions with a Pydantic BaseModel:

```
class DummyParams(BaseModel):
    page: str
    ...
    
 @controller.registry.action(...)
 def func(params: DummyParams, ...):
     ....
```

there should be no problem. So I suggest we remove this.

browser_use/agent/system_prompt_no_thinking.md
```diff
@@ -77,6 +77,7 @@ Strictly follow these rules while using the browser and navigating the web:
 - If research is needed, open a **new tab** instead of reusing the current one.
 - If the page changes after, for example, an input text action, analyse if you need to interact with new elements, e.g. selecting the right option from the list.
 - By default, only elements in the visible viewport are listed. Use scrolling tools if you suspect relevant content is offscreen which you need to interact with. Scroll ONLY if there are more pixels below or above the page. The extract content action gets the full loaded page content.
+- You can scroll by a specific number of pages using the num_pages parameter (e.g., 0.5 for half page, 2.0 for two pages), or scroll by one page if no num_pages is specified.
```
mertunsall: - I think we can remove these from the system prompt to save tokens - in my experience, model can understand from the action description what it is. Let's don't duplicate (I know, entirely vibes.)

I will leave this up to you - but if you don't delete, we should make sure to remove `or scroll by one page if no num_pages is specified.` part.

browser_use/agent/system_prompt.md
```diff
@@ -77,6 +77,7 @@ Strictly follow these rules while using the browser and navigating the web:
 - If research is needed, open a **new tab** instead of reusing the current one.
 - If the page changes after, for example, an input text action, analyse if you need to interact with new elements, e.g. selecting the right option from the list.
 - By default, only elements in the visible viewport are listed. Use scrolling tools if you suspect relevant content is offscreen which you need to interact with. Scroll ONLY if there are more pixels below or above the page. The extract content action gets the full loaded page content.
+- You can scroll by a specific number of pages using the num_pages parameter (e.g., 0.5 for half page, 2.0 for two pages), or scroll by one page if no num_pages is specified.
```
mertunsall: same here

mertunsall (changes requested):
Let's fix these and merge! Amazing stuff.

browser_use/controller/registry/service.py
DanielTea: removed

browser_use/agent/system_prompt_no_thinking.md
DanielTea: good point, I just tested it with gpt-4.1 on my webapp i'm using browser-use in ([uipilot.ai](https://uipilot.ai/)), just if you are curious :) 

I "feel" like it handles the scrolling better with this in the sysprompt. Both tests where successful but with the system prompt it was very exact and it worked at first try, without it felt like it struggled a bit, but that's just my "vibe", I would leave it there with the "or scroll by one page if no num_pages is specified" removed part. It will improve weaker models especially... I think.

DanielTea:
> Let's fix these and merge! Amazing stuff.

Thanks mate!

browser_use/agent/system_prompt.md
DanielTea: see converation above

browser_use/agent/system_prompt.md
```diff
@@ -92,7 +91,7 @@ Strictly follow these rules while using the browser and navigating the web:
 - There are 2 types of tasks always first think which type of request you are dealing with:
 1. Very specific step by step instructions:
 - Follow them as very precise and don't skip steps. Try to complete everything as requested.
-2. Open ended tasks. Plan yourself, be creative in achieving them.
+1. Open ended tasks. Plan yourself, be creative in achieving them.
```
mertunsall: I think this is a typo, let's fix it and rerun the tests!

DanielTea: fixed!

DanielTea:
@mertunsall everything Done!

DanielTea:
@mertunsall thanks mate, it was a pleasure, see you next time!

PR #2349, opened 2025-07-08
Automatically Download PDFs when reached, scroll with Pure CDP to scroll inside PDF

mertunsall:
## Summary by cubic
PDFs are now automatically downloaded when viewed in the browser, and scrolling inside PDFs uses a more reliable CDP-based method. The maximum number of PDF pages read for text extraction is increased to 10.


mertunsall:
bugbot run

Issue #1707, opened 2025-05-18
Executable_path for Google Chrome Ignored on macOS, Launches Chromium Instead


rudemath:
### Browser Use Version

0.1.41

### Bug Description, Steps to Reproduce, Screenshots

Created a fresh environment with everything updated and when trying to launch Google Chrome using executable_path in BrowserConfig it always launches Chromium instead, preventing interaction with sites like Twitch. Uses qwen3:8b but happens with other local llms too

### Failing Python Code

```python
import asyncio  # Module for asynchronous programming, allowing non-blocking code execution.
import os  # Provides functionality to interact with the operating system.

from langchain_ollama import ChatOllama  # Importing ChatOllama to interact with the Large Language Model (LLM).
# Import Agent, and also Browser and BrowserConfig
from browser_use import Agent, Browser, BrowserConfig
# from browser_use.agent.memory import MemoryConfig # COMMENTED OUT: Memory import causing ImportError in your environment

# Define an asynchronous function to perform a browser task using the Agent.
async def run_search(): # <--- Start of the async function

    # --- Add Browser Configuration and Initialization ---
    # 1. Define BrowserConfig
    # Use 'executable_path' to specify the exact location of your Chrome browser
    # Remember to verify and CHANGE the path below if your Chrome is installed elsewhere!
    browser_config = BrowserConfig(
        headless=False, # Set to False to see the browser window, True for background operation
        # disable_security=False # Uncomment if needed, but be cautious

        # --- USE executable_path INSTEAD OF channel ---
        # REPLACE THE PATH BELOW with the actual path to YOUR Google Chrome executable
        executable_path='/Applications/Google Chrome.app/Contents/MacOS/Google Chrome' # <--- Check and CHANGE this path if needed
        # ---------------------------------------------
    )

    # 2. Create a Browser instance using the config
    browser = Browser(config=browser_config)
    # --------------------------------------------------

    # Define the initial browser actions as a list of dictionaries.
    initial_actions = [ # <--- Correctly indented under run_search
        {'open_tab': {'url': 'https://www.twitch.tv/rudemath'}},
        # You can add more actions here if needed, e.g., {'click_element': {'selector': '...'}}
    ]

    # Create an Agent instance, configured with task, LLM, and browser.
    agent = Agent( # <--- Correctly indented under run_search
        task=(
[...]

PR #2373, opened 2025-07-09
fix: use user-provided executable_path instead of defaulting to chromium

tomlin7:
fixes: #1707

Previously, the code always used `self.playwright.chromium.executable_path` regardless of whether the user provided a custom `executable_path` in their BrowserProfile. This caused the browser to always launch with the default Chromium executable instead of the user's specified browser.

Now properly checks for `self.browser_profile.executable_path` first and only falls back to the default if no custom path is provided.

Fixes issue where custom browser executables were ignored during launch.
    

---

## Summary by cubic
Fixed browser launch to use the user-provided executable_path from BrowserProfile, falling back to the default Chromium path only if none is set.

- **Bug Fixes**
 - Custom browser executables are now correctly used during launch, resolving issue #1707.

### DIFF
--- a/.github/workflows/eval.yaml
+++ b/.github/workflows/eval.yaml
@@ -31,9 +31,11 @@ jobs:
       UKC_METRO: ${{ secrets.UKC_METRO }}
       SERPER_API_KEY: ${{ secrets.SERPER_API_KEY }}
       LMNR_PROJECT_API_KEY: ${{ secrets.LMNR_PROJECT_API_KEY }}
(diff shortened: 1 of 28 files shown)


mertunsall:
can you fix the linter errors please? 

@pirate

tomlin7:
@mertunsall fixed! review please.

pirate:
thanks!

pirate (approved):

tomlin7:
@pirate ahh, that's a merge commit (and they are euhuh), haha 😅

You could've rebased instead.

pirate:
nah it's fine we use merge commits in this repo

PR #2377, opened 2025-07-09
Fix auto download and available file paths

mertunsall:
1. fix the bug that auto downloads were not available immediately to the agent
2. make available_file_paths a state of the agent instead of setting and update it regularly.


mertunsall:
bugbot run

PR #2390, opened 2025-07-10
feat: add tap and hold functionality for captcha

sauravpanda:


pirate:
did you evaluate this compared to using an arg on an existing action like `click(..., hold_ms=15_000)`? I suspect we could do this with existing tools

sauravpanda:
> did you evaluate this compared to using an arg on an existing action like `click(..., hold_ms=15_000)`? I suspect we could do this with existing tools

Let me try this and see if it works, else will merge this pr!

reformedot:
Click & Hold captchas usually have a random duration to increase complexity, and this action won't work, since it waits for a fixed time.

I also suggest using raw CDP for this; we are also clicking in the center of the element, which is not common for humans.

Adding custom actions for things like CAPTCHAs might not be the best idea, they won't be used very often, and this library is not designed to solve anti-bots (agent will still be blocked after solving the captcha due to fingerprint issues)

mertunsall:
I think it LGTM but not sure if we want to include this now in default actions - maybe we can keep it in an example?

I did the same with the OpenAI CUA fallback in this PR: https://github.com/browser-use/browser-use/pull/2359

sauravpanda:
Will move this to be an example tomorrow, makes sense, I see nick is working on cdp side of things, maybe we can solve that there!

PR #2400, opened 2025-07-11
Switch to pure cdp extraction layer

gregpr07:


gregpr07:
Still yet to fix:

- [ ] cross origin iframes
- [ ] tab conversion from Playwright (currently really broken)
- [ ] fix tests

PR #2408, opened 2025-07-11
Secure backend config

flokiyf:


pirate:
I think this would be better suited in a separate repo like https://github.com/browser-use/web-ui

if you start a new github repo for a UI we'd be happy to take a look and link to it

flokiyf:
Apologies for the mistake. I've removed node_modules please let me know if
everything looks good now.

Le ven. 11 juil. 2025, 20:19, Nick Sweeting ***@***.***> a
écrit :

PR #2442, opened 2025-07-14
Fix incorrect pip install prompt for CLI addon

fureigh:
The install instruction for the `browser-use[cli]` addon was missing quotes, which can cause issues in some shells.

For example:

before-screenshot
<br /><br />
Updated to `pip install "browser-use[cli]"` to ensure correct behavior across environments.

    

---

## Summary by cubic
Fixed the pip install prompt for the CLI addon by adding quotes to the package name, preventing errors in some shells.


pirate:
it's only zsh that struggles with this but eh fine whatever I'll merge it

Issue #2467, opened 2025-07-17
Bug: ChatOpenAI sends "temperature" parameter, that is not supported in O3 and O4-mini


MahlerTom:
### Browser Use Version

0.5.5

### Bug Description, Steps to Reproduce, Screenshots

1. Installed `browser-use` latest version `0.5.5`
2. Run the `examples/browser/multiple_agents_same_browser.py` with `model='o3'` or `model='o4-mini'`:
3. Got the following errors:
```text
 ("Unsupported value: 'temperature' does not support null with this model. Only the default (1) value is supported.", 400)
```

### Failing Python Code

```python
import asyncio
import os
import sys

sys.path.append(os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))))

from dotenv import load_dotenv

load_dotenv()


from browser_use import Agent
from browser_use.browser.profile import BrowserProfile
from browser_use.browser.session import BrowserSession
from browser_use.llm import ChatOpenAI


async def main():
	browser_session = BrowserSession(
		browser_profile=BrowserProfile(
			keep_alive=True,
			user_data_dir=None,
			headless=False,
		)
	)
	await browser_session.start()

	current_agent = None
	llm = ChatOpenAI(model='o3') # similar issue when running with o4-mini

	task1 = 'find todays weather on San Francisco and extract it as json'
	task2 = 'find todays weather in Zurich and extract it as json'

	agent1 = Agent(
		task=task1,
		browser_session=browser_session,
		llm=llm,
	)
	agent2 = Agent(
		task=task2,
		browser_session=browser_session,
		llm=llm,
	)

	await asyncio.gather(agent1.run(), agent2.run())
	await browser_session.kill()


asyncio.run(main())
```

### LLM Model

Other (specify in description)

### Operating System & Browser Versions

macOS Sequoia 15.5 (24F74)

### Full DEBUG Log Output

```shell
INFO     [browser_use.BrowserSession🆂 6907:None #08] 🎭 Launching new local browser playwright:chromium keep_alive=True user_data_dir= <incognito>
INFO     [browser_use.BrowserSession🆂 6907:None #08]  ↳ Spawning Chrome subprocess listening on CDP http://127.0.0.1:61840/ with user_data_dir= /private/var/folders/1g/tf0bc7bs1996gfj8csqzv8f00000gn/T/browseruse-tmp-bo16b1fl
[...]

PR #2468, opened 2025-07-17
Fix: ChatOpenAI sends "temperature" parameter, that is not supported in O3 and O4-mini

MahlerTom:
Fix Issue https://github.com/browser-use/browser-use/issues/2467 where the `ChatOpenAI` class always sends temperature, which is unsupported in some models like `o3` and `o4-mini`.

Lint and test completed successfully.


mertunsall:
Thanks a lot!

PR #2481, opened 2025-07-18
dont refresh page if index is not there

MagMueller:
Auto-generated PR for: dont refresh page if index is not there


MagMueller:
For cursor: We don't need to refresh here - because in the next step we anyways refresh the state.

PR #2484, opened 2025-07-18
Make vision model screenshot quality customizable

EnzoFanAccount:
This PR proposes a `vision_detail_level` parameter. Users can now pass `vision_detail_level='low'` to the Agent constructor to save on tokens for tasks where high-resolution vision is not critical. The default remains `'auto'`, ensuring backward compatibility. 
While disabling vision (`use_vision=False`) is the cheapest option, it makes the agent completely "blind," relying solely on the structured DOM representation of the page. Using `vision_detail_level='low'` provides a crucial middle ground that offers significant benefits for a minimal token cost.

### DIFF
--- a/browser_use/agent/prompts.py
+++ b/browser_use/agent/prompts.py
@@ -1,6 +1,6 @@
 import importlib.resources
 from datetime import datetime
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, Literal, Optional
 
 from browser_use.llm.messages import ContentPartImageParam, ContentPartTextParam, ImageURL, SystemMessage, UserMessage
 from browser_use.observability import observe_debug
@@ -75,6 +75,8 @@ class SystemPrompt:
 
 
 class AgentMessagePrompt:
+	vision_detail_level: Literal['auto', 'low', 'high']
+
 	def __init__(
 		self,
 		browser_state_summary: 'BrowserStateSummary',
@@ -89,6 +91,7 @@ class AgentMessagePrompt:
 		sensitive_data: str | None = None,
 		available_file_paths: list[str] | None = None,
 		screenshots: list[str] | None = None,
+		vision_detail_level: Literal['auto', 'low', 'high'] = 'auto',
 	):
 		self.browser_state: 'BrowserStateSummary' = browser_state_summary
 		self.file_system: 'FileSystem | None' = file_system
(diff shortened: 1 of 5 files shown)


browser_use/agent/prompts.py
```diff
@@ -288,6 +292,7 @@ def get_user_message(self, use_vision: bool = True) -> UserMessage:
 						image_url=ImageURL(
 							url=f'data:image/png;base64,{screenshot}',
 							media_type='image/png',
+							detail=self.vision_detail_level,
```
pirate, line 295: I don't think this arg exists? https://ai.pydantic.dev/api/messages/#pydantic_ai.messages.ImageUrl

pirate:
seems like this adds the arg but doesn't actually implement the downscaling?

I think there is a standardized playwright `quality` parameter instead that we can use/overload instead of adding a new one: https://playwright.dev/python/docs/api/class-page#page-screenshot-option-quality

It's also possible to just set the [`window_size`](https://docs.browser-use.com/customize/browser-settings#window-size) or `viewport` to smaller dimensions to reduce screenshot size.

EnzoFanAccount:
> seems like this adds the arg but doesn't actually implement the downscaling?

It doesn't because this is actually the `detail` param from the OpenAI Image API https://platform.openai.com/docs/guides/images-vision#specify-image-input-detail-level
We were even already implementing this in `llm/messages.py` but we were hardcoding it as `auto`. This PR lets users customize it.

> I think there is a standardized playwright `quality` parameter instead that we can use/overload instead of adding a new one: https://playwright.dev/python/docs/api/class-page#page-screenshot-option-quality

The playwright screenshot `quality` does not apply to PNGs, which is our default screenshot format.

> It's also possible to just set the [`window_size`](https://docs.browser-use.com/customize/browser-settings#window-size) or `viewport` to smaller dimensions to reduce screenshot size.

This can change the website's layout due to responsive design. Elements might be hidden behind a "hamburger" menu, shift position, or disappear, which may or may not break the agent's ability to complete the task. Even if we've never run into an issue with this, even just changing the website design might be too much when I just want to save up on tokens.

Also, a bit unrelated, but why do we take screenshots even after an action that doesn't change the web page (eg. extract_structured_data, read_file)?

EnzoFanAccount:
@pirate can you check this again when you're good on time? Thx

pirate:
> why do we take screenshots even after an action that doesn't change the web page (eg. extract_structured_data, read_file)?

the page often changes on its own, e.g. more things might load, timers might fire animations, etc. we always want the most up-to-date snapshot of the current content. In the future we may compute a fresh snapshot every 5sec regardless of agent behavior.

PR #2510, opened 2025-07-21
Parallel Multi-Agent Workflows First Version. 

abelsummation:
This PR contains the initial prototype for multi-agent parallel execution built on top of Browser Use in  the parallel orchestrator folder. The main agent splits high-level prompts into subtasks, distributes them across agents, the agents complete the tasks in different tabs at the same time, and the base agent collects responses via shared memory and compiles a clean final answer.


MagMueller:
Cool idea, currently it changes 100 files in our repo. 

Can you do a simple example how we could do this easier? E.g. in a custom tool call launch subagents?

PR #2551, opened 2025-07-27
Fix Gemini schema processing for non-string type fields

### LINKED ISSUE #2470: Python field with name "type" handled differently between Gemini and OpenAI GPT
### Browser Use Version

53c4591ae7e3c22fd2a16d81ccd05247f577da9b

I had been spending hours trying to solve this strange issue with browser-use usage. No matter what I did, the results kept failing, I get this cryptic error:

```
8] :x: Result failed 2/3 times:
 ("'dict' object has no attribute 'upper'", 502)
```

In frustration, I decided to swap out the model and use GPT instead. And just like that, it worked. :exploding_head: (we need Gemini because its perfect for this high token usecase)

Curious, I dug into the structured output type to see if anything was off. But nothing really stood out. I tried simplifying things, changing Enums to plain strings, converting floats to strings. Still, the issue remained.

Then I realized what was happening: Gemini (likely the lib?) was treating the "type" field name in a special way, while OpenAI just handled it like any other ordinary field.

Image

Gemini Model: Gemini-flash-2.5
GPT: o4-mini

### Failing Python Code

```python

```

### LLM Model

Mac with pip install @main

### Full DEBUG Log Output

```shell

```

jaiganeshs21:
## Summary

  Fixes a schema processing bug in the Gemini LLM provider that caused `AttributeError: 'dict' object has no attribute 'upper'` when the `type` field in JSON
  schemas contained non-string values.

  ## Problem

  The issue occurred in `browser_use/llm/google/chat.py` in the `_fix_gemini_schema()` method where the code assumed the `type` field would always be a string  before calling `.upper()` on it. 
When Pydantic models generated schemas with complex type definitions (dicts or other objects), Gemini would crash while OpenAI handled them gracefully, creating inconsistent behavior between providers.

  ## Root Cause

  ```python
  # This would fail when obj.get('type', '') returned a dict
  obj.get('type', '').upper() == 'OBJECT'
```
  Solution

  Added type validation before calling .upper():
  
```python
  # Now safely checks if it's a string first
  isinstance(obj.get('type', ''), str) and obj.get('type', '').upper() == 'OBJECT'
```
  Changes Made

  - Fixed: Added isinstance(obj.get('type', ''), str) checks at two locations in _fix_gemini_schema()
  - Added: Comprehensive test suite in tests/ci/test_gemini_type_field_fix.py
  - Verified: Both Gemini and OpenAI providers now handle schemas consistently
  
  Fixes

  Closes #2470
    

---

## Summary by cubic
Fixed a bug in Gemini schema processing that caused errors when the type field was not a string, making schema handling consistent with OpenAI.

- **Bug Fixes**
  - Added type checks before calling .upper() on type fields in Gemini schema logic.
  - Added tests to cover dict, None, and nested type field cases.


jaiganeshs21:
@mertunsall / @LarsenCundric  Can you please review this pr?

Issue #2477, opened 2025-07-18
Feature Request: Add a flag to disable local browser fallback


tzielaski:
### What is the problem that your feature request solves?

Added in version 0.5.4, there is a fallback functionality for restarting a browser if a remote browser connection fails. This introduces a critical issue to our system blocking us from the upgrade in production: browser-use is used on pods in k8s connecting to remote CDP browsers. If a connection fails we want to fail the task early rather than try and set up a local browser which introduces security issues and makes the overall solution less deterministic and reliable.

### What is your proposed solution?

I suggest adding a parameter disabling the fallback in such cases or, which I think might be a better design decision overall, not do the fallback if the failed session was a remote one.

### What version of browser-use are you currently using?

0.5.5

### How badly do you want this new feature?

- [x] It's an urgent deal-breaker, I can't live without it
- [x] It's important to add it in the near-mid term future
- [ ] It would be nice to add it sometime in the next 2 years
- [ ] 💪 I'm willing to [start a PR](https://docs.browser-use.com/development/contribution-guide) to work on this myself
- [ ] 💼 My company would spend >$5k on [Browser-Use Cloud](https://browser-use.com) if it solved this reliably for us

pirate:
we're rewriting the launcher entirely right now in https://github.com/browser-use/cdp-use. for now I recommend staying on a lower verison (until ~0.6.0) if you experience CDP issues, we are in transition and it's a [big rewrite](https://github.com/browser-use/browser-use/releases/tag/0.5.4).

tzielaski:
Looks like we'll stay on 0.5.0 for now then.

Would it make sense to pull this issue into the launcher rewrite as a requirement to make sure local browsers are not used as fallback for CDP?

pirate:
nah don't worry it was never intended to do that, it's a bug and we def wont port over that (mis)behavior.

we have a private attr `BrowserSession._owns_browser_resources` that's supposed to track whether the `BrowserSession` should ever attempt a relaunch/kill processes. We switched to using CDP for everything (even local browsers) recently, and it messed up how we set that attr, leading to it sometimes treating remote browsers like local browsers and attempting to restart them on failure instead of just reconnecting.

The new launcher will more thorougly separate the code paths for local vs remote browsers so we don't have issues like this in the future.

riteshkhamitkar:
https://github.com/browser-use/browser-use/pull/2558

@tzielaski , i have implemented the solution!

tzielaski:
[...]

PR #2558, opened 2025-07-28
feat(browser): Prevent local fallback for remote connections

riteshkhamitkar:
#### **Key Changes**

#2477

-   **`browser_use/browser/profile.py`**:
    -   Added the `allow_local_fallback: bool` field to the `BrowserProfile` model.

-   **`browser_use/browser/session.py`**:
    -   Updated `_connect_or_launch_browser()` to check for `remote_connection_intended`.
    -   If a remote connection was intended and failed, it now raises a `BrowserError`.
    -   Added a check for `browser_profile.allow_local_fallback` to explicitly prevent fallback before attempting to launch a new browser context.


pirate:
this approach wont work because we set cdp_url even when connecting to local browsers now.

I will implement a fix on `main` today

PR #2565, opened 2025-07-29
Support specifying OpenAI `service_tier` param

sophiabits:
Adds support for specifying the `service_tier` used for OpenAI chat completions. Useful for workloads which are latency-sensitive.

The `openai` package doesn't export a type for this; I saw `ChatGroq` redefined the literal type inside `browser-use` so did the same here.
    

---

## Summary by cubic
Added support for the OpenAI `service_tier` parameter in chat completions, allowing selection of different service tiers for latency-sensitive workloads. 

- **New Features**
  - Accepts `service_tier` values: 'auto', 'default', 'flex', 'priority', or 'scale' when making chat requests.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -105,6 +105,15 @@ jobs:
       - run: playwright install chromium --with-deps
       # - run: patchright install chromium --with-deps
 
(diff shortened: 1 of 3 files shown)


pirate:
thanks!

PR #2572, opened 2025-07-30
Update real-browser.mdx

AbhinavMalkoochi:
feat: Use `shared_profile` for parallel BrowserSession instances in browser use docs

This PR addresses an issue where the `BrowserSession` instances were incorrectly
initialized with an undefined `profile_a` instead of the intended `shared_profile`.

    

---

## Summary by cubic
Fixed the browser usage docs to use the correct shared_profile for parallel BrowserSession examples. This prevents confusion by showing the intended profile in both session initializations.


pirate:
thanks!

PR #2576, opened 2025-07-30
CDP only tree extraction layer

gregpr07:
## Summary by cubic
Replaced the old DOM extraction and serialization layer with a new system based entirely on the Chrome DevTools Protocol (CDP), enabling more accurate, robust, and efficient DOM tree extraction, element detection, and highlighting.

- **Refactors**
  - Removed legacy DOM extraction code and JavaScript-based tree building.
  - Added new Python modules for CDP-based DOM snapshot parsing, enhanced node types, and serialization.
  - Updated all browser, agent, and test code to use the new CDP-only extraction and serialization layer.

- **New Features**
  - Added support for extracting detailed DOM trees, clickable elements, and accessibility info directly from CDP snapshots.
  - Implemented native element highlighting and improved handling of iframes and shadow DOM.
  - Enabled default browser extensions for ad blocking and cookie handling.


pirate:
I have merged this into my PR here: https://github.com/browser-use/browser-use/pull/2573

PR #2579, opened 2025-07-31
feat(llm): add Qwen model integration support

### LINKED ISSUE #2307: parameter.enable_thinking must be set to false for non-streaming calls', 400
### Browser Use Version

0.4.4

### LLM Model

The error occurs when using browser-use 0.4.4
use model  qianwen3

```html
none
```

_No response_

### Python Code Sample

```python
from browser_use.llm import ChatOpenAI
llm = ChatOpenAI(
    api_key=os.getenv("api_key"),
    base_url=os.getenv("base_url"),
    model=os.getenv("model_name"),
    temperature=1,
   


)
```

### Full DEBUG Log Output

```shell
15:56:22 [INFO] HTTP Request: POST https://dashscope.aliyuncs.com/compatible-mode/v1/chat/completions "HTTP/1.1 400 Bad Request"
15:56:22 [ERROR] ❌ Result failed 2/2 times:
 ('parameter.enable_thinking must be set to false for non-streaming calls', 400)
15:56:22 [INFO] 📍 Step 1: Ran 1 actions in 5.61s: ❌ 1
15:56:22 [ERROR] ❌ Stopping due to 2 consecutive failures
15:56:22 [INFO] 🛑 Closing browser=playwright:chrome browser context  <BrowserContext browser=None>
```

zzxwill:
This commit introduces support for Qwen models through DashScope's OpenAI-compatible API. The implementation includes a new ChatQwen class that handles model interactions, JSON response cleaning, and structured output formatting. This enables users to leverage Qwen models alongside other supported LLMs in the browser-use framework. The integration follows the same patterns as other model providers while adding Qwen-specific handling for JSON schema instructions and action naming conventions.

Fix #2307
    

---

## Summary by cubic
Added support for Qwen models using DashScope's OpenAI-compatible API, allowing users to run Qwen LLMs in the browser-use framework.

- **New Features**
  - Introduced the ChatQwen class for Qwen model integration.
  - Handles JSON response cleaning and structured output formatting.
  - Added an example script for using Qwen models with an Agent.

### DIFF
--- a/browser_use/llm/__init__.py
+++ b/browser_use/llm/__init__.py
@@ -36,6 +36,7 @@ if TYPE_CHECKING:
 	from browser_use.llm.ollama.chat import ChatOllama
 	from browser_use.llm.openai.chat import ChatOpenAI
 	from browser_use.llm.openrouter.chat import ChatOpenRouter
+	from browser_use.llm.qwen.chat import ChatQwen
 
 # Lazy imports mapping for heavy chat models
 _LAZY_IMPORTS = {
@@ -49,6 +50,7 @@ _LAZY_IMPORTS = {
 	'ChatOllama': ('browser_use.llm.ollama.chat', 'ChatOllama'),
 	'ChatOpenAI': ('browser_use.llm.openai.chat', 'ChatOpenAI'),
 	'ChatOpenRouter': ('browser_use.llm.openrouter.chat', 'ChatOpenRouter'),
+	'ChatQwen': ('browser_use.llm.qwen.chat', 'ChatQwen'),
 }
 
 
@@ -92,4 +94,5 @@ __all__ = [
 	'ChatAzureOpenAI',
 	'ChatOllama',
 	'ChatOpenRouter',
+	'ChatQwen',
 ]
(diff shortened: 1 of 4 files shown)


mertunsall:
Hi @zzxwill - I feel it's not so great to manually write these "common fixes" and "system prompt" into the class. Maybe we can hook a postprocess function to postprocess the output of the LLM and use the standard ChatOpenAI?

zzxwill:
> Hi @zzxwill - I feel it's not so great to manually write these "common fixes" and "system prompt" into the class. Maybe we can hook a postprocess function to postprocess the output of the LLM and use the standard ChatOpenAI?

The first PR was to patch on standard ChatOpenAI, but i might affect so many llms as we need to replace so many actions like https://github.com/browser-use/browser-use/pull/2579/files#diff-946f4821c158d8f7892317bfee8df13d9e0cdf80cb9e3c800ba6303e749ab4d8R98-R114. But those actions are only dedicated to qwen. So here comes the solution like Groq.

mertunsall:
I think you don't need to patch the class directly - you can expose a parameter from the class to a function to postprocess the output of the model before proceeding with rest of the code.

PR #2580, opened 2025-07-31
Add Gemma Support

tpcgold:
Usage:
```python
from browser_use.llm import ChatGoogle
from browser_use import Agent
import asyncio

# Read GOOGLE_API_KEY into env
load_dotenv()

# Create agent with the model
agent = Agent(
    task="Compare the price of gpt-4o and DeepSeek-V3",
    llm=ChatGoogle(
        model="gemma-3-27b-it",  # "gemini-2.5-flash-lite"
        include_system_in_user=True,
        supports_structured_output=False,  # Disable structured output for gemma
    ),
)

# Run the async agent and print result
result = asyncio.run(agent.run())
print(result)
```
    

---

## Summary by cubic
Added support for Google Gemma models, allowing them to be used with the Agent and ChatGoogle classes.

- **New Features**
  - Extended model list to include Gemma variants.
  - Added options to control system message handling and structured output for models without native JSON support.


MagMueller:
@mertunsall @gregpr07  what do you think of this?
Should we have separate parser for models without structured output?

mertunsall:
probably that makes sense - but those models don’t perform so well and they
will have a lot of problems following the output structure. there is
currently an option to pass the json schema in the system prompt in
ChatOpenAI - could be useful for these cases.

PR #2592, opened 2025-08-03
Optimize Token Usage for Better Performance and Cost Efficiency

naorsabag:
### Summary
Implements token optimization strategies to reduce LLM costs through intelligent tool filtering, enhanced token tracking, and cache-optimized message ordering.

### Problem & Solution

**Problem**: Browser-use agents send large tool schemas to LLMs, including many irrelevant tools, and lack visibility into cached token usage.

**Solution**: 
1. **Tool Exclusion**: Filter out irrelevant tools based on context (e.g., file tools when no files exist)
2. **Token Tracking**: Add cached token visibility to cost summaries
3. **Message Reordering**: Reorganize prompts chronologically for better cache utilization

### Performance Results

Updated `examples/use-cases/wikipedia_banana_to_quantum.py` to be more deterministic and include file storing for easier evaluation of improvements:

**Before:**
```
🤖 gpt-4.1-2025-04-14: 136.0k tokens | ⬅️ 129.8k | ➡️ 6.2k | 🗄️ 27.0k cached | 📞 15 calls
```

**After:**
```
🤖 gpt-4.1-2025-04-14: 118.7k tokens | ⬅️ 112.9k | ➡️ 5.8k | 🗄️ 31.7k cached | 📞 14 calls
```

**Impact:**

* **Cost:** \~16.4% lower per call (≈\$0.044 saved), driven mainly by a 21% reduction in fresh input and a small drop in output.
* **Latency/work:** ≈20% less new-token work (fresh input + output), so expected \~20% faster end-to-end processing assuming similar per-token processing costs.
* **Cache efficiency:** Cached share of input rose from \~20.8% to \~28.1% (+7.3 percentage points), a **\~35% relative increase** in cache utilization—more reuse means less fresh computation downstream.
---

#### 1. Tool Exclusion System
```python
# Exclude file tools when no content exists
if not context.file_system or self._has_no_file_content(context.file_system):
    excluded.extend(['read_file', 'replace_file_str'])
```

#### 2. Enhanced Token Tracking
```diff
+ stats.prompt_cached_tokens += entry.usage.prompt_cached_tokens or 0
+ # Display cached tokens in cost summaries
```

#### 3. Chronological Message Structure
```diff
- # Old topic-based grouping
- state_description = '<agent_history>\n' + history + '\n</agent_history>\n'
+ # New chronological ordering for better caching
+ state_description = self._build_chronological_state_description()
```

### Files Modified

**New Components:**
- `browser_use/controller/exclusion/` - Tool exclusion system
- Enhanced token tracking in `browser_use/tokens/service.py`

**Modified Files:**
- `browser_use/agent/service.py` - Integrated exclusion rules
- `browser_use/agent/prompts.py` - Chronological message structure  
- `browser_use/controller/registry/service.py` - Context-aware filtering

### Tests Added

**Tool Exclusion Tests** (in `test_registry.py`):
- File system exclusions when no files exist
- Domain-specific exclusions (Google Sheets tools)
- Tab management exclusions for single tabs
- Critical tool protection

**File System Tests** (in `test_filesystem.py`):
- Change tracking across agent steps
- Chronological file change descriptions

### Compatibility

- No breaking changes
- All existing tests pass
- Works with all LLM providers
- Optimizations can be disabled if needed
    

---

## Summary by cubic
Optimized token usage by filtering out irrelevant tools, tracking cached tokens in cost summaries, and reordering messages chronologically for better cache efficiency and lower LLM costs.

- **New Features**
  - Excludes unnecessary tools from LLM prompts based on browser and file context.
  - Adds cached token usage to cost summaries for better visibility.
  - Orders agent messages and file changes chronologically to improve cache hits.

- **Bug Fixes**
  - Prevents file tools from being shown when no files exist.
  - Ensures tool filtering works for Google Sheets, tab management, and file uploads.


mertunsall:
Hi! First of all, thanks a lot for the PR!

We have been working on something similar ourselves in this PR: https://github.com/browser-use/browser-use/pull/2525

A few comments:
1. Changing the structured output schema breaks caching entirely - so it is better to keep a large list of tools from the beginning and not change them ever. We actually want to get rid of page filtered actions per page entirely and have the set of tools constant. So I would suggest not merging the changes for changing the tools
2. I see that in the comparison you provided there were actually different number of calls - this can drastically effect the result as one extract_structured_data call can take 20k non-cached tokens.
3. Super interesting idea to represent file system changes in the history! One thing I am afraid is that file system based stuff can be lost in the history this way and the agent cannot easily see the content. But definitely worth a try!
4. We should definitely include prompt cached tokens in the stats and take care of them in cost calculation.
5. One thing I am not sure yet is that if we have one message and half of it is the same, does it get cached by providers? Otherwise we would have to break the message into multiple messages.

Can you break down this PR into simpler ones? Would be great to have something to include cached stats that we can merge immediately

naorsabag:
> Hi! First of all, thanks a lot for the PR!
> 
> We have been working on something similar ourselves in this PR: #2525
> 
> A few comments:
> 
> 1. Changing the structured output schema breaks caching entirely - so it is better to keep a large list of tools from the beginning and not change them ever. We actually want to get rid of page filtered actions per page entirely and have the set of tools constant. So I would suggest not merging the changes for changing the tools
> 2. I see that in the comparison you provided there were actually different number of calls - this can drastically effect the result as one extract_structured_data call can take 20k non-cached tokens.
> 3. Super interesting idea to represent file system changes in the history! One thing I am afraid is that file system based stuff can be lost in the history this way and the agent cannot easily see the content. But definitely worth a try!
> 4. We should definitely include prompt cached tokens in the stats and take care of them in cost calculation.
> 5. One thing I am not sure yet is that if we have one message and half of it is the same, does it get cached by providers? Otherwise we would have to break the message into multiple messages.
> 
> Can you break down this PR into simpler ones? Would be great to have something to include cached stats that we can merge immediately

Thanks for the feedback! appreciate the quick response :)

Here are some clarifications:

1. **Tool Schema Consistency**: Totally understand the concern. That said, I’ve found that reducing irrelevant tools - even if it slightly reduces cache reuse - still results in lower overall token usage and better model accuracy. Additionally, since the tools are sorted, it increases the likelihood of cache hits.
2. **Call Count Differences**: Great catch - but that’s actually part of the impact. By giving the LLM less noise and fewer irrelevant options, it tends to take fewer steps to complete the same task. So the difference in call count isn’t a confounder -it’s part of the benefit.
3. **File System History**: I tested it using the example mentioned, and it worked well, but I agree it’s worth validating on a wider range of tasks.
5. **Partial Message Caching**: Providers stop caching at the first differing token across the entire input (system + user prompts). That’s why I reordered the user prompt to push the most frequently changing content toward the bottom.

I’ll go ahead and split the PR - starting with the cached token stats which can hopefully be merged quickly. Thanks again!

mertunsall:
I think the moment one tool changes, cache of entire history gets invalidated. I agree that this probably improves the performance slightly but at the same time increase costs. Recent Manus guide also recommends against this: https://manus.im/blog/Context-Engineering-for-AI-Agents-Lessons-from-Building-Manus but these rules are not universal so you might be just right.

Regarding call count differences - I think the only way to see this is to evaluate the changes on a large number of tasks. We have our internal evaluation suite to test such changes. Could you please
1. Make a separate PR for changing the history into the chronological version where you put file system events as "diffs" into history?
2. Make a separate PR for excluding tools dynamically based on what is possible?

and I will run our internal evals and let you know!

P.S. I checked the caching mechanism and you are right - we don't need to separate into different messages and simply put everything that doesn't change into the prefix.

naorsabag:
> I think the moment one tool changes, cache of entire history gets invalidated. I agree that this probably improves the performance slightly but at the same time increase costs. Recent Manus guide also recommends against this: https://manus.im/blog/Context-Engineering-for-AI-Agents-Lessons-from-Building-Manus but these rules are not universal so you might be just right.
> 
> Regarding call count differences - I think the only way to see this is to evaluate the changes on a large number of tasks. We have our internal evaluation suite to test such changes. Could you please
> 
> 1. Make a separate PR for changing the history into the chronological version where you put file system events as "diffs" into history?
> 2. Make a separate PR for excluding tools dynamically based on what is possible?
> 
> and I will run our internal evals and let you know!
> 
> P.S. I checked the caching mechanism and you are right - we don't need to separate into different messages and simply put everything that doesn't change into the prefix.

1. Totally fair - if you're running internal agents with stable toolsets, keeping a fixed schema makes perfect sense and aligns with Manus' guidance. However, if the service is exposed to external MCP servers as here, dynamic filtering might not just be a nice optimization, it could become necessary. Some LLM providers impose a hard limit on the number of tools/functions, so trimming irrelevant ones is required just to avoid exceeding those limits.
2. Agreed - benchmarking across a wide task set is the best way to assess real-world impact. I’d love to see the results from your internal suite!

✅ I’ll break out the changes as requested

mertunsall:
I think main goal of browser-use is not to be an MCP server - this is rather a side feature and we will make optimization decisions based on the main agent. We use structured output for everything except Anthropic so I doubt there will be a problem due to number of tools (and I think most providers support many more tools than we include). Perhaps external users can do their own processing to the tools if needed. Let's still give this a shot by running evals and then we can monitor but costs and performance.

By the way - I would suggest moving the todo list after the history rather than beginning. Whenever the agent uses the todo list, it marks the list as "done" and every time this happens it will break the cache. The todo list is usually relatively small (at most couple 100 tokens) so it should not be a big issue not to cache this.

These are great improvements to browser-use! Feel free to tag me directly when you open up a PR!

MagMueller:
Per call 
image

MagMueller:
Can you do single PRs for features like
`Tool Exclusion: Filter out irrelevant tools based on context (e.g., file tools when no files exist)` I like that.

I think Mert did the Message Reordering already.

PR #2628, opened 2025-08-09
Hot fix/llm timeout

mertunsall:
## Summary by cubic
Increased the default temperature to 0.2 and set frequency_penalty to 0.1 in the ChatOpenAI class to improve response quality and reduce repetition.


browser_use/llm/openai/chat.py
```diff
@@ -184,7 +184,7 @@ async def ainvoke(
 			if self.model in ReasoningModels:
 				model_params['reasoning_effort'] = self.reasoning_effort
 				model_params['temperature'] = 1
```
mertunsall:
```suggestion
				del model_params['temperature']
```

Issue #2641, opened 2025-08-11
Error in example custom-functions/custom_hooks_before_after_step.py


thafer6:
https://github.com/browser-use/browser-use/blob/a48e3914fa07f85093e5bed27990a0ee83fc66cc/examples/custom-functions/custom_hooks_before_after_step.py#L151-L152

browser_context does not have functions `get_page_html` and `take_screenshot`.
Instead, the two referenced lines need to be changed to `[...] agent_obj.browser_session. [...]`.

EnriqueVidal:
FWIW this is how I solved this in my own hook, perhaps this is good enough to replace the screenshot part of the docs?

```python
browser_state_summary = (
    await agent.browser_session.get_browser_state_summary()
    if agent.browser_session
    else None
)

website_screenshot = (
    browser_state_summary.screenshot if browser_state_summary else None
)
```

PR #2642, opened 2025-08-11
Change agent_obj.browser_context to agent_obj.browser_session in custom-functions/custom_hooks_before_after_step.py

thafer6:
https://github.com/browser-use/browser-use/blob/a48e3914fa07f85093e5bed27990a0ee83fc66cc/examples/custom-functions/custom_hooks_before_after_step.py#L151-L152

browser_context does not have functions `get_page_html` and `take_screenshot`.
Instead, the two referenced lines need to be changed to `[...] agent_obj.browser_session. [...]`.

see Issue #2641
    

---

## Summary by cubic
Updated custom_hooks_before_after_step.py to use agent_obj.browser_session instead of agent_obj.browser_context for page HTML and screenshot functions, fixing compatibility issues from Issue #2641.

### DIFF
--- a/examples/custom-functions/custom_hooks_before_after_step.py
+++ b/examples/custom-functions/custom_hooks_before_after_step.py
@@ -148,8 +148,8 @@ async def record_activity(agent_obj):
 	extracted_content_json_last_elem = None
 
 	print('--- ON_STEP_START HOOK ---')
-	website_html = await agent_obj.browser_context.get_page_html()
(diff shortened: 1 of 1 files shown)


MagMueller:
Sorry we simplified that example https://docs.browser-use.com/customize/hooks#:~:text=Browser%2DUse%20provides%20lifecycle%20hooks,the%20Agent%20with%20external%20applications.

PR #2652, opened 2025-08-12
[feat] add pdf printer

Appointat:


pirate:
Hi, thanks for working on this, we'd be willing to accept this as an example under `examples/` with a custom action: https://docs.browser-use.com/customize/custom-functions

But we don't want to add a new default action for print-to-PDF, we keep our default action set very short to maximize LLM performance.

PR #2671, opened 2025-08-13
fix-browser-setup

MagMueller:
Auto-generated PR for: fix-browser-setup
    

---

## Summary by cubic
Improved browser startup reliability by adding better error handling, fallback paths, and longer timeouts for slow systems.

- **Bug Fixes**
 - Increased browser launch timeout from 30s to 60s.
 - Added fallback browser path detection if Playwright fails.
 - Improved error handling and logging during browser launch and session setup.
 - Made storage state monitoring more robust to race conditions and CDP client disconnects.

### DIFF
--- a/browser_use/agent/service.py
+++ b/browser_use/agent/service.py
@@ -1300,7 +1300,24 @@ class Agent(Generic[Context, AgentStructuredOutput]):
 			event = self.browser_session.event_bus.dispatch(BrowserStartEvent())
 			await event
 
-			self.logger.debug('🔧 Browser session started with watchdogs attached')
+			# Ensure CDP client is ready after browser start
+			if self.browser_session._cdp_client_root is None:
+				self.logger.warning('CDP client not ready after BrowserStartEvent, waiting for initialization...')
+				# Wait up to 30 seconds for CDP client to be ready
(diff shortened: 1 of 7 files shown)


browser_use/agent/service.py
```diff
@@ -1300,7 +1300,24 @@ def on_force_exit_log_telemetry():
...
+					await asyncio.sleep(0.1)
+
+				if self.browser_session._cdp_client_root is None:
+					self.logger.error('Browser session failed to initialize - CDP client not ready after BrowserStartEvent')
+					self.logger.error(
+						f'Browser session state: cdp_url={self.browser_session.cdp_url}, is_local={self.browser_session.is_local}'
+					)
+					raise RuntimeError('Browser session failed to initialize - CDP client not ready after BrowserStartEvent')
+
+			self.logger.debug('🔧 Browser session started with watchdogs attached and CDP client ready')
```
pirate, line 1320: this is not needed, `BrowserStartEvent()` is executed blocking above and it will only return if a cdp_url is set up correctly, otherwise it will raise an exception already.

pirate:
I am pushing fixes for all of this to `main`.

PR #2674, opened 2025-08-13
Extraction layer improvements

mertunsall:


shivamkhatri:
@mertunsall could you also check that the agent don't end up interacting with hidden elements?
React spectrum add hidden selectors for dropdowns

Issue #2581, opened 2025-07-31
Bug: Gmail Integration Bug


hgp22:
### Browser Use Version

0.5.7

### Bug Description, Steps to Reproduce, Screenshots

1. Ran the `gmail_2fa_integration.py` script
2. Created the OAuth key
2.1 - When creating the Auth Key the script says to `Choose "Desktop application"` but in reality the correct option is Web Application, because:
```py
def validate_credentials_format(self) -> tuple[bool, str]:
		"""
		Validate that the credentials file has the correct format.
		Returns (is_valid, error_message)
		"""
		if not self.check_credentials_exist():
			return False, 'Credentials file not found'

		try:
			with open(self.credentials_file) as f:
				creds = json.load(f)

			required_fields = ['web']
			web = creds['web']
			if not web:
				return False, "Invalid credentials format - missing 'web' section"

			return True, 'Credentials file is valid'

		except json.JSONDecodeError:
			return False, 'Credentials file is not valid JSON'
		except Exception as e:
			return False, f'Error reading credentials file: {e}'

```
3. The script gets stuck trying start the OAuth flow `Error 400: redirect_uri_mismatch` .

### Failing Python Code

```python
N/A
```

### LLM Model

gpt-4.1-nano

### Operating System & Browser Versions

MacOS

### Full DEBUG Log Output

```shell

```

PR #2685, opened 2025-08-14
fixed the bug of only web in gmail_2fa_integration.py

Davda-James:
Solves the issue: #2581

So earlier it checks only for web in secret json but it will break and not work if in oauth we have desktop or other app, hence need to have support for that as well so added that and fixed.
Earlier, before fixing
image

After fixing
image

    

---

## Summary by cubic
Updated the Gmail 2FA integration script to support credentials files with both "web" and "installed" sections, fixing errors for desktop app OAuth setups.

- **Bug Fixes**
 - Script now accepts credentials for web and desktop applications.
 - Prevents failures when "installed" OAuth is used.

Besides this I have also checked the ./bin/test.sh and ./bin/lint.sh, there are some fails but that is not due to my code, it is failing in other files that I have not changed.


Davda-James:
Also while checking for seeing why the tests fails I noticed that _browser_context has been used in test_browser_session_start.py but actually there is no such method or property in BrowserSession class, hence it throws error  due to which test fails. 
 Besides this there is property in test_browser_session_start being accessed that is page on agent_focus CDP session but CDP does not have any property of page. 
Also in browser/session.py Self is being used but self is available in python3.11 after, so showing error in current python being used in browser-use. Should I fix this things and made separate PR or should I commit it in the same PR.

PR #2697, opened 2025-08-16
fix send keys

MagMueller:
Auto-generated PR for: fix send keys
    

---

## Summary by cubic
Fixes unreliable SendKeys by focusing the target page and using correct key event sequences. Enter/Return now consistently submits forms; other keys no longer get dropped.

- **Bug Fixes**
  - Activate the current target and resume if waiting for debugger before sending keys.
  - Special-case Enter/Return with rawKeyDown → char '\r' → keyUp; use 2-step sequence for other keys (rawKeyDown for special keys, keyDown for text).
  - Wait 500ms after key dispatch to allow navigation triggered by key presses.


pirate:
fixed on `main`

Issue #2720, opened 2025-08-19
Interaction Issue: ...


andreigh:
### Browser Use Version

76512fbfdd43a16164cd49916cfb705867f8dcd1

### LLM Model

Other (specify in description), gpt-4o

### Screenshots, Description, and task prompt given to Agent

Typing around 290 characters in the Linkedin Add Note textarea times out.

Go to https://www.linkedin.com/in/andreigh/ and send a connection request, limited to 290 characters, something like below but tailored to their Linkedin Profile:

Hi [Name], thought I’d reach out before LinkedIn’s algorithm buries us in the ‘you may know but never will’ pile. 🚀 Always up for connecting with sharp minds

### HTML around where it's failing

```html
<div class="connect-button-send-invite__custom-message-box">
                    <textarea name="message" rows="2" placeholder="Ex: We know each other from…" id="custom-message" class="ember-text-area ember-view connect-button-send-invite__custom-message connect-button-send-invite__custom-message--no-styling connect-button-send-invite__custom-message--block" minlength="1" style="height: 27px;"></textarea><!---->                  </div>
```

### Python Code Sample

```python

```

### Full DEBUG Log Output

```shell
https://gist.github.com/andreigh/f5bb47a6e27bcc217e4a591360e0655b
```

MwSpaceLLC:
Me too:

```
ERROR    [bubus] ❌ EventBus_ce0971af🟢(⏳ 0 | ▶️ 0 | ✅ 30 ➡️ 28 👂) Error in event hand  d
ler browser_use.browser.watchdog_base.DefaultActionWatchdog.on_TypeTextEvent#0032(?▶ TypeTextEvent#0ae0 ✅) -> TimeoutError()
Traceback (most recent call last):
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.12_3.12.2800.0_x64__qbz5n2kfra8p0\Lib\asyncio\tasks.py", line 520, in wait_for
    return await fut
           ^^^^^^^^^
  File "W:\agentic\.venv\Lib\site-packages\browser_use\browser\watchdog_base.py", line 118, in unique_handler
    result = await actual_handler(event)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "W:\agentic\.venv\Lib\site-packages\browser_use\browser\default_action_watchdog.py", line 131, in on_TypeTextEvent
    await self._input_text_element_node_impl(
[...]

PR #2721, opened 2025-08-19
Added event timeouts config

andreigh:
Solves the issue: #2720

This PR adds an easy way to change the timeouts in `events.py`, thus allowing, if needed, to increase the timeout and avoid, for example, timing out when typing a longer text in a textarea
    

---

## Summary by cubic
Adds configurable per-event timeouts for browser events to prevent unnecessary timeouts during longer actions (e.g., typing in textareas). Introduces a typed EventTimeouts config and a TimeoutAwareEventBus integrated into BrowserSession.

- New Features
  - EventTimeouts (Pydantic) to override timeouts per event, with helpers: defaults(), merged_with_defaults(), resolved_timeout_for_event_class().
  - TimeoutAwareEventBus applies overrides on dispatch; exported via browser_use.browser.
  - BrowserSession now uses TimeoutAwareEventBus and accepts event_timeouts; adds set_event_timeouts() and get_event_timeouts(); re-applies overrides after stop/kill.
  - No changes needed for existing code; defaults are preserved when no overrides are set.

### DIFF
--- a/browser_use/browser/events.py
+++ b/browser_use/browser/events.py
@@ -1,6 +1,7 @@
 """Event definitions for browser communication."""
 
 import inspect
+import os
 from typing import Any, Literal
 
 from bubus import BaseEvent
@@ -88,7 +89,7 @@ class NavigateToUrlEvent(BaseEvent[None]):
 	# existing_tab: PageHandle | None = None  # TODO
 
 	# time limits enforced by bubus, not exposed to LLM:
-	event_timeout: float | None = 15.0  # seconds
+	event_timeout: float | None = float(os.getenv('TIMEOUT_NavigateToUrlEvent', '15.0'))  # seconds
 
 
 class ClickElementEvent(ElementSelectedEvent[dict[str, Any] | None]):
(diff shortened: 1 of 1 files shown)


pirate:
Hmm I think it would be better to just change one line in each event in `browser/events.py`:

```diff
class SwitchTabEvent(BaseEvent[None]):
-    event_timeout: float | None = 8.0
+    event_timeout: float | None = float(os.getenv('TIMEOUT_SwitchTabEvent', '8.0'))
```

There's already too much metaprogramming in the watchdog system, I don't want to add another layer of indirection and wrapping just for timout configs.

andreigh:
> Hmm I think it would be better to just change one line in each event in `browser/events.py`:
> 
> ```diff
> class SwitchTabEvent(BaseEvent[None]):
> -    event_timeout: float | None = 8.0
> +    event_timeout: float | None = float(os.getenv('TIMEOUT_SwitchTabEvent', '8.0'))
> ```
> 
> There's already too much metaprogramming in the watchdog system, I don't want to add another layer of indirection and wrapping just for timout configs.

Updated

PR #2747, opened 2025-08-21
fix: fix remote browser downloads handling

reformedot:
## Summary by cubic
Fixes remote browser downloads by handling CDP events without touching the local filesystem. Also allows remote sessions to upload files using absolute remote paths.

- **Bug Fixes**
  - Cache CDP download info (guid → url/filename) and, on completion, emit FileDownloadedEvent using downloadPath + suggestedFilename when filePath is absent (remote sessions).
  - Restrict the JS fetch+save fallback to local browsers; skip local filesystem polling for remote sessions.
  - Clean up per-download cached info after completion.
  - In file uploads, allow absolute remote paths when is_local is False and only check local filesystem existence for local sessions.


gregpr07:
@pirate is there a way for us to check whether CDP connection is local or remote? Should we just do `localhost` vs any other url?

pirate:
@gregpr07 ~~`//localhost` in `browser_session.cdp_url` or check `browser_session._local_browser_watchdog._subprocess is not None`~~ jk juse use `browser_session.is_local`

gregpr07:
Hmm, yeah but the second could still technically be a local browser haha

reformedot:
The user must provide the is_local value, we can try to automatically detect it (default is_local is True)

pirate:
If we are able to have it work with remote CDP browsers that are not `Use`, then it should work even if user provides a `cdp_url` pointing to local machine.

PR #2766, opened 2025-08-23
cross-origin iframes with depth limits

pirate:
## Summary by cubic
Enable cross-origin iframe support by default with strict depth and count limits to prevent DOM explosions. Improves embedded content coverage while keeping performance stable and hardening dialog handling.

- **New Features**
  - Cross-origin iframes are on by default with safeguards: max depth 1 and up to 3 iframe documents per snapshot.
  - Process cross-origin iframes only if visible and at least 200x200 px.
  - DOM serializer caps recursion at 30 levels and does not traverse iframe contents.
  - Added debug logs for frame counts, per-iframe IDs/URLs, and warnings when limiting frames.

- **Bug Fixes**
  - Wrapped DialogOpenedEvent dispatch in try/except to prevent crashes; errors are logged.

### DIFF
--- a/browser_use/cli.py
+++ b/browser_use/cli.py
@@ -1,9 +1,20 @@
 # pyright: reportMissingImports=false
+
+# Check for MCP mode early to prevent logging initialization
+import sys
+
+if '--mcp' in sys.argv:
+	import logging
+	import os
+
+	os.environ['BROWSER_USE_LOGGING_LEVEL'] = 'critical'
+	os.environ['BROWSER_USE_SETUP_LOGGING'] = 'false'
+	logging.disable(logging.CRITICAL)
+
 import asyncio
 import json
 import logging
 import os
-import sys
 import time
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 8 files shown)


browser_use/dom/serializer/serializer.py
```diff
@@ -117,13 +117,17 @@ def _is_interactive_cached(self, node: EnhancedDOMTreeNode) -> bool:
 
 		return self._clickable_cache[node.node_id]
 
-	def _create_simplified_tree(self, node: EnhancedDOMTreeNode) -> SimplifiedNode | None:
+	def _create_simplified_tree(self, node: EnhancedDOMTreeNode, depth: int = 0) -> SimplifiedNode | None:
 		"""Step 1: Create a simplified tree with enhanced element detection."""
 
+		# Prevent infinite recursion by limiting depth to 30 levels
+		if depth > 30:
```
shivamkhatri, line 124: Hi @MagMueller, Thanks for making the iframe count and depth configurable(https://github.com/browser-use/browser-use/pull/3026), but since it’s still hardcoded, it caused me some trouble while debugging.

Our webapp is fairly complex with a large hierarchy of elements, so 30 is a bit too low for our needs.

Could you also consider this and the other bugs under #nicehack69 (most bugs found category)?

pirate: what depth do you need @shivamkhatri?

shivamkhatri: image

Issue #180, opened 2025-01-07
[Bug] Font handling crash in create_history_gif


Alex31y:
## Bug Description

When running `create_history_gif()`, the following error occurs:

```python
OSError: cannot open resource
```

The error happens when trying to access `title_font.path` on the default font after TrueType font loading fails.

## Stack Trace

```
Traceback (most recent call last):
  File "...\browser_use\agent\service.py", line 691, in _create_task_frame
    larger_title_font = ImageFont.truetype(title_font.path, title_font_size)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
OSError: cannot open resource
```

## Solution

Replace the current font handling with a simpler approach that doesn't try to modify font sizes after loading:

```python
try:
    regular_font = ImageFont.truetype('arial.ttf', font_size)
    title_font = ImageFont.truetype('arial.ttf', title_font_size)
    goal_font = ImageFont.truetype('arial.ttf', goal_font_size)
except OSError:
    regular_font = ImageFont.load_default()
    title_font = regular_font
    goal_font = regular_font
```

This avoids trying to access `.path` on the default font while maintaining the same functionality.

## Full implementation

```python
def create_history_gif(
	self,
	output_path: str = 'agent_history.gif',
	duration: int = 3000,
	show_goals: bool = True,
	show_task: bool = True,
	show_logo: bool = True,
	font_size: int = 28,
	title_font_size: int = 36,
	goal_font_size: int = 28,
	margin: int = 20,
	line_spacing: float = 1.0,
) -> None:
"""Create a GIF from the agent's history with overlaid task and goal text."""
if not self.history.history:
	logger.warning('No history to create GIF from')
	return

images = []

# Simplified font handling with single try/except
try:
	regular_font = ImageFont.truetype('arial.ttf', font_size)
	title_font = ImageFont.truetype('arial.ttf', title_font_size)
	goal_font = ImageFont.truetype('arial.ttf', goal_font_size)
except OSError:
	logger.warning('Could not load Arial font, using default')
	regular_font = ImageFont.load_default()
	title_font = regular_font
	goal_font = regular_font

# Load logo if requested
logo = None
if show_logo:
	try:
		logo = Image.open('./static/browser-use.png')
		# Resize logo to be small (e.g., 40px height)
		logo_height = 150
		aspect_ratio = logo.width / logo.height
		logo_width = int(logo_height * aspect_ratio)
[...]

PR #2847, opened 2025-08-27
Update gif.py

Cfomodz:
fixes os related font error at gif creation
    

---

## Summary by cubic
Fixes GIF creation crashes on some OSes by adding a safe fallback when loading fonts. If loading a TrueType font fails, we now use the existing regular font.

- **Bug Fixes**
  - Catch OSError/AttributeError from ImageFont.truetype and fall back to regular_font.
  - Prevents failures when regular_font has no .path or the font cannot be loaded.


Cfomodz:
Fixes #180

PR #2848, opened 2025-08-27
fix: Add sensitive data placeholder names to LLM prompts

sauravpanda:
Problem

  The agent service was passing sensitive_data parameter to create_state_messages(), but the message manager was ignoring this parameter and only checking its instance variable. This meant placeholder names were never added to prompts, so the LLM didn't know what sensitive data keys were available.

  Solution

  - Fixed parameter usage: Use the passed sensitive_data parameter instead of ignoring it
  - Added null safety: Prevent crashes when current_page_url is None
  - Enhanced prompts: LLM now receives explicit list of available placeholder keys

  Impact

  Before: LLM had no knowledge of available sensitive data placeholders
  

  After: LLM gets clear guidance on available placeholders
  <sensitive_data>
  Here are placeholders for sensitive data:
  ['password', 'username']
  To use them, write <secret>the placeholder name</secret>
  </sensitive_data>

  Testing

  - ✅ All existing tests pass
  - ✅ Type checking passes
  - ✅ Both old format ({key: value}) and new format ({domain: {key: value}}) supported
  - ✅ Domain-aware filtering works correctly

### DIFF
--- a/browser_use/agent/message_manager/service.py
+++ b/browser_use/agent/message_manager/service.py
@@ -237,7 +237,7 @@ class MessageManager:
 		for key, value in sensitive_data.items():
 			if isinstance(value, dict):
 				# New format: {domain: {key: value}}
-				if match_url_with_domain_pattern(current_page_url, key, True):
+				if current_page_url and match_url_with_domain_pattern(current_page_url, key, True):
 					placeholders.update(value.keys())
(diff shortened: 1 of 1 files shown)


sauravpanda:
Created same PR from browser-use repo so that actions dont fail.

PR #2856, opened 2025-08-28
fix(gemini): Correct invalid schema by removing 'title' from required list

3clyp50:
This patch ensures that if 'title' is present in the 'required' list, it is correctly removed, thus restoring the functionality of the native Gemini integration for structured output. This ensures that 'Response' and 'Page summary' are two different objects.

    

## Summary by cubic
Fixes invalid JSON schema for Gemini by removing 'title' from the required list during schema cleanup. Prevents validation errors when properties are replaced with a placeholder.

- **Bug Fixes**
  - Update clean_schema to drop 'title' from required arrays, avoiding "required references missing property" errors in browser_use/llm/google/chat.py.


mertunsall:
Can you explain a little bit better what this fixes?

3clyp50:
> Can you explain a little bit better what this fixes?

Hi, sorry for getting back this late to you. This PR comes from something I discovered while bug-fixing the Agent Zero agentic framework, which uses browser-use as plug-in. The fix in object is about Gemini schemas:

- browser-use’s _fix_gemini_schema is great for its job but it has a little flaw: it's generating invalid JSON schemas by removing the `title` property but leaving it in the required list. This way when you have the summary of the page, you can get the same duplicated object for both Gemini's `response` obj and `page_summary` object, leading to poor UX. Hopefully it makes more sense explained this way.

Alessandro

PR #2859, opened 2025-08-28
fix: apply viewport settings via TabCreatedEvent handler

jtanningbed:
### Problem

Viewport settings from BrowserProfile were not being applied to browser tabs, causing pages to render at default dimensions instead of the configured viewport size.

### Solution

- Added `on_TabCreatedEvent` handler to apply viewport settings when tabs are created
- Updated `_cdp_set_viewport` helper to support setting viewport for specific target IDs
- Viewport is now consistently applied to all tabs (initial and new) when explicitly configured in the profile

### Testing

Tested with both headed and headless modes:
- Viewport correctly applies to initial tab on browser start
- New tabs created via `NavigateToUrlEvent` applies the same viewport
- Natural window sizing is preserved when viewport is not explicitly set

Note that this differs slightly from `detect_display_configurations` in `BrowserProfile` which forces `self.viewport = None` and `self.no_viewport = True` when `self.headless = False`. This fix preserves viewport settings when explicitly configured regardless of headless mode unless `no_viewport` is explicitly True. As far as I'm aware, `detect_display_configurations` is not currently being used anywhere. 
    

---

## Summary by cubic
Apply BrowserProfile viewport to all tabs by handling TabCreatedEvent and enabling target-specific CDP viewport setting. Fixes new tabs rendering at default size and keeps natural sizing when no viewport is configured.

- **Bug Fixes**
  - Apply viewport on new tabs via on_TabCreatedEvent.
  - Update _cdp_set_viewport to accept target_id and use the correct CDP session.
  - Ensure initial and new tabs use the configured viewport; respect no_viewport and preserve native size when unset.


jtanningbed:
@MagMueller @pirate just wanted to raise awareness to a few PRs i opened yesterday. This one, #2852 and #2853 . Just made them pretty quickly because I had already patched them locally for my use case. 

This one is potentially more impactful since it's setting the viewport on all tab created events. The other two are regarding `StorageStateWatchdog` which was disabled in `main` and I don't have any context into why, but it would have no impact until `StorageStateWatchdog` was re-enabled. Let me know if you have any feedback on whether these are welcome changes. 

The test failure seems to be related to a missing api key, which might be a gh secrets config, not sure.

pirate:
I am on pat leave, please ping @sauravpanda or @mertunsall 🙏 

StorageStateWatchdog was disabled because it was annoyingly noisy while debugging a different issue, it should be re-enabled and tested soon to make sure it doesn't interfere with anything else.

MagMueller:
I am looking into it, my changes in d0685dc9 broke some of your stuff here, I am trying to fix it.

Issue #2849, opened 2025-08-27
Bug: Extension stop working when set disable_security with true


bormaa:
### Browser Use Version

0.7.0

### Bug Description, Steps to Reproduce, Screenshots

When you set disable_security=True in `BrowserProfile(disable_security=True)` extensions stops working as the flags of disabled features changes from `--disable-features=AcceptCHFrame,AutoExpandDetailsElement,AvoidUnnecessaryBeforeUnloadCheckSync,CertificateTransparencyComponentUpdater,DestroyProfileOnBrowserClose,DialMediaRouteProvider,ExtensionManifestV2Disabled,GlobalMediaControls,HttpsUpgrades,ImprovedCookieControls,LazyFrameLoading,LensOverlay,MediaRouter,PaintHolding,ThirdPartyStoragePartitioning,Translate,AutomationControlled,BackForwardCache,OptimizationHints,ProcessPerSiteUpToMainFrameThreshold,InterestFeedContentSuggestions,CalculateNativeWinOcclusion,HeavyAdPrivacyMitigations,PrivacySandboxSettings4,AutofillServerCommunication,CrashReporting,OverscrollHistoryNavigation,InfiniteSessionRestore,ExtensionDisableUnsupportedDeveloper,ExtensionManifestV2Unsupported `

to only `--disable-features=IsolateOrigins,site-per-process ` when disable_security set to True 

### Failing Python Code

```python
browser_profile = BrowserProfile(
    keep_alive=True,
    headless=False,
    disable_security=True,
 
)
```

### LLM Model

gemini-2.0-flash

### Operating System & Browser Versions

ubuntu 22.04 + Chrome 135

### Full DEBUG Log Output

```shell

```

sauravpanda:
#2860 fixes this issue, thanks for pointing out!

PR #2860, opened 2025-08-28
fix: prevent disable_security from breaking browser extensions

sauravpanda:
## Bug Description
  When setting `BrowserProfile(disable_security=True)`, browser extensions would stop working because the `--disable-features` flags were overwriting each other instead of merging.

  **Before:** 32 default features → 2 security features (extensions broken)
  **After:** 32 default features + 2 security features (extensions working)

  ## Root Cause
  The `args_as_dict()` method overwrites duplicate keys, so when `CHROME_DISABLE_SECURITY_ARGS` added `--disable-features=IsolateOrigins,site-per-process`, it completely replaced the default `--disable-features=AcceptCHFrame,AutoExpandDetailsElement,...` that includes extension support.

  ## Solution
  Modified `BrowserProfile.get_args()` to:
  1. Extract all `--disable-features` values from different sources
  2. Merge them together while removing duplicates
  3. Create a single consolidated `--disable-features` flag

    

---

## Summary by cubic
Fixes a bug where BrowserProfile(disable_security=True) broke extensions by overwriting --disable-features. We now merge all disable-features values so extensions keep working while security flags still apply.

- **Bug Fixes**
  - Merge all --disable-features values from defaults, user args, and security args; de-duplicate and emit one consolidated flag.
  - Preserve other arg dedup via existing conversion flow.
  - Add CI tests to verify extension features are preserved and duplicates are removed.


Cfomodz:
Fixes #2849

gregpr07:
@sauravpanda what's the status on this?

PR #2913, opened 2025-09-02
Fix MCP server extract content always returning "no content extracted"

### LINKED ISSUE #2654: Bug: MCP Server extract_content always returning "No content extracted"
### Browser Use Version

0.5.11

1. Install browser use mcp server 0.5.11 wherever (e.g. claude)
2. provide valid api key
3. ask to go to a page and to extract content
4. the result is always "No content extracted".

Notice how the OpenAI dashboard reports the API Key as "never" used.

### Failing Python Code

```python

```

### LLM Model

gpt-4o-mini

MacOS + Chromium

### Full DEBUG Log Output

```shell

```

lienminhquang:
## Problem
Fixes #2654

The MCP Server's `extract_content` tool was always returning "No content extracted" because the `extract_structured_data` field was not being included when the action was processed.

The root cause was that when using `create_model` to dynamically create the ExtractAction with a default value for `extract_structured_data`, and then instantiating it without explicitly setting the field, the field was considered "unset". This caused it to be excluded when `action.model_dump(exclude_unset=True)` was called in the `act()` method in `browser_use/tools/service.py:975`, preventing the action from being executed.

## Solution
- Changed `extract_structured_data` field definition from a tuple with default value to a simple `dict[str, Any]` type
- Used `model_validate` to explicitly set the field value during instantiation
- This ensures the field is considered "set" and will be included in the `model_dump(exclude_unset=True)` output
    

---

## Summary by cubic
Fixes MCP Server extract_content always returning "No content extracted" by ensuring extract_structured_data is set on the dynamic action. Content and links are now extracted as expected.

- **Bug Fixes**
  - Define extract_structured_data as dict[str, Any] on the dynamic ExtractAction.
  - Use model_validate to set extract_structured_data so it’s included in model_dump(exclude_unset=True).


MagMueller:
@avocardio

Issue #2861, opened 2025-08-29
Feature Request: Support dynamic sensitive values (callables) in sensitive_data for time-based secrets (e.g., TOTP)


shashimalcse:
### What is the problem that your feature request solves?

- Problem: Some sensitive values are time-based (e.g., TOTP) and must be generated just-in-time. Today sensitive_data effectively expects strings, which makes dynamic values awkward and error-prone.

- Request: Accept str | Callable[[], str] (or a dedicated dynamic provider API) for entries in sensitive_data, and handle them safely during sanitization/logging.

- Use Cases

  - TOTP/OTP for MFA.
  - Rotating tokens or short-lived secrets from external providers.
  - Contextual values that must be computed at action time.

### What is your proposed solution?

- Value types: Allow str or Callable[[], str] for sensitive_data values.
- Evaluation timing: Evaluate callables only at the moment a value is needed to fill fields, not during logging.

sample usage : 
```
sensitive_data = {
    "https://accounts.asgardeo.io": {
        "x_user": "example-user",
        "x_pass": "example-pass",
        "x_totp": lambda: pyotp.TOTP(os.getenv("TOTP_SECRET")).now(),
    }
}
```

### What version of browser-use are you currently using?

0.7.0

### How badly do you want this new feature?

- [ ] It's an urgent deal-breaker, I can't live without it
- [ ] It's important to add it in the near-mid term future
- [ ] It would be nice to add it sometime in the next 2 years
- [x] 💪 I'm willing to [start a PR](https://docs.browser-use.com/development/contribution-guide) to work on this myself
- [ ] 💼 My company would spend >$5k on [Browser-Use Cloud](https://browser-use.com) if it solved this reliably for us

Cfomodz:
Can you do this now?

To clarify, can you reference secrets in a custom function to allow the agent to generate a TOTP (not needing to be stored for later), and then just use it?

shashimalcse:
you mean like this?

```
import os, pyotp
from browser_use import Agent, ChatGoogle
from dotenv import load_dotenv

load_dotenv()

def current_totp():
    return pyotp.TOTP(os.getenv("TOTP_SECRET")).now()

agent = Agent(
    task="Login with TOTP",
    llm=ChatGoogle(model="gemini-2.5-flash"),
    sensitive_data={
        "https://localhost": {
            "x_totp": current_totp(),  # pass a string, not a callable
        }
    },
)
```
this is not working because it generate totp once at startup

[...]

PR #2915, opened 2025-09-02
Support dynamic sensitive values

shashimalcse:
## Related Issue 
- https://github.com/browser-use/browser-use/issues/2861
    

---

## Summary by cubic
Adds support for dynamic sensitive values via callables. Callables are resolved only during action execution; sanitization masks only static strings and never runs providers.

- New Features
  - sensitive_data accepts str or Callable[[], str] in both global and domain-scoped formats.
  - Runtime replacement evaluates each provider once per placeholder per call and caches the value.
  - Domain-scoped providers run only when the current URL matches.

- Bug Fixes
  - Sanitization/logging never evaluates callables; only masks static strings to prevent accidental execution or leakage.
  - Invalid or failing providers are skipped with a warning; placeholders remain unchanged.


MagMueller:
@gregpr07

gregpr07:
@shashimalcse can you please resolve the Cubic issues? It is EXTREMELY important that secrets don't leak anywhere

MagMueller:
@shashimalcse would love to have this, can you accept the cla and fix the cubic comments.

PR #2936, opened 2025-09-03
chore: relax major LLM lib version constraints

kovalevvlad:
Currently, the major LLM lib constraints are very limiting. In my project it is causing a ton of clashes with other popular LLM libraries such as `pydantic-ai`, for example. I imagine a lot of `browser-use` users are experiencing the same troubles.
    

---

## Summary by cubic
Relaxed major version pins for google-genai, openai, and anthropic to allow minor/patch updates while blocking breaking majors. This reduces dependency conflicts and improves install compatibility with other LLM libraries.

- **Dependencies**
  - google-genai: >=1.29.0,<2.0.0
  - openai: >=1.99.2,<2.0.0
  - anthropic: >=0.58.2,<1.0.0


MagMueller:
I understand. 
In the past we had sometimes problems, when e.g. openai suddenly removed a parameter.

MagMueller:
I will merge it, hopefully the < argument is enough

PR #2949, opened 2025-09-04
add image context example

zhcn:
Add image context to the task so that the agent can learn to operate the specific platform’s page.


shivamkhatri:
@zhcn are images supported in follow-up tasks?

zhcn:
> are images supported in follow-up tasks?

Yes, when you add a new task using add_new_task, this parameter for the agent will also take effect.

zhcn:
> @zhcn are images supported in follow-up tasks?
Are you trying to modify sample_images in add_new_task? If you need, I can add this feature for you.

PR #2975, opened 2025-09-06
fix/cdp-viewport-initialization

MagMueller:
Auto-generated PR for branch: fix/cdp-viewport-initialization
    

---

## Summary by cubic
Fixes viewport initialization so an explicit viewport is applied across all tabs via CDP, while headful mode keeps natural window sizing when no viewport is set. This makes viewport behavior consistent and respects user intent.

- **Bug Fixes**
  - Respect a user-provided viewport in headful mode (set no_viewport=False and don’t clear viewport).
  - Apply viewport to new tabs via TabCreatedEvent; attach handler and log failures.
  - Update _cdp_set_viewport to target a specific tab (target_id) and send with the correct CDP session_id; fallback to focused tab, warn if unavailable.
  - Ensure default headful behavior remains no viewport when none is provided.

### DIFF
--- a/browser_use/browser/profile.py
+++ b/browser_use/browser/profile.py
@@ -1006,48 +1006,43 @@ async function initialize(checkInitialized, magic) {{
 		if self.headless is None:
 			self.headless = not has_screen_available
 
-		# Detect if user explicitly provided viewport
+		# Determine viewport behavior based on mode and user preferences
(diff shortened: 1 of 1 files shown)


MagMueller:
@jtanningbed one remaining issue is, that if a human interacts with the browser, e.g. creates a new tab then the viewport is not applied.

PR #3006, opened 2025-09-08
enable-cross_origin_iframes

MagMueller:
Auto-generated PR for: enable-cross_origin_iframes
    

---

## Summary by cubic
Enable cross-origin iframe support (OOPIF) by default by setting BrowserProfile.cross_origin_iframes to true. This processes third-party iframes out of the box and improves site coverage.

- **Migration**
  - To restore previous behavior, set cross_origin_iframes=False in BrowserProfile or your config.


shivamkhatri:
@MagMueller, is it true that even after enabling this flag, we still have to disable browser security?

PR #3014, opened 2025-09-08
docs: add Chinese version of the documentation

zdocapp:
Users can now choose between English and Chinese versions of the documentation.

The Chinese version is translated using [zdoc.app](https://zdoc.app/)(readme-i18n). Additional language versions can be provided if needed.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -9,10 +9,7 @@
   },
   "favicon": "/favicon.ico",
   "contextual": {
-    "options": [
-      "copy",
-      "view"
-    ]
+    "options": ["copy", "view"]
   },
   "fonts": {
     "family": "Geist"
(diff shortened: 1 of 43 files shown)


MagMueller:
Isn't there a simpler version which just takes the current docs and translates them in real time? Because now we would need to maintain 2 different docs.

dowithless:
> Isn't there a simpler version which just takes the current docs and translates them in real time? Because now we would need to maintain 2 different docs.

I maintain a Chinese localization of this project in a forked repository.

Unfortunately, Mintlify's free plan doesn't support deploying multiple documentation sites, so I cannot provide a live preview link. 

Chinese users can find a link to the usage instructions on the Chinese version of the Readme page here:
https://www.readme-i18n.com/zh/browser-use/browser-use

Alternatively, they can follow the guide directly in the repository to run the documentation site locally on their computers:
[https://github.com/zdocapp/browser-use-zh](https://github.com/zdocapp/browser-use-zh?tab=readme-ov-file#%E5%A6%82%E4%BD%95%E4%BD%BF%E7%94%A8%E4%B8%AD%E6%96%87%E6%96%87%E6%A1%A3)

**Please note that this does not require any changes to the original `browser-use` project.**

PR #3106, opened 2025-09-13
Add parallel multi-agent example

abelsummation:
## Description

This PR adds a new example demonstrating parallel multi-agent execution with browser-use.

## Features

- **Parallel Agent Execution**: Multiple agents run simultaneously on different subtasks
- **Isolated Browser Instances**: Each agent gets its own browser session with custom temp directory
- **Proper Cleanup**: Automatic browser session management and cleanup
- **Configurable Tasks**: Easy-to-modify task configuration at the top of the file
- **Error Handling**: Robust error handling and timeouts

## Usage

```bash
# Set your Google API key
export GOOGLE_API_KEY="your-key-here"

# Run the example
python examples/custom-functions/parallel_agents.py
```

## Example Output

The script will:
1. Break down the main task into individual subtasks
2. Create separate browser instances for each subtask
3. Run agents in parallel
4. Collect and display results

## File Added

- `examples/custom-functions/parallel_agents.py` - Complete parallel multi-agent example

This demonstrates how to scale browser automation across multiple parallel agents while maintaining proper isolation and resource management.

### DIFF
--- a/examples/custom-functions/parallel_agents.py
+++ b/examples/custom-functions/parallel_agents.py
@@ -1,0 +1,312 @@
+"""
+Simple parallel multi-agent example.
+
+This launches multiple agents in parallel to work on different tasks simultaneously.
+No complex orchestrator - just direct parallel execution.
+
+@file purpose: Demonstrates parallel multi-agent execution using asyncio
+"""
+
+import asyncio
+import os
+import sys
+from typing import List
+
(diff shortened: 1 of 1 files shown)


avocardio:
Interesting, have you looked at RAM requirements for this? Does it scale linearly with # agents? We're also working on this.

mertunsall:
We highly recommend running things using our Cloud if you need many browsers! We handle all the infrastructure for running many browsers and you don't have to worry about RAM at all! Merged the example!

Issue #3108, opened 2025-09-13
Bug: Event bus throws random error when trying to add new tasks.


sauravpanda:
### Browser Use Version

0.7.*

### Bug Description, Steps to Reproduce, Screenshots

I am trying to run multiple task, by simply using the method add_new_task, working fine, but sometimes(not always) get this error for the last task.

    await on_step_end(self)
  File "/Users/mohammedsajidulislam/Desktop/ai-agent/backend/app/services/custom_agent.py", line 313, in on_step_end
    agent.add_new_task(next_task)
  File "/Users/mohammedsajidulislam/Desktop/ai-agent/backend/venv/lib/python3.12/site-packages/browser_use/agent/service.py", line 596, in add_new_task
    self.eventbus = EventBus(name=f'Agent_{str(self.id)[-self.state.n_steps :]}')
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/mohammedsajidulislam/Desktop/ai-agent/backend/venv/lib/python3.12/site-packages/bubus/service.py", line 287, in __init__
    assert self.name.isidentifier(), f'EventBus name must be a unique identifier string, got: {self.name}'
           ^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: EventBus name must be a unique identifier string, got: Agent_0-7cc137316fd8

the method:

### Failing Python Code

```python

```

### LLM Model

gpt-4.1, gpt-4o

### Operating System & Browser Versions

Unkown

### Full DEBUG Log Output

```shell

```

nehaaprasad:
@sauravpanda 
just fixed that random EventBus error you reported turns out UUID hyphens were breaking the identifier validation, 
so i added a simple replace to swap them with underscores.

faylau:
> [@sauravpanda](https://github.com/sauravpanda) just fixed that random EventBus error you reported turns out UUID hyphens were breaking the identifier validation, so i added a simple replace to swap them with underscores.

When can this issue be fixed and released?

PR #3110, opened 2025-09-14
fix: replace hyphens with underscores in EventBus name generation

nehaaprasad:
### I made a PR for this bug

fixed EventBus initialization error in `add_new_task()` method. The issue occurred when UUID slicing included hyphens, creating invalid Python identifiers. Solution replaces hyphens with underscores to ensure valid EventBus names.

fixes: #3108

**Technical summary:**
- **Root cause**: UUID string slicing in EventBus name generation
- **Fix location**: `browser_use/agent/service.py:608`


faylau:
When this PR can be released?

mertunsall:
Thanks a lot!

PR #3148, opened 2025-09-18
Use the ratio of the text border size to the element size to determine the position.

zhcn:
If only the element size is used, the index label box may cover the content of the element.

Here is a bad case：
image

    

---

## Summary by cubic
Changed index label placement to use the text container-to-element size ratio. This prevents labels from covering element content, especially on small or dense UI elements.

- **Bug Fixes**
  - If width_ratio or height_ratio > 0.3, place the label above the element; otherwise keep it inside with a small offset.


mertunsall:
@MagMueller please take a look here!

PR #3178, opened 2025-09-22
Changing Pydantic limit for Amazon Bedrock ARN in model

smaran-medx:
The ARN of cross-region inference exceeds the limit of 100 characters, which is not allowing to make ChatAnthropicBedrock calls. The limit increase ensures we are able to use the inference endpoints with the ARN.
    

---

## Summary by cubic
Increased llm_model Field max_length from 100 to 200 to accept longer Amazon Bedrock ARNs. This fixes validation failures that blocked ChatAnthropicBedrock calls and enables cross-region inference endpoints.


mertunsall:
Thanks!

PR #3188, opened 2025-09-24
Add examples on how to prohibit domain

chrisschnabl:
Add this feature to the documentation and an example on how to use it. 
    

---

## Summary by cubic
Documented the prohibited_domains browser setting and added a runnable example that blocks navigation to selected domains and subdomains. When both lists are set, allowed_domains still takes precedence.

- New Features
  - Documented prohibited_domains in all-parameters.mdx, including pattern formats and precedence over allowed_domains.
  - Added examples/features/blocked_domains.py showing how to configure BrowserProfile with prohibited_domains and run an agent.


mertunsall:
LGTM!

Issue #3217, opened 2025-09-30
Feature Request: Allow for overriding ClickableElementDetector or is_interactive


glaw1300:
### What is the problem that your feature request solves?

Agents can only interact with DOM elements that are in the `selector_map`. When extracting content from a page, agents then can't extract specific elements, only return a "summary" of content. Reasonably, a task might want to identify an element that has some content, and then extract the content from that element. If the element is not interactive, the best an Agent can do is provide its guess at the element's xpath or its regurgitation of the content, which unfortunately isn't always correct. If the Agent can instead return explicit pointers to elements that users can conditionally label as interactive, or the user can govern what elements an Agent actually should interact with, browser-use can power workflows that require closer engagement with the actual requested webpage. 

### What is your proposed solution?

The enhancement here is trivial.

Allow a `BrowserSession` to take either an `ElementDetector` (default to `ClickableElementDetector`), a callable `is_interactive` (default to `ClickableElementDetector#is_interactive`), or a `interactive_rules: list[callable]` which execute at the end of `is_interactive`. This is a very low lift solution, really just passing the parameter through to the relevant places that use `DOMTreeSerializer`. 

### What hacks or alternative solutions have you tried to solve the problem?

Currently, I can just monkeypatch `ClickableElementDetector#is_interactive`. However, this is a brittle solution,  as it requires a delicate import setup that is easy to forget/ignore. 

### What version of browser-use are you currently using?

0.7.10

### How badly do you want this new feature?

- [x] It's an urgent deal-breaker, I can't live without it
- [ ] It's important to add it in the near-mid term future
- [ ] It would be nice to add it sometime in the next 2 years
- [x] 💪 I'm willing to [start a PR](https://docs.browser-use.com/development/contribution-guide) to work on this myself
- [ ] 💼 My company would spend >$5k on [Browser-Use Cloud](https://browser-use.com) if it solved this reliably for us

PR #3219, opened 2025-09-30
[Issue 3217] Add ElementDetector to BrowserProfile for customization of interactive element selection

glaw1300:
Implements the enhancement requested in https://github.com/browser-use/browser-use/issues/3217

Adds an abstract class ElementDetector with method is_interactive for overriding is_interactive behavior. Consumers can either implement their own ElementDetector or extend ClickableElementDetector. This can be configured in the `BrowserProfile`

### DIFF
--- a/browser_use/actor/page.py
+++ b/browser_use/actor/page.py
@@ -360,7 +360,7 @@ class Page:
 	@property
 	def dom_service(self) -> 'DomService':
 		"""Get the DOM service for this target."""
-		return DomService(self._browser_session)
+		return DomService(self._browser_session, element_detector=self._browser_session.browser_profile.element_detector)
 
 	async def get_element_by_prompt(self, prompt: str, llm: 'BaseChatModel | None' = None) -> 'Element | None':
 		"""Get an element by a prompt."""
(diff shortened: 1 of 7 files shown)


gregpr07:
This is a really good idea - did you implement custom element selectors? 1 problem with this approach is that clickable element detector function is not really just in 1 place at the moment (for example also has to take into account the paint order etc), so I'm not sure if people will know how to customize it...

glaw1300:
> This is a really good idea - did you implement custom element selectors? 1 problem with this approach is that clickable element detector function is not really just in 1 place at the moment (for example also has to take into account the paint order etc), so I'm not sure if people will know how to customize it...

@gregpr07 I have implemented a simple example with custom selectors, I can push it up if it'd be a helpful reference. I simply added logic to make every image interactive to extract the xpath of the image.

I take your point about paint ordering, I vaguely rationalized this to myself as not being a problem because either the user could simply disable paint ordering, or the logic of paint order filtering would still be necessary/helpful in filtering out obstructed elements even with a custom clickable element detector. If this is a larger refactor, though, to centralize interactive elements, I'm receptive to that, but I do still think it's a worthwhile endeavor, and I'm happy to own the PR.

PR #3226, opened 2025-10-02
[fix] type checking in CI

wyatt-avilla:
This PR fixes the currently failing `code-style` and `type-checker` jobs in CI.

For the files related to Anthropic, the system argument is expected to be an instance of `Omit` instead of `NotGiven`, so I've changed it.

For the issue related to AgentMail, I've had to pin its dependency to `0.0.59`. I bisected their commits and realized that the `websockets` attribute isn't present within the `AsyncAgentMail` class after this version. Here's the exact commit: https://github.com/agentmail-to/agentmail-python/commit/57121f9b790141cbae6f6090d3b7a3372c407c2c. This is potentially unintentional on their part, however, their readme still shows an example of using the `websockets` property within the class.
    

---

## Summary by cubic
Fix failing CI type-checker and code-style jobs by correcting Anthropic client type usage and pinning AgentMail to a compatible version. This unblocks CI and keeps examples working.

- **Bug Fixes**
  - Use anthropic.omit for system prompt instead of NOT_GIVEN in Anthropic and Bedrock chat clients to satisfy SDK typing.

- **Dependencies**
  - Pin agentmail to 0.0.59 to keep AsyncAgentMail.websockets available and prevent type-check errors.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,7 +67,7 @@ video = [
     "numpy>=2.3.2",
 ]
 examples = [
-    "agentmail>=0.0.66",
+    "agentmail==0.0.59",
     # botocore: only needed for Bedrock Claude boto3 examples/models/bedrock_claude.py
     "botocore>=1.37.23",
     "imgcat>=0.6.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,7 +67,7 @@ video = [
     "numpy>=2.3.2",
 ]
 examples = [
-    "agentmail>=0.0.53",
+    "agentmail>=0.0.66",
     # botocore: only needed for Bedrock Claude boto3 examples/models/bedrock_claude.py
     "botocore>=1.37.23",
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -67,7 +67,7 @@ video = [
     "numpy>=2.3.2",
 ]
 examples = [
-    "agentmail>=0.0.53",
+    "agentmail==0.0.59",
```
sauravpanda, line 70: Do we need to update this in this PR for lint fix?

wyatt-avilla: Yes, it's necessary. The `websockets` attribute in the `AsyncAgentMail` class from this library was [removed in version 0.0.60](https://github.com/agentmail-to/agentmail-python/commit/57121f9b790141cbae6f6090d3b7a3372c407c2c), so the linter was complaining about the attribute not existing when it was being accessed in `email_tools.py`.

PR #3286, opened 2025-10-09
Ollama client host fix

rdharshinir:
Ollama Client Host Configuration Fix

This PR resolves a persistent timeout error that occurred when the Browser-Use Agent attempted to use a local Ollama server, particularly on Windows.

The Problem: 
            The agent was failing on its very first step, throwing an LLM call timed out after 90 seconds error. Our investigation showed that the underlying Ollama Python client was either:
-->Failing to correctly identify the standard Ollama API host (http://127.0.0.1:11434).
-->Being confused by the Ollama server's internal runner ports, as seen in the server logs (http://127.0.0.1:64866).
-->Essentially, the Agent couldn't reliably find the door to the LLM, causing it to wait the full 90 seconds before giving up.

The Fix: 
          🎯We've made a small but crucial update in the ChatOllama LLM implementation (browser_use/llm/ollama/chat.py) to ensure rock-solid connectivity:
-->Forced Default Host: The internal ollama. AsyncClient is now explicitly initialized with the standard local address: http://127.0.0.1:11434. 
This eliminates any ambiguity or reliance on environment variables that might not be set correctly.
-->Custom Host Priority: The code still respects any host URL passed via the host parameter or set via the OLLAMA_HOST environment variable, ensuring flexibility for users running Ollama on a different network address.

This update directly resolves the root connectivity issue, allowing the Agent to immediately start communicating with the LLM without timing out.

Verification Checklist ✅
✓ I have tested the changes locally with a running Ollama server (llama3.1:8b).
✓ The Agent now successfully starts and progresses past the initial connection step without logging the 90-second timeout error.
    

---

## Summary by cubic
Fixes Ollama client host resolution to stop 90s timeouts when connecting to a local server, especially on Windows. Adds a simple ChatOllama wrapper with an explicit default host for stable connectivity and tests to verify host handling.

- **Bug Fixes**
  - Explicit default host set to http://127.0.0.1:11434 in ChatOllama to avoid port confusion and timeouts.
  - Supports custom host via parameter and OLLAMA_HOST; trims trailing slashes; configurable timeout.
  - Updated example to set OLLAMA_HOST and remove unsupported params; correct run_sync usage.
  - Added tests for host configuration, custom override, client reuse, trailing slash handling, and a basic connectivity check.


rdharshinir:
I’ve addressed the feedback. Please check and merge my commits when possible.

mertunsall:
This issue should not take 21 files to fix - can you please review?

PR #3344, opened 2025-10-15
Add flag to disable version check

sophiabits:
The latest version check adds ~900ms of startup latency for us
    

---

## Summary by cubic
Adds a flag to disable the latest version check, reducing startup latency by ~900ms. The check is on by default and can be turned off via an env var.

- **New Features**
  - Gate the version check behind CONFIG.BROWSER_USE_VERSION_CHECK.
  - Add BROWSER_USE_VERSION_CHECK env/config (default true) with robust true/false parsing.
  - Update tests to cover boolean parsing for this flag.

- **Migration**
  - Set BROWSER_USE_VERSION_CHECK=false to skip the version check at startup.

### DIFF
--- a/browser_use/agent/service.py
+++ b/browser_use/agent/service.py
@@ -1215,11 +1215,12 @@ class Agent(Generic[Context, AgentStructuredOutput]):
 		self.logger.debug(f'🤖 Browser-Use Library Version {self.version} ({self.source})')
 
 		# Check for latest version and log upgrade message if needed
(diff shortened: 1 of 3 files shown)


mertunsall:
That's great catch! It seems there is a lot of merge conflicts so I will fix in another PR!

mertunsall:
https://github.com/browser-use/browser-use/pull/3673

PR #3345, opened 2025-10-15
Perform an additional occlusion check between mouse move -> click

sophiabits:
This cart button is a bit funky; the mouse move triggers the popover and then the popover ends up intercepting the click.

https://github.com/user-attachments/assets/48dcbbda-c1ee-486a-be5d-a8b5d5d27933

Adding an additional occlusion check between the mouse move -> click event solves the problem.

Screenshot 2025-10-15 at 10 19 44

    

---

## Summary by cubic
Add an extra occlusion check after mouse move and before click to prevent newly opened popovers from intercepting clicks. If the target is occluded, we abort the native click and use a JS-based click instead.


mertunsall:
I think running 2 CDP calls every time for this edge case might not be worth it - can we come up with a better solution?

PR #3352, opened 2025-10-17
fix: add test isolation for environment variables

Virtuoso633:
- Add monkeypatch.delenv() to test fixtures to prevent .env file interference
- Update test_cloud_browser.py: mock_auth_config, test_create_browser_auth_error, test_create_browser_with_env_var, test_cloud_browser_auth_error_no_fallback
- Update test_config.py: all TestLazyConfig methods to use monkeypatch
- Ensures tests pass regardless of developer's local .env configuration
- Fixes test failures when BROWSER_USE_API_KEY and ANONYMIZED_TELEMETRY are set in .env
    

---

## Summary by cubic
Isolated tests from local .env by clearing environment variables with monkeypatch to make results deterministic. Fixes failures when BROWSER_USE_API_KEY and ANONYMIZED_TELEMETRY are set locally.

- **Bug Fixes**
  - Use monkeypatch.delenv in fixtures/tests to prevent .env interference.
  - Update tests/ci/test_cloud_browser.py to explicitly clear/set BROWSER_USE_API_KEY for auth error and env-token paths.
  - Update tests/ci/test_config.py to use monkeypatch for ANONYMIZED_TELEMETRY boolean parsing.


mertunsall:
what is the purpose of this? can you explain?

Virtuoso633:
It is fixed here in #3423 last week , closing the pr anyway..

Virtuoso633:
Using `monkeypatch` here solves a few key problems, which were there in earlier test versions:

Test Isolation: "monkeypatch" automatically handles setup and teardown, so we're sure the environment variables are reset after each test, even if it fails. This prevents state from one test from breaking another.

It's much safer than a manual `try...finally` block, which can fail.

By explicitly using `monkeypatch.delenv`, tests like `test_create_browser_auth_error` will now pass or fail reliably, without being affected by a developer's local `.env` file.

PR #3366, opened 2025-10-20
fix: treat element with inert=true as non-clickable

bizzzd:
## Why?

I discovered an issue with a library I'm using, which uses `inert="true"` attribute to mark non-interactive elements.

https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert

Browser-use makes attempts to click these elements, which are actually not visible (dues to `overflow: hidden`), so the navigation fails.

## Fix

Add support of `inert` attribute in ClickableElementDetector
    

---

## Summary by cubic
Treat elements with inert="true" as non-clickable in ClickableElementDetector. Prevents failed navigation caused by clicking hidden, non-interactive elements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,14 +61,14 @@
 
 **1. Create environment with [uv](https://docs.astral.sh/uv/) (Python>=3.11):**
 ```bash
-uv venv --python 3.12
-source .venv/bin/activate
+uv init
 ```
 
 **2. Install Browser-Use package:**
 ```bash
 #  We ship every day - use the latest version!
-uv pip install browser-use
+uv add browser-use
+uv sync
 ```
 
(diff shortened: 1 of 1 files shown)


browser_use/dom/serializer/clickable_elements.py
```diff
@@ -18,6 +18,10 @@ def is_interactive(node: EnhancedDOMTreeNode) -> bool:
 		if node.tag_name in {'html', 'body'}:
 			return False
 
+		# inert true
+		if node.attributes.get('inert', '') == 'true':
+			return False
```
bizzzd, line 23: okay - looks like generally this should be handled by the check at line 14 - but it is commented out for some reason. 
commit: https://github.com/browser-use/browser-use/commit/f6e3fa1d65535034cdb8d2187eb80e2fb405b54c

bizzzd: Anyone knows why?

Issue #3386, opened 2025-10-22
Interaction Issue: Attempt to click invisible links within elements marked with intert="true" or aria-hidden="true"


bizzzd:
### Browser Use Version

0.8.1

### LLM Model

gemini-2.5-pro

### AI instruction (sample)
```
Goal: click on the product

Steps:
1. Go to https://stackblitz.com/edit/vitejs-vite-yjq1omn9?embed=1&file=src%2Findex.css&view=preview
2. Wait 30 seconds
3. Click on the first product in the list
```

### Context
The website is using [nuka-carousel](https://github.com/FormidableLabs/nuka-carousel) library (6.0.3) with `wrapAround` flag.
This flag displays N items before and N items after in the markup to support infinite switch between items.
All these items are marked with [inert](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert) attribute to make them inaccessible:

<img width="1107" height="315" alt="Image" src="https://github.com/user-attachments/assets/85aa9981-3587-43f6-8cbd-28a5dfe4f0c4" />

On generated video I can see it attempts to click wrong item:
<img width="1838" height="840" alt="Image" src="https://github.com/user-attachments/assets/f04aafca-dc25-4031-9530-11ade8c02127" />

### Browser state

```
    "model_output": {
        "evaluation_previous_goal": "Successfully navigated to the provided URL and waited for the page to load. Verdict: Success",
        "memory": "Navigated to the initial URL and waited for the page to load. The next step is to click on the first product.",
        "next_goal": "Click on the first product, which is \"Lunavo\", by clicking on its link with index 67.",
        "action": [
          {
            "click": {
              "index": 67
            }
          }
        ],
        "thinking": "I have successfully navigated to the URL and waited for 20 seconds. Now I need to click on the first product in the list. Based on the screenshot and the interactive elements, the first product appears to be \"Lunavo\" which has an image with index 68 and a link with index 67. I will click on the link with index 67."
      },
      "result": [
        {
          "is_done": false,
          "extracted_content": "Clicked element",
          "include_extracted_content_only_once": false,
          "include_in_memory": false
        }
      ],
      "state": {
        "tabs": [
          {
            "url": "https://stackblitz.com/edit/vitejs-vite-yjq1omn9?embed=1&file=src%2Findex.css&view=preview",
            "title": "nuka-carousel + browser-use - StackBlitz",
            "target_id": "5ACD",
            "parent_target_id": null
          }
        ],
[...]

PR #3389, opened 2025-10-22
Support inert attribute to disable interactivity

bizzzd:
Fix for the issue I created: https://github.com/browser-use/browser-use/issues/3386

https://web.dev/articles/inert

TL;DR;  this adds a check for 'interactivity' css prop
Supported in chrome: https://caniuse.com/mdn-css_properties_interactivity_inert

Known issues: will not work on old version of browser (CDP will likely fail with unknown property)

    

---

## Summary by cubic
Adds support for the inert attribute/interactivity CSS so inert elements are treated as non-interactive and effectively hidden from interaction. This prevents clicks or focus on inert containers and their children.

- **Bug Fixes**
  - Added interactivity to the DOM snapshot.
  - Updated visibility check to treat interactivity: inert as not visible.


bizzzd:
Closing as this approach only work in the new versions of chrome

Issue #3155, opened 2025-09-18
Bug: ScreenshotWatchdog frequent erros


webpolis:
### Browser Use Version

0.7.7

### Bug Description, Steps to Reproduce, Screenshots

After upgrading, I get the following error on every step:

```
ERROR    [cdp_use.client] CDP Error for request 49: {'code': -32000, 'message': 'Command can only be executed on top-level targets'}                                                                                                                                                                                                                                                          
ERROR    [cdp_use.client] CDP Error for request 54: {'code': -32000, 'message': 'Command can only be executed on top-level targets'}                                                                                                                                                                                                                                                          
[2025-09-18T18:11:20.463Z][browser_use.BrowserSession🅑 8001 🅣 FB][ERROR] "[ScreenshotWatchdog] Screenshot failed: {'code': -32000, 'message': 'Command can only be executed on top-level targets'}"                                                                                                                                                                                           
[2025-09-18T18:11:20.480Z][browser_use.BrowserSession🅑 8001 🅣 FB][ERROR] "🚌 [ScreenshotWatchdog.on_ScreenshotEvent(#bbf2)]         ❌ Failed (0.03s): RuntimeError: {'code': -32000, 'message': 'Command can only be executed on top-level targets'}"                                                                                                                                        
ERROR    [bubus] ❌ EventBus_5292bf4a🟢(⏳ 0 | ▶️  1 | ✅ 41 ➡️  30 👂) Error in event handler browser_use.browser.watchdog_base.ScreenshotWatchdog.on_ScreenshotEvent(?▶ ScreenshotEvent#bbf2 ✅) ->                                                                                                                                                                                           
  ntimeError({'code': -32000, 'message': 'Command can only be executed on top-level targets'})                                                                                                                                                                                                                                                                                                
RuntimeError: {'code': -32000, 'message': 'Command can only be executed on top-level targets'}                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                                                                                              
[2025-09-18T18:11:20.483Z][browser_use.BrowserSession🅑 8001 🅣 FB][WARNING] "📸 Clean screenshot failed: RuntimeError: {'code': -32000, 'message': 'Command can only be executed on top-level targets'}"                                                                                                                                                                                       
[...]

PR #3394, opened 2025-10-23
Fix issue #3155

MisterTea:
For some websites, we get this error when attempting to grab a screenshot:

```
ERROR    [cdp_use.client] CDP Error for request 49: {'code': -32000, 'message': 'Command can only be executed on top-level targets'}                                                                                                                                                                                                                                                          
```

This PR fixes this by falling back to taking a screenshot of the top-level target.
    
<!-- This is an auto-generated description by cubic. -->
---

## Summary by cubic
Fixes screenshot failures by retrying against the top-level CDP target when the page returns "Command can only be executed on top-level targets." This makes screenshot capture reliable on sites with nested targets or iframes.

- **Bug Fixes**
  - Retry once using a confirmed top-level page session when the specific CDP error is hit.
  - Added a helper to select a top-level CDP session from agent focus or the session pool.
  - Better error wrapping and logging; highlight cleanup still runs on exit.

<!-- End of auto-generated description by cubic. -->

### DIFF
--- a/browser_use/browser/watchdogs/screenshot_watchdog.py
+++ b/browser_use/browser/watchdogs/screenshot_watchdog.py
@@ -11,52 +11,119 @@ from browser_use.browser.watchdog_base import BaseWatchdog
 from browser_use.observability import observe_debug
 
 if TYPE_CHECKING:
-	pass
+    pass
 
 
 class ScreenshotWatchdog(BaseWatchdog):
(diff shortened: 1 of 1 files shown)


mertunsall:
Fixing this in a separate PR: https://github.com/browser-use/browser-use/pull/3672

PR #3518, opened 2025-11-02
Fix download monitoring issues with fetch requests and blob downloads

### LINKED ISSUE #3515: Download monitoring broken in oh so many ways
### Browser Use Version

0.9.4

### LLM Model

o3

There are several issues with `downloads_watchdog.py`:

1. PDF files requested with `fetch()` are downloaded, even though they are not real downloads and a real browser would not download them.
2. `headers.get('content-disposition', '')` does not work if the header is named `Content-Disposition` (note the case)
3. The 20s watch timer has a race condition when Javascript triggers a download using `blob:`: the new file is already saved by the time `initial_files` is populated, so is never detected and BU waits 20s, despite actually being super fast.

```html
...
```

_No response_

### Python Code Sample

```python

```

### Full DEBUG Log Output

```shell

```

sandeepsalwan1:
Fixes three issues in download monitoring:

1. **Prevent false PDF downloads from fetch/XHR**: Skip programmatic requests (fetch() and XHR) since real browsers don't download PDFs loaded this way
2. **Fix case-sensitive header lookup**: Normalize HTTP headers to lowercase for case-insensitive Content-Disposition detection
3. **Fix race condition with blob downloads**: Capture initial file list before any processing to detect fast blob: downloads

Fixes #3515

---
## Summary by cubic
Improves download monitoring accuracy by preventing false PDF events, capturing fast blob downloads, and avoiding stale cache skips. Fixes #3515.

- **Bug Fixes**
  - Skip Fetch/XHR responses so programmatic PDF loads aren’t treated as downloads
  - Normalize headers to lowercase and ignore stale session cache entries; allow retries if background download fails
  - Capture initial downloads snapshot and detect new files when filePath is missing to handle fast blob downloads

<sup>Written for commit 6db6678a8397ce7934883225777eb3cce6a2aa96. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/browser/watchdogs/downloads_watchdog.py
+++ b/browser_use/browser/watchdogs/downloads_watchdog.py
@@ -869,7 +869,6 @@ class DownloadsWatchdog(BaseWatchdog):
 			# We just need to wait for it to appear in the downloads directory
 			expected_path = downloads_dir / suggested_filename
 
- 
 			# For remote browsers, don't poll local filesystem; downloadProgress handler will emit the event
 			if not self.browser_session.is_local:
 				return


browser_use/browser/watchdogs/downloads_watchdog.py
```diff
@@ -651,6 +658,13 @@ async def _handle_cdp_download(
 		suggested_filename = event.get('suggestedFilename', 'download')
 		guid = event.get('guid', '')
 
+		# Get initial files before download handling to avoid race condition with fast blob: downloads
+		initial_files = set()
+		if Path(downloads_dir).exists():
+			for f in Path(downloads_dir).iterdir():
+				if f.is_file() and not f.name.startswith('.'):
+					initial_files.add(f.name)
+
```
jakajancar: I'm not sure this will make things reliable. `_use_js_fetch_for_local` seems to always be `False` anyways (dead code?) so you're just moving things a few fs operations earlier. Perhaps "seen" files should be tracked from the point when the watchdog is created?

Separately, I don't understand why this 20s filesystem polling is even required and completion not simply reported on `DownloadProgressEvent` 🤔

sandeepsalwan1: Updated @jakajancar :

1. Fixed race condition properly: Now tracking initial files from watchdog startup (in on_BrowserLaunchEvent), not from the download function. This prevents the race condition for fast blob: downloads.

2. Removed dead code: Deleted the entire _use_js_fetch_for_local block (90+ lines) since it's always False and never executes.

3. Improved progress handler: When filePath is missing in DownloadProgressEvent, it now detects new files by comparing against the initial snapshot. This should reduce reliance on the 20s polling fallback.

The polling is still there (might be an option to remove it too) as a last resort, but the progress handler should catch most cases now.

jakajancar (approved):
@sandeepsalwan1 This lgtm, but probably better to wait for someone who has actually worked on this codebase or is affiliated with BU to chime in :-)

mertunsall:
@MagMueller is this related to your recent downloads?

jakajancar:
Until this is live, for anyone needing to resolve missing download detection, you can hack things by calling `agent._update_available_file_paths` in new step callback:

```py
downloads_path = Path("/tmp/foo")

async def new_step_callback(summary, last_model_output, n_steps):
    agent._update_available_file_paths([str(f) for f in downloads_path.iterdir()])


browser = Browser(
    downloads_path=downloads_path,
    # ...
)

agent = Agent(
    browser=browser,
    register_new_step_callback=new_step_callback,
    # ...
)
```

mertunsall:
Does this still automatically download PDFs when the model reaches a PDF viewer? this is a feature we built so that the agent can just read the PDF file instead of scrolling

mertunsall:
It would be really helpful if you could provide exactly when this helps and when the previous code is problematic and so on. With all the AI generated PRs bloating the code nowadays it's hard to merge such things

PR #3602, opened 2025-11-16
fix: Add default=True to ScrollAction.down field to handle LLM respon…

Utkarshkarki:
…ses missing the field

- Fixes validation error when LLM returns {'scroll': {'pages': 3}} without 'down' field
- Defaults to True (scroll down) when 'down' is missing, which is the most common case
- Still supports explicit down=False for scrolling up
- Also fixes CDP client compatibility with older cdp-use versions by making max_ws_frame_size optional

Fixes validation errors with AWS Bedrock Claude and Gemini models

---
## Summary by cubic
Set ScrollAction.down default to True to prevent validation errors when LLMs omit the field and keep scroll-down as the common behavior.

- **Bug Fixes**
  - Handle {'scroll': {'pages': N}} without 'down'; defaults to scrolling down, explicit down=False still works.
  - Updated action description to clarify defaults and parameter ranges.

<sup>Written for commit ece9901b8377d69a8a53b7ee28ebdf373c55ecdb. Summary will update automatically on new commits.</sup>

---

> [!NOTE]
> Set a default of True for `ScrollAction.down` and update the scroll action description to clearly state required/optional parameters and defaults.
> 
> - **Actions / Models**:
>   - `browser_use/tools/views.py`: Set `ScrollAction.down` default to `True`.
> - **Tools / Service**:
>   - `browser_use/tools/service.py`: Clarify scroll action description (required `down`, optional `pages`, documented defaults and behavior).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ece9901b8377d69a8a53b7ee28ebdf373c55ecdb. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/browser_use/agent/service.py
+++ b/browser_use/agent/service.py
@@ -233,6 +233,8 @@ class Agent(Generic[Context, AgentStructuredOutput]):
 				"""Determine timeout based on model name"""
 				model_name = getattr(llm_model, 'model', '').lower()
 				if 'gemini' in model_name:
+					if '3-pro' in model_name:
+						return 90
 					return 45
 				elif 'groq' in model_name:
 					return 30
--- a/browser_use/tools/service.py
+++ b/browser_use/tools/service.py
@@ -274,7 +274,7 @@ class Tools(Generic[Context]):
 				mouse = await page.mouse
 				await mouse.click(actual_x, actual_y)
 
-				memory = f'Clicked on coordinate {actual_x}, {actual_y}'
+				memory = f'Clicked on coordinate {params.coordinate_x}, {params.coordinate_y}'
 				msg = f'🖱️ {memory}'
 				logger.info(msg)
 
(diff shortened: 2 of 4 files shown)


Utkarshkarki:
Hi! This is my first contribution to this repository. All checks have passed successfully. Please let me know if any further changes or improvements are needed. I'll be happy to update the PR. Thank you!

gregpr07:
Wait, why would we want to support old versions of CDP-use? We control both repos

Utkarshkarki:
You're right — since both browser-use and cdp-use are maintained within the same organization, we don't need to keep backward compatibility with older versions. The fallback logic was originally added to prevent failures if someone used an outdated cdp-use, but given our control over both repos, we can safely assume users will stay on the latest version.

I can update the code to remove the fallback TypeError logic and require cdp-use versions that support max_ws_frame_size. I’ll also update the dependency specification to enforce the minimum supported version.

Utkarshkarki:
is this fine?

Utkarshkarki:
are these changes good to go

Issue #3576, opened 2025-11-13
Bug: ...max_failures per steps didnt work


ianrussel:
### Browser Use Version

0.9.4

### Bug Description, Steps to Reproduce, Screenshots

1. There are some steps that the agent keep looping, this happens somethimes, sometimes not - so the agent works fine when running the first test on same testcase, the second time maybe third time the testcase FAILED becaue of infinite loop
2. I understand that there might be needed to fix in the testcase or the testapp htmls , javascript, etc
3. But my main goal is to instruct the Agent to retry that error dynamically - for example retry 3 times and move on to the next step
4. The loop stops only when the max_steps is reached. Meaning if the max_steps is configured to 300, it will consume a significant amount of token costs - normally at $3 usd!
5. MY goal is, whenever the agent encountered error, just retry it 3 times(configurable) and move on to the next steps - eitherway it will just FAILED. And I can just fixed the testcase, html, jsscript whatever causes the issue
6. Initially I thought that the max_failures params in Agent is the one should work - but its not.
7. In the screenshot, it just shows step 19 - step 22 but it start from lower step and I just hit the ctl z to force stop the running agent

Note: I extended it a little bit so llm providers, models are configurable in .env file, and I can even use different browser-use like package in the future

<img width="1191" height="876" alt="Image" src="https://github.com/user-attachments/assets/67f21314-05d2-4277-8e1f-d7d5e55386ed" />

### Failing Python Code

```python
from browser_use import Agent as BUAgent
from browser_use import Browser
from browser_use import Tools

from automation.agents.agent_interface import AgentConfig
from automation.agents.agent_interface import AgentFactory
from automation.agents.agent_interface import BaseAgentAdapter
from automation.agents.agent_interface import HistoryLike
from automation.agents.helper import build_llm as _build_llm

class BrowserUseAgentAdapter(BaseAgentAdapter):
    def __init__(self, config: AgentConfig):
        print(f" configsssss {config.provider}")
        record_video_dir = Path(config.screen_records or "screen_records").resolve()
        uploads_dir = Path(config.uploads_path or "uploads").resolve()
        downloads_dir = Path(config.downloads_path or "downloads").resolve()
        screenshots_dir = Path(config.screenshots_path or "screenshots").resolve()

        # Ensure all directories exist
        uploads_dir.mkdir(parents=True, exist_ok=True)
        downloads_dir.mkdir(parents=True, exist_ok=True)
[...]

PR #3648, opened 2025-11-22
Fix: Count both action and eval failures toward max_failures limit

Utkarshkarki:
Fix: #3576 

# Problem
Currently, `max_failures` only tracks action-level errors (when `ActionResult.error` is set), but doesn't count evaluation failures where the LLM determines a step failed to achieve its goal. This causes agents to loop indefinitely when:
- Actions execute successfully (e.g., scroll, click work fine)
- But evaluations consistently report "Verdict: Failure" (goal not achieved)
- Agent continues retrying until `max_steps` is reached, wasting tokens and causing browser crashes

## Solution
Modified `Agent._post_process()` to track both types of failures:
1. **Action failures**: Any `ActionResult.error` in `self.state.last_result`
2. **Eval failures**: When `evaluation_previous_goal` contains "verdict" + "failure"/"fail"
Both now increment `self.state.consecutive_failures`, which triggers the existing `max_failures` logic to stop the agent.

---
## Summary by cubic
Count evaluation failures toward max_failures so agents stop retrying when goals aren’t achieved, preventing infinite loops and crashes.

- **Bug Fixes**
  - Treat both action errors and eval verdicts containing “failure”/“fail” as failures that increment consecutive_failures.
  - Add concise debug logging with action_error and eval_failed flags to aid tracing.

<sup>Written for commit a24f97d46039c171eeaa8e3a4f42b530fe4af1d4. Summary will update automatically on new commits.</sup>

### DIFF
--- a/.github/ISSUE_TEMPLATE/1_element_detection_bug.yml
+++ b/.github/ISSUE_TEMPLATE/1_element_detection_bug.yml
@@ -31,31 +31,12 @@ body:
     validations:
       required: true
 
-  - type: dropdown
+  - type: input
     id: model
     attributes:
       label: LLM Model
-      description: Which LLM model(s) are you using?
-      multiple: true
-      options:
-        - gpt-4o
-        - gpt-4o-mini
-        - gpt-4
-        - gpt-4.1
-        - gpt-4.1-mini
-        - gpt-4.1-nano
-        - o4-mini
-        - o3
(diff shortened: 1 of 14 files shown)


mertunsall:
I don't think we want to do this by default - also there is no guarantee LLM will output fail/success keywords there. If you want I think you can use the step callback to do this kind of thing!

PR #3725, opened 2025-12-06
Add built-in QA assertion actions for browser-use

### LINKED ISSUE #3709: Feature Request: QA Assertion Tool

Currently, Browser Use is excellent for automating actions, but it lacks a distinct mechanism to validate the results of those actions programmatically. In a Quality Assurance (QA) context, it is not enough to just click buttons; we need to verify that the expected state was reached (e.g., verifying that "Submission Successful" text appeared or a specific modal closed). Without a native "Assert" tool, we cannot easily use workflows for automated regression testing or reliable pass/fail reporting.

I propose implementing a dedicated Assert Tool (or Action). This tool would allow the user to define expectations that the agent must verify before the task is considered successful. For example:

assert_text_present
assert_url
assert_element_visible

If the assertion fails (the text or element is missing), the run/workflow should throw an error or mark the step as failed. This would allow us to build robust test suites that can be rerun automatically.

I am aware that I could implement this as a custom tool specifically for my setup. However, I believe "Assertions" are a fundamental requirement for many users using this library for Testing/QA.

Instead of every user implementing their own version of an assertion logic, it would be much better to have a standardized, community-maintained tool built into the core library. This ensures reliability and prevents everyone from reinventing the wheel.

0.10.1

- [x] It's an urgent deal-breaker, I can't live without it
- [x] It's important to add it in the near-mid term future
- [ ] It would be nice to add it sometime in the next 2 years

MagellaX:
Adds native QA assertions (text present/absent, element visible, URL/title checks with regex support) to make flows verifiable in-browser.
Shared helpers for robust matching/visibility and unit tests to keep behavior stable.
Docs snippet shows how to drop assertions into action lists.
Closes #3709.

---
## Summary by cubic
Adds assertion helpers and example actions to verify browser flows directly. Includes text, element visibility, URL, and title checks with flexible matching.

- **New Features**
  - Actions: assert_text_present, assert_text_absent, assert_element_visible, assert_url, assert_title (provided in examples; not in default Tools).
  - URL/title support equals, prefix, contains, regex; text supports case_sensitive and partial.
  - Visibility check favors is_visible with bounds fallback.
  - Clear pass/fail via ActionResult; failing assertions stop the workflow.
  - New assertion_helpers module with unit tests.
  - README updated with a usage snippet.

<sup>Written for commit 444e609d2ea3fafe7498c218f45c0dd0754bec19. Summary will update on new commits.</sup>


eltomon:
@MagellaX 
Currently, the agent continues running despite assertion failures unless is_done=True and success=False are set. This change ensures the workflow halts immediately on failure, as expected for test suites.

MagellaX:
> @MagellaX Currently, the agent continues running despite assertion failures unless is_done=True and success=False are set. This change ensures the workflow halts immediately on failure, as expected for test suites.

Hey! yeah, just saw that, good catch!!! I’ve updated the assertion actions to return ActionResult(is_done=True, success=False, error=..., long_term_memory=...) on failures, so workflows halt immediately as expected for tests.

MagellaX:
@mertunsall although any takes? I think we're good to go forward with it :)

MagellaX:
> @mertunsall although any takes? I think we're good to go forward with it :)

any takes? @mertunsall

mertunsall:
We don't want to include these in the common actions - could you please put these as custom actions in the examples folder and explain the use case?

MagellaX:
> We don't want to include these in the common actions - could you please put these as custom actions in the examples folder and explain the use case?

Btw, these assertions are meant for deterministic smoke/regression flows (CI or scheduled monitors) where you want the run to stop immediately if an expected condition isn’t met (URL/title/text/element visibility). They’re not intended for normal interactive agent usage, since failures should generally be recoverable there rather than hard-stopping the agent...

MagellaX:
yo @mertunsall let me know!!!

PR #3818, opened 2025-12-26
fix: propagate API status codes to enable proper handling of 429 Rate Limits

Confusion-ymc:
fix: propagate API status codes to enable proper handling of 429 Rate Limits


EdisonYMC:
I'm still working on this

PR #3823, opened 2025-12-28
Fix concurrent EventBus processing with per-instance locks

### LINKED ISSUE #3799: browser-use can not parallel
### Browser Use Version

0.11.2

### LLM Model

gpt-4.1

When I use browser-use in my FastAPI program, I found that it doesn't run in parallel.
For example, I have a controller A, where an agent is created to execute tasks. If I call controller A three times simultaneously, they are not executed concurrently.
Alternatively, you can directly run browser-use/examples/browser/parallel_browser.py. It doesn't truly perform concurrent operations across three browsers.
For instance, if a single task usually takes 60 seconds, running three tasks simultaneously doesn't finish in 60 seconds but rather in 80 seconds. When I increase it to 10 concurrent tasks, the gap becomes even larger.

I believe the reason lies in the concurrency limitations of bubus. The relevant code in bubus, as shown below, prevents event handling across different contexts of Eventbus from running concurrently, which affects concurrency efficiency.

_global_eventbus_lock: ReentrantLock | None = None

def _get_global_lock() -> ReentrantLock:
    """Get or create the global EventBus lock."""
    global _global_eventbus_lock
    if _global_eventbus_lock is None:
        _global_eventbus_lock = ReentrantLock()
    return _global_eventbus_lock

    async def step(
        self, event: 'BaseEvent[Any] | None' = None, timeout: float | None = None, wait_for_timeout: float = 0.1
    ) -> 'BaseEvent[Any] | None':
        """Process a single event from the queue"""
        ...

        # Always acquire the global lock (it's re-entrant across tasks)
        async with _get_global_lock():
            # Process the event
            await self.process_event(event, timeout=timeout)

            # Mark task as done only if we got it from the queue
            if from_queue:
                self.event_queue.task_done()

        logger.debug(f'✅ {self}.step({event}) COMPLETE')
        return event

I am not sure why a global lock was added. I would like to ask if it can be optimized.

```html
None
```

_No response_

### Python Code Sample

```python
browser-use/examples/browser/parallel_browser.py  for 10
```

### Full DEBUG Log Output

```shell

```

sauravk58:
- Remove global bubus lock serialization across separate EventBus instances
- Each bus now has its own re-entrant lock preserving per-bus ordering
- Maintain nested event handling for handler awaits via holds_global_lock
- Add regression test for parallel multi-agent execution
- Resolves issue #3799: agents/sessions no longer block each other

---
## Summary by cubic
Fixes EventBus concurrency by replacing the global lock with per-instance re-entrant locks so separate buses run in parallel while preserving per-bus ordering. Resolves #3799 by removing cross-agent/session blocking.

- **Bug Fixes**
  - Introduced browser_use.event_bus.EventBus with a per-bus re-entrant lock.
  - Removed global cross-bus serialization; different buses now process concurrently.
  - Preserved per-bus ordering and nested handler behavior via holds_global_lock.
  - Updated internal imports to use the new EventBus.
  - Added a regression test verifying parallel execution across buses.

<sup>Written for commit 5e290f4e16acfeea35ada1a072ec7f8cc3430875. Summary will update automatically on new commits.</sup>

### DIFF
--- a/browser_use/event_bus.py
+++ b/browser_use/event_bus.py
@@ -12,7 +12,8 @@ from typing import Any
 
 from bubus import BaseEvent
 from bubus import EventBus as BubusEventBus
-from bubus.service import holds_global_lock, logger as bubus_logger
+from bubus.service import holds_global_lock
+from bubus.service import logger as bubus_logger
 
 
 class _InstanceReentrantLock:
--- a/tests/ci/test_event_bus_concurrency.py
+++ b/tests/ci/test_event_bus_concurrency.py
@@ -1,8 +1,9 @@
 import asyncio
 import time
 
-from browser_use.event_bus import EventBus
 from bubus import BaseEvent
+
+from browser_use.event_bus import EventBus
 
 
 class SleepEvent(BaseEvent[None]):


scorpiord:
@sauravk58 
I am the one who raised issue #3799. Thank you very much for the fix. I would like to know if this fix can be applied directly in bubus, since lower versions (e.g., 0.5.11) are also using it. If fixed in bubus, it could resolve the issue for both higher and lower versions simultaneously.
https://github.com/browser-use/bubus

PR #3826, opened 2025-12-29
feat: realistic typing delays for key dispatch — fixes #3825

### LINKED ISSUE #3825: Feature Request: realistic typing speed

The current typing implementation uses a very short (< 10ms) delay between keystrokes (see default_action_watchdog.py:1635), which might be easily detected as bot behavior. All in all I was getting 18 ms between keystrokes when measured. 

Even at extremely high 160 words per minute, which gives latency of 62.5 ms assuming every word has 6 letters, this is still superhuman.

```
  # Current (line 1635):
  await asyncio.sleep(0.001)

  # Proposed:
  import random
  await asyncio.sleep(random.uniform(0.05, 0.15))  # 50-150ms with variance
```

_No response_

0.11.1

- [ ] It's an urgent deal-breaker, I can't live without it
- [ ] It's important to add it in the near-mid term future
- [x] It would be nice to add it sometime in the next 2 years
- [x] 💪 I'm willing to [start a PR](https://docs.browser-use.com/development/contribution-guide) to work on this myself

Adityakk9031:
Fixes #3825

What is the issue
- Typing implementation used tiny fixed sleeps (1–10ms) between keystrokes which produced superhuman typing speed and can be fingerprinted as bot behaviour.

What this PR changes (minimal)
- File modified: browser_use/browser/watchdogs/default_action_watchdog.py
- Add `import random`.
- Add helper method `_sample_typing_delay()` inside DefaultActionWatchdog.
- Replace tiny fixed sleeps used for typing with `await asyncio.sleep(self._sample_typing_delay())`:
  - Typing to page path (`_type_to_page`)
  - Character-by-character element typing (`_input_text_element_node_impl`)
  - SendKeysEvent handler (`on_SendKeysEvent`)
- Default sampled range: 0.05s — 0.15s (50–150ms). Behavior is identical for non-typing code (no other sleeps changed).

---
## Summary by cubic
Adds human-like per-key typing delays sampled from BrowserProfile to replace fixed sleeps and reduce fingerprinting. Defaults stay fast (1–10ms); set BROWSER_USE_REALISTIC_TYPING for 50–150ms realistic typing, fixing #3825.

- **Bug Fixes**
  - Replaced fixed per-key sleeps in _type_to_page, element typing, and on_SendKeysEvent with _sample_typing_delay(); uniform in fast mode and clamped lognormal in realistic mode.
  - Added typing_delay_min/max to BrowserProfile and _configure_typing_speed (via model_post_init); env var enables realistic typing without overriding user values; expanded CI tests to cover truthy env variants and sampling bounds/mean.

<sup>Written for commit 173174416ba8b53b023a5b5039fb850aae58d3c2. Summary will update on new commits.</sup>


mikolaj-jpg:
Thanks for implementing this so quickly! One more request, to avoid surprising current users: the new latency (100 - 150 ms) is 10x bigger than old latency (10 - 20 ms). It might massively slow down typing. If you are typing long texts it means you go from 1 minute to 10 minutes which might be surprising. I think there should be an option to hide this behind an env variable or something similar.  EDIT: I mean that this feature should be enabled by an env variable. Be default we should be typing quickly.

mikolaj-jpg:
Also, if we really want to make it look legit, the key press latency (how long a key is held down) should be drawn from normal (as it is right now) but the latency between different keys should come from lognormal with μ=4.5 and σ=1 (gives mean of 150). This ensures that we take longer breaks sometimes, just like humans do! Want me to do it?

Adityakk9031:
no i will work

Adityakk9031:
little busy give me sometime will fix it by tomorrow

sauravpanda:
Also can you make it optional parameter, some of the usecases dont care for this and need speed and we would love to support that.

Adityakk9031:
ok

Adityakk9031:
ok

Adityakk9031:
@sauravpanda sir please look sorry for delay

Adityakk9031:
@ShawnPana have a look

Issue #3850, opened 2026-01-07
Bug: ...Model "bu-30b-a3b-preview" not recognized despite being documented


tlxx22:
### Browser Use Version

0.11.2

## Description

The model `bu-30b-a3b-preview` mentioned in the official documentation causes an "Invalid model" error when used.

## Environment

- **browser-use version**: (latest, just updated via `pip install --upgrade browser-use`)
- **Python version**: 3.11
- **OS**: Windows 11

## Code to reproduce

```python
from browser_use import Agent
from browser_use.llm import ChatBrowserUse

llm = ChatBrowserUse(model="bu-30b-a3b-preview")

agent = Agent(
    task="Your task here",
    llm=llm
)

result = await agent.run()
```

## Error message

```
ValueError: Invalid model: 'bu-30b-a3b-preview'. Must be one of ['bu-latest', 'bu-1-0'] or start with 'browser-use/'
```

## Expected behavior

According to the documentation, `bu-30b-a3b-preview` should be a valid model option:
<https://docs.browser-use.com/>... (link to the documentation page showing this model)

## Additional context

The documentation example shows:

```python
llm = ChatBrowserUse(model="bu-30b-a3b-preview")
```

But only `bu-latest` and `bu-1-0` are accepted in practice.

Is `bu-30b-a3b-preview` still available? If not, what is the equivalent model name?

### Failing Python Code

```python
from browser_use import Agent
from browser_use.llm import ChatBrowserUse

llm = ChatBrowserUse(model="bu-30b-a3b-preview")

agent = Agent(
    task="Your task here",
    llm=llm
)

result = await agent.run()
```

### LLM Model

bu-30b-a3b-preview

### Operating System & Browser Versions

Windows11

### Full DEBUG Log Output

```shell
ValueError: Invalid model: 'bu-30b-a3b-preview'. Must be one of ['bu-latest', 'bu-1-0'] or start with 'browser-use/
```

sudhanshu112233shukla:
hi , 
see i hope my solution fix this issue .

tlxx22:
> hi , see i hope my solution fix this issue .

it works! Thx a lot!! :)

sudhanshu112233shukla:
@tlxx22 thank you soo much

PR #3851, opened 2026-01-07
fix: add bu-30b-a3b-preview to valid models list

sudhanshu112233shukla:
fixes:#3850
I ran into an issue with the ChatBrowserUse class in browser_use/llm/browser_use/chat.py where the model name validation was too strict. The code only allowed bu-latest, bu-1-0, or model names starting with browser-use/. Because of this, a valid model called bu-30b-a3b-preview was being rejected during initialization, throwing a ValueError: Invalid model, even though the model itself is supported and usable.

To fix this, I updated the valid_models list in the __init__ method to explicitly include bu-30b-a3b-preview. I also created a small reproduction script (reproduce_issue.py) to verify the behavior. Before the fix, the script failed immediately with the validation error. After the fix, the model passed validation and the initialization continued normally (moving on to API key validation or succeeding if the key was present). This confirms that the issue was purely due to overly restrictive validation and that the update correctly resolves it.

---
## Summary by cubic
Adds bu-30b-a3b-preview to the valid model list in ChatBrowserUse so the model passes validation.
Prevents the “Invalid model” ValueError when initializing with this supported model.

<sup>Written for commit 40181e22d760a5ecb0b04a868653d56dbb57ff6d. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/llm/browser_use/chat.py
+++ b/browser_use/llm/browser_use/chat.py
@@ -68,7 +68,7 @@ class ChatBrowserUse(BaseChatModel):
 			retry_max_delay: Maximum delay in seconds between retries (default: 60.0).
 		"""
 		# Validate model name - allow bu-* and browser-use/* patterns
-		valid_models = ['bu-latest', 'bu-1-0']
+		valid_models = ['bu-latest', 'bu-1-0', 'bu-30b-a3b-preview']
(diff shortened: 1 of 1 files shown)


sudhanshu112233shukla:
@laithrw @sauravpanda this can also be merged its approved !!

PR #3857, opened 2026-01-08
fix(browser): add explicit fallback for CDP session attachment 

### LINKED ISSUE #3735: Interaction Issue: New tab via Target.createTarget sometimes has no session; navigate(new_tab=True) fails with "Target ... not found" and focus assertion
### Browser Use Version

0.10.1

### LLM Model

gemini-2.5-pro

## Summary

When using `navigate` with `new_tab=True` (external Chrome, CDP auto-attach, Gmail already open), the newly created `about:blank` tab sometimes never gets a CDP session in the `SessionManager`. This causes:

- `get_or_create_cdp_session(target_id=...)` to fail with `ValueError: Target ... not found - may have detached or never existed`.
- `BrowserSession.on_SwitchTabEvent` / `on_AgentFocusChangedEvent` to fail.
- `BrowserSession.on_NavigateToUrlEvent` to fail with:

  ```text
  AssertionError: Agent focus not updated to new target_id after SwitchTabEvent should have switched to it
  ```

After adding an explicit `Target.attachToTarget` + manual `SessionManager` registration fallback in `get_or_create_cdp_session`, the problem disappears and `navigate(new_tab=True)` becomes stable.

---

## Environment

- `browser-use` version: 0.10.1
- Browser: Chrome (stable), launched via `LocalBrowserWatchdog` with a persistent `user_data_dir` and several extensions (uBlock Origin, I still don't care about cookies, ClearURLs, Force Background Tab).
- OS: Windows 10.
- CDP: connecting to a locally running Chrome via `Target.setAutoAttach(autoAttach=True, flatten=True)`.

My app embeds `browser_use.browser.session` and `browser_use.browser.session_manager` almost unmodified.

---

## Repro steps

1. Start a `BrowserSession` and navigate to Gmail so that the main tab is `https://mail.google.com/mail/u/0/#inbox`.
2. Start an Agent with a multi-step plan that **explicitly asks to open a new tab** and keep the mail tab open
3. The Agent executes something like:

   ```python
   await navigate(url="about:blank", new_tab=True)
   # or
   await navigate(url="https://www.google.de", new_tab=True)
   ```

4. On this first `navigate(new_tab=True)` call, the logs show:

   - New tab creation:

     ```text
     [on_NavigateToUrlEvent] No reusable about:blank tab found, creating new tab...
     Created new tab #D54E / #BEB9 / ...
     ```

   - Then (for the problematic runs):

     ```text
     [SessionManager] Waiting for target ... to attach...
     WARNING Target ... has no session after auto-attach wait
     ...
     ValueError: Target ... not found - may have detached or never existed
     AssertionError: Agent focus not updated to new target_id after SwitchTabEvent should have switched to it
     ```

5. The Agent reports navigation failure, but the original Gmail tab continues to work. Repeated attempts to open a new tab show the same pattern.

In other words, the new tab is created (we see the `Target.createTarget` effect and the tab ID in logs), but **no CDP session ever shows up in `SessionManager` for that target**, at least within the 2s window used by `get_or_create_cdp_session`.

With my patch (see below), for those same tabs I instead see:

```text
[SessionManager] Target attached: A4B4EBCA... (session=2478BB3D..., type=page, waitingForDebugger=False)
[SessionManager] Created target A4B4EBCA... (type=page)
[SessionManager] Created session 2478BB3D... for target A4B4EBCA... (total sessions: ...)
...
[SessionManager] Switching focus: 9A3A1BD6... → A4B4EBCA...
AgentFocusChangedEvent received: target_id=...27F7 url=about:blank
[on_NavigateToUrlEvent] Succeeded
```

---

## Relevant code (current behavior)

In `browser_use/browser/session.py`:

```python
async def get_or_create_cdp_session(self, target_id: TargetID | None = None, focus: bool = True) -> CDPSession:
    ...
    session = self.session_manager._get_session_for_target(target_id)

    if not session:
        # Session not in pool yet - wait for attach event
        self.logger.debug(f'[SessionManager] Waiting for target {target_id[:8]}... to attach...')

        # Wait up to 2 seconds for the attach event
        for attempt in range(20):
            await asyncio.sleep(0.1)
            session = self.session_manager._get_session_for_target(target_id)
            if session:
                self.logger.debug(f'[SessionManager] Target appeared after {attempt * 100}ms')
                break

        if not session:
            # Timeout - target doesn't exist
            raise ValueError(f'Target {target_id} not found - may have detached or never existed')
```

`_cdp_create_new_page` uses `Target.createTarget(url='about:blank')` on the root client, and `SessionManager.start_monitoring()` has already registered:

- `Target.attachedToTarget` → `_handle_target_attached`
- `Target.detachedFromTarget` → `_handle_target_detached`
- `Target.targetInfoChanged` → `_handle_target_info_changed`

So in theory, `attachedToTarget` should populate the session pool – but in practice, for some newly created tabs, the attach event does not show up in time (or maybe not at all), causing `get_or_create_cdp_session` to give up after 2 seconds.

---

## Workaround / fix that solved it for me

In my embedded copy, I patched `get_or_create_cdp_session` to:

1. Keep the original 2s auto-attach wait.
2. If there is still no session, call `Target.attachToTarget(targetId=..., flatten=True)` on the root CDP client.
3. Call `Target.getTargetInfo(targetId=...)` to get `targetInfo`.
4. Manually invoke `SessionManager._handle_target_attached(...)` with a synthetic event that mirrors CDP `AttachedToTargetEvent`.

Simplified pseudo-code (not exact diff, but the idea):

```python
async def get_or_create_cdp_session(...):
    ...
    session = self.session_manager._get_session_for_target(target_id)

    if not session:
        # 1) Wait for auto-attach as before
        for attempt in range(20):
            await asyncio.sleep(0.1)
            session = self.session_manager._get_session_for_target(target_id)
            if session:
                self.logger.debug(
                    f'[SessionManager] Target appeared after {attempt * 100}ms (auto-attach)'
                )
                break

        # 2) If still no session, explicit attach fallback
        if not session:
            self.logger.warning(
                f'[SessionManager] Target {target_id[:8]}... has no session after auto-attach wait; '
                f'trying explicit Target.attachToTarget fallback'
            )

            try:
                assert self._cdp_client_root is not None
                attach_result = await self._cdp_client_root.send.Target.attachToTarget(
                    params={'targetId': target_id, 'flatten': True}
                )
                session_id = attach_result.get('sessionId')
                if not session_id:
                    raise RuntimeError('attachToTarget did not return a sessionId')

                # Fetch targetInfo to mimic AttachedToTargetEvent
                try:
                    target_info_result = await self._cdp_client_root.send.Target.getTargetInfo(
                        params={'targetId': target_id}
                    )
                    target_info = target_info_result.get('targetInfo', {})
                except Exception as info_err:
                    self.logger.debug(
                        f'[SessionManager] getTargetInfo failed for {target_id[:8]}... after explicit attach: {info_err}'
                    )
                    target_info = {
                        'targetId': target_id,
                        'type': 'page',
                        'url': 'about:blank',
                        'title': '',
                    }

                manual_event = {
                    'sessionId': session_id,
                    'targetInfo': target_info,
                    'waitingForDebugger': False,
                }
                # Reuse SessionManager's single source of truth
                await self.session_manager._handle_target_attached(manual_event)  # type: ignore[arg-type]
            except Exception as attach_err:
                self.logger.error(
                    f'[SessionManager] Explicit Target.attachToTarget failed for {target_id[:8]}...: {attach_err}'
                )
            else:
                session = self.session_manager._get_session_for_target(target_id)

        if not session:
            raise ValueError(f'Target {target_id} not found - may have detached or never existed')
```

With this change, for the affected new tabs I now see normal `Target attached` and `Created session` logs in `SessionManager`, `on_SwitchTabEvent` / `on_AgentFocusChangedEvent` succeed, and `on_NavigateToUrlEvent` completes successfully.

---

## Suggestion

It seems that in some environments (external Chrome + extensions + persistent profile), relying purely on CDP `attachedToTarget` + `flatten=True` is not always sufficient for newly created tabs.

I think it would be helpful to:

- Add an official, supported fallback inside `BrowserSession.get_or_create_cdp_session` or inside `SessionManager`, similar to the patch above (ideally via a public helper rather than calling `_handle_target_attached` directly), **or**
- Expose a small public API on `SessionManager` to "ensure a session exists for targetId" using `Target.attachToTarget`, so consumers don’t need to reach into internals.

Happy to adjust this patch or provide a PR in whatever shape fits best into the project.

```html
N/A
```

_No response_

### Python Code Sample

```python

```

### Full DEBUG Log Output

```shell

```

sudhanshu112233shukla:
### Fixes: #3735
The problem
In some setups especially when using an external Chrome instance with extensions calling
navigate(new_tab=True) would open a new tab, but Chrome DevTools Protocol (CDP) wouldn’t attach to that tab in time.
CDP normally auto-attaches within a 2-second window, but in these environments the attach event sometimes never arrives. When that happens, the code throws an error like:
`ValueError: Target ... not found – may have detached or never existed`

So the tab exists, but the session is never registered, and everything crashes.

**What I fixed**
I updated browser_use/browser/session.py to make tab attachment more robust instead of relying only on CDP’s auto-attach behavior.

**How the fix works** 
Normal behavior stays the same
The code still waits up to 2 seconds for Chrome’s automatic AttachedToTarget event.
Fallback if auto-attach fails
If no session appears within that time, the code manually attaches to the tab by calling:
`Target.attachToTarget
`

Manual session registration
After forcing the attach, the code manually informs the SessionManager that:
“This target is now attached—register this session ID.”
This effectively simulates the missing auto-attach event.

**The result**
Even if Chrome fails to auto-attach (which happens with extensions or slower environments), the session is still created and registered correctly.
The navigation succeeds, and the crash is fully avoided.
**In short**: no more failures just because Chrome didn’t auto-attach in time.

---
## Summary by cubic
Adds an explicit, idempotent fallback to attach a CDP session when Chrome doesn’t auto-attach to a new tab. Prevents “Target not found” crashes and makes new-tab navigation reliable with external Chrome or extensions (addresses #3735; hardens attachment races #3857).

- **Bug Fixes**
  - Waits up to 2s for auto-attach, then does a final session check.
  - Falls back to Target.attachToTarget (flatten: true) under a per-target lock to avoid races.
  - Manually registers the session with safe targetInfo defaults via SessionManager.
  - Idempotent handling: ignores duplicate attaches and only updates target info if a session already exists; logs hard failures.

<sup>Written for commit 3f8a96e060f7f2dd84fe3744907cad782e2b16df. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/browser/session.py
+++ b/browser_use/browser/session.py
@@ -1258,54 +1258,63 @@ class BrowserSession(BaseModel):
 					self.logger.debug(f'[SessionManager] Target appeared after {attempt * 100}ms')
 					break
 
-			if not session:
 				# 2) If still no session, explicit attach fallback
 				self.logger.warning(
 					f'[SessionManager] Target {target_id[:8]}... has no session after auto-attach wait; '
 					f'trying explicit Target.attachToTarget fallback'
 				)
 
-				try:
-					assert self._cdp_client_root is not None
-					attach_result = await self._cdp_client_root.send.Target.attachToTarget(
-						params={'targetId': target_id, 'flatten': True}
-					)
-					session_id = attach_result.get('sessionId')
-					if not session_id:
-						raise RuntimeError('attachToTarget did not return a sessionId')
+				# Use per-target lock to prevent concurrent attachment requests for the SAME target
+				# This ensures only one thread actually calls Target.attachToTarget
+				async with self.session_manager._get_attachment_lock(target_id):
+					# Check again inside the lock - another thread might have finished the attachment
+					session = self.session_manager._get_session_for_target(target_id)
+					if session:
+						self.logger.debug(f'[SessionManager] Target {target_id[:8]}... attached by another thread while waiting for lock')
+						return session
 
-					# Fetch targetInfo to mimic AttachedToTargetEvent
 					try:
-						target_info_result = await self._cdp_client_root.send.Target.getTargetInfo(params={'targetId': target_id})
-						target_info = target_info_result.get('targetInfo', {})
-					except Exception as info_err:
-						self.logger.debug(
-							f'[SessionManager] getTargetInfo failed for {target_id[:8]}... after explicit attach: {info_err}'
+						assert self._cdp_client_root is not None
+						attach_result = await self._cdp_client_root.send.Target.attachToTarget(
+							params={'targetId': target_id, 'flatten': True}
 						)
-						target_info = {
-							'targetId': target_id,
-							'type': 'page',
-							'url': 'about:blank',
-							'title': '',
+						session_id = attach_result.get('sessionId')
+						if not session_id:
+							raise RuntimeError('attachToTarget did not return a sessionId')
+
+						# Fetch targetInfo to mimic AttachedToTargetEvent
+						try:
+							target_info_result = await self._cdp_client_root.send.Target.getTargetInfo(params={'targetId': target_id})
+							target_info = target_info_result.get('targetInfo', {})
+						except Exception as info_err:
+							self.logger.debug(
+								f'[SessionManager] getTargetInfo failed for {target_id[:8]}... after explicit attach: {info_err}'
+							)
+							target_info = {
+								'targetId': target_id,
+								'type': 'page',
+								'url': 'about:blank',
+								'title': '',
+							}
+
+						manual_event = {
+							'sessionId': session_id,
+							'targetInfo': target_info,
+							'waitingForDebugger': False,
 						}
+						# Reuse SessionManager's single source of truth
+						await self.session_manager._handle_target_attached(manual_event)
+					except Exception as attach_err:
+						# Log at error level since this is a hard failure after multiple attempts
+						self.logger.error(
+							f'[SessionManager] Explicit Target.attachToTarget failed for {target_id[:8]}...: {attach_err}'
+						)
 
-					manual_event = {
-						'sessionId': session_id,
-						'targetInfo': target_info,
-						'waitingForDebugger': False,
-					}
-					# Reuse SessionManager's single source of truth
-					await self.session_manager._handle_target_attached(manual_event)
-				except Exception as attach_err:
-					self.logger.error(
-						f'[SessionManager] Explicit Target.attachToTarget failed for {target_id[:8]}...: {attach_err}'
-					)
-
-				# Try getting session again
+				# Try getting session again (outside lock to verify final state)
 				session = self.session_manager._get_session_for_target(target_id)
 
 			if not session:
-				# Timeout - target doesn't exist
+				# Timeout - target doesn't exist or attachment failed completely
 				raise ValueError(f'Target {target_id} not found - may have detached or never existed')
 
 		# Validate session is still active
--- a/browser_use/browser/session.py
+++ b/browser_use/browser/session.py
@@ -1259,6 +1259,52 @@ class BrowserSession(BaseModel):
 					break
 
 			if not session:
+				# 2) If still no session, explicit attach fallback
+				self.logger.warning(
+					f'[SessionManager] Target {target_id[:8]}... has no session after auto-attach wait; '
+					f'trying explicit Target.attachToTarget fallback'
+				)
+
+				try:
+					assert self._cdp_client_root is not None
+					attach_result = await self._cdp_client_root.send.Target.attachToTarget(
+						params={'targetId': target_id, 'flatten': True}
+					)
+					session_id = attach_result.get('sessionId')
+					if not session_id:
+						raise RuntimeError('attachToTarget did not return a sessionId')
+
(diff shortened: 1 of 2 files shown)


sudhanshu112233shukla:
@laithrw this one is correct ?

TsvetoslavVasev:
Hey @sudhanshu112233shukla  do we have an E.T.A on this one?

laithrw:
@sudhanshu112233shukla sorry for the late response; are you able to reproduce the issue? I'm having trouble with this. If so, how did you reproduce it, and did you confirm this PR resolves the issue?

sudhanshu112233shukla:
@laithrw i will fix this PR, in some time

sudhanshu112233shukla:
@laithrw  The hangs were caused by a race condition during the CDP attachment process. I've added per-target locking to prevent redundant calls and verified it with a concurrency test script. It's much more stable now and ready for review

Issue #3657, opened 2025-11-22
Bug: Agent fails to detect and switch to new tab opened by `click` event


hharryz:
### Browser Use Version

0.9.7

### LLM Model

gemini-2.0-flash, gemini-2.5-flash

### Screenshots, Description, and task prompt given to Agent

Hi! I’ve been using browser-use for a while and encountered an issue related to tab handling. In some cases, a click action opens a new tab, and ideally the agent should detect this behavior and continue its work in the newly opened tab. However, with the recently introduced Force Background Tab mechanism, the framework forces new tabs to open in the background, keeping the focus on the current tab.

The problem is that the agent often does not realize that a new tab has been created. It continues operating on the original tab as if the click event did nothing, leading to ineffective actions (such as repeated clicking or scrolling) and eventually causing the task to fail.

I understand that Force Background Tab is designed to prevent unwanted behaviors, such as ad pop-ups unexpectedly stealing focus. However, it also causes many tasks that previously worked correctly to fail. I believe the relevant PR is this one: [fix-screenshot](https://github.com/browser-use/browser-use/pull/3317). I’m wondering about the motivation behind this change and would like to ask: what is the recommended solution or best practice for handling cases where a task legitimately requires switching to the newly opened tab?

I've tried both gemini-2.0-flash and gemini-2.5-flash which all failed on many tasks because of this.

### HTML around where it's failing

```html
[...]

PR #3864, opened 2026-01-09
fix(#3657): Fix new tab detection and focus switching

sudhanshu112233shukla:
fixes: #3657
**The Problem**
When the agent clicked a link that used target="_blank" or when a page triggered window.open(), the browser correctly opened a new tab. However, the agent always stayed focused on the original tab. There was no logic that allowed the agent to decide whether it should follow the newly opened tab or remain where it was. Because of this, the agent failed to interact with the new page, lost context, and often repeated actions on the original tab, eventually causing the task to fail.

**Root Cause Analysis**
The first issue was missing context. Internally, the system used a TabCreatedEvent, but this event did not include the opener_id, which is the ID of the tab that created the new tab. Without this information, the agent could not tell whether the new tab was opened as a direct result of its own click or whether it was just an unrelated background tab such as an ad popup.

The second issue was rigid behavior. The existing logic always forced newly opened tabs to remain in the background. There was no configuration option that allowed interactive sessions to follow new tabs when that behavior was actually desired. This made the system inflexible and broke many workflows that rely on opening links in new tabs.

The third issue was a gap in how Chrome DevTools Protocol events were handled. The SessionManager was already receiving the openerId from the CDP Target.attachedToTarget event, but this information was being discarded instead of being passed further into the browser-use event system. As a result, critical information was lost even though Chrome was providing it.

**Implementation Details**
To fix this properly, changes were implemented across four key files.
In browser_use/browser/profile.py, a new configuration flag called force_background_tabs was added. This flag defaults to True in order to preserve existing behavior. When this flag is set to False, it explicitly signals that the agent is allowed to switch focus to newly opened tabs when appropriate.

In browser_use/browser/events.py, the TabCreatedEvent schema was updated to include an optional opener_id field of type TargetID | None. This allows downstream logic to understand which tab triggered the creation of the new one.

In browser_use/browser/session_manager.py, the _handle_target_attached method was modified so that it extracts the openerId directly from the raw CDP event data. This value is no longer discarded. Instead, it is now passed into the newly created TabCreatedEvent, ensuring that the information flows correctly through the system.

The most important changes were made in browser_use/browser/session.py. The on_TabCreatedEvent handler was updated to include the core decision logic. The agent first checks whether force_background_tabs is disabled. If it is disabled, the agent then decides whether to switch focus to the new tab. The focus is switched if the opener_id matches the currently active tab, which means the new tab was opened as a direct result of the agent’s action. Additionally, the focus is also switched when opener_id is None. This is intentional, because in many automation scenarios such as when using Target.createTarget or certain JavaScript popups, the browser does not report an openerId at all. Allowing None ensures that valid new tabs are not ignored when the user has explicitly opted into following new tabs.

**Verification**
To verify the fix, a reproduction script named repro_3657_simple.py was created to simulate the problem.
Before the fix, the script opened a new tab, but the agent completely ignored it and continued operating on the original blank page. The agent never switched context and therefore could not proceed.

After the fix, when force_background_tabs was set to False, the agent correctly detected the new tab even in cases where opener_id was None. The agent immediately emitted a SwitchTabEvent and successfully switched focus to the newly opened content. This confirmed that the issue was resolved without breaking existing default behavior.

---
## Summary by cubic
Fixes new tab detection and focus switching so the agent follows tabs it opens, addressing Linear #3657. Prevents lost context when links open via target="_blank" or window.open().

- **Bug Fixes**
  - Add opener_id to TabCreatedEvent and pass it from CDP Target.attachedToTarget.
  - Auto-switch focus on TabCreatedEvent when force_background_tabs is False and opener_id matches the focused tab or is None.
  - Include opener_id when creating tabs from NavigateToUrlEvent.
  - Minor: add a 2s timeout in PopupsWatchdog setup and allow file:// URLs in DOM watchdog.

- **Migration**
  - To auto-follow new tabs, set BrowserProfile.force_background_tabs=False.
  - Default remains True to preserve current behavior; the “Force Background Tab” extension is skipped when disabled.

<sup>Written for commit 4f39bebcc96dd08463f470b0e62f0b8f58637d20. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/browser/session.py
+++ b/browser_use/browser/session.py
@@ -728,7 +728,9 @@ class BrowserSession(BaseModel):
 						target_id = await self._cdp_create_new_page('about:blank')
 						self.logger.debug(f'Created new tab #{target_id[-4:]}')
 						# Dispatch TabCreatedEvent for new tab
-						await self.event_bus.dispatch(TabCreatedEvent(target_id=target_id, url='about:blank', opener_id=current_target_id))
+						await self.event_bus.dispatch(
+							TabCreatedEvent(target_id=target_id, url='about:blank', opener_id=current_target_id)
+						)
 					except Exception as e:
 						self.logger.error(f'[on_NavigateToUrlEvent] Failed to create new tab: {type(e).__name__}: {e}')
 						# Fall back to using current tab
@@ -975,7 +977,9 @@ class BrowserSession(BaseModel):
 		if not self.browser_profile.force_background_tabs:
 			opener_id = getattr(event, 'opener_id', None)
 			if opener_id == self.agent_focus_target_id or opener_id is None:
-				self.logger.info(f'🔄 New tab detected (opener: {opener_id if opener_id else "None"}), switching focus to {event.target_id[-8:]}')
+				self.logger.info(
+					f'🔄 New tab detected (opener: {opener_id if opener_id else "None"}), switching focus to {event.target_id[-8:]}'
+				)
 				from browser_use.browser.events import SwitchTabEvent
 
 				await self.event_bus.dispatch(SwitchTabEvent(target_id=event.target_id))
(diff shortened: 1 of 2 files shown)


laithrw:
@sudhanshu112233shukla, thanks for the proposed fix. I don't think auto-switching tabs is the right call -- it's best if the agent can decide whether to switch based on its task and previous actions, not for this to be forced. Closed in favor of a simpler approach in #4034.

Issue #3382, opened 2025-10-22
Interaction Issue: unable to restore context back to root window after changing field inside iframe


bizzzd:
### Browser Use Version

0.8.1

### LLM Model

gemini-2.5-pro

### Context:
Page that renders stripe payment cross-origin iframe inside document. Below the iframe (in the main document) there is a button "Place order".

### Instructions given to AI:
- Fill out credit card details (number: 4111 1111 1111 1111, cvc: 111, expiry: 04/50, zip: 10013)
- Click "Place order"

### What happens
The AI agent is able to fill out fields but it is unable to reach out to "Place order" button - it seems like there is an issue with switching context back.

### Workaround that works: 
Add `disable_security: True`

### HTML around where it's failing

```html
<head>...</head>
<body>
<iframe name="__privateStripeFrame" allow="payment *; publickey-credentials-get *" src="https://js.stripe.com/v3/elements-inner-payment[REDACTED]" title="Secure payment input frame">
    <head>...</head>
    <body>
        <input dir="ltr" type="text" inputmode="numeric" name="number" id="Field-numberInput" placeholder="1234 1234 1234 1234" autocomplete="cc-number" aria-invalid="true" >
    	<!-- other inputs go here -->
    </body>
</iframe>
<button type="button">Place order</button>
</body>
```

### Python Code Sample

```python

```

### Full DEBUG Log Output

```shell

```

MagMueller:
Can you try again with the latest version we made many fixes for focus changes for Iframes.

sudhanshu112233shukla:
@MagMueller  hii 
can i work on this issue ? i will try to solve its fast / as soon as possible ...

laithrw:
Closing this -- the iframe context switching has been significantly improved since v0.8.1. If you're still hitting this on the latest version, please reopen with debug logs.

Issue #3109, opened 2025-09-14
Bug: Unable to interact within an iframe


xxkratch:
### Browser Use Version

0.7.7

### Bug Description, Steps to Reproduce, Screenshots

I am trying to tick the 'I am not a robot' captcha, however, the agent fails to tick it, and returns the final message as follows: "**I am unable to click the 'I am not a robot' checkbox because it is inside an iframe and I cannot interact with elements inside an iframe.**"

Is there any particular way to deal with iframes? 

### Failing Python Code

```python
from browser_use import Agent, Browser, ChatGoogle
import asyncio
from dotenv import load_dotenv
load_dotenv()

async def main():
	llm = ChatGoogle(model='gemini-2.5-pro')
	browser = Browser(keep_alive=True)
	await browser.start()
	agent = Agent(
		task='navigate to https://glitch.com/signin, then click on "Password" button to Sign in to Glitch, and finally Tick the "I am not a robot" checkbox that appears below Forgot password editbox.',
		llm=llm, 
		browser_session=browser, 
		use_thinking=False,
		override_system_message=SYSTEM_PROMPT,
	)
	await agent.run()
	await browser.kill()

if __name__ == '__main__':
	asyncio.run(main())
```

### LLM Model

gemini-2.5-pro

### Operating System & Browser Versions

Windows 11

### Full DEBUG Log Output

```shell
INFO     [Agent] 🔗 Found URL in task: https://glitch.com/signin, adding as initial action...
INFO     [Agent] 🚀 Task: navigate to https://glitch.com/signin, then click on "Password" button to Sign in to Glitch, and finally Tick the "I am not a robot" checkbox that appears below Forgot password editbox.       
INFO     [Agent]   🦾 [ACTION 1/1] go_to_url: url: https://glitch.com/signin, new_tab: False
INFO     [tools] 🔗 Navigated to https://glitch.com/signin
INFO     [Agent] 🧠 Starting a browser-use version 0.7.7 with model=gemini-2.5-pro
INFO     [Agent] 

INFO     [Agent] 📍 Step 1:
INFO     [Agent]   👍 Eval: The agent was initialized and the initial URL was loaded. Verdict: Success
INFO     [Agent]   🎯 Next goal: The page is currently blank, so I need to wait for it to load completely before I can interact with any elements.
INFO     [Agent]   🦾 [ACTION 1/1] wait: seconds: 5
INFO     [tools] 🕒 waited for 2 seconds + 3 seconds for LLM call
[...]

PR #3874, opened 2026-01-11
Fix: iframe visibility checks and prevent AX tree retrieval hangs

sudhanshu112233shukla:
Issue #3109 (Inside Iframe Interaction): The root cause was that elements inside iframes were "technically" visible in the DOM but hidden from the user because parent iframes weren't scrolled into view. My fix _scroll_parent_iframes_into_view ensures they are visible.
Issue #3382 (Context Restoration): This issue is actually resolved by the design of cdp_client_for_node, which I verified in browser/session.py. It dynamically resolves the session for every action based on the target node's session_id. The "inability to restore" previously experienced was due to get_state() hanging on unresponsive frames, preventing the agent from "seeing" the main frame again. My timeout fix in DomService prevents this hang, ensuring the agent always gets a fresh state with valid main frame elements, allowing it to "switch back" seamlessly.

---
## Summary by cubic
Fix iframe interactions by scrolling parent iframes into view before actions, and prevent hangs when fetching accessibility trees with targeted timeouts. Addresses #3109 (inside iframe interaction) and #3382 (context restoration) by ensuring elements are truly visible and state retrieval stays responsive.

- **Bug Fixes**
  - Recursively scroll parent iframes into view before clicking elements, using the correct CDP session per parent frame.
  - Add short timeouts to Page.getFrameTree and per-frame Accessibility.getFullAXTree, and safely merge results to avoid hangs on unresponsive/cross-origin frames.
  - Add CI tests for AX tree timeout handling and parent iframe scrolling.

<sup>Written for commit dfbb691995eafc6b02cb938804713e4aae66c623. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/browser/watchdogs/default_action_watchdog.py
+++ b/browser_use/browser/watchdogs/default_action_watchdog.py
@@ -476,6 +476,28 @@ class DefaultActionWatchdog(BaseWatchdog):
 			self.logger.debug(f'Occlusion check failed: {e}, assuming not occluded')
 			return False
 
+	async def _scroll_parent_iframes_into_view(self, element_node) -> None:
+		"""Recursively scroll all parent iframes into view."""
+		current = element_node.parent
+		while current:
+			# If we find an iframe/frame element, scroll it into view
+			if current.tag_name and current.tag_name.upper() in ('IFRAME', 'FRAME'):
+				try:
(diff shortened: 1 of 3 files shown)


sudhanshu112233shukla:
@sauravpanda @Cheggin  it is reviewed , it can be merged now ?

Issue #2748, opened 2025-08-21
MCP stdio broken by INFO logs on stdout even with ANONYMIZED_TELEMETRY=false → SyntaxError: Unexpected token 'I' … is not valid JSON


beshkenadze:
### Browser Use Version

0.6.0

### Description

When running `browser-use` in MCP stdio mode, the server writes a non-JSON `INFO` line to **stdout**, which causes MCP clients (e.g., MCP Inspector / proxies) to parse it as JSON and fail with:

```
SyntaxError: Unexpected token 'I', "INFO     ["... is not valid JSON
```

**This occurs even when telemetry is explicitly disabled via `ANONYMIZED_TELEMETRY=false`.**
Therefore stdio is unusable out of the box because stdout is polluted with non-JSON logs.

### Steps to Reproduce

1. Verify version:

   ```bash
   uvx 'browser-use[cli]' --version
   # 0.6.0
   ```
2. Start the server via MCP Inspector **with telemetry disabled**:

   ```bash
   ANONYMIZED_TELEMETRY=false \
   bunx @modelcontextprotocol/inspector uvx 'browser-use[cli]' --mcp
   ```
3. Inspector (or any stdio MCP client) fails with:

   ```
   Error from MCP server: SyntaxError: Unexpected token 'I', "INFO     ["... is not valid JSON
   ```
4. Running directly also shows a non-JSON `INFO` line on stdout at startup.

### Expected Behavior

* In MCP stdio mode, **stdout** must contain **only** JSON-RPC messages.
* Logs/telemetry banners should be routed to **stderr** or suppressed (`--quiet`) **before any output**.

### Actual Behavior

* An `INFO` line is written to **stdout** during startup **even when `ANONYMIZED_TELEMETRY=false`**, breaking JSON parsing in stdio clients.

### Environment

* OS: macOS (ARM64)
* `uvx`: `/opt/homebrew/bin/uvx`
* Python (uv managed): 3.12
* `browser-use[cli]`: **0.6.0**
* Repro clients: MCP Inspector, any stdio MCP proxy/client

### Workarounds (partial)

* `BROWSER_USE_LOGGING_LEVEL=error` — reduces noise but does not guarantee JSON-only stdout.

sudhanshu112233shukla:
WORKING on it !

PR #3885, opened 2026-01-14
fix(mcp): redirect logs to stderr to prevent stdout pollution 

sudhanshu112233shukla:
Fixes: #2748
Summary of Changes: Redirects all application logs and critical CLI errors to sys.stderr instead of sys.stdout.
Reasoning: When running in MCP mode (stdio), any non-JSON output on stdout breaks the JSON-RPC protocol, causing clients like MCP Inspector to fail with SyntaxError. By moving logs to stderr, we keep stdout reserved for protocol messages, ensuring robust integration.

---
## Summary by cubic
Redirects all logs and critical CLI errors to stderr so stdout stays clean for JSON-RPC in MCP stdio mode, preventing client parse errors. Also triggers onchange/onblur when present to support frameworks that rely on these handlers.

- **Bug Fixes**
  - Logging: default and CDP logs now write to stderr; CLI import error prints to stderr. Stdout is reserved for protocol messages in MCP.
  - Browser automation: when updating elements, call element.onchange()/onblur() if defined; improve local browser launch by detecting early crashes and surfacing stderr, and simplify binary discovery across OSes.

<sup>Written for commit d612c4e0be5c576b17382a6e5ab591be6c08379e. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/browser/session.py
+++ b/browser_use/browser/session.py
@@ -153,6 +153,7 @@ class BrowserSession(BaseModel):
 		paint_order_filtering: bool | None = None,
 		max_iframes: int | None = None,
 		max_iframe_depth: int | None = None,
+		event_bus_timeout: float | None = None,
 	) -> None: ...
 
 	# Overload 2: Local browser mode (use local browser params)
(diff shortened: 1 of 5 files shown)


laithrw:
@sudhanshu112233shukla, it appears that some of the changes you made here don't relate to the issue you tagged. Did you mean to include these in a separate PR?

Issue #3718, opened 2025-12-04
Bug: Root CDP client not initialized


mdxlCJ:
### Browser Use Version

0.10.1

### Bug Description, Steps to Reproduce, Screenshots

I execute the same command on Windows, and remote Windows can run normally, but when I put it on my local Mac, the browser always crashes with the same command. I have tried multiple times, but with the help of AI, it still cannot be solved
** history
 1141  pwd
 1142  cd ..
 1143  ls 
 1144  cd ..
 1145  ls
 1146  rm -rf .venv
 1147  ls -la
 1148  python3 -m venv .venv
 1149  pip install uv
 1150  uv venv --python 3.12
 1151  uv venv --python 3.12 --clear
 1152  source .venv/bin/activate
 1153  uv pip install browser-use
 1154  uvx browser-use install
 1155  python -m pip install browser-use playwright uv
 1156  python main.py
**

### Failing Python Code

```python
INFO     [Agent] Starting a browser-use agent with version 0.10.1, with provider=browser-use and model=bu-1-0
ERROR    [BrowserSession] 🚌 [BrowserSession.on_BrowserStartEvent(#b869)]           ❌ Failed (3.02s): JSONDecodeError: Expecting value: line 1 column 1 (char 0)
ERROR    [BrowserSession] 🚌 [BrowserSession.on_BrowserStartEvent(#b869)]           ❌ CDP connected but failed to re-create CDP session after error "JSONDecodeError: Expecting value: line 1 column 1 (char 0)" in on_BrowserStartEvent(BrowserStartEvent#b869): due to AssertionError: Root CDP client not initialized

ERROR    [Agent] Agent run failed with exception: Expecting value: line 1 column 1 (char 0)
Traceback (most recent call last):
  File "/Users/admin/mayun/browser-use-single/.venv/lib/python3.12/site-packages/browser_use/agent/service.py", line 1876, in run
    await self.browser_session.start()
  File "/Users/admin/mayun/browser-use-single/.venv/lib/python3.12/site-packages/browser_use/observability.py", line 73, in async_wrapper
    return await func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/admin/mayun/browser-use-single/.venv/lib/python3.12/site-packages/browser_use/browser/session.py", line 555, in start
    await start_event.event_result(raise_if_any=True, raise_if_none=False)
[...]

PR #3898, opened 2026-01-16
fix(browser): return ws url from watchdog to prevent json error and a…

sudhanshu112233shukla:
**How it fixes the bug:**
Fixes:#3718
**Previously**: The watchdog verified the browser was up, but BrowserSession.connect would then make a second HTTP request to /json/version to get the WebSocket URL. This second request was flaky on macOS, often returning invalid data or closing too early, causing the JSONDecodeError.
**Now:** The watchdog parses the /json/version response itself during its startup check (which already retries until success). It passes the valid webSocketDebuggerUrl directly to the session. This completely bypasses the problematic redundant HTTP call on Mac, ensuring a stable connection every time.
**Verification:** I added a permanent regression test 
tests/unit/test_cdp_connection.py
 that mocks this flow to guarantee the watchdog correctly extracts and returns the URL.

---
## Summary by cubic
Return the WebSocket URL directly from the watchdog to remove the flaky second /json/version call in BrowserSession.connect. Fixes the macOS JSONDecodeError and stabilizes CDP initialization.

- **Bug Fixes**
  - Watchdog parses /json/version and returns webSocketDebuggerUrl during startup, so BrowserSession doesn’t re-query.
  - Monitors the browser subprocess and fails fast with exit code and stderr if it crashes.
  - Retries on partial/invalid JSON and when webSocketDebuggerUrl is missing; added regression tests for these cases.

<sup>Written for commit 2c24715c4483ac51bd8c06a6782268f2c21f2733. Summary will update on new commits.</sup>


laithrw:
@sudhanshu112233shukla, thanks -- this looks great.

One issue: `_wait_for_cdp_url` now accepts `process`, but `_launch_browser()` still calls it without passing the subprocess, so the crash/stderr monitoring is currently dead code. Can you update the call to pass the `asyncio.subprocess.Process`?

Also, can you make `webSocketDebuggerUrl` extraction defensive (avoid KeyError; retry/continue if missing)?
Docstring tweak to clarify it returns the WebSocket CDP URL would be nice too.

sudhanshu112233shukla:
@laithrw ok i will do it !!

sudhanshu112233shukla:
@laithrw i have made the changes as you said do check it !!

Aqib053:
Hi, noticed this PR has been open a while — is it still being worked on? Happy to pick it up and finish it if it's stalled, or open a fresh PR referencing the same issue if that's easier for maintainers.

Issue #3897, opened 2026-01-16
LLM response parsing fails with 'NoneType' object is not subscriptable when using non-OpenAI-compatible proxy


priy-anka17:
### Browser Use Version

0.11.2  (Installed via uv pip install browser-use, confirmed in logs: Starting a browser-use agent with version 0.11.2)

### Bug Description, Steps to Reproduce, Screenshots

When using browser_use.Agent with a valid LLM accessed through a custom proxy (LN proxy), the agent crashes during execution with the error:

ModelProviderError: 'NoneType' object is not subscriptable

This happens even though:

The LLM model is valid and reachable

The same model works perfectly when called directly via HTTP

Credentials and API keys are correct

The issue appears to be caused by browser_use assuming an OpenAI-style chat completion response schema (choices[].message.content).
The proxy returns a valid response but with a different JSON structure, which causes response parsing to fail inside browser-use.

Steps to Reproduce

Install browser-use:

uv pip install browser-use

Install Playwright browser:

playwright install chromium --with-deps

Set environment variables:

BASE_URL=https://example.com/login
VALID_USERNAME=example
VALID_PASSWORD=example
BROWSER_USE_API_KEY=****

Run the Python script using Agent(task=...) (see code below)

Observe that:

Browser launches correctly

Navigation succeeds

Agent fails on the first reasoning step with repeated LLM errors

After 4 retries, execution stops

Observed Error
WARNING  [Agent] ⚠️ LLM error (ModelProviderError: 'NoneType' object is not subscriptable) but no fallback_llm configured
WARNING  [Agent] ❌ Result failed 1/4 times: 'NoneType' object is not subscriptable
...
ERROR    [Agent] ❌ Stopping due to 3 consecutive failures

Expected Behavior

The agent should either:

Handle non-OpenAI-compatible response schemas gracefully, or

Provide a clear validation error indicating unsupported LLM response format

### Failing Python Code

```python

```

### LLM Model

gpt-41-2025-04-14-ondemand_zalexissign_2971_nonprod (Accessed via a custom enterprise proxy, not OpenAI directly)

### Full DEBUG Log Output

```shell
INFO     [Agent] Starting a browser-use agent with version 0.11.2
INFO     [Agent] 🔗 Found URL in task: https://example.com/login
INFO     [Agent] ▶️ navigate: url: https://example.com/login
WARNING  [BrowserSession] Page readiness timeout (4.0s)
INFO     [Agent] 📍 Step 1:
[...]

PR #3899, opened 2026-01-16
fix(llm): handle missing choices in openai proxy response

sudhanshu112233shukla:
Fixes: #3897
**The Problem**
When using browser-use with certain custom LLM proxies (configured via 
ChatOpenAI
), the agent would crash unexpectedly during the execution loop. The error traceback pointed to 
browser_use/llm/openai/chat.py
 with TypeError: 'NoneType' object is not subscriptable.

This happened because some proxies return a JSON response where the choices field is either missing or explicitly set to None (likely due to internal errors or non-standard formatting). The existing code assumed response.choices always contained a valid list and attempted to access response.choices[0] directly, causing the crash when the list was None.

**The Fix**
This PR modifies 
browser_use/llm/openai/chat.py
 to add robust validation for the LLM response structure.

**Identified the Failure Point:** Located the exact lines in 
ainvoke
 where response.choices[0] was accessed for both standard string completions and structured output.
Added Guard Clauses: Implemented an explicit check: if not response.choices:.
Improved Error Handling: Instead of letting the application crash with a confusing Python TypeError, the code now raises a 
**ModelProviderErro**r
 with a specific message: "Model response missing 'choices' field. This can happen with incompatible proxies."
This change allows the agent to handle these malformed proxy responses gracefully, providing users with a clear error message that points to the proxy configuration rather than a bug in the library code.

**Verification/ testing** 
**Reproduction**: Created a reproduction script that mocked AsyncOpenAI to return a ChatCompletion object with choices=None, successfully replicating the TypeError.
**Validation**: Ran the same reproduction script with the fix applied; confirmed that 
ModelProviderError
 was raised with the correct message.
**Regression Testing**: Verified that standard OpenAI usage remains unaffected.

---
## Summary by cubic
Fix crash when some OpenAI-compatible proxies return chat completions without choices. We now validate choices in ainvoke and raise a clear ModelProviderError (with base_url hint) instead of a TypeError; structured-output calls no longer depend on schema injection.

- **Bug Fixes**
  - Validate response.choices in both text and structured paths; if missing/empty, raise ModelProviderError (502) with a base_url hint and preserve status/message.
  - Use the validated first choice for content and finish_reason; keep the structured output parse check.
  - Structured-output request runs even when add_schema_to_system_prompt is false; removed a temporary debug regression test and reverted the prior CI timeout change.

<sup>Written for commit 668f51ead96d9e8057cb345eb743dc741d08b41b. Summary will update on new commits.</sup>


priy-anka17:
it is throwing the same error after these changes.

sudhanshu112233shukla:
Hi @priy-anka17 , thanks for testing this!

I have verified the changes and the fix explicitly guards against this error. The code now checks if not response.choices: and raises a ModelProviderError before it ever tries to access response.choices[0].

If you are still seeing the original TypeError: 'NoneType' object is not subscriptable, it is highly likely that your environment is not running the modified code yet.

Could you please check the following?

Reinstall/Refresh: If you installed via pip install -e ., please restart your Python interpreter or kernel to ensure the changes are loaded.

Traceback: If it persists, could you please paste the full traceback here? It is possible the same error is happening in a different file that we missed, and the traceback will pinpoint exactly where.

Thanks!

PR #3915, opened 2026-01-22
fix(cli): enable chaining multiple run commands on same session

BouquetAntoine:
## Summary

- Persist agent across `run` commands in the same session
- Use `agent.add_new_task()` for subsequent runs instead of creating new agents
- Aligns CLI behavior with documented Python API pattern for task chaining

## Problem

Currently, each `browser-use run` command creates a **new Agent instance**, which loses the context and browser state from previous runs. This causes subsequent runs on the same session to fail:

```bash
browser-use --session test run "go to wikipedia.org"     # ✅ done: true, steps: 2
browser-use --session test run "search for France"       # ❌ done: false, steps: 0
browser-use --session test run "get first paragraph"     # ❌ done: false, steps: 0
```

The second and third runs return `steps: 0` because the new agent doesn't have the context from the previous agent's work.

## Solution

This fix aligns the CLI with the [documented Python API pattern for task chaining](https://docs.browser-use.com/examples/templates/follow-up-tasks):

```python
# Documented pattern - reuse agent with add_new_task()
agent = Agent(task='first task', browser_session=browser)
await agent.run()

agent.add_new_task('follow up task')  # Preserves context!
await agent.run()
```

**Changes:**
1. Add `agent` field to `SessionInfo` to persist the agent across runs
2. On first run: create agent and store in session
3. On subsequent runs: call `session.agent.add_new_task(task)` then run

## After this fix

```bash
browser-use --session test run "go to wikipedia.org"     # ✅ done: true, steps: 2
browser-use --session test run "search for France"       # ✅ done: true, steps: 2
browser-use --session test run "get first paragraph"     # ✅ done: true, steps: 2
```

## Testing

Tested locally with multiple chained runs - all complete successfully with `done: true`.

---
## Summary by cubic
Enables chaining multiple CLI run commands in the same session by reusing the agent. Preserves context and browser state so follow-up tasks complete instead of returning steps: 0.

- **Bug Fixes**
  - Persist agent in SessionInfo and reuse it across runs.
  - First run creates and stores the agent; subsequent runs use agent.add_new_task(task) before run, aligning the CLI with the documented Python API pattern.

<sup>Written for commit 900a027acf94d45614cc676c3467c26973d6ba77. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/skill_cli/commands/agent.py
+++ b/browser_use/skill_cli/commands/agent.py
@@ -40,15 +40,21 @@ async def handle(session: SessionInfo, params: dict[str, Any]) -> Any:
 				'error': 'No LLM configured. Set OPENAI_API_KEY, ANTHROPIC_API_KEY, or GOOGLE_API_KEY',
 			}
 
-		# Create and run agent
-		agent = Agent(
-			task=task,
-			llm=llm,
-			browser_session=session.browser_session,
-		)
-
-		logger.info(f'Running agent task: {task}')
-		result = await agent.run(max_steps=max_steps)
+		# Reuse existing agent or create new one (enables chaining multiple run commands)
+		if session.agent is not None:
+			# Chain: add new task to existing agent (preserves context and browser state)
+			logger.info(f'Chaining agent task: {task}')
+			session.agent.add_new_task(task)
+			result = await session.agent.run(max_steps=max_steps)
+		else:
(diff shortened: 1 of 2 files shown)


BouquetAntoine:
would be a good feature (and something that any developer would expect... seems like required for multi steps agentic workflow) I think

PR #4017, opened 2026-02-03
fix(session): force new window creation if no targets exist (#3908)

### LINKED ISSUE #3908: Target.createTarget fails with 'no browser is open' when all tabs are closed due to newWindow=false
## Description

When the last tab is closed (e.g., by SecurityWatchdog closing a disallowed URL like `file://`), `_recover_agent_focus` in SessionManager tries to create a new tab via `_cdp_create_new_page()`. This fails with:

```
RuntimeError: {'code': -32000, 'message': 'Failed to open new tab - no browser is open'}
```

## Root Cause

`_cdp_create_new_page()` in `session.py:2852` defaults to `new_window=False`:

```python
async def _cdp_create_new_page(self, url: str = 'about:blank', background: bool = False, new_window: bool = False) -> str:
    ...
    params={'url': url, 'newWindow': new_window, 'background': background}
```

When the last tab closes, Chrome closes the **window** (but keeps the process and CDP server alive). With `newWindow=false`, Chrome cannot create a tab because there's no window to put it in.

## Reproduction

```python
import asyncio
import websockets
import json
import urllib.request

async def test():
    # Assuming Chrome is running with --remote-debugging-port=9222
    version = json.loads(urllib.request.urlopen("http://localhost:9222/json/version").read())
    ws_url = version['webSocketDebuggerUrl']
    
    targets = json.loads(urllib.request.urlopen("http://localhost:9222/json").read())
    page = next((t for t in targets if t.get('type') == 'page'), None)
    target_id = page['id']
    
    async with websockets.connect(ws_url) as ws:
        # Close the only tab
        await ws.send(json.dumps({"id": 1, "method": "Target.closeTarget", "params": {"targetId": target_id}}))
        print(await ws.recv())  # success
        
        await asyncio.sleep(0.3)
        
        # Try with newWindow=false (like browser_use does)
        await ws.send(json.dumps({
            "id": 2, 
            "method": "Target.createTarget", 
            "params": {"url": "about:blank", "newWindow": False}
        }))
        print(await ws.recv())  # ERROR: "Failed to open new tab - no browser is open"
        
        # Try WITHOUT newWindow param
        await ws.send(json.dumps({
            "id": 3, 
            "method": "Target.createTarget", 
            "params": {"url": "about:blank"}
        }))
        print(await ws.recv())  # SUCCESS - Chrome auto-creates window

asyncio.run(test())
```

**Output:**
```
{"id":1,"result":{"success":true}}
{"id":2,"error":{"code":-32000,"message":"Failed to open new tab - no browser is open"}}
{"id":3,"result":{"targetId":"..."}}
```

## Environment

- browser-use version: 0.10.x
- OS: macOS (also reproducible on other platforms)
- Chrome: 131.x

## Suggested Fix

Either:
1. Don't pass `newWindow` parameter at all when creating tabs (let Chrome decide)
2. Or check if page targets count is 0 before creating and use `new_window=True` in that case

**Option 1 (minimal change):**
```python
async def _cdp_create_new_page(self, url: str = 'about:blank', background: bool = False, new_window: bool | None = None) -> str:
    params = {'url': url, 'background': background}
    if new_window is not None:
        params['newWindow'] = new_window
    ...
```

**Option 2 (explicit handling):**
```python
async def _cdp_create_new_page(self, url: str = 'about:blank', background: bool = False, new_window: bool = False) -> str:
    # If no windows exist, force new window creation
    if not new_window:
        page_targets = self.session_manager.get_all_page_targets()
        if len(page_targets) == 0:
            new_window = True
    ...
```

yttrium400:
## Description
Fixes #3908 by modifying `_cdp_create_new_page` in `browser_use/browser/session.py`.

When all browser tabs are closed, `Target.createTarget` fails if `newWindow=False` because there is no window to attach the new tab to. This PR detects when no targets exist and forces `newWindow=True` in that case, allowing recovery.

## Changes
- Updated `_cdp_create_new_page` to check `self.session_manager.get_all_page_targets()`.
- Added regression test `tests/ci/test_issue_3908.py`.

## Verification
- Ran reproduction test `tests/ci/test_issue_3908.py` which failed before fix and passed after fix.

---
## Summary by cubic
Force a new window when no page targets exist so Target.createTarget succeeds after all tabs are closed. This prevents the "no browser is open" CDP error (fixes #3908).

- **Bug Fixes**
  - Update _cdp_create_new_page to check session_manager.get_all_page_targets() and set new_window=True when none are found.
  - Add regression test tests/ci/test_issue_3908.py to cover the all-tabs-closed scenario.

<sup>Written for commit 5b179bc86ce4b0d165ec88dfbee6b13e5ff44d6f. Summary will update on new commits.</sup>

### DIFF
--- a/tests/ci/test_issue_3908.py
+++ b/tests/ci/test_issue_3908.py
@@ -40,7 +40,7 @@ async def test_reproduce_create_target_no_browser_failure():
 			await asyncio.sleep(0.1)
 
 		page_targets_now = session.session_manager.get_all_page_targets()
-		
+
 		# If auto-recovery happened (using our fix), we have pages.
 		# If not, we try to create one manually to verify the fix works in '0 pages' state.
 		if len(page_targets_now) == 0:
@@ -48,9 +48,7 @@ async def test_reproduce_create_target_no_browser_failure():
 			page_targets_now = session.session_manager.get_all_page_targets()
 
 		# The fix is verified if we have pages now and didn't crash with "no browser is open"
-		assert len(page_targets_now) > 0, f"Should have active pages (recovered or created), but found {len(page_targets_now)}"
-
-
+		assert len(page_targets_now) > 0, f'Should have active pages (recovered or created), but found {len(page_targets_now)}'
 
 	finally:
 		await session.stop()


laithrw:
Thanks for the PR. I'm closing it in favor of a simpler fix (#4283). Chrome auto-creates a window when needed if newWindow isn't specified, so there's no need to query the page on every call.

PR #4034, opened 2026-02-05
fix(#3657): agent not switching tabs after new opened

laithrw:
Include in agent context that new tab was opened

---
## Summary by cubic
Notify the agent when a click opens a new browser tab so it knows to switch to it. Addresses #3657 where the agent stayed on the old tab after actions that opened a new page.

- **Bug Fixes**
  - Detect new tabs by comparing tab IDs before and after clicks in _click_by_coordinate and _click_by_index.
  - Append a note to the action memory with the new tab’s short ID, prompting the agent to switch.

<sup>Written for commit 99be7b1b48dff53980f820ff56db7bd0f40eff96. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/browser/session.py
+++ b/browser_use/browser/session.py
@@ -1359,6 +1359,7 @@ class BrowserSession(BaseModel):
 		from browser_use.browser.watchdogs.default_action_watchdog import DefaultActionWatchdog
 		from browser_use.browser.watchdogs.dom_watchdog import DOMWatchdog
 		from browser_use.browser.watchdogs.downloads_watchdog import DownloadsWatchdog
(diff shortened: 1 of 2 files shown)


laithrw:
@bensaru this works for me. Are you still having issues with tab switching?

ekmekovski:
The issue still continues

laithrw:
Hey @ekmekovski, are you using proxies with auth? If so, this causes the tab tracking handler to be overwritten and has been fixed in #4220. It'll be included in the next release.

If not, could you please provide a script to reproduce this?

ekmekovski:
Yes I was usng proxy with authentication so I believe that's the reason. Thank you.

PR #4036, opened 2026-02-05
feat(mcp): save screenshots to disk instead of base64 in browser_get_state

### LINKED ISSUE #3934: MCP server: Return file path instead of base64 for screenshots
**What is the problem that your feature request solves?**

The `browser_get_state` tool in the MCP server currently returns screenshots as base64 strings embedded in the JSON response. This makes the responses very large and makes it difficult to work with screenshots programmatically (e.g., saving them, displaying them, or passing them to other tools).

When `include_screenshot: true` is passed, the response includes a `screenshot` field containing a base64-encoded string, which can be hundreds of KB or even MB in size. This bloats JSON responses and makes it harder to work with the screenshots.

**What is your proposed solution?**

Save screenshots to disk and return the file path instead of base64. This would:

1. Make JSON responses much smaller (file path string vs large base64 blob)
2. Make screenshots easier to work with programmatically
3. Allow screenshots to persist on disk for later use
4. Be consistent with how `BrowserStateHistory` already handles screenshots (it uses `screenshot_path`)

Implementation approach:
- Save screenshots to a temp directory (e.g., `~/.browser-use-mcp-screenshots/` or system temp dir)
- Return the absolute file path in the `screenshot` field
- Include error handling to fallback to base64 if file save fails (non-breaking)

**What hacks or alternative solutions have you tried to solve the problem?**

Currently decoding base64 manually and saving to files, but this is inefficient and requires extra processing. The MCP server should handle this automatically.

**What version of browser-use are you currently using?**

latest (main branch)

**Additional Notes:**

I'm willing to work on a PR for this. The change would be in `browser_use/mcp/server.py` in the `_get_browser_state` method. I've already started implementing this locally and can submit a PR once the approach is approved.

ashish-cheruku:
This PR implements the feature requested in #3934 - returning file paths instead of base64 for screenshots in the MCP server.

## Problem

The `browser_get_state` tool with `include_screenshot: true` returns screenshots as base64 strings embedded in JSON. This makes responses huge (hundreds of KB to MB) and harder to work with.

## Solution

Modified `_get_browser_state()` to:
1. Save screenshots to `~/.browser-use-mcp/screenshots/` 
2. Return `screenshot_path` field with the absolute file path
3. Fallback to base64 if file save fails (non-breaking)

## Benefits

- **Much smaller JSON responses** - file path string vs MB of base64
- **Screenshots persist on disk** for later use
- **Easier to work with** programmatically
- **Consistent** with how `BrowserStateHistory` uses `screenshot_path`

## Example Response

Before:
```json
{
  "screenshot": "iVBORw0KGgoAAAANSUhEU... (1MB+)"
}
```
After:
```json
{
  "screenshot_path": "/Users/john/.browser-use-mcp/screenshots/screenshot_20260205_183050_a1b2c3d4.png"
}
```
Closes #3934

---
## Summary by cubic
Save browser_get_state screenshots to disk and return screenshot_path instead of base64. This cuts JSON size and makes screenshots easier to reuse, with a safe fallback to base64 if saving fails.

- **New Features**
  - Saves files to ~/.browser-use-mcp/screenshots/
  - Adds screenshot_path with an absolute file path in the response
  - Uses async file I/O and cleans up old screenshots (keeps last 50)
  - Consistent with BrowserStateHistory’s screenshot_path pattern

<sup>Written for commit 9e9b71a6d52204352248b6c944c12c91cd23392f. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/mcp/server.py
+++ b/browser_use/mcp/server.py
@@ -823,21 +823,34 @@ class BrowserUseServer:
 				import base64
 				import uuid
 
+				import anyio
+
 				# Create screenshots directory
 				screenshots_dir = Path.home() / '.browser-use-mcp' / 'screenshots'
-				screenshots_dir.mkdir(parents=True, exist_ok=True)
+				await anyio.Path(screenshots_dir).mkdir(parents=True, exist_ok=True)
 
 				# Generate unique filename with timestamp
 				timestamp = time.strftime('%Y%m%d_%H%M%S')
 				filename = f'screenshot_{timestamp}_{uuid.uuid4().hex[:8]}.png'
 				screenshot_path = screenshots_dir / filename
 
-				# Decode base64 and save to file
+				# Decode base64 and save to file asynchronously
 				screenshot_data = base64.b64decode(state.screenshot)
-				screenshot_path.write_bytes(screenshot_data)
+				async with await anyio.open_file(screenshot_path, 'wb') as f:
+					await f.write(screenshot_data)
 
 				result['screenshot_path'] = str(screenshot_path)
 				logger.debug(f'Screenshot saved to: {screenshot_path}')
+
+				# Cleanup old screenshots - keep only last 50 files
+				try:
+					screenshot_files = sorted(screenshots_dir.glob('screenshot_*.png'), key=lambda p: p.stat().st_mtime)
+					if len(screenshot_files) > 50:
+						for old_file in screenshot_files[:-50]:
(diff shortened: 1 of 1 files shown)


laithrw:
@ashish-cheruku, thanks for the PR. You return file paths for screenshots, but MCP has a native ImageContent type for this. Returning ImageContent directly means it works across all transports (stdio, SSE, in-memory) without the client needing filesystem access to the server.

Closing in favor of #4255.

PR #4037, opened 2026-02-05
Add XPath Attribute Normalization Utilitycommit

### LINKED ISSUE #4003: Bug: not created correct locator
### Browser Use Version

Browser_use[video]=0.11.0-browser_use[video]=0.11.5

Not creating correct locator repeated for such scenarios not sure for reason 
Gpt-5.1

these locators for these step not coming correct(both versions) not able to figure out
locators
//div[@role='columnheader']//div[contains(@class,'oxd-table-header-cell-text') and normalize-space(.)='Job Titles']---->step:Verify that Job Tiles Column header is equal to Job Titles
//div[@role='columnheader']//div[contains(@class,'oxd-table-header-cell-text') and normalize-space(.)='Job Description']----->step:Verify that Job Description Column header is equal to Job Description

### Failing Python Code

```python
task="""
Perform the following manual steps and create locators
A. Locator Strategy (CRITICAL)
        Your first job for each interactive element is to find the most stable, unique locator by executing JavaScript in the live DOM (e.g., document.querySelector, getAttribute).
        -Priority Order:
            -Stable ID: Use only if the id is semantic, static, and non-dynamic (e.g., "username"). Avoid IDs with random numbers, timestamps, or session values.
            -Data Attributes: data-testid, data-qa, data-cy.
            -CSS Selector: Robust combinations of tag, class, and other static attributes.
            -ARIA Attributes: aria-label, aria-describedby.
            -Name Attribute: Use only if no better option from above exists.
            -XPath: Simplified last resort.
        -Critical Locator Rules:
            -Text Matching: For XPath text matching, use contains(., 'text') (not contains(text(), ...)), especially for <i> tags or nested text.
            -Case Sensitivity: Match the exact case of attributes and text as they appear in the live DOM.
        Simplicity: Prefer a simple, working locator over a complex "perfect" one. Avoid unnecessary parent/child relationships or Shadow DOM unless required.

Manual steps
Navigate to https://opensource-demo.orangehrmlive.com/web/index.php/auth/login,
Enter Username as Admin
Enter Password as admin123
Click on Login button
Verify that header is equal to Dashboard
Click on Admin from left side bar
Verify that header contains Admin
Click on Job menu option
Click on Job Titles
Verify that sub header is equal to Job Titles
Verify that Job Tiles Column header is equal to Job Titles
Verify that Job Tiles Column value is not empty
Verify that Job Description Column header is equal to Job Description
Verify that Job Description Column value is empty"""
```

### LLM Model

gpt 5.1 with flash_mode True and flash_mode false both

_No response_

### Full DEBUG Log Output

```shell

```

Adityakk9031:
Problem
Fixes #4003

LLMs sometimes generate XPath locators with incorrectly capitalized attribute names when users create custom prompts asking for XPath generation. For example:

@ROLE instead of @role
@Class instead of @class
@ID instead of @id
@ARIA-LABEL instead of @aria-label
Since XPath is case-sensitive and standard HTML attributes are lowercase, these locators fail to find elements.

Root Cause
This is not a bug in Browser-Use. Browser-Use correctly preserves HTML attribute case throughout the entire pipeline (extraction → storage → serialization → LLM output). The issue occurs when:

Users create custom prompts explicitly asking the LLM to generate XPath locators
The LLM hallucinates incorrect attribute casing in the generated XPath
The case-sensitive XPath expressions fail
Solution
Added a defensive utility function 
normalize_xpath_attributes()
 to fix LLM-generated XPath with incorrect attribute casing.

Why Add This Fix?
While the proper solution is to use Browser-Use's native action system or improve prompts, this utility provides:

Defensive programming for users who need custom XPath
Easy migration for existing codebases with LLM-generated XPath
Better DX - users don't need to debug obscure XPath failures
Changes
Core Implementation
File: 
browser_use/utils.py

Added 
normalize_xpath_attributes()
 function:

python
def normalize_xpath_attributes(xpath: str) -> str:
    """
    Normalize XPath attribute names to lowercase to fix LLM hallucination errors.
    
    Examples:
        >>> normalize_xpath_attributes("//div[@ROLE='button']")
        "//div[@role='button']"
        >>> normalize_xpath_attributes("//div[@Class='test'][@ID='myid']")
        "//div[@class='test'][@id='myid']"
    """
Features:

Converts attribute names to lowercase: @ROLE → @role
Preserves attribute values: @role='BUTTON' stays @role='BUTTON'
Handles ARIA attributes: @ARIA-LABEL → @aria-label
Handles data attributes: @DATA-ID → @data-id
Uses regex pattern r'@([A-Za-z][\w-]*)' for robust matching
Testing
File: 
tests/test_xpath_normalization.py

Added comprehensive test suite with 17 test cases:

Single and multiple attributes
Mixed case normalization
 ARIA and data-* attributes
 Value preservation
 Complex XPath with 
contains()
, 
text()
Real-world example from issue #4003
Manual Verification
bash
$ python -c "from browser_use.utils import normalize_xpath_attributes; \
print(normalize_xpath_attributes('//div[@ROLE=\"button\"][@Class=\"test\"]'))"
//div[@role="button"][@class="test"]
Successfully normalizes attribute names while preserving structure.

---
## Summary by cubic
Add normalize_xpath_attributes to auto-lowercase attribute names in XPath so LLM-generated locators stop failing due to case sensitivity. Fixes #4003 and improves reliability when users provide custom XPath.

- **New Features**
  - normalize_xpath_attributes(xpath): lowercases attribute names (e.g., @ROLE → @role), preserves values, and handles aria-* and data-*.
  - Test suite (16 cases) covering single/multiple attributes, complex predicates (contains, text), and a real example from #4003.

<sup>Written for commit 1de5980ea1817b8f1510c1a5730a830f78650a53. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/utils.py
+++ b/browser_use/utils.py
@@ -812,4 +812,3 @@ def normalize_xpath_attributes(xpath: str) -> str:
 		return f'@{match.group(1).lower()}'
 
 	return re.sub(pattern, lowercase_attribute, xpath)
-


Adityakk9031:
@ShawnPana please into this pr

laithrw:
@Adityakk9031, thanks for the PR. #4003 is not a Browser-Use interaction bug; it’s about LLM-generated XPath strings from a custom prompt. Also, hardcoding heuristics like this can be problematic (for example, the current regex could break valid camelCase attributes).

PR #4043, opened 2026-02-06
removed default step info from the system prompt

sauravpanda:
## Summary by cubic
Removed step counter and max_steps logic from prompts to avoid exposing step budgets. Agent state now includes only the current date (as <current_date>), and prompts instruct agents to wrap up only when explicitly told.

<sup>Written for commit a7bd106f8d45616bac00776b1a6c7f3415042a4a. Summary will update on new commits.</sup>


sauravpanda:
Doesnt improve the agent!

PR #4063, opened 2026-02-09
fix: disable proxy for localhost CDP requests (#4050)

### LINKED ISSUE #4050: CDP startup fails on Windows with proxy env: JSONDecodeError on /json/version
### Browser Use Version
0.11.9

#### Description
On Windows, `browser-use open ...` fails during startup with:

```text
JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

The crash happens while resolving CDP WebSocket URL from `http://127.0.0.1:<port>/json/version`.

#### Steps to Reproduce
1. Open terminal on Windows 11.
2. Run:
   ```bash
   browser-use open https://example.com
   ```
3. Observe startup failure before navigation.

#### Actual Result
CLI returns JSON decode error and does not open target page.

#### Expected Result
CLI should resolve the CDP endpoint and navigate successfully.

#### Screenshots / Evidence
- For the same local DevTools endpoint:
  - `Invoke-WebRequest http://127.0.0.1:<port>/json/version` returns valid JSON (includes `webSocketDebuggerUrl`).
  - Python `httpx.get("http://127.0.0.1:<port>/json/version")` returns `502` with empty body in this environment.
- Traceback points to:
  - `browser_use/browser/session.py`
  - `version_info.json()['webSocketDebuggerUrl']`

### Failing Python Code
```bash
browser-use open https://example.com
```

### LLM Model
N/A (CLI startup path)

Windows 11 + Playwright Chromium (browser-use default)

### Full DEBUG Log Output
```text
INFO [server] Dispatch: open
ERROR [server] Error dispatching open: Expecting value: line 1 column 1 (char 0)
  ...
  File ".../browser_use/browser/session.py", line 1531, in connect
    self.browser_profile.cdp_url = version_info.json()['webSocketDebuggerUrl']
  ...
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

### Proposed Fix
1. Use `httpx.AsyncClient(trust_env=False)` when querying localhost CDP endpoints.
2. Add short retries for `/json/version` during browser startup.
3. Improve JSON parse failure logs with HTTP status code + response length.

Chase-Xuu:
## Summary

Fixes #4050 - CDP startup fails on Windows with proxy env causing JSONDecodeError on /json/version.

## Problem

When `HTTP_PROXY` or `HTTPS_PROXY` environment variables are set, `httpx.AsyncClient()` by default routes **all** HTTP requests through the proxy, including localhost requests to the Chrome DevTools Protocol endpoint (`http://127.0.0.1:<port>/json/version`).

On Windows, this causes the CDP endpoint query to return a 502 response with an empty body, which then fails JSON parsing with:
```
JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

## Solution

Add `trust_env=False` to the `httpx.AsyncClient()` call when querying the CDP `/json/version` endpoint. This prevents httpx from using proxy environment variables for this specific localhost request.

## Changes

- Modified `browser_use/browser/session.py`: Added `trust_env=False` parameter to httpx.AsyncClient() in the connect() method

## Testing

- Verified the change allows CDP connection to succeed when proxy env vars are set
- The fix is minimal and targeted, only affecting the CDP endpoint query

---
## Summary by cubic
Disable proxy usage only for local CDP /json/version requests (localhost, 127.0.0.1, ::1) to fix Windows startup 502s and JSONDecodeError. Local requests use httpx.AsyncClient with trust_env=False; remote CDP URLs continue to honor HTTP_PROXY/HTTPS_PROXY.

<sup>Written for commit 9932f065000941dbdac3421bac291338ab84c04f. Summary will update on new commits.</sup>


Chase-Xuu:
Thanks for the review feedback! I've updated the fix to only disable `trust_env` for localhost addresses (`localhost`, `127.0.0.1`, `::1`). Remote CDP URLs will continue to respect proxy environment variables.

```python
is_localhost = parsed_url.hostname in ('localhost', '127.0.0.1', '::1')
async with httpx.AsyncClient(trust_env=not is_localhost) as client:
```

This way, Windows users with system proxies can connect to local CDP endpoints without 502 errors, while remote CDP connections still work through corporate proxies.

PR #4127, opened 2026-02-20
fix: prevent judge from corrupting structured output

### LINKED ISSUE #4103: Bug: _run_simple_judge corrupts structured_output when output_model_schema is used
### Browser Use Version

0.11.9

When using `output_model_schema` for structured output, `_run_simple_judge()` appends text to `extracted_content`, corrupting the JSON. Both `history.structured_output` and `history.final_result()` return the corrupted string, causing `model_validate_json()` to fail with "trailing characters".
## Root Cause
In `agent/service.py`, `_run_simple_judge()` unconditionally mutates `extracted_content`:
```python
# service.py line ~1506
if last_result.extracted_content:
    last_result.extracted_content += f'\n\n{note}'
```
The `done` action handler for structured output stores clean JSON in `extracted_content`:
```python
# tools/service.py - done action with output_model
output_dict = params.data.model_dump(mode='json')
return ActionResult(
    extracted_content=json.dumps(output_dict, ensure_ascii=False),
    ...
)
```
After the judge runs, `extracted_content` becomes:
```
{"contacts": [...]}\n\n[Simple judge: Task requirements not fully met]
```
The `structured_output` property then calls `final_result()` and tries `model_validate_json()` on this corrupted string, which fails.
## Reproduction
```python
from pydantic import BaseModel, Field
from browser_use import Agent, Browser
class MyOutput(BaseModel):
    items: list[str] = Field(default_factory=list, description="Extracted items")
agent = Agent(
    task="Go to https://news.ycombinator.com and extract the titles of the top 3 posts",
    llm=your_llm,
    output_model_schema=MyOutput,
)
history = await agent.run()
# This fails with ValidationError "trailing characters" when the simple
# judge disagrees with the agent and appends its note to extracted_content
output = history.structured_output
```
The bug is intermittent — it only triggers when the simple judge overrides `success` to `False` (i.e., `result.is_correct == False`). When the judge agrees with the agent, `extracted_content` is not modified and parsing works fine.
## Expected Behavior
`history.structured_output` should always return a valid parsed Pydantic model when `output_model_schema` is set and the agent produced valid structured output. The judge's assessment should not corrupt the structured data.
## Suggested Fix
`_run_simple_judge()` should store its note separately rather than appending to `extracted_content`, or at minimum skip the append when `output_model_schema` is set. For example:
```python
# Option A: use a separate field
last_result.judge_note = note
# Option B: skip mutation for structured output
if last_result.extracted_content and self._output_model is None:
    last_result.extracted_content += f'\n\n{note}'
```

### Failing Python Code

```python

```

### LLM Model

_No response_

_No response_

### Full DEBUG Log Output

```shell

```

hztBUAA:
## Summary
Fixes #4103

When using `output_model_schema` for structured output, `_run_simple_judge()` could corrupt the JSON stored in `extracted_content`, causing `model_validate_json()` to fail with "trailing characters".

The existing fix stored the judge note in the `metadata` dict, but this PR improves on that by:

- Adding a dedicated `judge_note` field to `ActionResult` for clearer semantics and discoverability
- Always storing the judge note in `judge_note` regardless of output mode
- Only appending to `extracted_content` in plain-text mode (no `output_model_schema`), preserving backward compatibility
- Keeping `metadata` for its intended purpose (observability data like click coordinates)

## Changes
- `browser_use/agent/views.py`: Add `judge_note: str | None` field to `ActionResult`
- `browser_use/agent/service.py`: Refactor `_run_simple_judge()` to use `judge_note` and only append to `extracted_content` for plain-text mode
- `tests/ci/test_simple_judge_structured_output.py`: New regression tests covering judge rejection with structured output, plain-text append, approval pass-through, and early-exit paths
- `tests/ci/test_sandbox_structured_output.py`: Add test verifying `structured_output` parses correctly after judge override

## Test plan
- [x] Judge rejection with `output_model_schema` keeps `extracted_content` as valid JSON
- [x] Judge rejection without `output_model_schema` still appends note to `extracted_content`
- [x] Judge approval leaves content untouched
- [x] Judge skips when `is_done=False` or `success=False`
- [x] `history.structured_output` parses correctly after judge override
- [x] All 14 related tests pass

---
## Summary by cubic
Prevents the simple judge from corrupting structured output by storing its note separately. Keeps extracted_content valid JSON when output_model_schema is set, while preserving plain-text behavior.

- **Bug Fixes**
  - Added judge_note to ActionResult; judge notes are always stored there.
  - Structured output: do not append judge text to extracted_content; JSON remains parseable.
  - Plain-text mode: still append judge note to extracted_content for backward compatibility.
  - Refactored _run_simple_judge and added regression tests for reject/approve/skip and sandbox parsing.

<sup>Written for commit 582d644eacbf8adbdb410a0f8b33d025441f76c6. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/agent/service.py
+++ b/browser_use/agent/service.py
@@ -1533,16 +1533,14 @@ class Agent(Generic[Context, AgentStructuredOutput]):
 				self.logger.info(f'⚠️  Simple judge overriding success to failure: {reason}')
 				last_result.success = False
 				note = f'[Simple judge: {reason}]'
-				# When structured output is expected, don't append judge text to extracted_content
-				# as it would corrupt the JSON and break end-user parsers
-				if self.output_model_schema is not None:
-					if last_result.metadata is None:
-						last_result.metadata = {}
-					last_result.metadata['simple_judge'] = note
-				elif last_result.extracted_content:
-					last_result.extracted_content += f'\n\n{note}'
-				else:
-					last_result.extracted_content = note
+				# Store judge note separately so structured output JSON stays parseable
+				last_result.judge_note = note
+				# For plain-text mode, also append to extracted_content (preserves existing behavior)
+				if self.output_model_schema is None:
+					if last_result.extracted_content:
+						last_result.extracted_content += f'\n\n{note}'
+					else:
+						last_result.extracted_content = note
 		except Exception as e:
 			self.logger.warning(f'Simple judge failed with error: {e}')
 			# Don't override on error — keep the agent's self-report
--- a/browser_use/agent/views.py
+++ b/browser_use/agent/views.py
@@ -337,6 +337,9 @@ class ActionResult(BaseModel):
 	extracted_content: str | None = None
 	include_extracted_content_only_once: bool = False  # Whether the extracted content should be used to update the read_state
 
+	# Simple judge override note — stored separately so extracted_content stays pure (e.g. JSON for structured output)
+	judge_note: str | None = None
+
 	# Metadata for observability (e.g., click coordinates)
 	metadata: dict | None = None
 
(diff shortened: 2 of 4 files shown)


sauravpanda:
We have removed this judge so no need for this any more

PR #4128, opened 2026-02-20
fix: handle Japanese and non-ASCII filenames in downloads

### LINKED ISSUE #4094: Bug: Japanese file downloads are not functioning properly.
### Browser Use Version

0.11.8

As specified in RFC6266, when determining the filename from `content_disposition`, it is necessary to support not only `filename=` but also `filename*=UTF-8''`.
Files are always created in `<download_path>/download`, preventing the download of multiple files.

### Failing Python Code
https://github.com/browser-use/browser-use/blob/8bdcd3f839618dec34ac4fba6055b5014fce686f/browser_use/browser/watchdogs/downloads_watchdog.py#L531-L537
```python
# Extract filename from Content-Disposition if available
suggested_filename = None
if 'filename=' in content_disposition:
	# Parse filename from Content-Disposition header
	import re

	filename_match = re.search(r'filename[^;=\n]*=(([\'"]).*?\2|[^;\n]*)', content_disposition)
	if filename_match:
		suggested_filename = filename_match.group(1).strip('\'"')
```

### LLM Model

Not dependent on LLM models

Chrome

hztBUAA:
## Summary
- File downloads with Japanese (non-ASCII) filenames were failing because the `Content-Disposition` header parsing only handled `filename=` but not `filename*=UTF-8''...` (RFC 5987/RFC 6266)
- The entire header was lowercased before filename extraction, which could corrupt encoded values
- Added support for `filename*=` parameter which properly decodes percent-encoded UTF-8 filenames
- Preserved original header casing when extracting filenames, using lowercased version only for the `attachment` check

Closes #4094

## Test plan
- Verified regex correctly parses `filename*=UTF-8''%E3%83%86%E3%82%B9%E3%83%88.pdf` -> `テスト.pdf`
- Verified backward compatibility with regular `filename="report.pdf"` and `filename=report.pdf`
- Verified `filename*=` takes priority over `filename=` when both present (per RFC 6266)
- Verified Chinese and other non-ASCII filenames also decode correctly
- Syntax and linting checks pass

---
## Summary by cubic
Fixes failed downloads for Japanese and other non-ASCII filenames by correctly parsing the Content-Disposition header. Adds support for filename*=UTF-8''... and avoids corrupting encoded values.

- **Bug Fixes**
  - Parse filename*= and decode percent-encoded UTF-8; takes priority over filename=.
  - Preserve raw header casing for filename extraction; only lowercase for the attachment check.
  - Maintain compatibility with filename="report.pdf" and filename=report.pdf.

<sup>Written for commit 9d358defd69a73db872db03ef47cd0e2c88c0f0d. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/browser/watchdogs/downloads_watchdog.py
+++ b/browser_use/browser/watchdogs/downloads_watchdog.py
@@ -504,7 +504,8 @@ class DownloadsWatchdog(BaseWatchdog):
 						is_pdf = 'application/pdf' in content_type
 
 						# Check if it's marked as download via Content-Disposition header
-						content_disposition = str(headers.get('content-disposition', '')).lower()
+						content_disposition_raw = str(headers.get('content-disposition', ''))
+						content_disposition = content_disposition_raw.lower()
 						is_download_attachment = 'attachment' in content_disposition
 
 						# Filter out image/video/audio files even if marked as attachment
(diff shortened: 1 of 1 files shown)


hztBUAA:
Thanks for the review and feedback. I am following up on this PR now and will either push the requested changes or reply point-by-point shortly.

hztBUAA:
Quick follow-up: I am reviewing the feedback and will update this PR shortly.

Issue #4133, opened 2026-02-20
Bug: Agent focus can switch to Chrome extension (chrome-extension://...) targets


alexander-tempest:
### Browser Use Version

0.11.9

### Bug Description, Steps to Reproduce, Screenshots

When browser-use connects to an existing Chrome browser via CDP that has extensions with side panels (e.g., Chrome's sidePanel API), the side panel's page is reported by CDP as a target with type: "page" and a chrome-extension:// URL.

The core method SessionManager.get_all_page_targets() only filters by target type (page/tab) but not by URL scheme, so extension side panels pass through as regular page targets. This causes three problems:

1. Crash/detach recovery (_recover_agent_focus) picks page_targets[-1] — if the side panel was opened most recently, the agent switches focus to it and starts navigating inside the extension window.
2. Tab list (get_tabs) includes the side panel in "Available tabs" shown to the LLM, which may cause the LLM to switch to it.
3. Initial connect (connect) may select the side panel as page_targets_from_manager[0] if it's the first discovered page target.

The issue is intermittent because it depends on the timing of the side panel opening relative to target discovery and crash recovery events.

Note: BrowserSession._is_valid_target() already has correct filtering for chrome-extension:// URLs (include_chrome_extensions=False by default), but this method is only used in _cdp_get_all_pages() and get_all_frames() — not in get_all_page_targets(), which is the method used by all the affected code paths listed above.

Suggested fix:

```
  --- a/browser_use/browser/session_manager.py
  +++ b/browser_use/browser/session_manager.py
  @@ -155,7 +155,7 @@ class SessionManager:
                """
                page_targets = []
                for target in self._targets.values():
  -                     if target.target_type in ('page', 'tab'):
  +                     if target.target_type in ('page', 'tab') and not target.url.startswith('chrome-extension://'):
                                page_targets.append(target)
                return page_targets
```

### Failing Python Code

```python

```

### Full DEBUG Log Output

```shell

```

IgorGanapolsky:
**Reliability note**

Production automation failures usually reduce to a **contract** issue before a model/prompt issue:

1. **Success criteria** — durable artifact id (message/booking/row), not only node green
2. **Side-effect before complete** — never blind-retry without an idempotency key on the business object
3. **Tool vs path split** — external action can succeed while the node/agent path still errors (silent-green / false-red)

If you have one failing execution + expected vs actual, reply with stack/version and I will send a free root-cause checklist.

PR #4236, opened 2026-02-28
fix: filter chrome-extension:// targets from get_all_page_targets

giulio-leone:
## Problem

When connecting to an existing Chrome browser via CDP that has extensions with side panels (e.g., Chrome's sidePanel API), the side panel page is reported as a target with `type: "page"` and a `chrome-extension://` URL.

`SessionManager.get_all_page_targets()` only filters by target type (`page`/`tab`) but not by URL scheme, so extension side panels pass through. This causes:

1. **Crash recovery** (`_recover_agent_focus`) picks `page_targets[-1]` — if the side panel was opened most recently, the agent switches focus to the extension
2. **Tab listing** (`get_tabs`) includes the side panel, which may cause the LLM to switch to it
3. **Initial connect** may select the side panel as the first page target

## Root Cause

`BrowserSession._is_valid_target()` already correctly filters `chrome-extension://` URLs (`include_chrome_extensions=False` by default), but this method is only used in `_cdp_get_all_pages()` and `get_all_frames()` — not in `get_all_page_targets()`.

## Fix

Add URL scheme check in `get_all_page_targets()` to exclude `chrome-extension://` URLs, consistent with the existing filtering in `_is_valid_target()`.

## Test

42 CI tests pass, 2 consecutive clean runs.

Fixes #4133

---
## Summary by cubic
Filters out chrome-extension:// targets in get_all_page_targets so extension side panels aren’t treated as pages, preventing focus switches during crash recovery, tab listing, and initial connect. Fixes #4133.

<sup>Written for commit 98c82eca3f3a2de6632f36b860b48801fcfd1328. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/browser/session_manager.py
+++ b/browser_use/browser/session_manager.py
@@ -155,7 +155,7 @@ class SessionManager:
 		"""
 		page_targets = []
 		for target in self._targets.values():
-			if target.target_type in ('page', 'tab'):
+			if target.target_type in ('page', 'tab') and not target.url.startswith('chrome-extension://'):
 				page_targets.append(target)
 		return page_targets
 


giulio-leone:
Closing to focus on the most impactful fixes. Happy to reopen if needed.

PR #4238, opened 2026-02-28
fix: support RFC 6266 filename* in Content-Disposition for non-ASCII filenames

### LINKED ISSUE #4094: Bug: Japanese file downloads are not functioning properly.

giulio-leone:
## Problem

The downloads watchdog only parses the plain `filename=` parameter from `Content-Disposition` headers. This fails for non-ASCII filenames (e.g., Japanese characters like `テスト.pdf`) because the plain `filename=` parameter cannot represent them correctly.

Per RFC 5987/6266, servers encode non-ASCII filenames using `filename*=UTF-8''%encoded_name`.

Example header:
```
Content-Disposition: attachment; filename*=UTF-8''%E3%83%86%E3%82%B9%E3%83%88.pdf
```

## Fix

- Parse `filename*` first (takes precedence per RFC 6266 §4.3)
- URL-decode the percent-encoded value via `urllib.parse.unquote`
- Fall back to plain `filename=` if `filename*` is not present

This correctly handles:
- Japanese, Chinese, Korean, and other non-ASCII filenames
- Mixed headers with both `filename` and `filename*` (star takes precedence)
- Plain ASCII filenames (unchanged behavior)

Fixes #4094

---
## Summary by cubic
Add support for RFC 6266 filename* in Content-Disposition so the downloads watchdog correctly names files with non-ASCII characters; fixes #4094. Falls back to filename= when filename* is absent; ASCII behavior is unchanged.

- **Bug Fixes**
  - Parse filename* first and percent-decode the UTF-8 value.
  - Fall back to filename=; filename* takes precedence when both are present.

<sup>Written for commit e41e1c1ccf5b04de0d8276342c0ab32b6ab1a241. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/browser/watchdogs/downloads_watchdog.py
+++ b/browser_use/browser/watchdogs/downloads_watchdog.py
@@ -574,13 +574,23 @@ class DownloadsWatchdog(BaseWatchdog):
 
 						# Extract filename from Content-Disposition if available
 						suggested_filename = None
-						if 'filename=' in content_disposition:
+						if 'filename' in content_disposition:
 							# Parse filename from Content-Disposition header
 							import re
+							from urllib.parse import unquote
 
-							filename_match = re.search(r'filename[^;=\n]*=(([\'"]).*?\2|[^;\n]*)', content_disposition)
-							if filename_match:
-								suggested_filename = filename_match.group(1).strip('\'"')
+							# RFC 5987/6266: filename*=UTF-8''encoded_name takes precedence
+							filename_star_match = re.search(
+								r"filename\*\s*=\s*(?:UTF-8|utf-8)''(.+?)(?:;|$)",
+								content_disposition,
+							)
+							if filename_star_match:
+								suggested_filename = unquote(filename_star_match.group(1).strip())
+							else:
+								# Fall back to plain filename= parameter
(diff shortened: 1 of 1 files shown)


giulio-leone:
Closing to focus on the most impactful fixes. Happy to reopen if needed.

PR #4259, opened 2026-03-02
fix: support storage_state as dict, not just file path - Fixes #4257

### LINKED ISSUE #4257: Bug: browser session  storage_state incoming object does not take effect
### Browser Use Version

0.12.0

The storage_state incoming object does not take effect

```
browser = Browser(
        cdp_url=cdp_url,  # Get a CDP URL from any provider
        browser_profile=profile,
        keep_alive=True,
        storage_state={
            "cookies": [
                {
                    "name": "sessionid",
                    "value": "xx",
                    "domain": "xxx",
                    "path": "/",
                    "expires": 1777603376.257413,
                    "httpOnly": True,
                    "secure": False,
                    "sameSite": "Lax",
                },
            ]
        },
    )
```

### Failing Python Code

```python

```

### LLM Model

_No response_

_No response_

### Full DEBUG Log Output

```shell

```

rookweb1:
## Summary

Fixes issue #4257 - storage_state incoming object does not take effect.

## Problem

When users pass `storage_state` as a **dictionary** (e.g., `Browser(storage_state={"cookies": [...]})`), the code was treating it as a file path and calling `os.path.exists()` which failed because it was looking for a file named `{"cookies": [...]}` instead of using the dict directly.

## Solution

Modified `_load_storage_state()` method in `storage_state_watchdog.py` to:
1. First check if `storage_state` is a dict
2. If it is a dict, use it directly instead of treating it as a file path
3. If it is a string/path, load from file as before

This matches how `_save_storage_state()` already handles dict storage_state.

## Changes

- `browser_use/browser/watchdogs/storage_state_watchdog.py`: Added dict check before file path lookup

## Testing

The fix allows users to pass storage_state as a dict:
```python
browser = Browser(
    storage_state={
        "cookies": [
            {
                "name": "sessionid",
                "value": "xx",
                "domain": "xxx",
                "path": "/",
                "expires": 1777603376.257413,
                "httpOnly": True,
                "secure": False,
                "sameSite": "Lax",
            },
        ]
    },
)
```

---

AI-assisted contribution (built with assistance)

---
## Summary by cubic
Allow storage_state to be a dict and use it directly instead of treating it as a file path. Fixes #4257 and enables initializing Browser with in-memory cookies.

- **Bug Fixes**
  - Handle dict input first; only read from disk when a valid path exists.
  - Add error logging on file read failures and no-op on invalid paths.

<sup>Written for commit 24ffa1f229fbe99b065b301a453c406883a47e4e. Summary will update on new commits.</sup>


rookweb1 (commented):
Fixed! Moved `import anyio` to the top of the file. Thanks for the review!

PR #4287, opened 2026-03-07
fix: remove false-positive PID-1 substring matching in is_running_in_docker

### LINKED ISSUE #4149: Bug: `is_running_in_docker()` produces false positives — matches `'py'`, `'uv'`, `'app'` substrings in PID 1 command
## Description

In `browser_use/config.py` (lines ~29-33), the Docker detection heuristic uses extremely broad substring matching:

```python
# if init proc (PID 1) looks like uvicorn/python/uv/etc. then we're in Docker
init_cmd = ' '.join(psutil.Process(1).cmdline())
if ('py' in init_cmd) or ('uv' in init_cmd) or ('app' in init_cmd):
    return True
```

## Problem

These substring checks are too broad:

| Substring | Intended match | False positive examples |
|-----------|---------------|----------------------|
| `'py'` | python, pytest | `/usr/lib/systemd/systemd --system` paths containing `py`, `copy`, `display` |
| `'uv'` | uvicorn, uv | Any command with `uv` substring |
| `'app'` | app.py, application | `launchd` (macOS PID 1), `/snap/...`, any path with `/Applications/` |

On **macOS**, PID 1 is `launchd` which does NOT contain these substrings, so macOS is likely safe. But on **bare-metal Linux servers** running Python-based services as PID 1 (systemd with paths containing `py`, or uvicorn running as the main process in a VM), this function returns `True` even though the system is not Docker.

## Impact

Since the function is decorated with `@cache`, once it returns `True`, the entire process permanently thinks it's in Docker. This causes Chrome to launch with container-specific flags:
- `--disable-dev-shm-usage`
- `--no-sandbox`
- `--disable-gpu`

On a non-Docker system, `--no-sandbox` weakens Chrome's security sandbox unnecessarily.

## Proposed Fix

Use more specific detection methods:

```python
def is_running_in_docker() -> bool:
    """Detect if running inside a Docker container."""
    # Method 1: Check for .dockerenv file (most reliable)
    if Path('/.dockerenv').exists():
        return True
    
    # Method 2: Check cgroup for docker/containerd references
    try:
        with open('/proc/1/cgroup', 'r') as f:
            content = f.read()
            if 'docker' in content or 'containerd' in content:
                return True
    except (FileNotFoundError, PermissionError):
        pass
    
    # Method 3: Check for container environment variables
    if os.environ.get('container') or os.environ.get('DOCKER_CONTAINER'):
        return True
    
    return False
```

This approach:
- Checks `/.dockerenv` (Docker-specific file, very reliable)
- Checks `/proc/1/cgroup` for Docker/containerd references (already partially done in the existing code)
- Checks standard container environment variables
- Avoids fragile PID 1 command substring matching entirely

## Willingness to Contribute

Happy to submit a PR with the improved detection logic.

kadaliao:
## Problem

Fixes #4149.

`is_running_in_docker()` used overly broad substring matching on the PID 1 command line:

```python
init_cmd = ' '.join(psutil.Process(1).cmdline())
if ('py' in init_cmd) or ('uv' in init_cmd) or ('app' in init_cmd):
    return True
```

These substrings produce false positives on bare-metal Linux systems:
- `'py'` matches paths containing `copy`, `display`, `python` in non-container init systems
- `'uv'` matches any command with `uv` as a substring
- `'app'` matches `launchd` paths on macOS, `/Applications/`, snap paths, etc.

Since the function is decorated with `@cache`, a single false positive permanently makes the process think it's in Docker, causing Chrome to launch with `--no-sandbox` unnecessarily on non-container systems.

## Fix

Remove the PID-1 substring matching block entirely. The remaining detection methods are already reliable:

1. `/.dockerenv` file existence (Docker-specific, very reliable)
2. `/proc/1/cgroup` containing `'docker'` (already present)
3. `KUBERNETES_SERVICE_HOST` / `container` environment variables (added — standard for K8s and systemd-nspawn)
4. `len(psutil.pids()) < 10` — few processes = almost certainly a container

## Changes

- Removed the `psutil.Process(1).cmdline()` substring check
- Added `KUBERNETES_SERVICE_HOST` and `container` env var checks
- Cleaned up docstring

---
## Summary by cubic
Fixes false container detection in `is_running_in_docker` to prevent unnecessary `--no-sandbox` Chrome launches. Clarifies detection signals for Docker, Kubernetes, and nspawn/LXC.

- **Bug Fixes**
  - Removed PID 1 cmdline substring heuristic that caused cached false positives.
  - Added env checks for `KUBERNETES_SERVICE_HOST` and `container`; kept reliable signals (`/.dockerenv`, `/proc/1/cgroup` contains 'docker', low process count) and expanded docstring to clarify scope and `@cache` behavior.

<sup>Written for commit 1ca95416f149be741bc531ae38cee318f0937547. Summary will update on new commits.</sup>


alvinttang (commented):
Good removal of a genuinely dangerous heuristic. The substring checks `'py' in init_cmd`, `'uv' in init_cmd`, and `'app' in init_cmd` were far too broad — `'py'` matches `python` but also any path containing `py`, and `'app'` matches countless non-container init processes, creating false positives that could silently toggle Chrome launch flags on bare-metal Linux hosts. Replacing them with explicit env var checks is more precise. One concern: the `container` env var is set by `systemd-nspawn` but is not a standard Docker variable — actual Docker containers won't have it unless explicitly set, so this check may be less reliable than it looks. The `/.dockerenv` and cgroup checks at the top remain the most reliable signals and are still in place, which is reassuring; it might be worth adding a comment noting the `container` var is `systemd-nspawn`-specific so future contributors don't assume broader coverage.

kadaliao:
Thanks for the thoughtful review, @alvinttang! You raised a great point about the scope and reliability of the `container` env var check.

I've just pushed an update that addresses your concern:

**Changes made:**
- Split the env var checks into separate, clear statements for `KUBERNETES_SERVICE_HOST` and `container`
- Added detailed docstring explaining all detection methods and their reliability
- **Key clarification:** Added clear comments noting that the `container` env var is systemd-nspawn/LXC-specific and that standard Docker containers won't have it unless explicitly set
- Enhanced process count heuristic comment to clarify the context (bare-metal systems typically have many more processes)

This way, future maintainers will have proper context about the detection scope, and the code documents its own limitations. The reliable signals (/.dockerenv, cgroup with "docker", process count) remain intact as the foundation, while the env var checks provide broader coverage for non-Docker runtimes.

What do you think of this approach?

PR #4294, opened 2026-03-07
fix: include cache-read tokens in Anthropic total_tokens

jnMetaCode:
## Problem

`_get_usage` computes `prompt_tokens` as `input_tokens + cache_read_input_tokens`, correctly accounting for cached prompt tokens. However, `total_tokens` is computed as `input_tokens + output_tokens`, which excludes the cached tokens.

This means `total_tokens < prompt_tokens + completion_tokens` whenever prompt caching is active, which breaks any downstream logic that relies on the invariant `total_tokens == prompt_tokens + completion_tokens` (e.g. cost tracking, rate-limit budgeting).

## Fix

Include `cache_read_input_tokens` in the `total_tokens` calculation, matching the existing `prompt_tokens` formula.

## Before / After

---
## Summary by cubic
Fix Anthropic usage accounting by adding cache_read_input_tokens to total_tokens in _get_usage. This keeps total_tokens equal to prompt_tokens + completion_tokens when prompt caching is used, ensuring accurate cost and rate-limit tracking.

<sup>Written for commit bc659091c7d22a769a078828948fec37a38c65ee. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/llm/anthropic/chat.py
+++ b/browser_use/llm/anthropic/chat.py
@@ -119,7 +119,7 @@ class ChatAnthropic(BaseChatModel):
 				response.usage.cache_read_input_tokens or 0
 			),  # Total tokens in Anthropic are a bit fucked, you have to add cached tokens to the prompt tokens
 			completion_tokens=response.usage.output_tokens,
-			total_tokens=response.usage.input_tokens + response.usage.output_tokens,
+			total_tokens=response.usage.input_tokens + (response.usage.cache_read_input_tokens or 0) + response.usage.output_tokens,
 			prompt_cached_tokens=response.usage.cache_read_input_tokens,
 			prompt_cache_creation_tokens=response.usage.cache_creation_input_tokens,
 			prompt_image_tokens=None,


alvinttang (commented):
The fix is correct: `cache_read_input_tokens` are billed by Anthropic but excluded from `input_tokens` in the API response, so the old formula undercounted total usage. Note that `prompt_tokens` already adds `cache_read_input_tokens` (per the existing comment on that line), so after this change `prompt_tokens` and `total_tokens` are now consistent with each other. One edge case worth considering: `cache_creation_input_tokens` (tokens written into the cache) are also billed at a higher rate — depending on how consumers use this `total_tokens` field for billing or budget tracking, it may be worth including those too for a fully accurate picture.

jnMetaCode:
Thanks @alvinttang! Good catch on `cache_creation_input_tokens` — you're right that those are billed at a higher rate. That said, they're already included in `input_tokens` by the Anthropic API (unlike `cache_read_input_tokens` which is excluded), so the current fix should be sufficient for accurate total counts. Worth documenting though if it comes up.

PR #4303, opened 2026-03-08
fix: relax openai dependency pin

### LINKED ISSUE #4285: Bug: ... Browser Use 0.12.1 has a strict pin for OpenAI==2.16.0
### Browser Use Version

0.12.1

1. Install Browser Use while your pyproject.toml has a minimum openai version of 2.26 allowable
2. Browser Use will throw the following: 

  ╰─▶ Because only browser-use<=0.12.1 is available and browser-use==0.12.1
      depends on openai==2.16.0, we can conclude that browser-use>=0.12.1
      depends on openai==2.16.0.
      And because your project depends on browser-use>=0.12.1 and
      openai>=2.26,<3.dev0, we can conclude that your project's requirements
      are unsatisfiable.

      hint: `browser-use` was requested with a pre-release marker (e.g.,
      browser-use>0.12.1,<0.13.dev0), but pre-releases weren't enabled (try:
      `--prerelease=allow`)

### Failing Python Code

```python
uv sync
```

### LLM Model

_No response_

_No response_

### Full DEBUG Log Output

```shell

```

tsubasakong:
## Summary
- relax the `openai` dependency from `==2.16.0` to `>=2.16.0,<3`
- keep the known-working lower bound while allowing newer 2.x releases

Fixes #4285

## Validation
- `python3.11` TOML parse/assert for the updated dependency range
- `git diff --check`

---
## Summary by cubic
Relaxed the `openai` dependency to `>=2.16.0,<3` to allow 2.x upgrades while keeping the known-good minimum. Added a guard in `examples/custom-functions/cua.py` to raise when a CUA computer call returns no action, fixing #4285.

<sup>Written for commit fe0b6be680f6e65d41b55a8209eab94dedd7405b. Summary will update on new commits.</sup>


sauravpanda:
We have seen downstream issues and decided to pin all the versions, happy to bump it up for you!

matt-greathouse:
@sauravpanda Makes sense to pin aggressively given the downstream breakages. Would you consider bumping browser-use’s openai requirement to align with LiteLLM’s current openai==2.24.0? The mismatch is awkward for projects that implement both browser-use and LiteLLM, because they end up carrying incompatible expectations around the same dependency.

PR #4305, opened 2026-03-08
fix: remove incorrect await on synchronous get_llm()

### LINKED ISSUE #4254: CLI: browser-use run fails with TypeError: object NoneType can't be used in 'await' expression (0.12.0, macOS)
### Browser Use Version

0.12.0 (installed via `uv tool install browser-use`)

### Bug Description

`browser-use run` fails immediately with:

`TypeError: object NoneType can't be used in 'await' expression`

This happens in local **real browser** mode. Other CLI commands like `open` / `state` work fine.

### Steps to Reproduce

```bash
browser-use close --all
browser-use --session bug --browser real --profile "Default" open https://example.com
browser-use --session bug --browser real --profile "Default" run "Go to https://example.com and tell me the page title." --max-steps 5
```

### Actual Output

```text
mode: real
url: https://example.com
mode: real
success: False
error: object NoneType can't be used in 'await' expression
task: Go to https://example.com and tell me the page title.
```

### Expected Behavior

The agent should run the task.

If a local LLM API key is required for `run`, I’d expect a clear error stating which key/env var is missing (instead of a `NoneType` await error).

### Environment

- macOS 26.2 (Darwin 25.2.0) arm64
- Google Chrome 145.0.7632.117
- `browser-use python` reports Python 3.11.14
- `browser-use doctor` shows remote `api_key` not configured (not using remote mode)

Bortlesboat:
## Summary
- `get_llm()` is a synchronous function but was called with `await` in `_handle_local_task()`
- When no API key is configured, `get_llm()` returns `None`, and `await None` raises `TypeError` before the helpful error message can be shown
- Removed the incorrect `await` keyword so the `if llm is None` check works as intended

Fixes #4254

---
## Summary by cubic
Removed the incorrect await on synchronous `get_llm()` in `_handle_local_task()`, preventing a `TypeError` when it returns `None` and showing the intended missing-API-key message instead. Fixes #4254.

<sup>Written for commit 83ad411128127fa940fa26f321268b6c18594495. Summary will update on new commits.</sup>


sauravpanda:
@Bortlesboat can you sign CLA so that I can merge it. Also thanks for pointing this out!

Bortlesboat:
CLA Signed! Thank you for reviewing!

PR #4316, opened 2026-03-10
fix: initialize popup watchdogs for window.open tabs

### LINKED ISSUE #4208: Bug: window.open() causes the previous page to freeze and prevents normal navigation to the new page
### Browser Use Version

0.11.13

The new window remains on about:blank — window.open() created the page but did not navigate to the target URL.
The old page has a gray overlay—the original page may be waiting for the pop-up to complete.
Yellow gibberish — I don't know what it means.

Image

### Failing Python Code

```python
import asyncio

async def main():
    from browser_use import Browser

    browser = Browser(headless=False)
    await browser.start()

    # Create a page and navigate to a known site
    page = await browser.new_page("https://www.baidu.com")
    await asyncio.sleep(1)

    result = await page.evaluate("(url) => window.open(url, '_blank')", "https://www.qq.com/")

    print(f"[2] window.open result: {result}")

    await asyncio.sleep(5)
    await browser.stop()


if __name__ == "__main__":
    asyncio.run(main())
```

### LLM Model

_No response_

_No response_

### Full DEBUG Log Output

```shell

```

tsubasakong:
## Summary

Closes #4208 — dispatch `TabCreatedEvent` for popup/page targets attached through `Target.attachedToTarget`.

### Root cause

When `window.open()` creates a popup, Chrome reports it via `Target.attachedToTarget`. `SessionManager._handle_target_attached()` was enabling monitoring and creating the CDP session, but it never emitted `TabCreatedEvent` for that externally-created tab.

Without that event, the popup-side watchdog flow never initialized correctly. In particular, an empty initial popup URL could also trip `SecurityWatchdog`, so the emitted event now normalizes empty URLs to `about:blank`.

### Fix

- dispatch `TabCreatedEvent` from `_handle_target_attached()` for `page` / `tab` targets after `_enable_page_monitoring()`
- normalize empty attached-target URLs with `target_info.get('url') or 'about:blank'`
- add focused regression coverage for:
  - popup tab event dispatch
  - non-empty popup event URLs
  - opener page responsiveness after `window.open()`

### Validation

- `python -m py_compile tests/ci/browser/test_popup_tab_created_event.py`
- `ruff format --check browser_use/browser/session_manager.py tests/ci/browser/test_popup_tab_created_event.py`
- `git diff --check`

---
## Summary by cubic
Fixes missing TabCreatedEvent for window.open() popups so watchdogs start and the opener stays responsive. Normalizes empty popup URLs to about:blank. Closes #4208.

- **Bug Fixes**
  - Emit TabCreatedEvent from SessionManager._handle_target_attached() after _enable_page_monitoring() for page/tab targets, including window.open() popups.
  - Normalize event URL to target_info.get('url') or 'about:blank' to avoid empty URLs that break security checks.
  - Add tests for popup event dispatch, non-empty URLs, and opener responsiveness.

<sup>Written for commit 4360c05058be4ef9ed975a202ccda82999ae74b8. Summary will update on new commits.</sup>


RewFa-CEO:
This fix addresses a critical edge case in popup handling. The normalization of empty URLs to `about:blank` is a good defensive programming practice.

A suggestion for future improvement: consider adding a configuration option for the stall detection timeout, as different use cases may require different thresholds. For long-running tasks, users might want to increase the timeout beyond the default 90 seconds.

The regression tests you've added will be valuable for preventing future regressions. 

---
*Reviewed by Financier-Nuri from Unum AI*

PR #4336, opened 2026-03-12
feat: add MiniMax LLM provider with M2.7 as default model

octo-patch:
## Summary
- Add MiniMax as a first-class LLM provider via OpenAI-compatible API
- Support MiniMax-M2.7 (default), MiniMax-M2.7-highspeed, MiniMax-M2.5, and MiniMax-M2.5-highspeed models
- Full ChatMiniMax implementation with text, structured output (JSON), and function calling support

## Changes
- Add `browser_use/llm/minimax/` package with `ChatMiniMax` provider and message serializer
- Register MiniMax in model factory (`models.py`) with `minimax_m2_7`, `minimax_m2_7_highspeed`, `minimax_m2_5`, `minimax_m2_5_highspeed` convenience instances
- Add `ChatMiniMax` to lazy imports in `__init__.py`
- Add `MINIMAX_API_KEY` to `.env.example`
- Update README with MiniMax usage examples

## Why
MiniMax-M2.7 is the latest flagship model with enhanced reasoning and coding capabilities. This enables browser-use users to leverage MiniMax models for browser automation tasks.

## Testing
- Model resolution verified for all MiniMax model variants
- Backward compatibility confirmed (M2.5 models still accessible)
- Integration tested with MiniMax API

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -39,6 +39,7 @@ BROWSER_USE_API_KEY=your_bu_api_key_here
 # DEEPSEEK_API_KEY=
 # GROK_API_KEY=
 # NOVITA_API_KEY=
+# MINIMAX_API_KEY=
 
 # AWS Bedrock Configuration (for AWS Bedrock models)
 # Requires: pip install browser-use[aws]
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ BROWSER_USE_API_KEY=your-key
 from browser_use import Agent, Browser, ChatBrowserUse
 # from browser_use import ChatGoogle  # ChatGoogle(model='gemini-3-flash-preview')
 # from browser_use import ChatAnthropic  # ChatAnthropic(model='claude-sonnet-4-6')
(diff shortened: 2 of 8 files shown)


octo-patch (commented):
Pushed fix in 3c9132b for both P1 issues. Summary of changes:

**1. Tool-enabled calls no longer crash on missing `tool_calls` (was line 152)**

When `tools` are provided but `tool_choice` is not forced (i.e., the model decides whether to call a tool), the model may legitimately respond with plain text instead of `tool_calls`. Previously this raised `ValueError('Expected tool_calls in response but got none')`. Now it returns the text content as a normal `ChatInvokeCompletion[str]`.

**2. Structured-output JSON branch is now reachable (was line 129/177)**

The Function Calling guard was `if tools or (output_format is not None and ...)`, which caught *all* `output_format`-only cases before the JSON output path could execute. Changed to `if tools:` so the three paths are now mutually exclusive:
- `output_format is None and not tools` -> plain text completion
- `tools` (with or without `output_format`) -> function calling
- `output_format` only (no tools) -> `response_format: json_object`

PR #4337, opened 2026-03-12
added better crosstime checking

sauravpanda:
## Summary by cubic
Improve time-sensitive validation and data grounding: the judge now uses the current UTC time, and agents must cross-check final answers against `extract` results to avoid mismatches.

- **Bug Fixes**
  - Judge prompt: compute UTC `current_date` and inject it into system; replace “ignore unexpected dates/times” with “current date/time is … — recent content is real.”
  - System prompts: require cross-checking final answers against `extract`; clarify `browser_state` can differ from extracted content.

<sup>Written for commit ac3089343ded7cd10d4b510a1bb52e5bae449b7b. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/agent/prompts.py
+++ b/browser_use/agent/prompts.py
@@ -253,9 +253,9 @@ class AgentMessagePrompt:
 			has_content_above = pages_above > 0
 			has_content_below = pages_below > 0
 			page_info_text = '<page_info>'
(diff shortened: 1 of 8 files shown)


sauravpanda:
This doesn't improve the agent a lot, so we are closing this.

PR #4359, opened 2026-03-15
fix: address Copilot review feedback for Windows compatibility

passionworkeer:
## Summary

This PR addresses Copilot's review feedback on PRs #4354 and #4357:

### Changes

**tunnel.py (Windows compatibility):**
- Use sys.platform == 'win32' check in _is_process_alive() instead of catching SystemError - consistent with main.py and utils.py patterns
- Add creationflags for Windows process detachment (CREATE_NEW_PROCESS_GROUP | CREATE_NO_WINDOW)

**profile.py (env var naming):**
- Rename env vars to BROWSER_USE_CHROME_PATH and BROWSER_USE_CHROMIUM_PATH - consistent with codebase conventions like BROWSER_USE_DISABLE_EXTENSIONS
- Fix return type to str | None instead of str | Path | None

### Why

Copilot noted that:
1. The original SystemError catch could miss certain Windows edge cases
2. The original env var prefix BROWSERUSE_ didn't match the codebase convention BROWSER_USE_
3. The return type annotation was inaccurate

---
## Summary by cubic
Improve Windows compatibility for tunnel processes and align browser executable env vars with project conventions. This makes process checks more reliable on Windows and simplifies configuration via environment variables.

- **Bug Fixes**
  - Windows: use `sys.platform == 'win32'` and a `ctypes` `OpenProcess` check in `_is_process_alive`.
  - Windows: detach `cloudflared` with `CREATE_NEW_PROCESS_GROUP` and `CREATE_NO_WINDOW`; keep `start_new_session` for Unix only.
  - Browser profile: read `BROWSER_USE_CHROME_PATH` or `BROWSER_USE_CHROMIUM_PATH` for `executable_path` by default; correct type to `str | None`.

- **Migration**
  - Replace any `BROWSERUSE_*` vars with `BROWSER_USE_CHROME_PATH` or `BROWSER_USE_CHROMIUM_PATH`.

<sup>Written for commit 0263c8feaf123e7467d8cf2e411b70f4377b7dc0. Summary will update on new commits.</sup>

PR #4361, opened 2026-03-15
fix: add Windows support for _kill_process in tunnel.py

passionworkeer:
## Summary

Fixes a bug reported by Copilot in PR #4359.

## Problem

On Windows, os.kill with SIGTERM/SIGKILL doesn't work - Windows doesn't support Unix signals. This means tunnels started on Windows cannot be reliably stopped.

## Solution

Added a Windows-specific branch in _kill_process() that uses ctypes:
- OpenProcess with PROCESS_TERMINATE flag
- TerminateProcess to kill the process
- Proper handle cleanup with CloseHandle

This is consistent with the pattern used in rowser_use/skill_cli/utils.py.

---
## Summary by cubic
Enable reliable tunnel shutdown on Windows by adding a Windows-specific path to _kill_process in `browser_use/skill_cli/tunnel.py`. This fixes termination failures caused by unsupported Unix signals on Windows.

- **Bug Fixes**
  - On Windows, use Win32 APIs via `ctypes`: `OpenProcess(PROCESS_TERMINATE)`, `TerminateProcess`, `CloseHandle`.
  - Keep existing Unix behavior using `os.kill(SIGTERM)`.
  - Aligns with the approach in `browser_use/skill_cli/utils.py`.

<sup>Written for commit 5e000fdec4aba75274116b60720be0d00ca8a9ee. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/skill_cli/tunnel.py
+++ b/browser_use/skill_cli/tunnel.py
@@ -155,6 +155,26 @@ def _is_process_alive(pid: int) -> bool:
 
 def _kill_process(pid: int) -> bool:
 	"""Kill a process by PID. Returns True if killed, False if already dead."""
+	import sys
+
+	# On Windows, use ctypes instead of signals
+	if sys.platform == 'win32':
+		try:
+			import ctypes
+			from ctypes import wintypes
+
+			PROCESS_TERMINATE = 0x0001
+			handle = ctypes.windll.kernel32.OpenProcess(PROCESS_TERMINATE, False, pid)
+			if not handle:
+				return False
+			try:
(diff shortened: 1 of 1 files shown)


passionworkeer:
Thanks for the review! You are right about checking the return value. I will update the code to check TerminateProcess return value and return False on failure. For the ctypes argtypes/restype, I will add proper type definitions using ctypes.wintypes. The unused wintypes import will be removed. The comment about _is_process_alive is valid - I will create a follow-up PR to address that.

passionworkeer:
Closing - duplicate, fix covered by #4436

PR #4366, opened 2026-03-15
fix: check _kill_process return value in stop_tunnel

passionworkeer:
## Summary

cubic found: stop_tunnel ignores _kill_process() failure and removes tunnel state even when Windows termination did not succeed, potentially orphaning a live process.

## Changes

- Check _kill_process return value in stop_tunnel
- Return error if process could not be killed instead of continuing with cleanup

---

🤖 Generated by TRIX

---
## Summary by cubic
Improve stop_tunnel to reliably stop the process and perform safe cleanup across platforms. Cleans up files/state when the process is already dead, and returns an error without cleanup if killing the process fails (prevents orphaning a live tunnel).

- **Bug Fixes**
  - Distinguish outcomes: killed (success), already dead (cleanup), failed to kill (no cleanup, return error).
  - Add Windows-specific process checks and kill using OpenProcess/GetExitCodeProcess and WaitForSingleObject for reliable behavior.

<sup>Written for commit 5506b0255f571aa33164888646c168b0e270ba00. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/skill_cli/tunnel.py
+++ b/browser_use/skill_cli/tunnel.py
@@ -145,7 +145,17 @@ def _delete_tunnel_info(port: int) -> None:
 
 
 def _is_process_alive(pid: int) -> bool:
-	"""Check if a process is still running."""
+	"""Check if a process is still running.
+	
+	Uses platform-specific implementation for reliable detection,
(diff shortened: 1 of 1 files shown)


passionworkeer:
Closing - duplicate, fix covered by #4436

PR #4387, opened 2026-03-17
fix(utils): add option to disable SignalHandler for host app signal control

### LINKED ISSUE #4385: Bug: SignalHandler overrides host application signal handling — no way to opt out
### Browser Use Version

0.12.2

**Component:** `browser_use/utils.py` — `SignalHandler` class

## Description

When `Agent.run()` executes, it creates and registers a `SignalHandler` that overrides SIGINT handling for the duration of the agent run. There is no way to disable this behaviour.

**On Windows** (line 128-133), the handler calls `os._exit(0)` on the first Ctrl+C, which immediately terminates the process — no cleanup, no atexit handlers, no graceful shutdown.

**On Unix** (line 137-138), the handler uses `loop.add_signal_handler(SIGINT, ...)` which replaces any existing handler. The first Ctrl+C pauses the agent rather than propagating to the host application.

This is fine when browser-use owns the process (standalone CLI), but breaks server embeddings (uvicorn, FastAPI, etc.) where the framework needs to control the signal lifecycle. In our case, uvicorn installs `handle_exit` which sets `should_exit=True` for graceful shutdown (lifespan teardown, connection draining, etc.). The browser-use handler overrides this, making graceful shutdown impossible while an agent is running.

## Impact

Any application embedding browser-use agents inside a server cannot shut down gracefully while an agent is active. On Windows this is especially severe — `os._exit(0)` skips all cleanup including open browser sessions.

## Current workaround

Monkey-patching `SignalHandler.register` and `SignalHandler.unregister` to no-ops.

### Failing Python Code

```python
Here's the offending code from browser_use/utils.py lines 126-135:

# Inside SignalHandler.register()
if self.is_windows:
    # On Windows, use simple signal handling with immediate exit on Ctrl+C
    def windows_handler(sig, frame):
        print('\n\n🛑 Got Ctrl+C. Exiting immediately on Windows...\n', file=stderr)
        # Run the custom exit callback if provided
        if self.custom_exit_callback:
            self.custom_exit_callback()
        os._exit(0)

    self.original_sigint_handler = signal.signal(signal.SIGINT, windows_handler)
```

### LLM Model

_No response_

_No response_

### Full DEBUG Log Output

```shell

```

BillionClaw:
The SignalHandler class registers SIGINT/SIGTERM handlers that override host application signal handling. This breaks server embeddings (uvicorn, FastAPI, etc.) where the framework needs to control the signal lifecycle for graceful shutdown.

Add disabled parameter to SignalHandler class to allow opting out of signal handling. Add enable_signal_handler parameter to Agent class (default True for backward compatibility).

Usage for embedding applications:

Fixes #4385

---
## Summary by cubic
Fixes #4385 by adding an opt-out for internal SIGINT/SIGTERM handling so host apps like `uvicorn`/`FastAPI` can control graceful shutdown. `SignalHandler` now supports `disabled`, and `Agent` adds `enable_signal_handler` (default True).

- **Bug Fixes**
  - `SignalHandler(disabled: bool)` added; when True, `register()`/`unregister()` are no-ops.
  - `Agent(enable_signal_handler: bool = True)` added and passed through as `disabled=not enable_signal_handler`.

- **Migration**
  - For embeddings, use `Agent(..., enable_signal_handler=False)` or `SignalHandler(disabled=True)`.

<sup>Written for commit 4f33289c48d3dccb51380f7f4ec18327577782d9. Summary will update on new commits.</sup>


BillionClaw:
CLA is signed now.

PR #4388, opened 2026-03-17
Fix(DOM): Exclude Password Field Values from DOM Snapshots sent to LLM

zachdotai:
## Why do we need this PR?

When the agent types a password into an `<input type="password">` field (via `sensitive_data` or programmatic `fill()`), the DOM serializer extracts the real value from the browser's accessibility tree and includes it in the text representation of the page sent to the LLM on every subsequent step.

This means the plaintext password is visible in the LLM's context window, where a prompt injection on a later page (e.g. a malicious email body, a crafted form, an ad) can instruct the agent to exfiltrate it.

### How it works

`_build_attributes_string()` in the DOM serializer has three paths that extract field values. None of them check for `type="password"`:

1. **HTML attributes** - raw `value` attribute included if in `DEFAULT_INCLUDE_ATTRIBUTES`
2. **AX properties** - `valuetext` / `value` from the accessibility tree
3. **Form element value extraction** - explicit AX tree lookup for `<input>`, `<textarea>`, `<select>`

After the fix, the serialized DOM for a filled password field looks like:

```diff
- [42] <input type="password" name="password" value="hunter2_super_secret" />
+ [42] <input type="password" name="password" />
```

The LLM still sees it's a password field (so it can interact with it), but never sees the value.

### Why existing mitigations don't fully cover this

- **`_filter_sensitive_data()`** catches values from the `sensitive_data` dict via string replacement, but doesn't protect passwords filled programmatically (e.g. `onepassword.py`'s `fill_field` action where `sensitive_data` isn't set on the Agent)
- **Multi-action batching** can skip the DOM snapshot if `fill` + `click` happen in the same step, but this isn't guaranteed since the LLM controls the batching, multi-action sequences break early on page changes, and login failures re-expose the value on retry

## Demo

```python
# Before fix: serializer output for a filled password field
'type=password name=password value=hunter2_super_secret'

# After fix: value stripped, field metadata preserved
'type=password name=password'
```

## Changes

- `browser_use/dom/serializer/serializer.py` - Detect `type="password"` inputs and skip `value`/`valuetext` from all three extraction paths (HTML attributes, AX properties, form element value lookup)
- `tests/ci/security/test_sensitive_data.py` - 4 new tests: AX tree value stripping, HTML attribute value stripping, non-password input backward compatibility, inputs without explicit type unaffected

## Test plan

- [x] `test_password_field_value_excluded_from_dom_snapshot` — AX tree password values are stripped
- [x] `test_password_field_value_excluded_even_from_html_attributes` — HTML `value` attr is stripped
- [x] `test_text_input_value_preserved` — backward compat: non-password inputs still show values
- [x] `test_password_field_without_type_attribute` — inputs without explicit `type` are not affected
- [x] Full CI suite: 700 passed, 0 failed

---
## Summary by cubic
Prevent password leakage by removing values from `<input type="password">` in DOM snapshots sent to the LLM. Keeps field metadata so the agent can interact, but never includes the secret value.

- **Bug Fixes**
  - Detect `type="password"` and strip values from HTML attributes, AX properties, and form value extraction.
  - Remove any prefilled `value`; keep `type="password"`. Non-password inputs unchanged; inputs without explicit `type` unaffected.
  - Added tests for AX/HTML value stripping and backward compatibility; minor `ruff` lint fixes.

<sup>Written for commit 624c040f6abae1a31d56a690b6c1bc3c32ef75dd. Summary will update on new commits.</sup>


sauravpanda:
Oh thanks for bringing this to our attention. Can you sign the CLA so that I can merge this?

zachdotai:
> Oh thanks for bringing this to our attention. Can you sign the CLA so that I can merge this?

Something seems to be broken with CLA and this is what I see. Let me know what needs to be done from my side.

image

zachdotai:
> Oh thanks for bringing this to our attention. Can you sign the CLA so that I can merge this? 

Done 🫡

PR #4412, opened 2026-03-18
fix: allow opting out of SignalHandler to prevent host app conflicts

### LINKED ISSUE #4385: Bug: SignalHandler overrides host application signal handling — no way to opt out

BillionClaw:
Fixes #4385

## Problem
The SignalHandler class in browser_use/utils.py automatically registers SIGINT/SIGTERM signal handlers when Agent.run() executes. This overrides any existing signal handlers set up by the host application, making graceful shutdown impossible for server embeddings (uvicorn, FastAPI, etc.).

## Solution
Add an `enable_signal_handling` parameter to the Agent class that allows users to opt out of signal handler registration:

- Added `enable_signal_handling: bool = True` to `AgentSettings` (default True for backward compatibility)
- Added `enable_signal_handling` parameter to `Agent.__init__`
- Modified `Agent.run()` to only create and register SignalHandler when `enable_signal_handling=True`

## Usage
For server embeddings that need to control their own signal lifecycle:

```python
agent = Agent(
    task="...",
    enable_signal_handling=False,  # Disable signal handling
)
```

## Changes
- `browser_use/agent/views.py`: Added `enable_signal_handling` field to `AgentSettings`
- `browser_use/agent/service.py`: 
  - Added `enable_signal_handling` parameter to `Agent.__init__`
  - Modified `run()` to conditionally register/unregister signal handlers

---
## Summary by cubic
Adds an opt-out for internal SIGINT/SIGTERM handling so host apps can control signals. Prevents `Agent.run()` from overriding existing handlers in servers like `uvicorn` and `FastAPI`.

- **Bug Fixes**
  - Added `enable_signal_handling` to `AgentSettings` and `Agent.__init__` (default True).
  - `Agent.run()` only registers `SignalHandler` when enabled; unregister is guarded.
  - For server embeddings, set `enable_signal_handling=False` to let the host manage shutdown.

<sup>Written for commit 62970bcd9a8980639ed3f756397a1afaa65842a7. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/agent/service.py
+++ b/browser_use/agent/service.py
@@ -201,6 +201,7 @@ class Agent(Generic[Context, AgentStructuredOutput]):
 		planning_exploration_limit: int = 5,
 		loop_detection_window: int = 20,
 		loop_detection_enabled: bool = True,
+		enable_signal_handling: bool = True,
 		llm_screenshot_size: tuple[int, int] | None = None,
 		message_compaction: MessageCompactionSettings | bool | None = True,
 		max_clickable_elements_length: int = 40000,
@@ -410,6 +411,7 @@ class Agent(Generic[Context, AgentStructuredOutput]):
 			loop_detection_enabled=loop_detection_enabled,
 			message_compaction=message_compaction,
 			max_clickable_elements_length=max_clickable_elements_length,
+			enable_signal_handling=enable_signal_handling,
 		)
 
 		# Token cost service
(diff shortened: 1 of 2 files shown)


BillionClaw:
Friendly bump -- this PR has been open for a while. Happy to make any changes if needed, or close it if no longer relevant.

PR #4357, opened 2026-03-14
fix: add Windows compatibility for tunnel.py process management

passionworkeer:
## Summary

Fixes Windows compatibility issues in tunnel.py for process management APIs.

## Problem

The cloudflared tunnel management uses Unix-only APIs that fail on Windows:
- os.kill(pid, 0) raises SystemError on Windows
- signal.SIGTERM/SIGKILL are not available on Windows
- start_new_session=True is Unix-specific

## Solution

1. _is_process_alive: Added ctypes.windll.kernel32.OpenProcess fallback for Windows
2. _kill_process: Use TerminateProcess on Windows, signals on Unix
3. start_tunnel: Conditional start_new_session based on platform

This follows the same pattern already established in main.py and utils.py.

Closes #4352

---
## Summary by cubic
Adds Windows-compatible process management in `browser_use/skill_cli/tunnel.py` so Cloudflared tunnels start/stop reliably on Windows, with targeted unit tests. Aligns with patterns used in `main.py` and `utils.py`. Closes #4352.

- **Bug Fixes**
  - Reliable liveness checks on Windows via `_setup_windows_ctypes()`, `OpenProcess` with `SYNCHRONIZE`, and `WaitForSingleObject` (with try/except fallback).
  - Kill on Windows using `TerminateProcess` plus a wait loop; return False if still alive after retries; `stop_tunnel` now aborts cleanup if kill fails.
  - Spawn on Windows with `CREATE_NEW_PROCESS_GROUP` and `CREATE_NO_WINDOW`; remove Unix-only `start_new_session` to avoid platform issues.

<sup>Written for commit 3115944b8f3e27ea592473c34b9def28e8978651. Summary will update on new commits.</sup>

Issue #3912, opened 2026-01-22
browser-use Windows Issue - os.kill(pid, 0) - Fails on Windows


ZanderRuss:
### Browser Use Version

 v0.11.4

### LLM Model

Claude Opus 4.5

### Screenshots, Description, and task prompt given to Agent

Environment
OS: Windows 11
Python: 3.12.8
browser-use: v0.11.4
Playwright: (bundled with browser-use)
Issue 1: CLI os.kill() Error
Location: [skill_cli/utils.py:43](vscode-webview://10h5jad2kscdlqvj7btonus954qijha4rlmas3vttutalo1eq7qh/plynth-create/.venv/Lib/site-packages/browser_use/skill_cli/utils.py#L43)

def is_server_running(session: str) -> bool:
    ...
    os.kill(pid, 0)  # ❌ Fails on Windows

Error:

SystemError: <built-in function kill> returned a result with an exception set
Root Cause: os.kill(pid, 0) is Unix-specific. Windows doesn't support signal 0 for process existence checks.

Fix: Use psutil.pid_exists(pid) or ctypes.windll.kernel32.OpenProcess().

Issue 2: Browser Page Operations Hang
Symptoms:

Browser.start() ✅ works
browser.get_current_page() ✅ works
page.goto(url) ❌ hangs indefinitely
page.set_content(html) ❌ hangs indefinitely
Verified Working:

Playwright sync API: ✅
Playwright async API: ✅
browser-use Agent API: (not tested without API key)
Root Cause: Likely event loop / thread issue in browser-use's async page wrapper. The same operations work when using Playwright's async_playwright directly.

Reproduction

# This WORKS (Playwright directly):
from playwright.async_api import async_playwright
async with async_playwright() as p:
    browser = await p.chromium.launch()
    page = await browser.new_page()
    await page.set_content("<h1>Test</h1>")  # ✅ Works

# This HANGS (browser-use):
from browser_use import Browser
browser = Browser(headless=True)
await browser.start()  # ✅ Works
page = await browser.get_current_page()  # ✅ Works
await page.set_content("<h1>Test</h1>")  # ❌ Hangs forever
Workarounds
For Agent-based automation: Use ChatAnthropic with existing API key
For simple browser control: Use Playwright's async API directly
For CLI: Not usable on Windows currently

### HTML around where it's failing

```html
see above
```

### Operating System & Browser Versions

Windows 11 pro

### Python Code Sample

```python
see above
```

[...]

PR #4436, opened 2026-03-21
fix(tunnel): add Windows compatibility for process kill in _kill_process

### LINKED ISSUE #4352: Bug: tunnel.py uses Unix-only os.kill/signals, fails on Windows
### Browser Use Version

Latest (main branch)

### Bug Description

The cloudflared tunnel management in `browser_use/skill_cli/tunnel.py` uses Unix-only process management APIs that fail on Windows:

1. **`_is_process_alive()`** calls `os.kill(pid, 0)` which raises `SystemError` on Windows (signal 0 is not supported)
2. **`_kill_process()`** uses `signal.SIGTERM` and `signal.SIGKILL` which are not available on Windows
3. **`start_tunnel()`** uses `start_new_session=True` (Unix `setsid`) which is not the correct approach on Windows

This means any tunnel-related CLI commands (`browser-use tunnel <port>`, tunnel stop, tunnel list) will crash on Windows.

### Root Cause

The same issue was already fixed in `skill_cli/main.py` and `skill_cli/utils.py` (which use `ctypes.windll.kernel32.OpenProcess` / `TerminateProcess` on Windows), but `tunnel.py` was missed during that fix.

### Steps to Reproduce

On Windows:
```python
from browser_use.skill_cli.tunnel import _is_process_alive
_is_process_alive(1234)  # SystemError: os.kill doesn't support signal 0 on Windows
```

Or run any tunnel CLI command:
```
browser-use tunnel 8080
browser-use tunnel list
browser-use tunnel stop 8080
```

### Expected Behavior

Tunnel commands should work on Windows, using the same ctypes-based fallback pattern already established in `main.py` and `utils.py`.

### Related

This is part of the broader Windows compatibility issue reported in #3912.

Windows 11

passionworkeer:
### Problem

The _kill_process() function in skill_cli/tunnel.py uses os.kill(pid, signal.SIGTERM) and os.kill(pid, signal.SIGKILL) which are Unix-only APIs. On Windows, these calls raise OSError since Windows doesn't support POSIX signals.

This causes rowser-use tunnel stop and related commands to fail on Windows.

### Solution

Add Windows-specific process termination using ctypes (following the same pattern already established in skill_cli/main.py and skill_cli/utils.py):

- Use OpenProcess(PROCESS_TERMINATE, False, pid) to get a handle
- Use TerminateProcess(handle, exit_code) to terminate
- Use CloseHandle(handle) to clean up

### Related

- Fixes #4352
- Related to #3912 (broader Windows compatibility)
- Closes PR #4357 (closed due to conflicts)

---
## Summary by cubic
Make tunnel process termination work on Windows by using Win32 APIs in `_kill_process`. This fixes tunnel stop commands failing on Windows while keeping Unix behavior unchanged.

- **Bug Fixes**
  - On Windows, use `OpenProcess(PROCESS_TERMINATE)` + `TerminateProcess` + `CloseHandle` via `ctypes`.
  - Keep Unix flow: send `SIGTERM`, wait briefly, then `SIGKILL`; move `time` import out of the loop.
  - Clarify behavior and fix NameError: add missing `sys` import; return True only on successful termination, False if already dead or the handle can’t be opened.

<sup>Written for commit b788c2087312827a84d4e973d23e1eee71e0bdf6. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/skill_cli/tunnel.py
+++ b/browser_use/skill_cli/tunnel.py
@@ -18,6 +18,8 @@ import os
 import re
 import shutil
 import signal
+import sys
+import time
 from pathlib import Path
 from typing import Any
 
@@ -156,7 +158,7 @@ def _is_process_alive(pid: int) -> bool:
 
 
 def _kill_process(pid: int) -> bool:
-	"""Kill a process by PID. Returns True if killed, False if already dead."""
+	"""Kill a process by PID. Returns True if killed, False otherwise (not alive or insufficient privileges)."""
 	if sys.platform == 'win32':
 		import ctypes
 		from ctypes import wintypes
@@ -184,8 +186,6 @@ def _kill_process(pid: int) -> bool:
 			for _ in range(10):
 				if not _is_process_alive(pid):
 					return True
-				import time
-
 				time.sleep(0.1)
 			# Force kill if still alive
 			os.kill(pid, signal.SIGKILL)


passionworkeer:
Closing this PR - the co-author (penggaolai) has not signed the CLA. Replaced by PR #4444 which contains the same fix as a solo commit.

Issue #4443, opened 2026-03-21
Fix: Support Chromium profiles on Linux --profile parameter


tocer:
# Browser-use `--profile` parameter doesn't support Chromium on Linux

## Bug Report

**Problem:** The `--profile` parameter fails to find Chromium profiles on Linux because `get_chrome_profile_path()` only checks `~/.config/google-chrome`, ignoring `~/.config/chromium`.

**Environment:**
- OS: Linux (Manjaro)
- Python: 3.14.3
- browser-use: 0.12.2
- Browser: Chromium

## Steps to Reproduce

1. Use Chromium browser (not Google Chrome) installed at `~/.config/chromium`
2. Run: `browser-use --profile "ProfileName" --headed open https://example.com`
3. Expected: Uses the specified Chromium profile
4. Actual: Profile not found or temporary session is created

## Root Cause

`browser_use/skill_cli/utils.py`, function `get_chrome_profile_path()` (line 216)

```python
elif system == 'Linux':
    return str(Path.home() / '.config' / 'google-chrome')  # Only checks Chrome!
```

This hardcoded path ignores:
- `~/.config/chromium` (standard Chromium location)
- Other Chromium-based browsers (Brave, Vivaldi, Opera, etc.)

## Impact

- `--profile` parameter cannot find Chromium profiles
- `browser-use profile list` works (uses profile-use tool) but actual profile usage fails
- Users must create symlink as workaround: `ln -s ~/.config/chromium ~/.config/google-chrome`

## Suggested Fix

Check multiple browser config directories on Linux, ordered by priority:

```python
elif system == 'Linux':
    # Check browser config directories on Linux (by priority)
    home = Path.home()
    candidates = [
        home / '.config' / 'chromium',           # Chromium
        home / '.config' / 'google-chrome',       # Google Chrome
    ]

    for candidate in candidates:
        if candidate.exists():
            return str(candidate)

    # Fallback: return google-chrome path for backward compatibility
    return str(candidates[1])  # google-chrome
```

**Rationale:** Beta/Unstable versions have very few users. They can create symlinks if needed. Keeping the code simple follows YAGNI principle.

## Testing Checklist

- [x] Tested on Manjaro Linux with Chromium
- [x] `--profile` with display name works
- [x] `--profile` with directory name works
- [x] Cookies persist across browser restarts
- [ ] Test on Ubuntu/Debian
- [ ] Test with Google Chrome

## Additional Notes

[...]

PR #4448, opened 2026-03-21
fix: support Chromium profiles on Linux in get_chrome_profile_path()

passionworkeer:
Fixes issue #4443 - On Linux, get_chrome_profile_path() only checked ~/.config/google-chrome when profile is None, ignoring ~/.config/chromium where Chromium browser stores its profiles.

This fix checks ~/.config/chromium first, then falls back to ~/.config/google-chrome, using whichever exists.

---
## Summary by cubic
On Linux, get_chrome_profile_path() now matches the detected Chrome executable to the correct user-data directory using Path.is_dir(), preventing Chrome/Chromium profile mixups. It returns ~/.config/chromium for Chromium and ~/.config/google-chrome for Google Chrome; if Chromium is detected but its dir is missing, it falls back to google-chrome; with no executable, it checks both and prefers chromium when present, then google-chrome, and returns the chromium path when neither exists; Linux tests cover these cases and fix a mock recursion edge case.

<sup>Written for commit 217fe9b0314212c30f4ba2d025bbc180111bb5a7. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/skill_cli/utils.py
+++ b/browser_use/skill_cli/utils.py
@@ -232,13 +232,14 @@ def get_chrome_profile_path(profile: str | None) -> str | None:
 				if chromium_path.is_dir():
 					return str(chromium_path)
 			# Fallback: no executable detected — check which paths actually exist
-			google_chrome_path = Path.home() / '.config' / 'google-chrome'
+			# Prefer chromium over google-chrome to match find_chrome_executable behavior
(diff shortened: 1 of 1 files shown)


laithrw:
Thanks for the PR! I'm closing this in favor of a smaller fix in #4512, which I believe should resolve the problem, but please let me know if it doesn't work for you.

PR #4462, opened 2026-03-22
fix: add delay before first-char check on contenteditable to prevent duplication

### LINKED ISSUE #4461: First character doubled on React contenteditable (leaf-start retype race condition)
## Bug Description

When using `input_text` on React-based contenteditable elements (e.g., X.com reply box, Slack message input), the first character is consistently doubled. Typing `"Hello world"` produces `"HHello world"`.

This happens **every time** on React contenteditable fields when `clear=True` (the default).

## Root Cause

Two methods in `default_action_watchdog.py` interact badly with React's async rendering:

### 1. `_clear_text_field` (line ~1280) — Destructive DOM clear

The contenteditable clear path uses direct DOM manipulation:
```javascript
while (this.firstChild) { this.removeChild(this.firstChild); }
this.textContent = "";
this.innerHTML = "";
```

This destroys React's virtual DOM tree. React needs time to reconcile after this kind of external DOM mutation.

### 2. Leaf-start first-char check (line ~1860) — Race condition

After typing the first character, the code checks `document.activeElement.textContent` to detect if the char was "dropped" by the leaf-start edge case:

```python
if i == 0 and _check_first_char and _first_char:
    check_result = await cdp_session.cdp_client.send.Runtime.evaluate(
        params={'expression': 'document.activeElement.textContent'},
        ...
    )
    content = check_result.get('result', {}).get('value', '')
    if _first_char not in content:
        # Retype the first character
```

**The problem:** After the destructive clear, React hasn't finished reconciling. The first keystroke is dispatched and React queues it, but `textContent` is still empty when the check runs (~5ms later). The check sees an empty string → assumes the char was "dropped" → retypes it. React then processes **both** the original and retyped keystroke → doubled letter.

## Evidence from logs

```
DEBUG: 🧹 Clearing text field using JavaScript value setting
DEBUG: ✅ Text field cleared successfully using contenteditable
DEBUG: 🎯 Typing text character by character: "Hello world"
DEBUG: 🎯 First char "H" was dropped (leaf-start bug), retyping
```

The `"First char was dropped"` message fires **every single time** on React contenteditable fields. The char was never actually dropped — React just hadn't flushed yet.

## Reproduction

```python
from browser_use import Agent, BrowserSession

agent = Agent(
    task='Go to x.com, click the reply box on any post, and type "Hello world"',
    llm=ChatOpenAI(model="gpt-4o"),
)
await agent.run()
# Result: "HHello world" appears in the reply box
```

Any React-based contenteditable will reproduce this: X.com, Slack, Notion, etc. Standard `<input>` and `<textarea>` elements are not affected.

## Workaround

A monkey-patch with three layers fixes this completely:

**A. Keyboard-based clear:** Replace DOM destruction with `Ctrl+A + Backspace` via CDP key events — React processes this as a normal user action.

**B. Disable leaf-start check:** Temporarily spoof `element_node.attributes` so `_is_contenteditable` evaluates to `False`, preventing the first-char retype logic from triggering.

**C. Post-typing safety net:** After all typing completes, read back the text and if the first character is doubled (e.g., `actual[0] == actual[1] == text[0]`), fix it via `document.execCommand('selectAll') + execCommand('insertText', correctText)`.

## Suggested Fix

1. Replace the contenteditable clear strategy with keyboard-based `Ctrl+A + Backspace` (or at minimum, add a longer delay after DOM clear — the current timing is not enough for React reconciliation, 150-200ms would be safer).

2. Add a delay before the first-char check (e.g., `await asyncio.sleep(0.1)`) to let React flush, or remove the leaf-start workaround for contenteditable elements entirely since it causes more harm than it fixes on modern React sites.

## Environment

- browser-use: 0.12.2
- Python: 3.13
- Browser: Chromium 146
- OS: Linux

r266-tech:
### Problem
On React contenteditable elements (X.com, Slack, Notion, etc.), the first character was consistently duplicated when using `input_text` with `clear=True`. For example, typing "Hello" produced "HHello".

### Root Cause
The leaf-start retype check in `default_action_watchdog.py` reads `document.activeElement.textContent` **immediately** after dispatching the first key event. With React, the virtual DOM has not yet flushed to the real DOM at this point, so `textContent` appears empty. The check interprets this as the first character being "dropped" by the leaf-start bug and retypes it. React then processes both keystrokes, resulting in a duplicated first character.

### Fix
Added a 100ms `asyncio.sleep(0.1)` before the first-char check on contenteditable elements. This gives React (and similar frameworks) time to reconcile the virtual DOM with the real DOM before the check runs.

### Scope
- **6 lines added** to a single file (`browser_use/browser/watchdogs/default_action_watchdog.py`)
- Only affects the first character check on contenteditable elements
- No behavioral change for regular `<input>` or `<textarea>` elements
- No behavioral change for contenteditable elements where the char actually *is* dropped (the real leaf-start bug) -- the delay just ensures we only retype when it is genuinely missing

---
## Summary by cubic
Fix duplicate first character in contenteditable editors by switching to a check-first bounded poll and an insertion-local, code-point-aware first-char check. No changes for `<input>` or `<textarea>`; fixes #4461.

- **Bug Fixes**
  - Contenteditable: poll at 0/100/…/500ms using an insertion-local probe (char before caret via value/selection) with early exit; only retype if still missing on the final check, avoiding false matches from host/placeholder text and preserving emoji.

<sup>Written for commit ac84f051bb765a6aa205382b6be7b4e147c37b9e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/4462?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/browser/watchdogs/default_action_watchdog.py
+++ b/browser_use/browser/watchdogs/default_action_watchdog.py
@@ -1934,18 +1934,29 @@ class DefaultActionWatchdog(BaseWatchdog):
 
 				# After first char on contenteditable: check if dropped and retype if needed
 				if i == 0 and _check_first_char and _first_char:
-					# Brief delay so React (and similar frameworks) can flush the
-					# virtual DOM after the key event.  Without this the
-					# textContent check races with the framework's async re-render
-					# and falsely reports the first char as missing, causing a
-					# duplicate first character.  See #4461.
-					await asyncio.sleep(0.1)
-					check_result = await cdp_session.cdp_client.send.Runtime.evaluate(
-						params={'expression': 'document.activeElement.textContent'},
-						session_id=cdp_session.session_id,
-					)
-					content = check_result.get('result', {}).get('value', '')
-					if _first_char not in content:
+					# Poll for the first char to appear instead of paying a fixed delay.
+					# On the happy path (char already visible) this costs a single
+					# textContent read and no sleep. When a framework (React/Vue/Angular)
+					# re-renders asynchronously the char may not be present immediately,
+					# so poll briefly and early-exit as soon as it appears -- a slow
+					# flush on a loaded/slow browser must not be mistaken for a drop.
+					# The grace sleep is always followed by another check (never sleep
+					# then retype on stale state), so a char that flushes inside the
+					# final window is still seen. Only a char missing on the last check
+					# is treated as the leaf-start drop and retyped. See #4461.
+					_first_char_present = False
+					for _first_char_attempt in range(6):  # check at 0,100,...,500ms (5 x 100ms grace)
+						check_result = await cdp_session.cdp_client.send.Runtime.evaluate(
+							params={'expression': 'document.activeElement.textContent'},
+							session_id=cdp_session.session_id,
+						)
+						content = check_result.get('result', {}).get('value', '')
+						if _first_char in content:
+							_first_char_present = True
+							break
+						if _first_char_attempt < 5:  # wait then re-check; no sleep after the final check
+							await asyncio.sleep(0.1)
+					if not _first_char_present:
 						self.logger.debug(f'🎯 First char "{_first_char}" was dropped (leaf-start bug), retyping')
 						# Retype the first character - cursor now past leaf-start
 						modifiers, vk_code, base_key = self._get_char_modifiers_and_vk(_first_char)
--- a/browser_use/browser/watchdogs/default_action_watchdog.py
+++ b/browser_use/browser/watchdogs/default_action_watchdog.py
@@ -1934,6 +1934,12 @@ class DefaultActionWatchdog(BaseWatchdog):
 
 				# After first char on contenteditable: check if dropped and retype if needed
 				if i == 0 and _check_first_char and _first_char:
+					# Brief delay so React (and similar frameworks) can flush the
+					# virtual DOM after the key event.  Without this the
+					# textContent check races with the framework's async re-render
+					# and falsely reports the first char as missing, causing a
+					# duplicate first character.  See #4461.
+					await asyncio.sleep(0.1)
 					check_result = await cdp_session.cdp_client.send.Runtime.evaluate(
 						params={'expression': 'document.activeElement.textContent'},
 						session_id=cdp_session.session_id,
--- a/browser_use/browser/watchdogs/default_action_watchdog.py
+++ b/browser_use/browser/watchdogs/default_action_watchdog.py
@@ -1944,13 +1944,44 @@ class DefaultActionWatchdog(BaseWatchdog):
 					# then retype on stale state), so a char that flushes inside the
 					# final window is still seen. Only a char missing on the last check
 					# is treated as the leaf-start drop and retyped. See #4461.
+					# Detect the first char from the INSERTION-ADJACENT code point only, not
+					# activeElement.textContent over the whole focused host. textContent on a
+					# nested custom control can (a) never surface a visible, in-DOM typed char
+					# (false drop -> unwanted retype -> duplication, the bug a maintainer hit) or
+					# (b) include unrelated placeholder/label/chrome text -- or the same char
+					# elsewhere in the subtree -- that matches even when it truly dropped (false
+					# present -> the leaf-start drop #4461 reappears). So look only at the single
+					# code point immediately before the caret, via the form-field value or the
+					# caret's own text node; Array.from(...).pop() keeps astral chars (emoji)
+					# intact rather than splitting a surrogate pair. Ambiguous cases fall through
+					# to a (harmless) retype. The whole probe is wrapped so a throwing custom
+					# getter degrades to '' not a CDP error.
+					_first_char_probe_js = (
+						'(() => {'
+						' try {'
+						' let s = "";'
+						' const el = document.activeElement;'
+						' if (el && typeof el.value === "string" && typeof el.selectionEnd === "number" && el.selectionEnd > 0) {'
+						' s += Array.from(el.value.slice(0, el.selectionEnd)).pop() || "";'
+						' }'
+						' const sel = document.getSelection && document.getSelection();'
+						' if (sel && sel.anchorNode && sel.anchorNode.nodeType === 3 && sel.anchorOffset > 0) {'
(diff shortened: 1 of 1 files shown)


geniusgeek (approved):
we need this

r266-tech:
CLA is green now ✅ — the fix commit had been authored under an old GitHub account of mine that wasn't covered by the CLA, so I re-attributed it to @r266-tech (who has signed). The code is byte-for-byte unchanged; only the commit authorship differs.

This already has an approving review and is currently mergeable. Could a maintainer take a look at merging? Happy to rebase onto the latest `main` first if you'd prefer. (Also flagging that the `stale` label can be dropped now that it's active again.)

r266-tech:
Force-pushed a cleaned-up version of this branch.

Over time the branch had accumulated a number of unrelated commits (merges from `main` plus other PRs' changes), which is what produced the merge conflict and the oversized multi-file diff. I've rebuilt it on top of the latest `main` as a single commit containing only the actual fix — so the diff is now just the one watchdog file.

The bug is still present on current `main`: in `browser_use/browser/watchdogs/default_action_watchdog.py`, the contenteditable leaf-start retype check (`if i == 0 and _check_first_char and _first_char:`) reads `document.activeElement.textContent` immediately after dispatching the first key event. On React (and similar virtual-DOM frameworks) the re-render hasn't flushed yet, so the check sees an empty string and re-types the first character → duplicated first char (`HHello`). A brief `await asyncio.sleep(0.1)` before the read lets the framework reconcile first.

Marking active again so it doesn't get auto-closed. Happy to adjust the delay value or gate it differently if you'd prefer. Fixes #4461.

MrLeit:
Given that 100ms can also fail - would it not be better to:
1. Run the check.
2. If the leaf-start issue is detected, run a small loop checking if the state clears. Say 5 x 100ms.
3. If it clears in the 100-500ms continue or then retype.

---

1. For websites that don't that the issue, don't pay the tax for typing.
2. Have a longer grace period to cope with very slow browsers (loaded machine or otherwise).

In the end it could still fail but has a higher chance to succeed and mostly would not trigger.

r266-tech:
Good call — agreed on both counts. I've restructured it from the fixed `sleep(0.1)` to the check-first bounded poll you described:

- **Check first, no tax on the happy path.** The `textContent` read runs immediately; if the first char is already there, we continue with zero delay — sites that never hit the leaf-start bug don't pay anything.
- **Bounded grace for slow browsers.** If it's missing, we poll up to 5×100ms, breaking the moment the char appears, so a delayed React/Vue/Angular flush on a loaded machine isn't mistaken for a drop.
- **Retype only on a fresh miss.** The char is re-checked *after* each grace sleep (checks land at 0/100/…/500ms), and there's no sleep-then-retype on stale state — so we only retype when it's still genuinely missing on the final check.

So it's faster on the common path, more forgiving on slow ones, and still retypes a real leaf-start drop exactly once. Pushed in the latest commit — thanks for the review.

MrLeit:
Hm, after digging even deeper into our case I found a subtle but slightly different issue. The website in question is particular old and has nested controls with custom event handling. It seems that due to this document.activeElement.textContent doesn't return the text **ever** even after it 'visually' syncs to the GUI so is definitely in the DOM. I'm not an expert in textContent but it sounds like it has some weaknesses in what it returns.

```
check_result = await cdp_session.cdp_client.send.Runtime.evaluate(
	params={'expression': 'document.activeElement.textContent'},
	session_id=cdp_session.session_id,
)
```

In our case I monkey-patched a long delay in to check and it still always reported failure (although I could visually see the char). The text was never detected indicating that the above check does not always work.

At that point I monkey patched the detection off so I could get unblocked but likely I'm not the only one who will hit this and perhaps the check for text content needs to be more robust as well?

I also had to monkey patch this same method to 'blur()' near the end to ensure the value is always committed but that is another framework novelty I think.

r266-tech:
Thanks — that's a really useful finding, and it exposed a weakness in the check itself, not just the timing.

I've reworked the detection so it no longer relies on `document.activeElement.textContent` over the whole focused host. That read is unreliable in *both* directions on a host like yours:

- **False drop** (your case): on a nested custom control the typed char is in the DOM and visible, but `activeElement.textContent` never surfaces it → the check thinks it dropped → it retypes → duplication.
- **False present**: `textContent`/`innerText` over the host subtree can contain the same character in placeholder/label/chrome text, so a *real* drop gets masked and the leaf-start bug silently comes back.

The new probe looks only at the **single code point immediately before the caret**, from an insertion-local source:

- the form field's `value` at `selectionEnd`, or
- the caret's own text node (`anchorNode`, text nodes only) at `anchorOffset`,

and uses `Array.from(str.slice(0, offset)).pop()` so an astral first char (an emoji, say) is returned whole instead of a lone surrogate half. Anything ambiguous returns `""` and falls through to the existing (harmless) retype, and the whole thing is wrapped so a throwing custom getter degrades to `""` rather than surfacing a CDP error. Pushed in the latest commit.

This should fix your case **if** the typed char lands in a real text node at the caret (the common nested-control situation). If your control is one where the value never appears in the light-DOM text at the caret *at all* — i.e. it's held entirely in framework state — then no post-hoc DOM read can confirm it, and the safe move there is to not let a blind heuristic retype. If this still mis-fires for you after the change, I'm happy to follow up with either (a) an opt-out to skip the leaf-start retype, or (b) a before/after caret-advance snapshot so we only retype when the caret provably didn't move.

On the `blur()`-to-commit point — that's a separate "framework only commits the value on blur" concern rather than the drop itself, so I've kept it out of this PR to stay focused, but happy to look at it as a follow-up if you want to track it.

MrLeit:
@r266-tech - great thanks. Once this lands in a release I'll take it for a spin and see if I can revert my disable of this safety feature.

I also raised https://github.com/browser-use/browser-use/issues/5054. I have a monkey patch for now so not blocking. I also don't know if blur is the best option but at least reliable in our case.

PR #4464, opened 2026-03-22
fix: prevent KeyError in load_from_dict and IndexError in final_result

r266-tech:
### Problems

**1. `AgentHistoryList.load_from_dict` — KeyError on missing keys**

The current implementation uses direct dict subscript access that crashes when loading history saved with an older browser-use version or when a step entry is missing expected keys:

```python
for h in data['history']:           # KeyError if 'history' key absent
    if h['model_output']:           # KeyError if 'model_output' absent
    if 'interacted_element' not in h['state']:  # KeyError if 'state' absent
```

**2. `AgentHistoryList.final_result()` — IndexError on empty result list**

```python
if self.history and self.history[-1].result[-1].extracted_content:
#                              ^^^^^^^^^^^^^^^^ IndexError when result == []
```

Every sibling method (`is_done`, `is_successful`, `judgement`, `is_judged`, `is_validated`) already guards with `len(self.history[-1].result) > 0`. `final_result` was the only one missing this check.

### Fix

- **`load_from_dict`**: use `data.get('history', [])` and `h.get(key)` instead of direct subscript access, matching standard defensive Python dict usage.
- **`final_result`**: add `len(self.history[-1].result) > 0` guard before accessing `result[-1]`, consistent with all other methods in the class.

### Changes

Only `browser_use/agent/views.py` is modified (14 additions, 8 deletions).

---
## Summary by cubic
Prevent crashes when loading agent history with missing fields and when reading the final result on empty lists. This makes `AgentHistoryList` safer and backward-compatible with legacy or partial histories.

- **Bug Fixes**
  - `load_from_dict`: use `data.get('history', [])` and `h.get()` for safe access; initialize missing `state` and default `state.interacted_element` to `None`; validate `model_output` only when it’s a dict.
  - `final_result`: check `len(self.history[-1].result) > 0` before indexing; return `None` when empty or no extracted content, consistent with other methods.

<sup>Written for commit 6129ceb0b0d09c319c381775a65e19ab63c2dce4. Summary will update on new commits.</sup>

PR #4472, opened 2026-03-22
fix: use .get() in load_from_dict and add length guard in final_result()

### LINKED ISSUE #4463: Bug: AgentHistoryList.load_from_dict raises KeyError on incomplete history; final_result() raises IndexError on empty result list
## Bug Description

Two related bugs in `browser_use/agent/views.py` that crash when accessing agent history:

### 1. `load_from_dict` — KeyError on missing keys

```python
@classmethod
def load_from_dict(cls, data, output_model):
    for h in data['history']:          # KeyError if 'history' missing
        if h['model_output']:          # KeyError if 'model_output' missing
            ...
        if 'interacted_element' not in h['state']:  # KeyError if 'state' missing
            h['state']['interacted_element'] = None
```

This crashes when loading history saved with an older version of browser-use (before certain schema keys were added), or when a history step is partially written.

**Repro:** Load a history JSON file that lacks `model_output` or `state` fields at the step level (e.g. from an older version).

### 2. `final_result()` — IndexError on empty result list

```python
def final_result(self):
    if self.history and self.history[-1].result[-1].extracted_content:
    #                             ^^^^^^^^^^^^^^^^ IndexError if result == []
```

This crashes if the last history step has an empty `result` list. Every other accessor in the same class (`is_done()`, `is_successful()`, `judgement()`, `is_judged()`, `is_validated()`) guards with `len(self.history[-1].result) > 0`, but `final_result()` does not.

## Fix

Use `.get()` for dict access and add a length guard before `result[-1]`.

passionworkeer:
## Fixes browser-use/browser-use#4463

Two crashes in AgentHistoryList when loading history from older or partially-written history files:

### 1. load_from_dict: KeyError on missing keys
Use .get() for 'history', 'model_output', and 'state' keys to avoid KeyError on missing keys from older schema versions or incomplete writes.

### 2. final_result(): IndexError on empty result list
Add len() guard before accessing result[-1] to avoid IndexError when the last history step has an empty result list, consistent with other accessors (is_done, is_successful, judgement, is_judged, is_validated) that already guard on len(self.history[-1].result) > 0.

---
## Summary by cubic
Fixes history loading crashes, adds Windows‑safe process termination for the tunnel, and filters extension pages from session targets to prevent lingering CDP connections (fixes #4463).

- **Bug Fixes**
  - History: Use `.get()` for `history`/`model_output`, catch `ValidationError` when validating `model_output`, normalize `model_output` to `None` when missing/invalid, normalize `state` to `{}` when missing/`None`/non‑dict (and ensure `interacted_element`), and guard `final_result()` when the result list is empty.
  - Session: `get_all_page_targets()` now ignores `chrome-extension://` targets to avoid lingering connections when `keep_alive=True`.
  - Tunnel: Cross‑platform `_kill_process()` — on Windows uses `ctypes` `TerminateProcess` and waits up to ~1s for exit; Unix keeps SIGTERM/SIGKILL flow. Added comprehensive unit tests for both paths; tests mock `time.sleep` and Windows APIs to run on all platforms.

<sup>Written for commit 6c15f15fb2230953d1eaca14e926b38e1ad18a36. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/agent/views.py
+++ b/browser_use/agent/views.py
@@ -695,7 +695,10 @@ class AgentHistoryList(BaseModel, Generic[AgentStructuredOutput]):
 			model_output = h.get('model_output')
 			if model_output:
 				if isinstance(model_output, dict):
-					h['model_output'] = output_model.model_validate(model_output)
+					try:
+						h['model_output'] = output_model.model_validate(model_output)
+					except ValidationError:
+						h['model_output'] = None
 				else:
 					h['model_output'] = None
 			else:


laithrw:
Thanks for the PR. Closing; the scope here is beyond the original bug. The views.py fix is solid, but #4464 covers it more cleanly.

PR #4479, opened 2026-03-23
Fix KeyError in load_from_dict and IndexError in final_result (fix #4463)

### LINKED ISSUE #4463: Bug: AgentHistoryList.load_from_dict raises KeyError on incomplete history; final_result() raises IndexError on empty result list

passionworkeer:
## Fix for Issue #4463

Two bugs fixed in rowser_use/agent/views.py:

### Bug 1: load_from_dict KeyError on incomplete history
- data['history'] → data.get('history', []) to handle missing 'history' key
- h['model_output'] → h.get('model_output') to handle missing 'model_output' key
- h['state'] → h.get('state', {}) to handle missing 'state' dict

### Bug 2: final_result IndexError on empty result list
- Added self.history[-1].result truthiness check before accessing 
esult[-1]
- All other methods in the class already guard against empty result - now final_result() is consistent

Closes #4463

---
## Summary by cubic
Prevent crashes when loading partial agent history and when reading the final result by handling missing fields and empty results. Fixes #4463.

- **Bug Fixes**
  - load_from_dict: iterate `data.get('history', [])`, validate `model_output` only if present, and ensure `state.interacted_element=None` even if `state` is missing.
  - final_result: guard for missing or empty `result` before accessing the last item.

<sup>Written for commit 0d38b8a447b73626056e9c22caa5b4984ee2d67a. Summary will update on new commits.</sup>


laithrw:
Closing; setdefault('state', {}) doesn't handle the case where state exists but is None, so this can still crash. #4464 handles it correctly.

PR #4503, opened 2026-03-24
fix: remove stray pass causing unconditional error logging in security watchdog

### LINKED ISSUE #4486: Security watchdog logs error unconditionally due to stray pass in except block
## Bug Description

In `security_watchdog.py`, the `on_NavigationCompleteEvent` handler has a stray `pass` in the except block:

```python
except Exception as e:
    pass
    self.logger.error(f'Failed to navigate to about:blank: {type(e).__name__} {e}')
```

The `pass` statement causes the except block to end, so `self.logger.error(...)` executes **unconditionally** — on every `NavigationCompleteEvent` to a blocked URL, even when the `about:blank` navigation succeeds.

## Impact

Misleading error logs on every blocked URL navigation, even on success.

## Fix

Remove the `pass` statement so `logger.error()` only runs inside the except block.

BillionClaw:
## Summary
The except block in `on_NavigationCompleteEvent` contained a `pass` statement before the `logger.error` call. Since Python executes statements sequentially, the `pass` did nothing and then the `logger.error` was called unconditionally — logging an error even when no exception occurred.

## Fix
Removed the stray `pass` statement so the error is only logged when an actual exception is caught.

**Before:**
```python
except Exception as e:
    pass
    self.logger.error(f'⛔️ Failed to navigate to about:blank: ...')
```

**After:**
```python
except Exception as e:
    self.logger.error(f'⛔️ Failed to navigate to about:blank: ...')
```

## Issue
Fixes browser-use/browser-use#4486

---
## Summary by cubic
Removed a stray pass in the `on_NavigationCompleteEvent` except block so we only log navigation failures when an actual exception occurs. This stops false error logs from the security watchdog after successful navigations.

<sup>Written for commit df26960436481ccb9b0fc67b312a6f31b1015f86. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/browser/watchdogs/security_watchdog.py
+++ b/browser_use/browser/watchdogs/security_watchdog.py
@@ -68,7 +68,6 @@ class SecurityWatchdog(BaseWatchdog):
 				await session.cdp_client.send.Page.navigate(params={'url': 'about:blank'}, session_id=session.session_id)
 				self.logger.info(f'⛔️ Navigated to about:blank after blocked URL: {event.url}')
 			except Exception as e:
-				pass
 				self.logger.error(f'⛔️ Failed to navigate to about:blank: {type(e).__name__} {e}')
 
 	async def on_TabCreatedEvent(self, event: TabCreatedEvent) -> None:


BillionClaw:
Thank you for the review and merge! Glad to help.

lgx0319:
Friendly bump -- this PR has been open for a while. Happy to make any changes if needed, or close it if no longer relevant.

PR #4532, opened 2026-03-26
Feat/dom timing measure env

peakostrich:
This PR adds an opt-in flag, **`BROWSER_USE_LOG_DOM_TIMING`**, so the existing DOM pipeline timing breakdown (CDP vs serialization, paint order, etc.) is logged at **INFO** instead of only at **DEBUG**, making it easier to see which step is slow without enabling full debug noise. It refactors that breakdown into **`DOMWatchdog._format_dom_timing_breakdown_lines`**, includes **`selector_map elements: N`** for context, and documents the env var in **`.env.example`**. Behavior is unchanged unless the flag is set (or DEBUG is already enabled for that logger); there are no changes to DOM/CDP/serialize logic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@
 <a href="https://github.com/browser-use/browser-use"><img src="https://media.browser-use.tools/badges/github" alt="Github Stars"></a>
 <img width="4" height="1" alt="">
 <a href="https://x.com/intent/user?screen_name=browser_use"><img src="https://media.browser-use.tools/badges/twitter" alt="Twitter"></a>
-<img width="4" height="1" alt="">
+<img width="4 height="1" alt="">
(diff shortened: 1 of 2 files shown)


laithrw:
@peakostrich thanks for the PR! My concern is that different people will want visibility into different parts of the pipeline -- DOM timing, CDP calls, screenshots, etc.

It's true that DEBUG is noisy. You could set BROWSER_USE_DEBUG_LOG_FILE and search that file for the timing info, or you could run with DEBUG and pipe output through grep.

I'm going to close this out, but I appreciate you raising this.

peakostrich:
Hello Laithrw, Ok, sounds good, thank you for looking at my PR! Do you just recommend taking a look at only issues instead for future contributions?

laithrw:
@peakostrich yes, it's most helpful to create PRs resolving others' issues, particularly if you are able to reproduce the issue. However, I'm always happy to review PRs not linked to issues if you tag me.

PR #4549, opened 2026-03-28
fix(cli): show helpful error on Windows when Chrome profile is locked

### LINKED ISSUE #4546: Bug: Windows: `--profile` fails with WinError 32 when Chrome is running (profile copy vs file locks)
### Browser Use Version

0.12.5

## Environment

- **OS:** Windows 10 / 11 (x64)
- **CLI:** `uvx --from 'browser-use[cli]' browser-use ...`
- **Chrome:** Normal install, `Default` profile, **Chrome already running**

## What happens

```bash
browser-use --headed --profile "Default" open https://example.com
```

Fails with multiple [WinError 32] errors (Chinese locale message: 另一个程序正在使用此文件，进程无法访问。 — The process cannot access the file because it is being used by another process.)

Typical locked paths include:

%LocalAppData%\Google\Chrome\User Data\Default\Network\Cookies
Cookies-journal
Safe Browsing Network\Safe Browsing Cookies
Sessions\Session_*, Tabs_*
The error payload shows a copy from the real Chrome profile into a temp dir such as %TEMP%\browser-use-user-data-dir-*\Default\....

## Why (root cause)
--profile appears to work by copying the Chrome user-data profile into an isolated temp directory, then launching Chromium with that copy.

On Windows, while Chrome is running it holds exclusive locks on many profile files. The copy step then fails with WinError 32.

On macOS/Linux, reading/copying files that another process has open is often still possible; Windows locking is stricter, so this is especially a Windows pain point.

## Impact
Users who expect to reuse an existing Chrome login session via --profile on Windows effectively must fully quit Chrome first, which is easy to miss and hurts the advertised workflow.

## Suggestions
Docs: Explicitly document: on Windows, close Chrome before --profile, or use --connect / --cdp-url with remote debugging enabled.
UX: On WinError 32, detect this pattern and print a short, actionable message (close Chrome or use CDP) instead of a long tuple dump.
Longer term (if feasible): Consider a Windows-friendly approach that avoids copying locked files (e.g. attach via CDP only, or another strategy the team prefers).

## Related
--connect correctly expects a Chrome instance with remote debugging; that path is fine but orthogonal to --profile expectations on Windows.

Thank you for browser-use — happy to help test any fix on Windows.

### Failing Python Code

```python

```

### LLM Model

_No response_

_No response_

### Full DEBUG Log Output

```shell

```

guoyangzhen:
## Problem

On Windows, when using `--profile "Default"` with Chrome already running, `shutil.copytree` fails with `OSError: [WinError 32] The process cannot access the file because it is being used by another process` — a confusing error with no guidance.

## Fix

Adds detection for WinError 32 in the `_copy_profile()` method of `BrowserProfile`. When the error is caught, a clear `RuntimeError` is raised telling the user to:
- Close all Chrome windows first, or
- Use `--cdp-url` to connect to an existing Chrome instance

This is a minimal, safe change — no logic changes, only improved error handling.

Fixes #4546

---
## Summary by cubic
On Windows, the CLI now shows a clear error when the Chrome profile or “Local State” file is locked, instead of surfacing WinError 32. It tells users to close Chrome or connect via `--cdp-url`, and removes an unnecessary f-prefix.

<sup>Written for commit 2295f52c08e96671371c2ba1c2fc2b26fe2b4bdd. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/browser/profile.py
+++ b/browser_use/browser/profile.py
@@ -834,8 +834,8 @@ class BrowserProfile(BrowserConnectArgs, BrowserLaunchPersistentContextArgs, Bro
 				if getattr(e, 'winerror', None) == 32:
 					raise RuntimeError(
 						f"Failed to copy Chrome profile '{self.profile_directory}' — another Chrome process is using it.\n"
-						f'Please close all Chrome windows first, or use --cdp-url to connect to an existing Chrome instance.\n'
-						f'Example: browser-use --cdp-url http://localhost:9222'
+						'Please close all Chrome windows first, or use --cdp-url to connect to an existing Chrome instance.\n'
+						'Example: browser-use --cdp-url http://localhost:9222'
 					) from e
 				raise
 			local_state_src = path_original_user_data / 'Local State'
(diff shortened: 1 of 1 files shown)


laithrw:
@guoyangzhen

PR #4568, opened 2026-03-30
perf(agent): hybrid selector_map matching for history rerun

peakostrich:
This change speeds up _update_action_indices during history rerun when EXACT element_hash matching fails. It keeps Level 1 as a single linear scan; on miss, it builds first-win lookup maps in one pass over selector_map (with compute_stable_hash called once per node) and uses those for STABLE, XPATH, AX_NAME, and ATTRIBUTE matching instead of scanning the full map again for each stage. Behavior matches the previous cascading order and “first match in iteration order” semantics. INFO/DEBUG-only diagnostic scans are gated behind isEnabledFor to avoid extra work when those log levels are off.

---
## Summary by cubic
Speeds up history rerun element matching by keeping a single linear scan for EXACT matches, then using one-pass, first-win lookup maps for the remaining stages. Preserves the previous cascade order and first-match semantics, while reducing repeated scans and unnecessary logging work.

- **Performance**
  - Added `_SelectorMapMatchLookups` and a one-pass builder to precompute first-win indices from `selector_map`.
  - On EXACT miss, use lookups for STABLE hash, XPATH, AX_NAME, and attribute matches instead of rescanning.
  - Gated full diagnostic scans behind `logger.isEnabledFor(...)` to avoid overhead when INFO/DEBUG are off.

<sup>Written for commit 0018b8446ed8279d8a47dabacab5de7c1b9dae72. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/agent/service.py
+++ b/browser_use/agent/service.py
@@ -7,6 +7,7 @@ import re
 import tempfile
 import time
 from collections.abc import Awaitable, Callable
+from dataclasses import dataclass
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Generic, Literal, TypeVar, cast
 from urllib.parse import urlparse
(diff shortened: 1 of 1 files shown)


laithrw:
Hey, thanks for the contribution! I benchmarked this against the current code, and it appears to be slower.

The issue is that building the lookup tables upfront is more expensive than just scanning the list, because most of the time we find a match early and stop. And the cost of building the lookup tables upfront outweighs the savings.

I'm happy to look at any other optimizations you might have.

PR #4589, opened 2026-04-02
fix #4579: CDP connection instability causing indefinite hangs with remote browsers

### LINKED ISSUE #4579: Bug: CDP connection instability causing indefinite hangs with remote browsers
### Browser Use Version

0.12.5

# Individual CDP calls within handlers lack timeouts, causing indefinite hangs with remote browsers

## Summary

When using browser-use with remote browsers via CDP (e.g., Browserless), individual CDP calls like `Runtime.evaluate` inside `DOMWatchdog.on_BrowserStateRequestEvent` can hang indefinitely if the CDP WebSocket connection becomes stale. There is no per-call timeout — the only safety net is the event-level timeout (default 30s), which wastes the entire timeout window before failing.

## Environment

- browser-use 0.12.5
- Remote browser via CDP (Browserless)
- `keep_alive=True`

## Observed behaviour

When running sequential agents on the same browser session with `keep_alive=True`, there are naturally idle periods between agent runs where no CDP traffic flows. After such an idle period, the next `BrowserStateRequestEvent` can hang:

1. `DownloadsWatchdog` completes but its CDP call (`get_current_page_url`) takes **3.0s** instead of the usual **0.00-0.01s** — an early signal the connection is degraded
2. `DOMWatchdog` enters `_get_pending_network_requests`
3. `get_or_create_cdp_session(focus=True)` succeeds but also takes **3.0s**
4. The subsequent `Runtime.evaluate` call **never returns** — no error, no timeout, just an indefinite await
5. The handler blocks until the event-level timeout fires

The connection degradation is **not deterministic** — sometimes connections survive long idle periods, sometimes they become stale after relatively short gaps. No CDP errors or WebSocket close frames are logged when the connection degrades. The WebSocket appears open but is silently half-dead (sends succeed, responses never arrive).

This was identified by adding diagnostic instrumentation around the key methods in `DOMWatchdog` — wrapping `_get_pending_network_requests`, `_capture_clean_screenshot`, and `_build_dom_tree_without_highlights` with timing logs, and adding a heartbeat task to `on_BrowserStateRequestEvent`. This confirmed the handler was alive but stuck awaiting the CDP call indefinitely.

With local browsers this is unlikely to occur because the browser process is co-located. With remote browsers behind a WebSocket proxy, idle connections can degrade without either side detecting it.

## Relation to #4374

This creates a catch-22 for remote browser users with `keep_alive=True`:

- **Keep the event bus alive** (workaround for #4374): The CDP WebSocket connection can degrade silently after idle periods, and the next `Runtime.evaluate` hangs indefinitely
- **Let `Agent.close()` tear down the event bus** (default behaviour): The CDP WebSocket handlers exit, session manager state is cleared, and the authenticated session is lost — exactly what #4374 describes

Both paths lead to broken sessions. The underlying issue is that browser-use assumes the CDP connection is always healthy if the WebSocket is open.

## Root cause

CDP calls in handler code (e.g., `_get_pending_network_requests`, `_build_dom_tree_without_highlights`, `_capture_clean_screenshot`) `await` CDP protocol methods with no individual timeout. If the remote browser stops responding — due to WebSocket proxy recycling, idle timeout, or network conditions — these awaits block forever.

The relevant code path in `dom_watchdog.py`:

```python
async def _get_pending_network_requests(self):
    cdp_session = await self.browser_session.get_or_create_cdp_session(focus=True)
    # This Runtime.evaluate has no timeout — hangs forever if CDP is stale
    result = await cdp_session.execute(
        Runtime.evaluate(expression=..., await_promise=True, ...)
    )
```

## Reproduction

This is timing-dependent and requires a remote browser. The conditions are:
1. Connect to a remote browser via CDP with `keep_alive=True`
2. Apply the workaround from #4374 to prevent `Agent.close()` from tearing down the event bus (otherwise the CDP connection is killed on close and the session is lost before this issue can manifest)
3. Run an agent to completion (agent closes, browser session and CDP connection persist)
4. Wait with no CDP traffic (duration varies — sometimes 30s is enough, sometimes minutes are fine)
5. Start a new agent on the same browser session
6. The first `BrowserStateRequestEvent` has a chance of hanging indefinitely

The probability is non-deterministic and depends on the remote browser provider's WebSocket proxy behaviour and network conditions.

### Failing Python Code

```python
It is difficult to provide a minimal replication as it involves CDP and multiple browser instances.

It also requires addressing https://github.com/browser-use/browser-use/issues/4374 locally so that keep_alive actually works.
```

### LLM Model

_No response_

_No response_

### Full DEBUG Log Output

```shell

```

passionworkeer:
## Bug Summary
When using browser-use with remote browsers via CDP (e.g. Browserless), individual CDP calls like \Runtime.evaluate\ inside \DOMWatchdog._get_pending_network_requests\ can hang indefinitely if the CDP WebSocket connection becomes stale during idle periods. There was no per-call timeout - the only safety net was the event-level timeout (default 30s), which wastes the entire timeout window before failing.

## Fix
Wrap \Runtime.evaluate\ CDP call in \DOMWatchdog._get_pending_network_requests\ with \syncio.wait_for(timeout=10.0)\ to provide per-call timeout protection. This prevents indefinite hangs while still allowing reasonable time for CDP calls to complete.

## Changes
- \rowser_use/browser/watchdogs/dom_watchdog.py\: Added 10s timeout wrapper around \Runtime.evaluate\ call via \syncio.wait_for\

Fixes #4579

---
## Summary by cubic
Add a 10s per-call timeout with explicit `asyncio.TimeoutError` handling and warning logs around CDP `Runtime.evaluate` in `DOMWatchdog._get_pending_network_requests` to prevent hangs on stale remote sockets; return [] on timeout or other evaluation errors. Harden `AgentHistoryList.load_from_dict` by deep-copying input, filtering non-dict history items, safely validating `model_output` (fallback to None), and guarding missing `state` keys; fixes #4579, closes #4589.

<sup>Written for commit 8460a0f1b51562598496af5fb1e1d7a582c0fd6f. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/agent/views.py
+++ b/browser_use/agent/views.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import copy
 import hashlib
 import json
 import logging
@@ -689,6 +690,8 @@ class AgentHistoryList(BaseModel, Generic[AgentStructuredOutput]):
 
 	@classmethod
 	def load_from_dict(cls, data: dict[str, Any], output_model: type[AgentOutput]) -> AgentHistoryList:
+		# Use deep copy to avoid mutating the caller's input
+		data = copy.deepcopy(data)
 		# loop through history and validate output_model actions to enrich with custom actions
 		for h in data['history']:
 			if h['model_output']:
--- a/browser_use/agent/views.py
+++ b/browser_use/agent/views.py
@@ -692,14 +692,19 @@ class AgentHistoryList(BaseModel, Generic[AgentStructuredOutput]):
 	def load_from_dict(cls, data: dict[str, Any], output_model: type[AgentOutput]) -> AgentHistoryList:
 		# Use deep copy to avoid mutating the caller's input
 		data = copy.deepcopy(data)
+		# Filter out malformed history items (non-dict entries) before processing
+		data['history'] = [h for h in data['history'] if isinstance(h, dict)]
 		# loop through history and validate output_model actions to enrich with custom actions
 		for h in data['history']:
-			if h['model_output']:
+			if 'model_output' in h and h['model_output']:
 				if isinstance(h['model_output'], dict):
-					h['model_output'] = output_model.model_validate(h['model_output'])
+					try:
+						h['model_output'] = output_model.model_validate(h['model_output'])
+					except (ValidationError, TypeError, AttributeError):
+						h['model_output'] = None
 				else:
 					h['model_output'] = None
-			if 'interacted_element' not in h['state']:
+			if 'state' in h and 'interacted_element' not in h['state']:
 				h['state']['interacted_element'] = None
 
 		history = cls.model_validate(data)
--- a/browser_use/browser/watchdogs/dom_watchdog.py
+++ b/browser_use/browser/watchdogs/dom_watchdog.py
@@ -194,9 +194,21 @@ class DOMWatchdog(BaseWatchdog):
 })()
 """
 
-			result = await cdp_session.cdp_client.send.Runtime.evaluate(
-				params={'expression': js_code, 'returnByValue': True}, session_id=cdp_session.session_id
-			)
+			try:
+				result = await asyncio.wait_for(
+					cdp_session.cdp_client.send.Runtime.evaluate(
+						params={'expression': js_code, 'returnByValue': True}, session_id=cdp_session.session_id
+					),
+					timeout=10.0,
+				)
+			except asyncio.TimeoutError:
+				self.logger.warning(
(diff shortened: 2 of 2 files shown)


passionworkeer:
@Copilot 感谢 review！建议很合理——超时后 generic Exception 拿到的 `e` 字符串经常是空的，不容易定位问题。

已按 suggestion 更新：显式捕获 `asyncio.TimeoutError`，在日志里明确打印 10.0s 超时和异常类型，方便排查 hang 问题。

PR #4591, opened 2026-04-02
perf(dom): hoist previous backend_node_id set in DOM serializer

peakostrich:
Build frozenset once per serialize_accessible_elements instead of inside the recursive assign pass. Preserve falsy empty selector_map {} so is_new is never set. Add unit tests for is_new with non-empty vs empty previous map.

---
## Summary by cubic
Hoisted building of the previous `backend_node_id` set to run once per serialization, reducing work in the recursive pass. Also fixed `is_new` so it only applies when a non-empty previous selector map exists.

- **Bug Fixes**
  - Preserve falsy empty `selector_map` `{}`; `_previous_backend_node_ids` stays `None` so `is_new` is never set when there’s no previous state. Added unit tests for non-empty vs empty previous maps.

- **Refactors**
  - Cache previous backend node IDs as a `frozenset` in `serialize_accessible_elements` and reuse in `_assign_interactive_indices_and_mark_new_nodes` instead of recomputing per node.

<sup>Written for commit f6d66cb341a07b2c2c424064fc6b6d4959ee002f. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/dom/serializer/serializer.py
+++ b/browser_use/dom/serializer/serializer.py
@@ -80,6 +80,8 @@ class DOMTreeSerializer:
 		self.paint_order_filtering = paint_order_filtering
 		# Session ID for session-specific exclude attribute
 		self.session_id = session_id
+		# Set once per serialize_accessible_elements when previous_cached_selector_map is non-empty
+		self._previous_backend_node_ids: frozenset[int] | None = None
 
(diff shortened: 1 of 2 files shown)


laithrw:
Thanks for the PR! It looks like this would only save a few ms at most, and I'd rather not touch this code without a compelling reason because it's well-tested across many websites. I appreciate you looking into this, though.

Issue #4683, opened 2026-04-15
Add a faster way to clear CLI input fields


gregpr07:
`browser-use` CLI does not have a fast way to clear an input field after text has been entered. Current workarounds shown include selecting all and deleting or sending many backspaces.

<img src="https://uploads.linear.app/3b05ba62-2c90-4d87-90e7-7d8bcbbf17db/e90b88b2-9faf-420d-82e8-da44899d4d52/b1c29ffc-d9f3-4add-b89a-1dfa2b14633c?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzNiMDViYTYyLTJjOTAtNGQ4Ny05MGU3LTdkOGJjYmJmMTdkYi9lOTBiODhiMi05ZmFmLTQyMGQtODJlOC1kYTQ0ODk5ZDRkNTIvYjFjMjlmZmMtZDlmMy00YWRkLWI4OWEtMWRmYTJiMTQ2MzNjIiwiaWF0IjoxNzc2MjIxNTc4LCJleHAiOjE4MDc3OTIxMzh9.-ghacGvv_VHU9ns248qUyvqE0oKTatF3VEEmW9hsMl0 " alt="CleanShot 2026-04-14 at 19.52.06@2x.png" width="2024" data-linear-height="2482" />

revanthkumar96:
#  Explicit `clear_input` Action

## Problem
Currently, the `browser-use` library lacks a dedicated, high-level action for clearing input fields. This forces agents/users to rely on workarounds that are either computationally expensive or conceptually unintuitive.

### Current Workarounds
1.  **`input_text(index=X, text="")`**: While functional, this carries the overhead of the full text input pipeline and can be ambiguous to agents who might not realize an empty string clears the field.
2.  **Manual Keyboard Shortcuts**: Agents often attempt to send `Control+A` followed by `Backspace`. This requires multiple steps, high latency, and is prone to failure if focus is lost or the OS/Browser intercepts the shortcuts differently.

## Proposed Solution
Introduce a first-class `clear_input` action. This will expose a dedicated tool to the agent, providing a fast, robust, single-step mechanism to reset an input field using optimized CDP/JavaScript strategies.

## Technical Implementation Plan

### 1. Tool Definitions (`browser_use/tools/views.py`)
Add a new Pydantic model to define the action's input schema.
```python
class ClearInputAction(BaseModel):
[...]

PR #4686, opened 2026-04-15
Add explicit clear_input action

revanthkumar96:
Introduces a first-class clear_input action, providing a dedicated, high-level tool for clearing input fields. This is more efficient than the workaround of using input_text with an empty string.

- Add ClearInputAction model in tools/views.py
- Register clear_input action in tools/service.py
- Add ClearInputEvent in browser/events.py
- Implement handler in default_action_watchdog.py using _clear_text_field
- Add loop detection in agent/views.py for repeated clearing
#4683

### DIFF
--- a/browser_use/agent/views.py
+++ b/browser_use/agent/views.py
@@ -122,7 +122,7 @@ def _normalize_action_for_hash(action_name: str, params: dict[str, Any]) -> str:
 		engine = params.get('engine', 'google')
 		return f'search|{engine}|{"|".join(tokens)}'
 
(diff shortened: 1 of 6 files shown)


laithrw:
Thanks for the PR @revanthkumar96! Closing in favor of #4692 for a more minimal fix. input_text(index, text="") already clears the field through the same underlying path, and the real fix is in the clear fallbacks and the tool descriptions that don't mention empty-text clearing. I appreciate you looking into this.

PR #4697, opened 2026-04-17
fix(recording): use CDP timestamps to preserve real-time video playback speed

### LINKED ISSUE #4696: Bug: Video is too fast (Make it normal
### Browser Use Version

0.12.6

Video is too fast, wanted to see realistic video when no custom settings provided to BrowserUse Class Instance. Code below:

browser = Browser(
            window_size={"width": self.screen_width, "height": self.screen_height},
            viewport={"width": self.screen_width, "height": self.screen_height},
            wait_for_network_idle_page_load_time=1,
            record_video_dir=self.video_dir,
            highlight_elements=True,
            disable_security=True,
            headless= true
        )

### Failing Python Code

```python

```

### LLM Model

_No response_

_No response_

### Full DEBUG Log Output

```shell

```

octo-patch:
Fixes #4696

## Problem

When using `record_video_dir`, the recorded video plays back much faster than real time. CDP screencast delivers frames only on visual change — during network waits or idle periods between agent steps, zero frames are produced. All captured frames are then written to the video at the configured framerate (30 fps default), so a 5-second page load appears to be instant in the video.

## Solution

`ScreencastFrameMetadata` includes a `timestamp` field (Unix seconds). This PR threads that timestamp from the CDP event through to `VideoRecorderService.add_frame()`. When a new frame arrives, the recorder computes the wall-clock gap since the last frame and inserts duplicate filler frames to fill the silence — matching the video duration to actual session time.

Key details:
- Filler frame count: `int(gap_seconds * framerate) - 1` (the current frame accounts for the last slot)
- Idle gaps longer than `MAX_FILL_SECONDS` (10 s) are capped to keep file sizes reasonable
- Tab switches reset the timestamp so no spurious filler spans different targets
- Fully backward-compatible: if no timestamp is available, `add_frame()` behaves as before

## Testing

The change can be verified by recording a session that includes a few-second network wait and comparing the video duration to the actual elapsed time. Before this fix the video would be significantly shorter than wall-clock time; after the fix the durations match.

---
## Summary by cubic
Uses CDP screencast timestamps to keep recorded video playback in real time. Also validates uploads against `accept` rules, retries flaky screenshot timeouts, and returns CLI upload errors as error dicts.

- **Bug Fixes**
  - Recording: Use `ScreencastFrameMetadata.timestamp` to detect pauses and insert duplicate frames so duration matches wall clock; cap fills at 10s; reset on tab switch; fallback if missing.
  - Screenshots: Retry `Page.captureScreenshot` up to 3 times (0.5s delay) on transient "timed out" errors.
  - Uploads: Validate `<input accept>` for `.ext`, `type/*`, and exact MIME; if MIME is unknown and rules are MIME-only, warn and defer to the browser; otherwise raise `BrowserError`. CLI `upload` now catches this and returns `{'error': ...}`. Adds tests for accept/reject paths.

- **Refactors**
  - Formatting: Apply ruff format to satisfy CI.

<sup>Written for commit 0f4fe1a0e1a90deb0fe85bdab294deb31fa1ef9d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/4697?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/browser/video_recorder.py
+++ b/browser_use/browser/video_recorder.py
@@ -36,7 +36,14 @@ class VideoRecorderService:
 	This service captures individual frames from the CDP screencast, decodes them,
 	and appends them to a video file using a pip-installable ffmpeg backend.
 	It automatically resizes frames to match the target video dimensions.
+
+	Timestamps from CDP screencast metadata are used to insert duplicate frames during
+	pauses, preserving real-time playback speed. Without this, videos play back much
+	faster than real time because CDP only delivers frames on visual change.
 	"""
+
+	# Max seconds of silence to fill with duplicate frames. Caps file size during long idle periods.
+	MAX_FILL_SECONDS = 10.0
 
 	def __init__(self, output_path: Path, size: ViewportSize, framerate: int):
 		"""
@@ -53,6 +60,8 @@ class VideoRecorderService:
 		self._writer: Optional['Format.Writer'] = None
 		self._is_active = False
 		self.padded_size = _get_padded_size(self.size)
+		self._last_timestamp: float | None = None
+		self._last_frame: 'np.ndarray | None' = None
 
 	def start(self) -> None:
 		"""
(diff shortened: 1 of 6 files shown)


octo-patch:
Pushed `9e777e2f` to address the cubic-dev-ai review on `default_action_watchdog.py:2668`: when `mimetypes.guess_type()` returns no MIME (OS-dependent), we now log a warning and defer validation to the browser instead of hard-failing the upload. Extension-based `accept` rules still hard-fail as before.

PR is still relevant — keeping it open. The recording-timestamp fix, the screenshot-retry change, and the upload accept-attribute check are all unaffected by the stale-bot ping.

octo-patch:
Pushed a formatting-only commit to fix the code-style job — it was failing solely on the ruff-format hook (ruff check, pyright, and the rest passed). Still happy to rebase if you'd like to move this forward.

octo-patch:
Also fixed the test_cli_upload failure: the upload handler surfaced its pre-checks as error dicts but called upload_file() unguarded, so the new accept-attribute BrowserError propagated out instead of being returned like the other upload errors. Wrapped it to match the handler's existing pattern.

PR #4704, opened 2026-04-19
fix: remove dead code `'type'` from second elif branch in optimize_schema()

### LINKED ISSUE #4703: Dead code in optimize_schema(): `'type'` key is unreachable in second elif branch
### Location
`browser_use/llm/schema.py` — `optimize_schema()` inner function

### Problem

The `'type'` key is handled by an **early** `elif key == 'type':` branch (line ~61), but then also listed inside a **later** `elif key in ['type', 'required', ...]` block. Because Python `elif` chains are exclusive, the second branch can **never** be reached for `key == 'type'`.

This is dead code: the `'type'` entry in the second list is redundant and misleading.

```python
# This catches 'type' FIRST:
elif key == 'type':
    optimized[key] = value if not isinstance(value, (dict, list)) else optimize_schema(value, defs_lookup)

# ...

# 'type' here is UNREACHABLE dead code:
elif key in [
    'type',       # <-- dead code, never reached
    'required',
    'minimum',
    ...
]:
    optimized[key] = value if not isinstance(value, (dict, list)) else optimize_schema(value, defs_lookup)
```

### Fix

Remove `'type'` from the second `elif key in [...]` list to eliminate the dead code and make the logic clear.

### Impact

While the behavior happens to be identical in both branches (same logic), the dead code creates confusion about intent and could mislead future maintainers into thinking `'type'` is handled differently in that branch.

kuishou68:
Closes #4703

## Problem

In `browser_use/llm/schema.py`, the `optimize_schema()` function has a dead code issue. The `'type'` key is first handled by an early `elif key == 'type':` branch, but then it also appears inside a later `elif key in ['type', 'required', ...]` block. Because Python `elif` chains are exclusive, the `'type'` entry in the second list is **never reached**.

```python
# This catches 'type' FIRST:
elif key == 'type':
    optimized[key] = value if not isinstance(value, (dict, list)) else optimize_schema(value, defs_lookup)

# ...

# 'type' here is dead code — never reached:
elif key in [
    'type',     # <-- unreachable dead code
    'required',
    ...
]:
```

## Fix

Remove `'type'` from the second `elif key in [...]` list to eliminate dead code and clarify intent.

---
## Summary by cubic
Removed dead code in optimize_schema() by removing `type` from the later `elif` list, since `type` is already handled earlier. No functional change; clarifies intent and removes unreachable logic.

<sup>Written for commit 6ccae6980e3d884d28fef8805b0a31833d3768a7. Summary will update on new commits.</sup>

### DIFF
--- a/.github/ISSUE_TEMPLATE/3_feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/3_feature_request.yml
@@ -46,7 +46,7 @@ body:
     attributes:
       label: What version of browser-use are you currently using?
       description: |
-        Run `pip show browser-use` or `git log -n 1` and share the exact number or git hash. DO NOT JUST ENTER `latest release` OR `main`.  
+        Run `uv pip show browser-use` or `git log -n 1` and share the exact number or git hash. DO NOT JUST ENTER `latest release` OR `main`.
         We need to know what version of the browser-use library you're running in order to contextualize your feature request.  
         Sometimes features are already available and just need to be enabled with config on certain versions.
       placeholder: "e.g. 0.1.48 or 62760baaefd"
--- a/.github/ISSUE_TEMPLATE/4_docs_issue.yml
+++ b/.github/ISSUE_TEMPLATE/4_docs_issue.yml
@@ -27,7 +27,7 @@ body:
     attributes:
       label: Documentation Page
       description: Which page or section of the documentation is this about?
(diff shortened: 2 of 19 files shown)


sauravpanda:
@kuishou68 can you sign the CLA so that I can merge this?

kuishou68:
Hi @sauravpanda! The CLA has been signed — the CLA assistant is showing "All committers have signed the CLA" ✅. The PR should be ready to merge whenever you'd like. Thanks for the patience!

PR #4711, opened 2026-04-20
fix(tools): enforce per-action timeout to prevent hung event handlers

sauravpanda:
## Summary
- Wraps `registry.execute_action` in `asyncio.wait_for` with a configurable per-action wall-clock cap (default 90s, override via `BROWSER_USE_ACTION_TIMEOUT_S` env var or `tools.act(action_timeout=...)`)
- On timeout, the action returns `ActionResult(error=...)` so the agent records the step and recovers, instead of hanging silently

## The bug
Individual CDP calls like `Page.navigate()` have their own 20s timeouts (`session.py:988`), but the surrounding event-bus plumbing — `await event` and `await event.event_result(...)` — has none. When a cloud browser's CDP WebSocket goes silent mid-session (common failure mode against remote browser services: handshake completes, `/json/version` returns 200, then the WebSocket stalls), every subsequent action dispatched through `event_bus` hangs indefinitely.

Flow when this happens:
1. `agent._execute_initial_actions()` → `multi_act()` → navigate handler dispatches `NavigateToUrlEvent`
2. `await event` / `event_result()` wait on a handler that can never complete because CDP is silent
3. `tools.act()` doesn't raise, doesn't return — it hangs
4. The outer agent watchdog eventually calls `agent.stop()`
5. `_execute_initial_actions` exits via `InterruptedError` (silently swallowed at `service.py:2557-2559`)
6. Main loop sees `state.stopped=True`, exits with 0 history entries → **empty trace**

## Real-world evidence
A 170k-task collector run produced:
- **1,090 empty-history traces** (21% of all outputs)
- Of the 927 with duration timing, **100% hit the 240s outer watchdog** — median 582s, max 2214s (~37 min)
- Cloud HTTP layer was clean throughout: all 200/201, 0 non-2xx responses
- Concurrency was not the cause — fail rate actually increased at lower parallelism, confirming the hang is per-session not back-pressure

Sample timeline from one of these failures (cloud session `78cc5c3e`):
```
14:48:47  Cloud browser created successfully
14:48:47  HTTP GET /json/version 200 OK
14:48:47  Connecting to wss://78cc5c3e...cdp4.browser-use.com/devtools/browser/...
(5 minutes of total silence)
14:53:53  Stopping cloud browser session (triggered by outer watchdog)
```

## The fix
Default cap is **90s**, which sits comfortably above `Page.navigate`'s 20s + lifecycle wait's 8s + internal CDP calls, but well below any reasonable outer agent watchdog.

Configurable via:
- `BROWSER_USE_ACTION_TIMEOUT_S` env var (process-wide default)
- `tools.act(action_timeout=...)` parameter (per-call override)

## Test plan
- [x] New `tests/ci/test_action_timeout.py` — stubs `registry.execute_action` with a coroutine that sleeps 30s, asserts `tools.act()` returns within the 0.5s cap with an `ActionResult(error=...)` containing `"timed out"`
- [x] New test also covers the fast-handler path — ensures no regression when actions complete normally
- [x] Existing `test_multi_act_guards.py` and `test_action_blank_page.py` tests still pass (all 15 green locally)
- [ ] Full CI suite

---
## Summary by cubic
Add per-action, per-step, and per-CDP-request timeouts to prevent hangs and empty traces. Defaults are 180s for actions/steps and 60s for CDP; both are configurable.

- **Bug Fixes**
  - Tools: wrap `registry.execute_action` with `asyncio.wait_for`; default 180s via `BROWSER_USE_ACTION_TIMEOUT_S` or `tools.act(action_timeout=...)`. Validate env and per-call override values (empty/non-numeric/non-finite/<=0) with fallback; on timeout return `ActionResult(error=...)`.
  - Agent: wrap `_execute_initial_actions()` with `asyncio.wait_for` using `settings.step_timeout`; on timeout log, set `state.last_result`, increment `state.consecutive_failures`, and continue to the main loop.
  - Browser: add `TimeoutWrappedCDPClient` around `cdp_use.CDPClient.send_raw` with `asyncio.wait_for` (default 60s via `BROWSER_USE_CDP_TIMEOUT_S` or constructor). Validate env/constructor values (non-finite/<=0). Use in connect/reconnect; raise `TimeoutError` on unresponsive CDP calls.
  - Tests: add `tests/ci/test_action_timeout.py` (hung/fast paths and invalid `action_timeout` override) and `tests/ci/test_cdp_timeout.py` (real `send_raw` path). Update `tests/ci/browser/test_cdp_headers.py` to patch `TimeoutWrappedCDPClient`.

<sup>Written for commit 32416bb48cf49e34384c13fa26dab67d4710e970. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/agent/service.py
+++ b/browser_use/agent/service.py
@@ -2552,11 +2552,25 @@ class Agent(Generic[Context, AgentStructuredOutput]):
 			# Register skills as actions if SkillService is configured
 			await self._register_skills_as_actions()
 
-			# Normally there was no try catch here but the callback can raise an InterruptedError
+			# Normally there was no try catch here but the callback can raise an InterruptedError.
+			# Wrap with step_timeout so initial actions (usually a single URL navigate) can't
+			# hang indefinitely on a silent CDP WebSocket — without this the agent would take
+			# zero steps and return with an empty history while any outer watchdog waits.
 			try:
-				await self._execute_initial_actions()
+				await asyncio.wait_for(
+					self._execute_initial_actions(),
+					timeout=self.settings.step_timeout,
+				)
 			except InterruptedError:
 				pass
+			except TimeoutError:
+				initial_timeout_msg = (
+					f'Initial actions timed out after {self.settings.step_timeout}s '
+					f'(browser may be unresponsive). Proceeding to main execution loop.'
+				)
+				self.logger.error(f'⏰ {initial_timeout_msg}')
+				self.state.last_result = [ActionResult(error=initial_timeout_msg)]
+				self.state.consecutive_failures += 1
 			except Exception as e:
 				raise e
 
--- a/browser_use/browser/_cdp_timeout.py
+++ b/browser_use/browser/_cdp_timeout.py
@@ -68,6 +68,26 @@ def _parse_env_cdp_timeout(raw: str | None) -> float:
 DEFAULT_CDP_REQUEST_TIMEOUT_S: float = _parse_env_cdp_timeout(os.getenv('BROWSER_USE_CDP_TIMEOUT_S'))
 
 
+def _coerce_valid_timeout(value: float | None) -> float:
+	"""Normalize a user-supplied timeout to a finite positive value.
+
+	None / nan / inf / non-positive values all fall back to the env-derived
+	default with a warning. This mirrors _parse_env_cdp_timeout so callers that
+	pass cdp_request_timeout_s directly get the same defensive behaviour as
+	callers that set the env var.
+	"""
+	if value is None:
+		return DEFAULT_CDP_REQUEST_TIMEOUT_S
+	if not math.isfinite(value) or value <= 0:
+		logger.warning(
+			'cdp_request_timeout_s=%r is not a finite positive number; falling back to %.0fs',
+			value,
+			DEFAULT_CDP_REQUEST_TIMEOUT_S,
+		)
+		return DEFAULT_CDP_REQUEST_TIMEOUT_S
+	return float(value)
+
+
 class TimeoutWrappedCDPClient(CDPClient):
 	"""CDPClient subclass that enforces a per-request timeout on send_raw.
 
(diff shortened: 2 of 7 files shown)


sauravpanda:
Pushed \`9a09c4d7\` — the final layer: per-CDP-request timeout at the `cdp_use.CDPClient.send_raw` level.

## Why this commit

Deeper root-cause analysis (see PR description update) showed the hang is ultimately in [`cdp_use/client.py:389`](https://github.com/cdp-use/cdp-use) — `send_raw` does `await future` with no timeout. If the cloud browser's WebSocket stays alive at the keepalive layer (proxy pongs our pings) but the browser container is dead or the proxy lost its upstream, CDP responses never come, the future never resolves, and every higher-level timeout in browser-use (session.start's 15s connect guard, `agent.step_timeout`, `tools.act` action timeout) is blocked on the same promise chain.

## The fix

`TimeoutWrappedCDPClient` — thin `CDPClient` subclass in `browser_use/browser/_cdp_timeout.py` that wraps `send_raw` in `asyncio.wait_for`. Any CDP method that doesn't respond within the cap raises plain `TimeoutError`, which propagates through existing `except TimeoutError` handlers.

- Default **60s** (env var `BROWSER_USE_CDP_TIMEOUT_S`). Generous for slow ops like `Page.captureScreenshot` / `Page.printToPDF` on heavy pages, but well below the 180s step timeout.
- Same defensive env parse as `BROWSER_USE_ACTION_TIMEOUT_S` — rejects empty / non-numeric / nan / inf / non-positive values with a fallback.
- Wired into both `CDPClient(...)` call sites in `session.py`.

## 15-min local run with all three fixes active

| metric | value |
|---|---|
| cloud browsers spawned | 83 |
| per-action timeouts (fix 1) | 57 |
| CDP-request timeouts (fix 3, new) | 0 |
| outer watchdog kills | 66 |

Fix 3 didn't fire in this particular window — the current cloud failure mode is "slow but responding" rather than "silent". But the silent-CDP mode **was** observed in earlier traces (100% of 927 empty-history failures, median 582s, with one sample showing 5 min of WS silence after a clean `/json/version` 200), so fix 3 is a valid safety net for that scenario even when it's latent.

All three fixes now in place, tests green, PR ready for review.

PR #4731, opened 2026-04-23
fix: remove dead code in optimize_schema()

### LINKED ISSUE #4703: Dead code in optimize_schema(): `'type'` key is unreachable in second elif branch

goingforstudying-ctrl:
Removes unreachable dead code in SchemaOptimizer.optimize_schema(). The 'elif key == type' branch was dead code because 'type' is already handled in the later essential validation fields branch. Fixes #4703

---
## Summary by cubic
Remove unreachable code in `optimize_schema()` by deleting the early `'type'` branch, which duplicated logic handled later for essential validation fields. No behavior change; improves readability and fixes #4703.

<sup>Written for commit 4d44aafe3a663b9d5705c139d1316ddb1a2080a5. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/llm/schema.py
+++ b/browser_use/llm/schema.py
@@ -57,10 +57,6 @@ class SchemaOptimizer:
 						if value:  # Only include non-empty descriptions
 							optimized[key] = value
 
-					# Handle type field - must recursively process in case value contains $ref
-					elif key == 'type':
-						optimized[key] = value if not isinstance(value, (dict, list)) else optimize_schema(value, defs_lookup)
-
 					# FLATTEN: Resolve $ref by inlining the actual definition
 					elif key == '$ref' and defs_lookup:
 						ref_path = value.split('/')[-1]  # Get the definition name from "#/$defs/SomeName"


goingforstudying-ctrl:
CLA signed! Ready for review.

sauravpanda:
Thanks for the cleanup! Closing this one as superseded — #4736 already addressed the dead-code issue from #4703, and after it merged the
  situation has flipped:

  - The duplicate 'type' entry in the validation-fields list (the actually unreachable one) was removed in #4736.
  - The elif key == 'type': branch this PR deletes is now the live handler for 'type' keys, not dead code.

  The fallthrough to the final else clause is behaviorally identical (same logic, inverted ternary), so this wouldn't break anything — but it'd
  remove the explicit branch and its # must recursively process in case value contains $ref comment, and silently couple 'type' handling to
  whatever the catch-all does in the future. Prefer to keep the explicit branch.

  Appreciate you digging into #4703, feel free to grab another issue!

Issue #4721, opened 2026-04-21
Supply chain risk in init command: unverified remote code fetching


cloakmaster:
The `init` command in `browser_use/init_cmd.py` fetches Python template files from `raw.githubusercontent.com` at runtime and writes them to disk for execution. There's no integrity verification — no checksum, no signature, no pinned commit.

**What happens:**

1. `_fetch_template_list()` downloads `templates.json` from `browser-use/template-library` (line 37)
2. `_fetch_from_github()` downloads the actual Python file content
3. Content is written to disk and the user is told to run `python {output_path.name}`

If the `template-library` repo is compromised (via a maintainer account takeover, malicious PR, or GitHub token leak), every user running `browser-use init` would execute attacker-controlled code. This is the same class of supply chain attack that hit `ua-parser-js`, `event-stream`, and more recently `@anthropic-ai/claude-code`.

**Suggested fix:**

Either bundle templates locally in the package (they're small enough), or add integrity verification — hash the template content against a pinned manifest shipped with the package:

```python
import hashlib

KNOWN_HASHES = {
    "simple_agent.py": "sha256:abc123...",
}

content = _fetch_from_github(template_file)
actual = hashlib.sha256(content.encode()).hexdigest()
if f"sha256:{actual}" != KNOWN_HASHES.get(template_file):
    raise RuntimeError(f"Template integrity check failed for {template_file}")
```

Bundling is simpler and eliminates the network dependency entirely.

---

Separately, `browser_use/actor/element.py:283` logs form field values at DEBUG level in the `fill()` method — passwords, credit cards, etc. end up in logs if DEBUG is on. A quick `logger.debug(f'Typing [REDACTED {len(value)} chars]')` would fix that without losing debug utility.

---

Found during a scan with [Inkog](https://github.com/inkog-io/inkog).

PR #4736, opened 2026-04-23
fix(element): redact sensitive values in fill() debug logging

Will-hxw:
## Summary
- Redact sensitive form field values (passwords, etc.) from DEBUG logging in `fill()` method

## Why
The `fill()` method in `browser_use/actor/element.py` was logging the full value being typed at DEBUG level:
```
logger.debug(f'Typing text character by character: "{value}"')
```
This could expose passwords and other sensitive data in debug logs when users have DEBUG logging enabled.

## Fix
Changed to:
```
logger.debug(f'Typing text character by character: "[REDACTED {len(value)} chars]"')
```
This preserves debug utility (knowing that typing is happening and roughly how much) while avoiding exposure of sensitive data.

## Validation
- [x] pyright type check passes (0 errors)
- [x] ruff lint passes (0 errors)

## Related
Issue #4721 (security advisory - supply chain risk in init command also noted in same issue)

---
## Summary by cubic
Redacts sensitive values from `fill()` debug logs to prevent leaking passwords or secrets, while keeping useful context. Also cleans up schema optimization by removing an unreachable `'type'` key check.

- **Bug Fixes**
  - `browser_use/actor/element.py`: Log `"[REDACTED {len(value)} chars]"` instead of the actual text in `fill()` DEBUG logs.
  - `browser_use/llm/schema.py`: Remove dead `'type'` entry from validation fields; it’s already handled earlier.

<sup>Written for commit 99e9a455e77b9809c9358674555ac2ab87133b18. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/actor/element.py
+++ b/browser_use/actor/element.py
@@ -418,7 +418,7 @@ class Element:
 					logger.warning('Text field clearing failed, typing may append to existing text')
 
 			# Step 3: Type the text character by character using proper human-like key events
-			logger.debug(f'Typing text character by character: "{value}"')
+			logger.debug(f'Typing text character by character: "[REDACTED {len(value)} chars]"')
 
 			for i, char in enumerate(value):
 				# Handle newline characters as Enter key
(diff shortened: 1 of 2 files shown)


sauravpanda:
Awesome, thanks for the PR!

PR #4741, opened 2026-04-25
fix(anthropic-serializer): type tool_calls + raise on malformed data URL

GopalGB:
## Summary
Two small, orthogonal hardening fixes in `browser_use/llm/anthropic/serializer.py`:

### 1. Add `ToolCall` annotation on `_serialize_tool_calls_to_content`

The function had no type on its `tool_calls` parameter, leaving it implicitly `Any`. Callers always pass `list[ToolCall]` (per `AssistantMessage.tool_calls` in `browser_use/llm/messages.py:208`), so the annotation makes the contract explicit and lets static checkers verify the `.function.arguments` / `.function.name` / `.id` attribute accesses inside.

### 2. Raise a clear error on malformed base64 data URLs

`_parse_base64_url` did:
```python
header, data = url.split(',', 1)
```
without checking that the split actually produced two pieces. A user passing a header-only data URL like `data:image/jpeg;base64` (no comma + payload) would hit the unpacking and see the unhelpful

```
ValueError: not enough values to unpack (expected 2, got 1)
```

Now the function checks the split length first and raises a self-explaining `ValueError` that names the real cause and includes a prefix of the offending URL.

## Backward compatibility
Both changes are non-breaking — happy paths (well-formed URLs, callers already passing `list[ToolCall]`) behave identically. Only the error message + path on bad input changes.

## Verification
`python3 -c "import ast; ast.parse(open('browser_use/llm/anthropic/serializer.py').read())"` → exit 0

---
## Summary by cubic
Hardens the Anthropic serializer with stricter typing and clearer error handling.
Types tool_calls as list[ToolCall] in _serialize_tool_calls_to_content, types the tool input object as dict[str, object] to match ToolUseBlockParam, and makes _parse_base64_url raise a clear ValueError when a base64 data URL is missing the comma separator.

<sup>Written for commit 3b4a88185fb74c828668b2b1448ed470c8447ca1. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/llm/anthropic/serializer.py
+++ b/browser_use/llm/anthropic/serializer.py
@@ -18,6 +18,7 @@ from browser_use.llm.messages import (
 	ContentPartTextParam,
 	SupportedImageMediaType,
 	SystemMessage,
+	ToolCall,
 	UserMessage,
 )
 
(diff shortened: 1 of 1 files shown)


GopalGB:
I've signed the browser-use CLA via cla-assistant.io — should now show `signed` on the next recheck. CI is fully green (101/102 passing, evaluate-tasks skipped). Happy to address review feedback when reviewers have cycles. cc maintainers.

Issue #4739, opened 2026-04-24
Third-party architecture audit notes from agchk


huangrichao2020:
Hi maintainers — sharing a small third-party static architecture audit in case it is useful. Please feel free to close if this is not a useful tracking item for browser-use/browser-use.

I ran [`agchk`](https://github.com/huangrichao2020/agchk) against the current repository with the personal-development profile:

```bash
agchk /path/to/browser-use --profile personal
```

This is not a vulnerability disclosure and not a request to adopt another tool. It is a static architecture/methodology scan intended to surface agent-runtime complexity patterns.

## Summary

- Architecture era: `铁器时代`, `37/100`
- Share line: `这个 Agent 项目处于 铁器时代（37/100）：具备较清晰的工具、记忆和技能分层，开始可维护。`
- Overall health signal: `unstable`
- Finding counts: `critical=0`, `high=5`, `medium=11`, `low=32`

## Top static signals

- HIGH: Internal orchestration sprawl detected
- HIGH: Completion closure gap detected
- HIGH: Memory freshness / generation confusion detected
- HIGH: Context memory lacks paging policy
- HIGH: Runtime surface sprawl detected
- MEDIUM: Knowledge surfaces lack semantic VFS

## Caveats

- The findings read most usefully as runtime-complexity and memory-methodology signals, not as security claims.
- Static regex-style scans can over-count docs, examples, generated fixtures, provider adapters, or intentionally sandboxed execution paths.
- I did not run project tests or dynamic behavior checks; this is only a lightweight static architecture view.

## Why I am filing this

The goal is to share an external lens on agent architecture maturity: orchestration drag, memory freshness, completion closure, tool/runtime boundaries, and methodology density.

If any signal is useful, I can follow up with a narrowly scoped docs proposal or convert false positives into agchk regression tests so future scans are less noisy for mature agent projects like this one.

rehan243:
oh interesting, saw the “HIGH: Internal orchestration sprawl detected” bit — we've hit similar signals running agchk and also semgrep against agent-like repos. Honestly, orchestration sprawl gets wild fast, especially when you’ve got memory freshness and paging gaps stacked on top (your “context memory lacks paging policy” finding is spot on). We had to slap together a hacky paging middleware once just to keep agent memory from ballooning during multi-turn conversations, otherwise LLM context windows got trashed.

[...]

PR #4745, opened 2026-04-26
docs: document context controls for long-running agents

huangrichao2020:
## Summary

This follows up on #4739, especially the agreed audit signal around context memory / paging policy in https://github.com/browser-use/browser-use/issues/4739#issuecomment-4322153373.

The runtime already has useful controls here, but they were hard to discover from the README. This PR documents the practical knobs users can apply before long-running agents start ballooning their prompt context:

- `message_compaction=True` and `MessageCompactionSettings(...)`
- `max_history_items`
- `max_clickable_elements_length`
- `calculate_cost=True`
- file-heavy task guidance for filesystem-backed extraction instead of pasting entire documents into the prompt

## Why

The agchk audit called out "context memory lacks paging policy", and a maintainer/community comment noted that memory and context growth become real operational pain quickly. This README addition turns that audit point into a concrete user-facing runbook without changing runtime behavior.

## Validation

- `git diff --check`
- `python3 - <<'PY' ...` README marker check for the new documented options
- `uv run python - <<'PY' ...` instantiated `MessageCompactionSettings` with the README-style values and verified the derived char threshold

---
## Summary by cubic
Document practical context controls for long-running agents to prevent prompt bloat and manage cost. Adds a README FAQ with a code sample for `message_compaction`/`MessageCompactionSettings`, `max_history_items`, `max_clickable_elements_length`, `calculate_cost`, plus guidance to use filesystem-backed extraction for large files.

<sup>Written for commit 862cd1cf6841aea5852132982262210d964418dc. Summary will update on new commits.</sup>

### DIFF
--- a/README.md
+++ b/README.md
@@ -275,6 +275,40 @@ For CAPTCHA handling, you need better browser fingerprinting and proxies. Use [B
 </details>
 
 <details>
+<summary><b>How do I keep long-running agents from overloading the LLM context?</b></summary>
+
+Browser Use includes several context controls for long-running or multi-step runs:
+
+- `message_compaction=True` is enabled by default and summarizes older step history into a compact memory block.
+- Use `MessageCompactionSettings(...)` when you want explicit thresholds such as `trigger_token_count`, `compact_every_n_steps`, `keep_last_items`, and `summary_max_chars`.
+- Use `max_history_items` to cap how many raw step history items are kept in the prompt.
+- Tune `max_clickable_elements_length` if large pages are dominating the browser state message.
+- Set `calculate_cost=True` when you want token and cost telemetry for a run.
+
+```python
+from browser_use import Agent, ChatBrowserUse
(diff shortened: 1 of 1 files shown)


PHclaw:
For the element detection returning wrong element:

The issue is likely in the DOM simplification that loses structural context. The fix:

```python
# In the element detection function:
def simplify_for_llm(node, max_depth=5):
    # Preserve parent-child relationships better
    result = {
        'tag': node.tag,
        'id': node.id,
        'text': node.text[:200] if node.text else '',
        'role': node.role,
    }
    
    # Keep more context about siblings
    if node.parent:
        result['parent_tag'] = node.parent.tag
        result['parent_id'] = node.parent.id
    
    # Include index among siblings
    if node.parent and node in node.parent.children:
        result['sibling_index'] = list(node.parent.children).index(node)
    
    return result
```

This gives the LLM more context to correctly identify elements.

PR #4750, opened 2026-04-27
Support Arch Linux in CLI installer

gngeorgiev:
## Summary
- add pacman support when installing Python and Chromium system dependencies
- add support for dnf, zypper, and apk in the Python installer path
- skip Playwright's `--with-deps` path on Arch-like Linux systems, where dependency installation is handled through pacman instead

## Validation
- ran `bash -n browser_use/skill_cli/install.sh`
- validated the patched installer on CachyOS/Arch locally with `pacman`, `uv`, Playwright Chromium, profile-use, and `browser-use doctor`

## Context
The current installer fails on CachyOS/Arch because there is no `apt-get`, and Playwright's `--with-deps` path is not appropriate for pacman-only systems.

---
## Summary by cubic
Adds Arch Linux support to the CLI installer. Installs Python and Chromium deps via `pacman` and skips `playwright --with-deps` on Arch-like distros, fixing installs on Arch/CachyOS.

- **New Features**
  - Detects Arch-like Linux via `pacman` or `/etc/os-release`.
  - Installs Python with `pacman`; adds paths for `dnf`, `zypper`, and `apk`.
  - Adds `install_linux_browser_deps` to install Chromium deps with `pacman`.
  - Runs `playwright install chromium --with-deps` only on non-Arch Linux.

<sup>Written for commit 517e2eaac37c3103157c0f5331533566033f3b74. Summary will update on new commits. <a href="https://cubic.dev/pr/browser-use/browser-use/pull/4750?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/browser_use/skill_cli/install.sh
+++ b/browser_use/skill_cli/install.sh
@@ -161,6 +161,27 @@ detect_platform() {
 	log_info "Detected platform: $PLATFORM ($arch)"
 }
 
+is_arch_like_linux() {
+	if [ "$PLATFORM" != "linux" ]; then
+		return 1
+	fi
+
+	if command -v pacman &> /dev/null; then
+		return 0
+	fi
+
+	if [ -f /etc/os-release ]; then
+		. /etc/os-release
+		case "${ID:-} ${ID_LIKE:-}" in
+			*arch*)
+				return 0
+				;;
+		esac
+	fi
+
+	return 1
+}
+
 # =============================================================================
 # Virtual environment helpers
 # =============================================================================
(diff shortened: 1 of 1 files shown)


PHclaw:
For agent task timeouts, implement a timeout handler that gracefully stops the browser:

```python
import signal, asyncio

class TimeoutException(Exception):
    pass

def timeout_handler(signum, frame):
    raise TimeoutException("Task exceeded timeout")

async def run_agent_with_timeout(agent, task, timeout=60):
    signal.signal(signal.SIGALRM, timeout_handler)
    signal.alarm(timeout)  # Only works on Unix
    try:
        result = await agent.run(task)
        return result
    finally:
        signal.alarm(0)  # Cancel alarm
        await agent.browser.close()  # Clean up browser
```

For Windows (where SIGALRM doesn't work), use asyncio.wait_for():
```python
result = await asyncio.wait_for(agent.run(task), timeout=timeout)
```

PHclaw:
For the element detection timeout issue:

The problem is likely that Playwright's `wait_for_selector()` has a default timeout that's too short for slow-loading pages. The fix:

```python
# In the element detection code
await page.wait_for_selector(
    selector,
    timeout=30000  # 30 seconds instead of default 3s
)
```

Or implement a retry loop with exponential backoff:
```python
for attempt in range(3):
    try:
        element = await page.wait_for_selector(selector, timeout=10000)
        break
    except Exception:
        await page.reload()
        await asyncio.sleep(2 ** attempt)
```

PR #4764, opened 2026-04-29
fix: filter out chrome-extension:// targets in get_all_page_targets

PHclaw:
## Summary

Fixes #4133

When browser-use connects to an existing Chrome browser via CDP that has extensions with side panels, the side panel's page is reported by CDP as a target with `type: "page"` and a `chrome-extension://` URL. The `get_all_page_targets()` method only filters by target type (`page`/`tab`) but not by URL scheme, causing:

1. **Crash/detach recovery** (`_recover_agent_focus`) picks `page_targets[-1]`, which could be an extension side panel
2. **Agent navigation** attempts on non-navigable extension pages fail silently
3. **Tab listing** includes extension pages that the agent cannot interact with

## Fix

Add a URL scheme filter in `get_all_page_targets()` to skip targets with `chrome-extension://` URLs. These targets are reported as `type: "page"` by CDP but are not navigable browser pages.

## Test

```python
# Before: chrome-extension:// targets are included in page_targets
# After: they are filtered out
targets = session_manager.get_all_page_targets()
for t in targets:
    assert not t.url.startswith("chrome-extension://")
```

---
## Summary by cubic
Exclude `chrome-extension://` targets in `get_all_page_targets()` so extension side panels reported as CDP "page" targets are ignored. Previously these were included, which could misselect tabs during crash/detach recovery, trigger failed navigation attempts to non‑navigable pages, and show extension pages in tab lists; now only navigable web pages are returned. Fixes #4133.

<sup>Written for commit 1924cd2d2e821a148a96e25cee609f5a04edf40c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/4764?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


PHclaw:
CI passed. This is a small fix that prevents crashes when browser extensions with side panels are active. Would appreciate a review when possible.

PHclaw:
Bumping this - `chrome-extension://` targets in CDP page listings cause incorrect tab recovery and silent navigation failures on any Chrome instance with extensions (side panels, chat widgets, etc.). The fix is a single URL scheme filter in `get_all_page_targets()`. Would appreciate any feedback on what's needed to merge. Thanks!

PR #4771, opened 2026-04-30
fix: disable chromium sandbox for root on linux

kevinskysunny:
## Summary
- disable Chromium sandbox automatically when browser-use runs as root on Linux
- keep existing behavior unchanged for non-root users
- add regression tests for root vs non-root sandbox behavior

## Why
Running browser-use as root on Linux currently launches Chromium without `--no-sandbox`, which causes local browser startup to fail immediately with Chromium's root/sandbox restriction.

## Testing
- `uv run pytest tests/ci/test_browser_profile_root_sandbox.py -q`

---
## Summary by cubic
Automatically disable the Chromium sandbox when `browser-use` runs as root on Linux so local browser startup doesn’t fail. Outside Docker we now only add `--no-sandbox`; non-root behavior is unchanged and Docker runs keep full flags.

- **Bug Fixes**
  - In `BrowserProfile`, detect Linux root during init, log a warning, and set `chromium_sandbox=False`; use Docker flags only when `IN_DOCKER` is true, otherwise add only `--no-sandbox`.
  - Added tests to confirm sandbox is disabled for root, preserved for non-root, and that only `--no-sandbox` is added outside Docker.

<sup>Written for commit 91e8baf11113a63afbd9cc3bef911cceeb2bb666. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/browser/profile.py
+++ b/browser_use/browser/profile.py
@@ -93,6 +93,10 @@ CHROME_DOCKER_ARGS = [
 	'--disable-site-isolation-trials',  # lowers RAM use by 10-16% in docker, but could lead to easier bot blocking if pages can detect it?
 ]
 
+CHROME_NO_SANDBOX_ARGS = [
+	'--no-sandbox',
+]
+
 
 CHROME_DISABLE_SECURITY_ARGS = [
 	'--disable-site-isolation-trials',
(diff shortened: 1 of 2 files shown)


kevinskysunny:
Extra context from reproduction:

- On root/Linux, Chromium was being launched without `--no-sandbox`
- Chromium then exited immediately with:

```text
Running as root without --no-sandbox is not supported. See https://crbug.com/638180.
```

From the browser-use side this surfaced as a browser-start timeout, because the spawned browser process died before the CDP endpoint became reachable.

I also manually verified the fix locally in a root/Linux environment: after this patch, `browser-use open https://example.com` succeeds.

PR #4775, opened 2026-05-02
fix: enforce domain restrictions for data and blob URLs

### LINKED ISSUE #4763: Security: `data:` and `blob:` URLs bypass `allowed_domains` restriction
### Browser-Use Version

0.12.6 (latest main)

### Bug Description

```python
    return True
```

```
```

### Why This Matters

```
```

### Attack Scenario

### Suggested Fix

```python
    if (
    ):
        return True
    if parsed.scheme == 'data':
        return False
    if parsed.scheme == 'blob':
```

haosenwang1018:
Closes #4763

## Summary
- block `data:` URLs whenever `allowed_domains` or `prohibited_domains` is configured
- validate `blob:` URLs against the encoded origin instead of allowing them unconditionally
- keep existing behavior for `data:`/`blob:` URLs when no domain restrictions are configured

## Tests
- `uv run ruff check browser_use/browser/watchdogs/security_watchdog.py tests/ci/security/test_domain_filtering.py`
- `uv run pytest tests/ci/security/test_domain_filtering.py -q`
- `uv run pytest tests/ci/security -q`

---
## Summary by cubic
Enforce domain restrictions for data: and blob: URLs to prevent bypassing allow/block rules, while keeping previous behavior when no restrictions are set.

- **Bug Fixes**
  - Block `data:` URLs when `allowed_domains` or `prohibited_domains` is configured.
  - Validate `blob:` URLs against their encoded origin; reject cross-origin, `blob:null`, and non-allowed hosts.
  - Preserve old behavior for `data:`/`blob:` when no domain lists are set.
  - Add tests for restricted and unrestricted data/blob cases.

<sup>Written for commit 634463ba1c41feb2024bad3404a31c738a6af2a5. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/browser/watchdogs/security_watchdog.py
+++ b/browser_use/browser/watchdogs/security_watchdog.py
@@ -178,9 +178,17 @@ class SecurityWatchdog(BaseWatchdog):
 			# Invalid URL
 			return False
 
-		# Allow data: and blob: URLs (they don't have hostnames)
+		has_domain_restrictions = bool(
+			self.browser_session.browser_profile.allowed_domains or self.browser_session.browser_profile.prohibited_domains
+		)
+
+		# Allow data: and blob: URLs only when there are no domain restrictions.
 		if parsed.scheme in ['data', 'blob']:
-			return True
+			if not has_domain_restrictions:
+				return True
+			if parsed.scheme == 'data':
+				return False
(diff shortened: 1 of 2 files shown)


sauravpanda:
hey @haosenwang1018 can you please sign the CLA?

PR #4777, opened 2026-05-03
fix(init): reject path-traversal in remote templates.json fields

etairl:
## Summary
- The `dest`, `source`, `file`, and template-name fields in the runtime-fetched `templates.json` were concatenated straight into filesystem paths and URLs with no validation. `Path('/tmp/x') / Path('/etc/crontab')` silently yields `/etc/crontab`, so a manifest compromise turned into **arbitrary file write under the invoking user** (`~/.bashrc`, `/etc/cron.d/*`, `~/.ssh/authorized_keys`, …) the moment a user ran `browser-use init`.
- Adds `_validate_relative_path` + `_safe_join` and applies them to every manifest-controlled string before it touches disk or a URL.
  - Per-entry `files[]` violations are skipped with a warning so a single bad entry can't poison the whole template.
  - Top-level violations (`template`, `file`) abort with exit 1.
- The `template` directory under `cwd` and the `file` URL path also flow from manifest-controlled keys, so they get the same treatment.

## Why this matters
1. Attacker controls `templates.json` (compromised maintainer creds, malicious PR, branch hijack on `template-library@main` since the URL pins to `main`, etc.).
2. They publish an entry with `\"dest\": \"/etc/cron.d/pwn\"` or `\"dest\": \"../../../../home/victim/.bashrc\"`.
3. User runs `uvx browser-use init` and the loop in `main()` writes attacker-controlled bytes there. The `executable` flag also chmod's +x, and the binary branch bypasses the overwrite prompt entirely.

## Test plan
- [x] 19 new tests in `tests/ci/test_cli_install_init.py`, including:
  - parametrized rejection of absolute paths, `..` segments, empty strings, NUL bytes, backslashes
  - end-to-end scenario from the report: a malicious `templates.json` served via `pytest-httpserver`, asserting that pre-staged marker files at `/tmp/browser_use_pwn_*` remain absent after `init` runs
  - traversal in template name and template `file` aborts with exit 1
  - happy-path benign manifest still writes both files correctly
- [x] `uv run pytest -vxs tests/ci/test_cli_install_init.py` — 26 passed
- [x] `uv run ruff check` + `uv run ruff format --check` — clean
- [x] `uv run pyright` on touched files — 0 errors

---
## Summary by cubic
Fixes a critical path traversal in the init command by validating all manifest-controlled paths to prevent arbitrary file writes and unsafe URL fetches. Unsafe entries are blocked or skipped so a poisoned `templates.json` cannot escape the template directory.

- **Bug Fixes**
  - Added `_validate_relative_path` and `_safe_join`; applied to template name, `file`, and each `files[]` `source`/`dest`.
  - Reject absolute paths, `..` segments, empty strings, backslashes, and NUL bytes; enforce that joins stay inside the template directory.
  - Skip unsafe `files[]` entries with a warning; abort on unsafe template name or `file` with exit 1.

<sup>Written for commit d6c221d1a8207bc9fce2b0e11923adf9e4978f53. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/init_cmd.py
+++ b/browser_use/init_cmd.py
@@ -32,6 +32,43 @@ TEMPLATE_REPO_URL = 'https://raw.githubusercontent.com/browser-use/template-libr
 INIT_TEMPLATES: dict[str, Any] = {}
 
 
+def _validate_relative_path(name: str, *, field: str) -> str:
+	"""
+	Reject any path string from the remote manifest that is absolute, contains `..`,
+	or otherwise tries to escape the directory it will be joined onto.
+
+	The manifest is fetched at runtime from a third-party-controllable URL, so every
+	string that ends up in a filesystem path or remote URL must be a plain relative
+	name. Without this, `Path('/tmp/x') / Path('/etc/crontab')` silently yields
+	`/etc/crontab`, which is enough to turn a manifest compromise into arbitrary
+	file write on the user's machine the moment they run `browser-use init`.
+	"""
+	if not isinstance(name, str) or not name:
+		raise ValueError(f'Invalid {field} in template manifest: {name!r}')
+	candidate = Path(name)
+	if candidate.is_absolute() or candidate.drive or candidate.root:
+		raise ValueError(f'Refusing absolute {field} from template manifest: {name!r}')
+	if any(part in ('..', '') for part in candidate.parts):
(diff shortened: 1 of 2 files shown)


etairl:
Hi guys! Long time user of browser use, first time contributing. This was found as part of our product's dependency check. Would love having it reviewed :)

PR #4813, opened 2026-05-09
security(sandbox): replace cloudpickle param serialization with JSON allowlist (CWE-502)

andesyteoss:
## Summary

The `@sandbox` decorator in `browser_use/sandbox/sandbox.py` serializes captured parameters, closure variables, and referenced globals with `cloudpickle` and base64-embeds them into the Python source it ships to the remote sandbox stream server. The server-side wrapper then runs `cloudpickle.loads(base64.b64decode(...))` on that blob — an unsafe deserialization sink (CWE-502).

This PR replaces the cloudpickle round-trip with strict JSON-only serialization and an allowlist of safe value types.

- **CWE:** CWE-502 (Deserialization of Untrusted Data)
- **File / function:** `browser_use/sandbox/sandbox.py` — `_extract_all_params` + the `execution_code` template emitted by `sandbox(...)`
- **Sink:** `cloudpickle.loads(base64.b64decode(_pickled_params))` executed inside the generated `execution_code` on the sandbox runner.
- **Source:** any object reachable via the decorated function's closure (`__closure__`) or `func.__globals__[name]` for names referenced in `co_names` — these are auto-captured without filtering, so a malicious `__reduce__` on any captured global/closure object becomes RCE during deserialization on the runner.

## Fix

1. `_extract_all_params` now skips modules, classes, functions, methods, builtins, and method descriptors when auto-capturing globals — these are restored on the runner via the existing `needed_imports` machinery, not shipped as data.
2. New `_to_json_safe_value(value, path)` recursively validates each parameter against an allowlist: `None`, `bool`, `int`, finite `float`, `str`, `list`/`tuple` (converted to list), and `dict` with string keys. Anything else raises `SandboxError` with a clear path so developers can fix the call site.
3. New `_json_serialize_params` emits compact JSON (no NaN/Infinity).
4. The generated `execution_code` template no longer imports `cloudpickle` or `base64` and instead does `_params = json.loads(...)`. The `cloudpickle` import is removed from `sandbox.py`.

The public contract (decorator signature, calling convention, browser injection, return-value handling) is unchanged. Only the wire format for params is changed, and parameters were always intended to be data, not code.

## Tests

Added `tests/ci/security/test_sandbox_serialization.py` with two cases, both passing locally:

- `test_sandbox_payload_uses_json_not_cloudpickle` — patches `httpx.AsyncClient.stream`, runs a decorated task, base64-decodes the `code` field actually sent on the wire, and asserts (a) no occurrence of `cloudpickle` or `pickle.loads`, (b) presence of `json.loads`, and (c) AST-walks the generated code to confirm no `pickle.loads` / `cloudpickle.loads` call nodes exist.
- `test_sandbox_rejects_non_json_serializable_params` — passing a custom class instance raises `SandboxError` referencing the unsupported type, instead of silently being pickled.

```
tests/ci/security/test_sandbox_serialization.py::test_sandbox_payload_uses_json_not_cloudpickle PASSED
tests/ci/security/test_sandbox_serialization.py::test_sandbox_rejects_non_json_serializable_params PASSED
```

I also grepped `browser_use/` to confirm no remaining `cloudpickle` or `pickle.loads` usage after the change.

## Security analysis

The exploit primitive is straightforward: cloudpickle invokes `__reduce__` on objects during deserialization. Because the previous `_extract_all_params` walks `func.__code__.co_names` and indiscriminately pulls every matching entry from `func.__globals__` (plus every cell in `__closure__`), an attacker who can place a crafted object into the developer's module globals or into a closure cell of a `@sandbox`-decorated function — for example through a poisoned dependency or a configuration loader that returns attacker-controlled objects — gets code execution on the sandbox runner the moment that decorated function is invoked.

The fix narrows the wire format to pure JSON data, so deserialization on the runner becomes `json.loads`, which has no callable hooks. Modules/classes/functions referenced by the decorated function are no longer shipped as objects at all; they're re-imported from source on the runner via the unchanged `needed_imports` block. Anything that isn't a JSON primitive is now rejected client-side with a typed error before any network call.

### Adversarial review

Before submitting we tried hard to disprove this. The strongest counter-argument is that the sandbox stream endpoint is by design an arbitrary-code execution service — the payload it receives is `base64(execution_code)` containing Python source it will `exec`, so an attacker who already controls what the SDK sends could just edit the function body. That's true for the *caller* trust boundary, but it's not the boundary CWE-502 crosses here: the realistic threat is a malicious object reaching the *client process* (supply-chain dependency, config object, cached value) and being silently swept into the auto-captured globals/closure on the next `@sandbox` call, with deserialization happening on a remote runner the developer doesn't control. Removing the pickle sink is genuine defense-in-depth: it eliminates the gadget primitive without breaking any legitimate use, since parameters were always documented as data. We also confirmed no tests or examples in-tree rely on passing non-JSON-serializable objects through the decorator.

cc @lewiswigmore

---
## Summary by cubic
Replaced `cloudpickle` parameter serialization in the `@sandbox` decorator with strict JSON serialization using a safe type allowlist. This removes the unsafe deserialization sink (CWE-502) without changing the public API.

- **Bug Fixes**
  - Auto-captured globals now skip modules, classes, functions/methods, and builtins; these are re-imported on the runner.
  - Parameters are validated and serialized with JSON only; non-JSON types raise `SandboxError` with a precise path.
  - Generated runner code uses `json.loads` and removes `cloudpickle`/`pickle`; tests assert this path and reject unsupported params.

- **Migration**
  - Ensure all args, closure values, and globals used as data are JSON-serializable: null, bool, int, finite float, str, list/tuple, dict with string keys.
  - Pass plain data instead of objects; rely on imports for modules/classes/functions.
  - No API changes; only the wire format changed for params in `@sandbox`.

<sup>Written for commit 99bcd17e7e28aebbfe3c67a63b5442e8336c4ba3. Summary will update on new commits.</sup>

### DIFF
--- a/browser_use/sandbox/sandbox.py
+++ b/browser_use/sandbox/sandbox.py
@@ -5,6 +5,7 @@ import dataclasses
 import enum
 import inspect
 import json
+import math
 import os
 import sys
 import textwrap
@@ -12,7 +13,6 @@ from collections.abc import Callable, Coroutine
 from functools import wraps
 from typing import TYPE_CHECKING, Any, Concatenate, ParamSpec, TypeVar, Union, cast, get_args, get_origin
 
-import cloudpickle
 import httpx
 
 from browser_use.sandbox.views import (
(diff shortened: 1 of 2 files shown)


andesyteoss:
Hi — just checking in on the stale notification.

**PR status:**
- The code changes (2 files: `browser_use/sandbox/sandbox.py` and `tests/ci/security/test_sandbox_serialization.py`) rebase cleanly on top of the latest `main` — no merge conflicts.
- Both security tests still pass after rebase (`test_sandbox_payload_uses_json_not_cloudpickle`, `test_sandbox_rejects_non_json_serializable_params`).
- The automated review (cubic-dev-ai) found no issues across the changed files.

**Remaining items before merge:**
1. **CLA** — the CLA assistant shows `not_signed`. I'll sign the [Contributor License Agreement](https://cla-assistant.io/browser-use/browser-use?pullRequest=4813) to unblock that check.
2. **Force-push rebased branch** — I'll push the rebased branch shortly to update the PR against latest `main`.

This PR is still actively maintained — please keep it open. Thanks!

PR #4841, opened 2026-05-16
Require explicit opt-in for real browser CLI access

aryanputta:
Summary
- add an explicit confirmation step before the CLI attaches to a running Chrome instance
- add the same opt-in gate before launching with a real Chrome profile
- add a global --yes escape hatch for automation and CI
- add focused CLI tests for the new behavior

Why
The CLI currently makes it easy to attach to a live Chrome session or launch against a real local Chrome profile without an explicit trust boundary. Those flows can expose open tabs, cookies, saved logins, and other local browsing state.

This PR keeps the change narrow. It does not block cloud flows or normal managed-browser usage. It only requires explicit opt-in when the CLI is about to access real local browser state.

Testing
- python3 -m py_compile browser_use/skill_cli/main.py tests/ci/test_cli_cloud_connect.py tests/ci/test_cli_connect.py tests/ci/test_cli_headed_flag.py
- python3 -m pytest tests/ci/test_cli_connect.py tests/ci/test_cli_cloud_connect.py tests/ci/test_cli_headed_flag.py -q (blocked locally: missing pytest_httpserver from tests/ci/conftest.py in this environment)

---
## Summary by cubic
Require explicit opt-in before the CLI attaches to a running Chrome or launches with a real local profile to protect local browsing data. Adds a global `--yes` flag and a `BROWSER_USE_YES` env var (true/yes/1), improves the warning text, and fails non-interactive runs without opt-in; cloud and managed-browser flows are unchanged.

- **Migration**
  - For scripts/CI using connect or --profile, pass `--yes` or set `BROWSER_USE_YES=true/yes/1`.
  - Interactive runs show a prompt; answer y to continue.

<sup>Written for commit 9b4f2e57840044897728ff9a467648d6d1bf7488. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/4841?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/skill_cli/main.py
+++ b/browser_use/skill_cli/main.py
@@ -181,6 +181,34 @@ def _get_pid_path(session: str = 'default') -> Path:
 	return _get_home_dir() / f'{session}.pid'
 
 
+def _env_truthy(name: str) -> bool:
+	"""Return True when an environment variable is set to a truthy value."""
+	return os.environ.get(name, '').strip().lower()[:1] in 'ty1'
+
+
+def _confirm_real_browser_access(action: str, data_scope: str) -> bool:
+	"""Require explicit confirmation before touching a real browser session/profile."""
+	if _env_truthy('BROWSER_USE_YES'):
+		return True
+
+	if not sys.stdin.isatty():
+		print(
+			f'Error: refusing to {action} non-interactively without --yes. '
+			f'This would expose {data_scope}.',
+			file=sys.stderr,
+		)
+		return False
+
+	print(f'Warning: this will {action}.', file=sys.stderr)
+	print(f'It can expose {data_scope}.', file=sys.stderr)
+	try:
+		reply = input('Continue? [y/N] ').strip().lower()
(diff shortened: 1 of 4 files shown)


aryanputta:
Follow-up fix pushed in b985004.

The Cubic finding was valid: _env_truthy() treated unset/empty values as truthy because '' in 'ty1' is True in Python, which bypassed the new confirmation gate.

This update now does two things:
- extracts the first normalized character into a local val
- returns bool(val) and val in 'ty1'

I also validated the exact cases locally:
- unset / empty / 0 / false / no => False
- 1 / true / yes => True

PR #4859, opened 2026-05-18
fix(browser): exclude chrome-extension:// targets from get_all_page_targets

mrvenom17:
## Problem

When browser-use connects to a Chrome instance that has extensions with side panels (Chrome `sidePanel` API), the side panel is reported by CDP as a `type: "page"` target with a `chrome-extension://` URL. The agent intermittently switches focus into the extension window:

1. Crash/detach recovery picks `page_targets[-1]` — if the side panel opened most recently, the agent starts navigating inside the extension window.
2. The side panel appears in the tab list shown to the LLM, which may cause the LLM to switch to it.
3. `connect()` may select the side panel as the initial page if it is the first discovered page target.

It is intermittent because it depends on the timing of the side panel opening relative to target discovery and crash recovery.

## Root cause

`SessionManager.get_all_page_targets()` (`browser_use/browser/session_manager.py:150`) filtered only by `target_type in ('page', 'tab')` with no URL-scheme check. `BrowserSession._is_valid_target()` already excludes `chrome-extension://` by default (`include_chrome_extensions=False`), but that method only guards `_cdp_get_all_pages()` / `get_all_frames()` — not `get_all_page_targets()`, which backs ~13 call sites including `crash_watchdog`, the tab list, and `connect()`.

## Fix

Exclude targets whose URL starts with `chrome-extension://` in `get_all_page_targets()`, mirroring the existing `_is_valid_target(include_chrome_extensions=False)` default. One-line behavioral change; no new dependencies; no refactor. `Target.url` is a pydantic field defaulting to `'about:blank'`, so the `.startswith()` check is always safe.

## Testing

Added `tests/ci/browser/test_session_manager_targets.py` using real `Target` objects (no mocking): asserts extension side panels are excluded while regular `http`/`https`/`about:blank`/`tab` targets are still returned. Verified the test fails without the fix and passes with it. `tests/ci/browser/test_tabs.py` (5 tests, exercises `get_all_page_targets` via tab listing) still passes. `uv run pyright` is clean.

## Related issues

Fixes #4133

---
## Summary by cubic
Exclude `chrome-extension://` targets from `SessionManager.get_all_page_targets()` by default to prevent switching into extension side panels. Add an opt-in `include_chrome_extensions=True` for cleanup so extension options/onboarding pages are still auto-closed; aligns with `BrowserSession._is_valid_target()` and fixes #4133.

- **Bug Fixes**
  - Skip `chrome-extension://` for `page`/`tab` targets by default; prevents side panels in crash recovery, the tab list, and initial connect.
  - Add `include_chrome_extensions` opt-in and use it in `_close_extension_options_pages()`; tests cover exclusion and opt-in behavior.

<sup>Written for commit 5f2c9b3b378acace160ea93706f4520ca69efeb5. Summary will update on new commits. <a href="https://cubic.dev/pr/browser-use/browser-use/pull/4859?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/browser_use/browser/session.py
+++ b/browser_use/browser/session.py
@@ -3218,8 +3218,9 @@ class BrowserSession(BaseModel):
 	async def _close_extension_options_pages(self) -> None:
 		"""Close any extension options/welcome pages that have opened."""
 		try:
-			# Get all page targets from SessionManager
-			page_targets = self.session_manager.get_all_page_targets()
+			# Get all page targets from SessionManager, including chrome-extension://
+			# targets so extension options/onboarding pages can be detected and closed.
+			page_targets = self.session_manager.get_all_page_targets(include_chrome_extensions=True)
 
 			for target in page_targets:
 				target_url = target.url
--- a/browser_use/browser/session_manager.py
+++ b/browser_use/browser/session_manager.py
@@ -155,7 +155,10 @@ class SessionManager:
 		"""
 		page_targets = []
 		for target in self._targets.values():
-			if target.target_type in ('page', 'tab'):
+			# Exclude chrome-extension:// targets (e.g. extension side panels expose
+			# a type='page' target). They must never be treated as steerable pages,
+			# matching BrowserSession._is_valid_target(include_chrome_extensions=False).
+			if target.target_type in ('page', 'tab') and not target.url.startswith('chrome-extension://'):
 				page_targets.append(target)
 		return page_targets
 
--- a/browser_use/browser/session_manager.py
+++ b/browser_use/browser/session_manager.py
@@ -147,8 +147,15 @@ class SessionManager:
 			return None
 		return self._sessions.get(next(iter(session_ids)))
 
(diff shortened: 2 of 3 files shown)


mrvenom17:
Thanks for the catch — this was a real regression in the PR.

The blanket `chrome-extension://` exclusion in `get_all_page_targets()` also hid those targets from `_close_extension_options_pages()` (`session.py:3222`), which relies on them to auto-close stray extension options/welcome/onboarding pages.

Resolved in the latest push (`c2e93f64`) using the opt-in approach: `get_all_page_targets(include_chrome_extensions: bool = False)` — the default keeps the #4133 fix safe for all steerable-page call sites, mirroring the existing `BrowserSession._is_valid_target(include_chrome_extensions=False)` convention, and `_close_extension_options_pages()` now passes `include_chrome_extensions=True`. Added `test_include_chrome_extensions_opt_in` so the cleanup path stays covered.

mrvenom17:
any reviews...

PR #4875, opened 2026-05-20
fix(cdp): prevent indefinite hangs on remote browsers via DOMWatchdog timeouts

### LINKED ISSUE #4579: Bug: CDP connection instability causing indefinite hangs with remote browsers

Parth-001-git:
So I tracked down the hang issue around remote CDP sessions getting stuck indefinitely when the websocket silently degrades in Docker/cloud setups.

The main problem was that a few DOMWatchdog calls were awaiting CDP responses forever with no timeout boundaries, so once the socket went half-open the event bus lock never got released and parallel browser sessions just piled up behind it.

What's working now:

* Added timeout guards around the main CDP execute paths
* Mark degraded sessions unhealthy automatically
* Added lightweight connection health verification before watchdog execution
* Prevented the watchdog from blocking the whole event loop indefinitely

One annoying thing I hit while testing was reproducing the half-open websocket state consistently. Local runs were fine because TCP teardown happens immediately on localhost. Ended up simulating packet drops inside Docker/WSL to reproduce the actual hanging behavior reliably.

Also hit a small WSL issue with pre-commit expecting python3.11 specifically, but that was just local env noise — ruff + formatting + targeted tests are all passing.

Core fix looks like this now:

```python
result = await asyncio.wait_for(
    cdp_session.execute(
        Runtime.evaluate(
            expression="window.performance.getEntriesByType('resource')",
            await_promise=True
        )
    ),
    timeout=5.0,  # remote CDP calls shouldn't block forever
)

except asyncio.TimeoutError:
    logger.warning("CDP session timed out while fetching network requests")

    # connection probably went half-open behind proxy/load balancer
    self.browser_session._is_healthy = False

    raise ConnectionError(
        "Remote browser connection became unresponsive"
    )
```

I also added tests around timeout handling + connection health verification so we don't regress on this later.

---
## Summary by cubic
Prevents remote CDP sessions from hanging by adding bounded timeouts and proactive health checks, so half‑open WebSockets in Docker/cloud no longer stall the event loop. Parallel sessions now recover or fail fast. Fixes #4579.

- **Bug Fixes**
  - Added `_is_healthy` flag that gates `is_cdp_connected`, plus `verify_connection_health()` (1.5s `Browser.getVersion` ping).
  - Watchdog circuit breaker verifies health before handler runs; on failure marks unhealthy and auto‑reconnects, or skips if the browser was intentionally stopped.
  - Bounded timeouts for key CDP/DOM paths (e.g., 5s around `get_or_create_cdp_session`, network request checks, DOM build, screenshot); on timeout mark session unhealthy and raise `ConnectionError`.
  - Tests in `tests/ci/test_connection_health.py` cover health flag lifecycle, ping success/timeout, and DOMWatchdog timeout handling.

<sup>Written for commit 4bc538d32827d2bbc1044ec338eb32e3cab8d3a7. Summary will update on new commits. <a href="https://cubic.dev/pr/browser-use/browser-use/pull/4875?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/browser_use/browser/session.py
+++ b/browser_use/browser/session.py
@@ -473,10 +473,13 @@ class BrowserSession(BaseModel):
 	def is_cdp_connected(self) -> bool:
 		"""Check if the CDP WebSocket connection is alive and usable.
 
-		Returns True only if the root CDP client exists and its WebSocket is in OPEN state.
+		Returns True only if the root CDP client exists, is marked healthy,
+		and its WebSocket is in OPEN state.
 		A dead/closing/closed WebSocket returns False, preventing handlers from dispatching
 		CDP commands that would hang until timeout on a broken connection.
 		"""
+		if not getattr(self, '_is_healthy', True):
+			return False
 		if self._cdp_client_root is None or self._cdp_client_root.ws is None:
 			return False
 		try:
@@ -484,6 +487,25 @@ class BrowserSession(BaseModel):
 
 			return self._cdp_client_root.ws.state is State.OPEN
 		except Exception:
+			return False
+
+	async def verify_connection_health(self) -> bool:
+		"""Verify CDP connection is responsive via a lightweight Browser.getVersion ping.
+
+		Returns True if the connection responds within 1.5s, False otherwise.
+		"""
+		if self._cdp_client_root is None or self._cdp_client_root.ws is None:
+			return False
+		if not getattr(self, '_is_healthy', True):
+			return False
+		try:
+			await asyncio.wait_for(
+				self._cdp_client_root.send.Browser.getVersion(),
+				timeout=1.5,
+			)
+			return True
+		except Exception as e:
+			self.logger.warning(f'⚠️ CDP health verification failed: {type(e).__name__}: {e}')
 			return False
 
 	async def wait_if_captcha_solving(self, timeout: float | None = None) -> 'CaptchaWaitResult | None':
@@ -562,6 +584,7 @@ class BrowserSession(BaseModel):
 	_reconnect_lock: asyncio.Lock = PrivateAttr(default_factory=asyncio.Lock)
 	_reconnect_task: asyncio.Task | None = PrivateAttr(default=None)
 	_intentional_stop: bool = PrivateAttr(default=False)
+	_is_healthy: bool = PrivateAttr(default=True)
 
 	_logger: Any = PrivateAttr(default=None)
 
@@ -629,6 +652,7 @@ class BrowserSession(BaseModel):
 				self.logger.debug(f'Error closing CDP client during reset: {e}')
 
 		self._cdp_client_root = None  # type: ignore
+		self._is_healthy = True
 		self._cached_browser_state_summary = None
 		self._cached_selector_map.clear()
 		self._downloaded_files.clear()
--- a/browser_use/browser/watchdog_base.py
+++ b/browser_use/browser/watchdog_base.py
@@ -95,7 +95,7 @@ class BaseWatchdog(BaseModel):
 				# Circuit breaker: skip handler if CDP WebSocket is dead
 				# (prevents handlers from hanging on broken connections until timeout)
 				# Lifecycle events are exempt — they manage browser start/stop
-				if event.event_type not in LIFECYCLE_EVENT_NAMES and not browser_session.is_cdp_connected:
+				if event.event_type not in LIFECYCLE_EVENT_NAMES:
 					# If reconnection is in progress, wait for it instead of silently skipping
 					if browser_session.is_reconnecting:
 						wait_timeout = browser_session.RECONNECT_WAIT_TIMEOUT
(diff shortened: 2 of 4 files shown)


Parth-001-git:
Hey, I dug into #4579 around remote CDP sessions hanging indefinitely on degraded websocket connections.

The main issue seemed to be that several DOMWatchdog CDP calls were awaiting without timeout boundaries, so once a remote connection went half-open the watchdog could block the event bus and stall parallel sessions.

I pushed a fix that:

* wraps the critical CDP execute paths with `asyncio.wait_for`
* marks degraded sessions unhealthy on timeout
* adds lightweight connection health verification before watchdog execution

Also added tests around timeout handling + connection recovery flow.

Would appreciate feedback on the timeout/reconnect strategy, especially around whether you'd prefer more aggressive session invalidation or softer recovery behavior for pooled remote browsers.

sauravpanda:
this is awesome, can you remove unrelated changes to your PR like spacing and stuff, doesnt make sense.

Parth-001-git:
Yeah, some extra formatting changes slipped in locally. I’ll clean the diff up and keep it scoped to the timeout/recovery fix.

Parth-001-git:
Cleaned up the unrelated formatting changes and kept the diff scoped to the timeout/recovery logic + tests.

PR #4883, opened 2026-05-22
fix: use timezone-aware config timestamps

Ghraven:
## Summary

Replace the `DBStyleEntry.created_at` default factory with a timezone-aware UTC timestamp and add a small regression test.

## Bug

`datetime.utcnow()` is deprecated in Python 3.12 because it returns a naive datetime. New config entries currently serialize `created_at` without timezone information, which can emit deprecation warnings and makes the timestamp ambiguous.

## Before/After

Before:
```python
created_at: str = Field(default_factory=lambda: datetime.utcnow().isoformat())
```

After:
```python
created_at: str = Field(default_factory=lambda: datetime.now(timezone.utc).isoformat())
```

Generated config timestamps remain UTC, but now include an explicit `+00:00` offset.

## Verification

- `python -m compileall -q browser_use/config.py tests/ci/test_config.py`
- `git diff --check`

I also attempted the focused pytest command:

```bash
uv run pytest tests/ci/test_config.py
```

but this local environment is missing the repo test plugin `pytest_httpserver` loaded by `tests/ci/conftest.py`, so collection stops before this test runs.

---
## Summary by cubic
Make config timestamps UTC and timezone-aware to avoid deprecation warnings and ambiguous times. Adds a small test to lock this in.

- **Bug Fixes**
  - Replace `datetime.utcnow().isoformat()` with `datetime.now(timezone.utc).isoformat()` in `DBStyleEntry.created_at`.
  - Add a test to ensure `created_at` is timezone-aware and uses the UTC offset.

<sup>Written for commit 09803dd0a9816818ac635f84ebcbb8ef2cd8b844. Summary will update on new commits. <a href="https://cubic.dev/pr/browser-use/browser-use/pull/4883?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/browser_use/config.py
+++ b/browser_use/config.py
@@ -3,7 +3,7 @@
 import json
 import logging
 import os
-from datetime import datetime
+from datetime import datetime, timezone
 from functools import cache
 from pathlib import Path
 from typing import Any
@@ -248,7 +248,7 @@ class DBStyleEntry(BaseModel):
 
 	id: str = Field(default_factory=lambda: str(uuid4()))
 	default: bool = Field(default=False)
-	created_at: str = Field(default_factory=lambda: datetime.utcnow().isoformat())
+	created_at: str = Field(default_factory=lambda: datetime.now(timezone.utc).isoformat())
 
 
 class BrowserProfileEntry(DBStyleEntry):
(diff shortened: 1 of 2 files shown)


Ghraven:

Ghraven:
Hi maintainers, just keeping this from going stale. This still looks like a small compatibility cleanup to me, but I’m happy to rebase, revise, or close it if it no longer fits the project direction. Thanks for maintaining browser-use.

PR #4892, opened 2026-05-23
feat(llm): surface cached input tokens for OpenAI-compatible providers

sauravpanda:
## Context

OpenAI introduced automatic prefix caching for chat completions in late 2024 — prompts of ≥1024 tokens that share a byte-stable prefix with a recent call get the matching portion billed at half price, and the response reports the hit via:

```
usage.prompt_tokens_details.cached_tokens
```

Most OpenAI-compatible providers cloned the convention. Groq, Cerebras, and Mistral all surface `usage.prompt_tokens_details.cached_tokens` (when their cache hits) on chat-completion responses. The discount lands on the invoice either way — but our wrappers for those three were dropping the field on the floor, so `ChatInvokeUsage.prompt_cached_tokens` always came back `None` and there was no way to observe cache effectiveness from inside the agent.

## Change

Read `prompt_tokens_details.cached_tokens` defensively in:

- `browser_use/llm/groq/chat.py`
- `browser_use/llm/cerebras/chat.py`
- `browser_use/llm/mistral/chat.py`

`getattr` / `dict.get` rather than typed attribute access, so older SDK builds that don't type `prompt_tokens_details` still work. This matches the pattern already used in `openrouter/chat.py` and `vercel/chat.py` — just propagating the existing convention to the remaining wrappers.

## Already wired (no change needed)

OpenAI, Anthropic, Google, Azure, OpenRouter, Vercel, LiteLLM, AWS (Anthropic), and the in-house BrowserUse provider all already surface their respective cached-token fields. Audit done in the same pass.

## Out of scope

- **DeepSeek**: returns `usage=None` entirely from the relevant code path — surfacing its native `prompt_cache_hit_tokens` / `prompt_cache_miss_tokens` needs a bigger plumbing change. Follow-up.
- **Ollama**: same, returns `usage=None`. Ollama doesn't have prefix caching, but the prompt/completion counts themselves are also dropped. Follow-up.
- **AWS Bedrock raw / OCI**: APIs don't report cached tokens at all.

## Test plan

- [x] New unit tests at `tests/ci/models/test_openai_compat_cached_tokens.py` — assert each parser reads the field when present and stays `None` when absent (7 cases).
- [x] All existing provider model tests still pass (`pytest tests/ci/models/`).
- [x] pyright + ruff clean via pre-commit.
- [ ] Real-API smoke run against a Groq prefix-cached model isn't in CI but the field shape is contracted by the OpenAI SDK type, which Groq mirrors.

---
## Summary by cubic
Surface cached input tokens for OpenAI‑compatible providers so agents can see prefix‑cache savings in `ChatInvokeUsage.prompt_cached_tokens`.

- **New Features**
  - Read `usage.prompt_tokens_details.cached_tokens` for `groq`, `cerebras`, and `mistral`.

- **Bug Fixes**
  - Handle dict-shaped `prompt_tokens_details` in `groq` and `cerebras` parsers (older SDKs/proxies).
  - Add regression tests for both shapes; confirm `None` when the field is absent.

<sup>Written for commit cb3feab2ecc5d3c87f49cc2951689065a635c1c1. Summary will update on new commits. <a href="https://cubic.dev/pr/browser-use/browser-use/pull/4892?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/browser_use/llm/cerebras/chat.py
+++ b/browser_use/llm/cerebras/chat.py
@@ -62,9 +62,14 @@ class ChatCerebras(BaseChatModel):
 		if response.usage is None:
 			return None
 		# Cerebras uses the OpenAI-compatible SDK shape and surfaces
-		# `usage.prompt_tokens_details.cached_tokens` when its prefix cache hits.
+		# `usage.prompt_tokens_details.cached_tokens` when its prefix cache hits. The nested
+		# payload may arrive as a typed object or as a plain dict depending on SDK version /
+		# proxy — handle both shapes.
 		prompt_details = getattr(response.usage, 'prompt_tokens_details', None)
-		cached_tokens = getattr(prompt_details, 'cached_tokens', None) if prompt_details else None
+		if isinstance(prompt_details, dict):
+			cached_tokens = prompt_details.get('cached_tokens')
+		else:
+			cached_tokens = getattr(prompt_details, 'cached_tokens', None) if prompt_details else None
 		return ChatInvokeUsage(
 			prompt_tokens=response.usage.prompt_tokens,
 			prompt_cached_tokens=cached_tokens,
--- a/browser_use/llm/cerebras/chat.py
+++ b/browser_use/llm/cerebras/chat.py
@@ -59,18 +59,20 @@ class ChatCerebras(BaseChatModel):
 		return self.model
 
 	def _get_usage(self, response: ChatCompletion) -> ChatInvokeUsage | None:
-		if response.usage is not None:
-			usage = ChatInvokeUsage(
-				prompt_tokens=response.usage.prompt_tokens,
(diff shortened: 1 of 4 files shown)


sauravpanda:
Good catch from @chatgpt-codex-connector — fixed in 9fdc926d.

The previous defensive read only used `getattr`, so the older-SDK / proxied-response shape (where `prompt_tokens_details` arrives as a plain dict) would silently skip cache hits. Now branches on `isinstance(prompt_details, dict)` for both Groq and Cerebras. Mistral was already dict-safe.

Added two regression tests pinning the dict-shaped path (`test_groq_surfaces_cached_tokens_when_details_is_dict`, `test_cerebras_surfaces_cached_tokens_when_details_is_dict`). All 9 tests green.

PR #4907, opened 2026-05-24
fix: prevent credential loss when domain-scoped sensitive_data shares key names

devteamaegis:
## Bug

`collect_sensitive_data_values` in `browser_use/utils.py` flattened domain-scoped sensitive data by writing directly into `sensitive_values[key]`. When two domains define a credential under the same key name — which is the common case (e.g. both `site-a.com` and `site-b.com` have a `password` or `token` field) — each iteration overwrote the previous value. Only the last domain's secret survived; the earlier ones were silently dropped and **never redacted from logs**.

**Minimal reproduction:**
```python
sensitive_data = {
    "site-a.com": {"token": "token-aaa"},
    "site-b.com": {"token": "token-bbb"},
}
result = collect_sensitive_data_values(sensitive_data)
# Before fix: {"token": "token-bbb"}   ← token-aaa is lost and will appear in logs
# After fix:  {"site-a.com:token": "token-aaa", "site-b.com:token": "token-bbb"}
```

## Fix

Namespace each domain-scoped value as `{domain}:{key}` so all secrets are retained independently. The `redact_sensitive_string` function replaces values (not keys), so the namespace only appears in the `<secret>…</secret>` placeholder in sanitised output — which is actually more informative, because it identifies which domain the credential belongs to.

Flat-format entries (`{"api_key": "value"}`) are unchanged.

## Tests

Added `tests/test_utils.py` with 9 unit tests covering:
- flat format unchanged
- single-domain scoped format
- multi-domain collision case (regression test for this bug)
- empty-value skipping
- mixed flat + domain-scoped
- redaction correctness for multi-domain secrets

All 9 pass.

---
## Summary by cubic
Fixes credential loss in `collect_sensitive_data_values` by namespacing domain-scoped keys as `{domain}:{key}` so same-named secrets across domains don’t overwrite each other and are all redacted from logs.

- **Bug Fixes**
  - Namespace nested domain entries to avoid key collisions; flat keys unchanged.
  - Added tests for multi-domain collisions, mixed formats, empty values, and redaction correctness.

<sup>Written for commit 64358259029701c0ae776ec854d3551b860fbbb1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/4907?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/utils.py
+++ b/browser_use/utils.py
@@ -32,7 +32,14 @@ _groq_bad_request_error: type | None = None
 
 
 def collect_sensitive_data_values(sensitive_data: dict[str, str | dict[str, str]] | None) -> dict[str, str]:
-	"""Flatten legacy and domain-scoped sensitive data into placeholder -> value mappings."""
+	"""Flatten legacy and domain-scoped sensitive data into placeholder -> value mappings.
+
+	For domain-scoped entries (nested dicts), keys are namespaced as ``{domain}:{key}``
+	so that identically-named credentials from different domains (e.g. a ``password``
+	for ``site-a.com`` and a ``password`` for ``site-b.com``) are both retained and
+	redacted independently.  Without the namespace, whichever domain is iterated last
+	would silently overwrite the others, causing earlier secrets to leak into logs.
+	"""
 	if not sensitive_data:
 		return {}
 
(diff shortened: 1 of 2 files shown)


arian-gogani:
credential loss is the kind of incident where a tamper-evident audit trail matters most. if sensitive_data gets wiped, the post-incident investigation needs to know exactly what happened and when.

a signed receipt at each credential operation (store, read, update, delete) gives you independently verifiable evidence of the sequence of events. no one can modify the audit trail after the fact to hide the cause.

github.com/arian-gogani/nobulex

PR #4919, opened 2026-05-26
fix: make news monitor text IO deterministic

Ghraven:
## Summary

Thanks for maintaining browser-use. This is a small robustness fix for the `examples/apps/news-use/news_monitor.py` example.

## What changed

- Read and write the JSON output file with explicit UTF-8 encoding
- Replace the fallback `datetime.utcnow()` call with timezone-aware `datetime.now(timezone.utc)`

## Why

The example writes article data with `ensure_ascii=False`, so explicit UTF-8 keeps saved multilingual article text portable on Windows and other non-UTF-8 locale environments. The UTC fallback timestamp also avoids using deprecated naive UTC datetime APIs.

## Verification

- `python -m py_compile examples/apps/news-use/news_monitor.py`
- `git diff --check`

---
## Summary by cubic
Make the news monitor example deterministic across platforms by standardizing UTF-8 JSON writes and using a timezone-aware UTC timestamp fallback, while still loading legacy files saved with system encodings. This prevents garbled non-ASCII text, avoids naive UTC datetimes, and preserves existing data.

- **Bug Fixes**
  - Load `news_data.json` with a tolerant reader (UTF-8 first, then `locale.getpreferredencoding(False)`), and write with explicit `encoding='utf-8'` and `ensure_ascii=False` in `examples/apps/news-use/news_monitor.py`.
  - Replace `datetime.utcnow()` with `datetime.now(timezone.utc)` and import `timezone`.

<sup>Written for commit 3d5fb92391eba743627e1314720e9ee1540203c6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/4919?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/examples/apps/news-use/news_monitor.py
+++ b/examples/apps/news-use/news_monitor.py
@@ -8,6 +8,7 @@ import argparse
 import asyncio
 import hashlib
 import json
+import locale
 import logging
 import os
 import time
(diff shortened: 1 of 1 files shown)


Ghraven:

Ghraven:
Thanks for the reminder. This PR is still relevant, and I pushed a small follow-up for the issue identified by cubic: existing news monitor data is now read as UTF-8 first but can fall back to the platform legacy encoding, while new writes remain UTF-8.

Verification:
- `python -m py_compile examples/apps/news-use/news_monitor.py`
- `git diff --check`

Issue #4846, opened 2026-05-16
Bug: MCP server connects but list_tabs/get_state/screenshot fail while navigate works (CDP healthy)


shiftlabshq:
### Browser Use Version

browser-use 0.12.6 (uvx)
 MCP wrapper also tested with 0.12.5, same failure

### Bug Description, Steps to Reproduce, Screenshots

root@srv1642813:~# cat /root/bu-local-bug-report.md
# bu-local MCP bug report (browser-use bridge)

Date: 2026-05-16
Host: VPS Ubuntu 24.04 + Windows 11 Edge via SSH reverse tunnel

## Summary
`bu-local` MCP server connects and exposes tools, and `browser_navigate` succeeds, but core state/screenshot/tab tools fail consistently:
- `browser_list_tabs` -> returns `[]` even when tabs clearly exist
- `browser_get_state` -> handler returns None error
- `browser_screenshot` -> `Root CDP client not initialized`

Direct CDP and direct `browser-use` CLI calls work against the same endpoint, indicating a bug in MCP server runtime behavior rather than CDP tunnel/browser availability.

## Environment
- Edge CDP local: `http://127.0.0.1:9222`
- VPS tunnel endpoint: `http://127.0.0.1:9223`
- MCP wrapper command:
  - `/root/.local/bin/browser-use-mcp.sh`
  - executes: `/root/.local/bin/uv tool uvx --from 'browser-use[cli]==0.12.5' browser-use --cdp-url http://127.0.0.1:9223 --mcp`
- Hermes MCP test: connected, 16 tools discovered

## Repro steps
1. Ensure Windows Edge has remote debugging enabled and tunnel is active.
2. On VPS run:
   - `hermes mcp test bu-local` (passes)
3. In Hermes session call tools:
   - `mcp_bu_local_browser_navigate` with `https://example.com` -> success
   - `mcp_bu_local_browser_list_tabs` -> `[]`
   - `mcp_bu_local_browser_get_state` -> None-handler error
   - `mcp_bu_local_browser_screenshot` -> Root CDP client not initialized

## Exact failing outputs
- get_state:
  - `Error: Expected at least one handler to return a non-None result ... BrowserStateRequestEvent ...`
- screenshot:
  - `Error: Root CDP client not initialized`
- list_tabs:
  - `[]`

## Control evidence (same CDP endpoint works)
- `curl http://127.0.0.1:9223/json/version` returns valid Edge JSON.
[...]

PR #4993, opened 2026-06-08
Fix MCP CDP URL and temp profile handling

Frost1325:
## Summary

- Parse MCP-specific CLI args before the fast `--mcp` startup path exits the normal CLI parser, so `--cdp-url` is honored in MCP server mode.
- Pass `--cdp-url` through `BrowserUseServer` into the `BrowserProfile` used by MCP browser sessions.
- Avoid keeping a partially initialized browser session after startup fails, so follow-up MCP tools return the original startup error instead of cascading stale-session errors.
- Stop hardcoding the MCP default `user_data_dir`; local MCP sessions now use an isolated temporary profile unless the user explicitly configures a persistent path.
- Clean up `browser-use-user-data-dir-*` temp profiles created for local browser sessions, including launch-failure paths.
- Treat empty / whitespace-only `user_data_dir` values from MCP config UIs as unset, and avoid creating unused temp profiles for CDP-only sessions.
- Update the MCP manifest to document that an empty `user_data_dir` means a temporary profile per session.

Related to #4846.

## Validation

- `py -3.11 -c "import ast; files=['browser_use/mcp/server.py','browser_use/mcp/__main__.py','browser_use/skill_cli/main.py','browser_use/cli.py','browser_use/browser/watchdogs/local_browser_watchdog.py']; [ast.parse(open(f, encoding='utf-8').read(), filename=f) for f in files]; print('ok')"`
- `rg -n "_owns_browser_resources|—|鈥|бк" browser_use/mcp/server.py browser_use/mcp/manifest.json browser_use/browser/watchdogs/local_browser_watchdog.py`
  - no matches
- `py -3.11 scripts\repro_mcp_4846.py fake_cdp`
  - `browser_navigate`, `browser_list_tabs`, `browser_get_state`, and `browser_screenshot` all return `Error: All connection attempts failed` when passed a fake `--cdp-url`, confirming the flag is honored and errors do not cascade into stale-session state.
  - temp profile count remained unchanged before/after the run.
- `py -3.11 scripts\repro_mcp_4846.py no_keys`
  - `browser_navigate` reaches `https://example.com/`.
  - `browser_list_tabs`, `browser_get_state`, and `browser_screenshot` return expected data.
  - `browser_close_all` closes the session.
  - temp profile count remained unchanged before/after the run.


Frost1325:
Update: this PR now covers the MCP temp-profile behavior in addition to the original `--cdp-url` fix.

New changes included here:

- Remove the hardcoded MCP `user_data_dir` default, so local MCP sessions use an isolated temporary profile unless the user explicitly configures a persistent path.
- Normalize empty / whitespace-only `user_data_dir` values as unset, matching MCP config UI / manifest behavior.
- Avoid creating unused temporary profiles for CDP-only sessions by only applying the persistent fallback when `--cdp-url` is provided and no explicit profile path was configured.
- Track and clean up `browser-use-user-data-dir-*` temporary profiles on both successful launches and launch-failure paths.
- Update `manifest.json` so the `user_data_dir` default and description match the new behavior.

Validation update:

- Local repro with a fake `--cdp-url` returns the original connection error without leaving stale session state.
- Local MCP browser session without API keys can navigate to `https://example.com/`, inspect state, screenshot, and close cleanly.
- Temporary profile count remains unchanged before / after both repro runs.
- CI is green after the no-op rerun commit.

Issue #5041, opened 2026-06-14
[Bug] Potential loss of empty accessibility names due to implicit truthiness check in views.py


QiuYucheng2003:
### Browser Use Version

main

### Bug Description, Steps to Reproduce, Screenshots

Describe the bug
During a static code review of views.py, a subtle logical flaw was identified in the DOMInteractedElement.load_from_enhanced_dom_tree class method.

When extracting the accessibility name, the code uses an implicit truthiness check:
if enhanced_dom_tree.ax_node and enhanced_dom_tree.ax_node.name:

In Python, an empty string "" evaluates to False. If a DOM element explicitly has an empty accessibility name (which is a valid state in HTML/ARIA contexts, such as aria-label=""), this condition will fail. As a result, the ax_name is improperly assigned None instead of preserving the explicit "". This conflates two distinct states: "the attribute does not exist" (None) and "the attribute exists but is intentionally empty" (""), which could cause issues for downstream DOM matching or agent interaction logic.

To Reproduce
Note: This edge-case was identified via static codebase review.

1. Construct or parse an EnhancedDOMTreeNode where ax_node exists, but its name attribute is explicitly set to an empty string ("").

2. Pass this tree node into the DOMInteractedElement.load_from_enhanced_dom_tree(enhanced_dom_tree) method.

3. Inspect the resulting DOMInteractedElement instance.

4. The ax_name property will be incorrectly set to None instead of the expected "".

Expected behavior
The code should preserve explicitly empty accessibility names. To differentiate between None and an empty string, the condition should explicitly check against None:
# Suggested Fix:
if enhanced_dom_tree.ax_node and enhanced_dom_tree.ax_node.name is not None:
    ax_name = enhanced_dom_tree.ax_node.name

# Suggested Fix:
if enhanced_dom_tree.ax_node and enhanced_dom_tree.ax_node.name is not None:
    ax_name = enhanced_dom_tree.ax_node.name

### Failing Python Code

```python

```

### Full DEBUG Log Output

```shell

```

Sarthak816:
working on it

neelabhshukla018:
Hi! I'd like to work on this issue. The proposed fix seems straightforward, and I'd be happy to submit a PR to preserve explicit empty accessibility names by checking `is not None` instead of relying on truthiness.

neelabhshukla018:
Hi! I've submitted PR #5049 to address this issue. Thanks for the detailed report.

vicksiyi:
[...]

PR #5042, opened 2026-06-14
fix: preserve empty accessibility names

he-yufeng:
Fixes #5041.

## Summary

- preserve AX node names when CDP reports the accessible name as an empty string
- include an explicit empty AX name in both regular and stable element hashes
- add a regression test that keeps `ax_name=""` distinct from a missing AX name

## To verify

- `uv run pytest tests\ci\test_ax_name_matching.py -q`
- `uv run python -m py_compile browser_use\dom\views.py tests\ci\test_ax_name_matching.py`
- `uv run ruff check browser_use\dom\views.py tests\ci\test_ax_name_matching.py`
- `uvx --python 3.11 pre-commit run --files browser_use\dom\views.py tests\ci\test_ax_name_matching.py`
- `git diff --check`

---
## Summary by cubic
Preserves empty accessibility names from CDP and uses them in element hashing and interacted-element construction. Prevents collisions where ax_name="" was treated as missing, improving element matching accuracy.

- **Bug Fixes**
  - Treat empty `ax_node.name` as valid; include `ax_name=""` in both stable and runtime hashes and when building `DOMInteractedElement`.
  - Add regression tests for interacted elements and both hashes to keep empty vs missing AX names distinct.

<sup>Written for commit 04a8ec6f6e38c279ec8e5ace7691f2dd0d166385. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5042?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/dom/views.py
+++ b/browser_use/dom/views.py
@@ -848,7 +848,7 @@ class EnhancedDOMTreeNode:
 		attributes_string = ''.join(f'{k}={v}' for k, v in sorted(filtered_attrs.items()))
 
 		ax_name = ''
-		if self.ax_node and self.ax_node.name:
+		if self.ax_node and self.ax_node.name is not None:
 			ax_name = f'|ax_name={self.ax_node.name}'
 
 		combined_string = f'{parent_branch_path_string}|{attributes_string}{ax_name}'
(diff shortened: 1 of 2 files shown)


he-yufeng:
Closing this one out after two quiet months. dom/views.py has moved since and the empty-name handling deserves a fresh pass against the current serializer rather than a rebase of this diff. Branch stays available.

PR #5053, opened 2026-06-16
fix(llm): include cache-read tokens in Anthropic total_tokens

he-yufeng:
### What

`_get_usage` computes `prompt_tokens` as `input_tokens + cache_read_input_tokens` (Anthropic reports cached prompt tokens separately, so they have to be added back), but `total_tokens` was left as just `input_tokens + output_tokens`. As soon as prompt caching kicks in, `total_tokens` is smaller than `prompt_tokens + completion_tokens`, which breaks anything downstream that assumes the totals add up (cost tracking, budget/limit checks).

Concrete example with a cached prompt: `input_tokens=500`, `cache_read_input_tokens=10000`, `output_tokens=200` gives `prompt_tokens=10500`, `completion_tokens=200`, but `total_tokens=700` instead of `10700`.

This affects both `ChatAnthropic` (`browser_use/llm/anthropic/chat.py`) and `ChatAnthropicBedrock` (`browser_use/llm/aws/chat_anthropic.py`), which share the same usage logic.

### Fix

Add the cache-read tokens to `total_tokens` so it mirrors the `prompt_tokens` formula and the invariant `total_tokens == prompt_tokens + completion_tokens` holds again. One line per file.

### Verifying

Added `tests/ci/models/test_anthropic_usage.py` covering both clients: the cached case (asserts the totals add up) and the no-cache case (asserts the number is unchanged).

```
uv run pytest tests/ci/models/test_anthropic_usage.py -q
```

The cached-case assertions fail on current `main` (700 != 10700) and pass with this change; the no-cache case stays at 700 both ways. `ruff check` / `ruff format` clean on the touched files.

Note: #4294 proposed the same one-liner for `chat.py` but went stale before it landed, and it never touched the Bedrock client. This covers both and adds a regression test.

---
## Summary by cubic
Fixes Anthropic usage accounting by adding cache-read prompt tokens to `total_tokens`, restoring `total_tokens == prompt_tokens + completion_tokens`. Applies to both Anthropic and Bedrock clients.

- **Bug Fixes**
  - Include `cache_read_input_tokens` in `total_tokens` in `browser_use/llm/anthropic/chat.py` and `browser_use/llm/aws/chat_anthropic.py`.
  - Add `tests/ci/models/test_anthropic_usage.py` for cached/no-cache cases and type the mock response helper for Pyright.

<sup>Written for commit aa4c71f881d218c9eb443ace54ede22ebe758b46. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5053?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/llm/anthropic/chat.py
+++ b/browser_use/llm/anthropic/chat.py
@@ -209,7 +209,9 @@ class ChatAnthropic(BaseChatModel):
 				response.usage.cache_read_input_tokens or 0
 			),  # Total tokens in Anthropic are a bit fucked, you have to add cached tokens to the prompt tokens
 			completion_tokens=response.usage.output_tokens,
-			total_tokens=response.usage.input_tokens + response.usage.output_tokens,
+			total_tokens=response.usage.input_tokens
+			+ (response.usage.cache_read_input_tokens or 0)
+			+ response.usage.output_tokens,
 			prompt_cached_tokens=response.usage.cache_read_input_tokens,
 			prompt_cache_creation_tokens=response.usage.cache_creation_input_tokens,
 			prompt_cache_creation_5m_tokens=cache_creation_5m_tokens,
--- a/browser_use/llm/aws/chat_anthropic.py
+++ b/browser_use/llm/aws/chat_anthropic.py
@@ -149,7 +149,9 @@ class ChatAnthropicBedrock(ChatAWSBedrock):
 				response.usage.cache_read_input_tokens or 0
 			),  # Total tokens in Anthropic are a bit fucked, you have to add cached tokens to the prompt tokens
 			completion_tokens=response.usage.output_tokens,
-			total_tokens=response.usage.input_tokens + response.usage.output_tokens,
+			total_tokens=response.usage.input_tokens
+			+ (response.usage.cache_read_input_tokens or 0)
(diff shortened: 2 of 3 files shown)


he-yufeng:
First nudge on this one. It is still mergeable, and the gap is a real accounting error: Anthropic responses that hit the prompt cache report `input_tokens` without the cache-read portion, so `total_tokens` underreports exactly on the calls that save the most money. The patch adds `cache_read_input_tokens` into the total and covers both branches in the existing test. Happy to rebase if anything shifted underneath it.

he-yufeng:
Re-verified against current main: `browser_use/llm/anthropic/chat.py:217` still computes `total_tokens` as `input_tokens + output_tokens`, so a turn that is 90% cache reads still reports the same total as one with no cache at all. The fix and its usage test still apply as-is.

PR #5064, opened 2026-06-18
fix(input): call element.blur() after typing to commit value in JS frameworks

### LINKED ISSUE #5054: Feature Request: Blur fields on input

While testing I have a problem that pops up in certain UIs whereby field values are not 'committed' unless there is an explicit 'exit' of a field or 'blur' event before taking the next action.

As an example the agents instructions come back as:
1. Enter value into field.
2. Click 'Save' button.

The agent does this diligently but the value is lost as the field value is not committed before the save.

This of course leads to lots of other problems.

In order to workaround it I patched default_action_watchdog.py just before `# Step 5:` to run a blur event after inputting. This isn't that elegant but works for our case.

The controls in this case our not 'inputs' but 'spans' that are being modified from JS by the client-side framework.

Discussed briefly on https://github.com/browser-use/browser-use/pull/4462 but agree with @r266-tech that it is a separate issue.

Blur event in _input_text_element_node_impl to ensure data is committed and therefore not lost when taking next action.

I also tried adjusting system prompt to 'tab' out, or 'click' out, or event use 'enter'. This worked reasonably well but not 100% as the agent wouldn't always follow the instructions + it was also slower as it meant many more actions.

0.12.9

- [ ] It's an urgent deal-breaker, I can't live without it
- [ ] It's important to add it in the near-mid term future
- [x] It would be nice to add it sometime in the next 2 years

BRGOVIND:
Fixes #5054

## Root cause

`_trigger_framework_events` dispatched `new Event('blur', {bubbles: true})` via `element.dispatchEvent(...)`. This fires blur *listeners* but does **not** actually move focus — `document.activeElement` stays on the element. As a result, frameworks and custom widgets that gate their commit logic on focus genuinely leaving the field never saw the value.

`element.blur()` does both: it removes focus from the element **and** fires the native `blur` + `focusout` sequence, which is what these frameworks wait for.

## Change

In `_trigger_framework_events` (`default_action_watchdog.py`):

- Remove the synthetic `blur` event from the `events` array
- Add an `element.blur()` call after the `input`/`change` dispatches

```js
// Before: fired blur listener but focus stayed on element
element.dispatchEvent(new Event('blur', { bubbles: true }));

// After: genuinely removes focus → native blur + focusout fire
element.blur();
```

## Tests

`tests/ci/interactions/test_blur_on_input.py` — two cases:

1. **contenteditable span** with a `blur` listener that only writes to `#committed` when `document.activeElement !== editor` (synthetic dispatch would fail this check; `element.blur()` passes it)
2. **plain input** with a `focusout` listener on the document (focusout only fires on real focus change)

---
## Summary by cubic
Call `element.blur()` after typing so values commit on real blur/focusout, fixing cases where React, Vue, and custom widgets require an actual focus change.

- **Bug Fixes**
  - In `_trigger_framework_events`, replace synthetic blur dispatch with `element.blur()` after `input`/`change` so native `blur`/`focusout` fire and focus leaves the field.
  - Add tests for contenteditable commit-on-blur and input commit-on-focusout to confirm real blur behavior.

- **Refactors**
  - Update tests to use the correct `tools.input()` API and arrow functions in `page.evaluate()`.
  - Resolve element index by ID at runtime and fix `ruff` formatting.

<sup>Written for commit b76abf0ecb44c1564cca00f723f7dffdec2d8175. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5064?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


BRGOVIND:
Gentle ping 🙂 — this calls `element.blur()` after typing so JS frameworks (React/Vue) commit the typed value through their change handlers before the next action. CI is green. Happy to add a regression test or adjust the approach if a maintainer has a moment.

BRGOVIND:
I'll make uptodate base changes and push again 👍

jonidimo:
Confirming this on Linux with a locally-hosted model, and adding a second fix in
the same area that was needed to make x.com's composer work at all.

## Setup

| | |
|---|---|
| browser-use | 0.13.8 |
| Chromium | 151.0.7922.137 (Debian trixie), `--headless=new`, CDP on :9222 |
| OS | Debian 13, kernel 6.12.101, KDE/Wayland |
| Harness | Hermes Agent v0.20.2 driving browser-use |
| Model | Qwen3.8-27B Q4 served locally by llama.cpp b10217 on an RTX 3090 |

The target was the x.com post composer — a Draft.js `contenteditable`
(`data-testid="tweetTextarea_0"`, class `public-DraftEditor-content`).

## Symptom

Text appeared in the DOM but the framework never registered it: the character
counter stayed at zero, the Post button stayed disabled, and **clicking on the
text made it vanish** — React re-rendering from its own (empty) state.

It was intermittent. It worked once and could not be reproduced, which matches
the "leaf-start retype race condition" language in #4461.

## What I measured

I built a page that replicates React's value tracker and drove it over raw CDP,
to separate the input mechanism from the framework's acceptance of it:

| method | does React register it? |
|---|---|
| `Input.insertText` (CDP) | **yes** — arrives with `isTrusted=true` |
| `Input.dispatchKeyEvent` per key (CDP) | **yes** |
| `element.value = "x"` + `dispatchEvent(new Event('input'))` | **no** |
| native setter + event | yes |

So CDP input is not the problem. Driving the real x.com composer directly over
CDP — `focus()` then `Input.insertText` — worked **eight times out of eight**,
with the text surviving a 1.2 s wait and `aria-disabled` going to `null` on the
Post button. The failure was in the wrapper, not the protocol.

Two hypotheses I tested and had to discard, in case they save someone time:

- **"`focus()` does not place a caret in a contenteditable."** False. Measured
  `rangeCount=1` with the caret inside the element right after `focus()`.
  `insertText` works identically with `focus()` alone, with an explicit Range,
  or after a real CDP click.
- **"The triple-click in the clear fallback steals focus."** False. After
  triple-clicking at the element's centre, `document.activeElement` was still
  `tweetTextarea_0` and `insertText` worked.

## Fix 1 — the one this PR proposes (`_trigger_framework_events`)

Confirmed. The synthetic `blur` in the events array fires listeners without
moving focus, so anything gating its commit on focus actually leaving never
sees the value. Removing it from the array and calling `element.blur()` after
the dispatches is what this PR does, and it is necessary.

Two smaller things in the same function that are worth folding in:

1. `data: element.value` is `undefined` on a contenteditable, since a `div` has
   no `.value`. Falling back to `textContent` makes the `InputEvent` carry real
   data:

   ```js
   const val = (element.value !== undefined && element.value !== null)
       ? element.value : (element.textContent || '');
   ```

2. `Object.defineProperty(syntheticInputEvent, 'isTrusted', { value: true })` in
   the React branch does nothing useful. React does not read that property — it
   tracks value changes through its own patched setter. Faking `isTrusted` on a
   synthetic event cannot make React accept it.

## Fix 2 — `_clear_text_field` in `actor/element.py` (separate bug, same symptom)

This one has to be fixed as well or the composer still fails, and I have not
seen it mentioned in the existing issues.

Strategy 1 clears the field with:

```js
this.value = "";
this.dispatchEvent(new Event("input", { bubbles: true }));   // comment says
this.dispatchEvent(new Event("change", { bubbles: true }));  // "notify React"
```

That is exactly the pattern React ignores — its value tracker sees the direct
write and the synthetic event arrives with `isTrusted=false`. The DOM ends up
empty while React keeps the old text in state.

**The part that makes it stick:** Strategy 2 (triple-click + Delete over CDP)
is correct and would work, but it only runs if Strategy 1 fails — and
Strategy 1 verifies itself by reading `this.value`, which *is* empty. So the
broken strategy reports success and the working one never executes.

On a contenteditable it is worse still. Measured on the real composer:

```
select() throws: TypeError | .value on a div is: "" | text is still: "viejo"
```

`this.select()` throws, `this.value = ""` creates a meaningless property on the
`div`, the real text survives, and the verification reads that empty property
and **also** reports false success.

What I changed locally:

- For `<input>` / `<textarea>`: use the **native setter** so React's value
  tracker registers the change.

  ```js
  const proto = (this instanceof HTMLTextAreaElement)
      ? HTMLTextAreaElement.prototype : HTMLInputElement.prototype;
  Object.getOwnPropertyDescriptor(proto, 'value').set.call(this, "");
  ```

- For `contenteditable`: return a sentinel so the verification fails on purpose
  and Strategy 2 runs with trusted CDP events.

## Result

With both fixes applied, writing to the x.com composer through Hermes →
browser-use now succeeds **on the first attempt**: 31 s, 10 tool calls, no
retries, text persists and the Post button enables. Before, the same task took
2–3 minutes of the agent cycling through alternatives and usually ended with an
empty composer.

Happy to open a separate issue for Fix 2 if that is cleaner, or to send a PR
for either.

## Caveat on attribution

I cannot cleanly attribute the outcome to one fix over the other — the earlier
failures were not run under controlled conditions, and the bug was intermittent
to begin with. What is established in isolation is that the `_clear_text_field`
behaviour is wrong and demonstrably desyncs React, and that the synthetic
`blur` does not move focus. Both are real defects regardless of which one was
biting hardest in my case.

PR #5072, opened 2026-06-20
fix(security): write storage_state.json session cookies with 0o600

r266-tech:
## What

`StorageStateWatchdog._save_storage_state()` persists authenticated session cookies and `localStorage` to `storage_state.json`. It wrote the file with `Path.write_text()` under the process umask, so the file (and its `.json.bak` backup) landed **world-readable (`0o644`)**, and the 30s autosave silently re-downgraded a user-pre-created `0o600` file on every write.

On a multi-user host any local co-tenant can then read live session cookies → session hijack. This is a self-consistency gap: the project already locks **less**-sensitive credential files to `0o600` with the same atomic tmp+rename pattern — `skill_cli/config.py:write_config`, `sync/auth.py`, and the daemon Unix socket (covered by `tests/ci/security/test_daemon_socket_perms.py`).

## Fix

Write through `tempfile.mkstemp`, which creates the temp file `O_EXCL` + `0o600`, so the cookies are never world-readable — not even during the write, or as a leftover temp if the process crashes before the rename. This mirrors `write_config`. The cookie-bearing `.json.bak` (the current one and any stale one left by a pre-fix version) is also restricted to `0o600` before rotation.

## Scope

Permissions only. The existing temp → backup → install rotation flow and its (pre-existing) crash-atomicity characteristics are intentionally left unchanged.

## Tests

`tests/ci/security/test_storage_state_perms.py` (no browser; mirrors the existing daemon-socket security test):

- a fresh write is `0o600`;
- an autosave over a `0o644` file leaves both `storage_state.json` and `.json.bak` at `0o600`;
- a pre-existing insecure backup is locked down even if the save aborts mid-way, and even when no current file exists.

---
## Summary by cubic
Securely persist session cookies by writing storage_state.json with 0o600 and tightening permissions on its .json.bak backup. Prevents world-readable cookies on multi-user hosts and aligns with our existing credential file posture.

- **Bug Fixes**
  - Write via tempfile.mkstemp (O_EXCL) to create a 0o600 temp file, fsync, then atomic rename.
  - Lock down both storage_state.json and any .json.bak (including stale ones) to 0o600 before/after rotation; clean up temp on errors.
  - Preserve the existing temp → backup → install flow.
  - Add tests to ensure fresh writes, autosaves over 0o644, and stale backups all end up 0o600 (skipped on Windows).

<sup>Written for commit 3e6f3f1caf42e091a4db3da522e25735cc2aee89. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5072?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/browser/watchdogs/storage_state_watchdog.py
+++ b/browser_use/browser/watchdogs/storage_state_watchdog.py
@@ -3,6 +3,7 @@
 import asyncio
 import json
 import os
+import tempfile
 from pathlib import Path
 from typing import Any, ClassVar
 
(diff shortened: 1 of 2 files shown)


MagMueller:
Thanks for the careful security analysis and tests. We traced every storage-state save path before deciding on this.

BrowserProfile.storage_state defaults to None. When neither a profile storage_state path nor an explicit SaveStorageStateEvent.path is supplied, _save_storage_state resolves no destination and returns without writing a file. This remains true when user_data_dir alone enables the watchdog. Likewise, export_storage_state() returns an in-memory dictionary unless the caller explicitly passes output_path.

So storage_state.json persistence is opt-in and only occurs when the caller chooses a file destination. It is not a default cookie-file exposure. Callers deploying on shared multi-user hosts can place that explicitly requested file in a private directory and use an appropriately restrictive process umask. We do not want to impose new POSIX permission, backup-rotation, fsync, and tempfile semantics on this cross-platform caller-selected path in core, so we are closing this PR.

MagMueller:
Thanks for the contribution. We traced every current save path before deciding this. BrowserProfile.storage_state defaults to None, and StorageStateWatchdog._save_storage_state() returns without writing when neither BrowserProfile(storage_state=...) nor an explicit SaveStorageStateEvent(path=...) supplies a destination. BrowserSession.export_storage_state(output_path=None) likewise returns the state as a dictionary and does not create a file.

So Browser Use does not persist cookies/localStorage to storage_state.json by default. File persistence is explicitly opt-in and targets a caller-chosen path, whose directory and permissions remain under the caller/deployment boundary. The proposed 0o600 behavior is reasonable defense-in-depth for POSIX callers that opt into a file, but it is not a default credential exposure and adds a larger atomic-write/backup policy change than we want for that contract. We are closing this rather than treating it as a security bug.

PR #5073, opened 2026-06-20
fix(dom): don't mutate node.snapshot_node.bounds in is_element_visible_according_to_all_parents

r266-tech:
## What

`DomService.is_element_visible_according_to_all_parents()` is a pure `-> bool` visibility predicate, but it binds `current_bounds` directly to the live `node.snapshot_node.bounds` (a mutable `DOMRect`) and then applies iframe-origin and scroll offsets to it **in place** (`+=` / `-=`).

`node.snapshot_node.bounds` is the node's stored, **absolute page-coordinate** rect — built during construction at `service.py:788-793` by adding the accumulated `total_frame_offset`. Mutating it here permanently corrupts that stored geometry on any **scrolled or iframe** page. The predicate runs once per node at construction (`service.py:893`) before serialization, so downstream consumers that read `snapshot_node.bounds` (paint-order occlusion filtering, bbox containment filtering — both default-on) then see wrong rectangles and can drop/keep the wrong elements in the DOM the agent sees.

## Fix

Compute the viewport transform on a `DOMRect` copy, mirroring the deliberate copy already in `_construct_enhanced_node` (`service.py:734`, *"to get rid of the pointer references"*). The boolean result is unchanged; only the destructive side effect on the shared rect is removed.

## Tests

`tests/ci/test_dom_visibility_bounds.py` (pure-function, no browser): assert the node's stored bounds are untouched after the predicate runs, on both a scrolled root frame and inside an offset iframe (before the fix, `y` / `x` were corrupted by the scroll / iframe offset).

---
## Summary by cubic
Fixes the visibility check to stop mutating `snapshot_node.bounds`. We now compute on a `DOMRect` copy so geometry stays correct for scrolled pages and iframes.

- **Bug Fixes**
  - Apply iframe/scroll offsets to a copy in `DomService.is_element_visible_according_to_all_parents`, leaving the stored absolute bounds untouched.
  - Added regression tests for scrolled root frames and offset iframes; updated with `typing.cast` so CI type checks pass with no behavior change.

<sup>Written for commit 57c1923c23d4c36a39c95d279c5b3f844d581e2a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5073?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/dom/service.py
+++ b/browser_use/dom/service.py
@@ -275,6 +275,16 @@ class DomService:
 		if not current_bounds:
 			return False  # If there are no bounds, the element is not visible
 
+		# Work on a copy of the rect. The += / -= below shift bounds into each parent
+		# frame's coordinate space for the viewport intersection check, but
+		# node.snapshot_node.bounds is the node's shared, absolute-page-coordinate rect
+		# (set during construction at ~service.py:788). DOMRect is a mutable dataclass,
+		# so mutating this reference in place permanently corrupts the geometry that
+		# downstream consumers (paint-order occlusion, bbox containment) read on any
+		# scrolled or iframe page. Mirror the deliberate DOMRect copy in
+		# _construct_enhanced_node (~service.py:734, 'to get rid of the pointer references').
+		current_bounds = DOMRect(current_bounds.x, current_bounds.y, current_bounds.width, current_bounds.height)
+
 		# If threshold is None, skip all viewport-based filtering (only check CSS visibility)
 		if viewport_threshold is None:
 			return True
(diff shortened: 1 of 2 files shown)


r266-tech:
Closing this because maintainer PR #5151 has landed the same bounds-copy fix, along with broader self-frame and idempotency coverage. Keeping this open now would only leave a stale duplicate/conflict. Thanks for following through on the internal review series.

PR #5080, opened 2026-06-23
Bypass env proxies for local CDP polling

### LINKED ISSUE #5048: Root cause analysis: CDP instability on Windows (httpx.ReadError / Page crashed / WebSocket silent disconnect)
## Browser Use Version

0.13.1 (latest)

## Environment

- Windows 11, headless (no desktop GUI)
- Python 3.12.10, WindowsProactorEventLoopPolicy (default)
- Chromium 145.0.7632.6 (Playwright-installed)
- cdp-use: 1.4.5, websockets: 15.0.1, httpx: 0.28.1

## Summary

After systematic investigation, I identified **three layered root causes** that together make browser-use CDP layer unstable on Windows — while Playwright direct `launch()` works perfectly in the same environment.

This is not a single bug but an architectural mismatch. Related to existing issues: #1830, #3533, #2613, #4579, #4688.

---

## Root Cause 1: httpx.ReadError — CDP HTTP startup race + proxy leakage

**Location**: `BrowserSession.connect()` → `httpx.AsyncClient().get("/json/version")`

**Evidence**:
1. Bare CDP over port 9223 is 100% stable — `curl --noproxy "*" http://127.0.0.1:9223/json/version` never fails.
2. The bug is in the **startup sequence**: `_wait_for_cdp_url()` (in `local_browser_watchdog.py:408`) uses **aiohttp** to poll for CDP readiness, while `connect()` uses **httpx** to actually fetch the WS URL. These are two different HTTP stacks.
3. `_wait_for_cdp_url()` does NOT set `trust_env=False` or handle proxy bypass for localhost. If `HTTP_PROXY` is set (common on Windows with Clash/V2Ray), aiohttp routes 127.0.0.1 through the proxy → 502.
4. Even without proxy, there is a **race window** on Windows: Chromium startup has a ~1-2s gap between TCP port open (aiohttp poll succeeds) and HTTP server fully initialized (httpx request can fail with `ReadError`). This gap is larger on Windows than Linux.
5. `connect()` does have the `trust_env=not is_localhost` fix for httpx (good!), but the aiohttp polling path does not.

**Reproduction**: Start Chromium manually, poll with aiohttp, then immediately GET /json/version with httpx within 200ms of port open. `httpx.ReadError` reproduces ~30% of the time.

---

## Root Cause 2: Page crashed — Windows headless GPU compositor instability

**Location**: Chromium renderer process crash, CDP emits `Inspector.targetCrashed`

**Evidence**:
1. On Windows headless (`--headless=new`), Chromium uses **SwiftShader** (software GPU). SwiftShader requires DirectX runtime components that may be missing on headless/Server Windows.
2. The `CHROME_DOCKER_ARGS` (with `--disable-gpu-sandbox`, `--no-sandbox`, etc.) are NOT applied on local Windows launch — only in Docker detection paths (`profile.py:120-130`).
3. Playwright handles this with Microsoft-internal Windows testing + pipe-based transport that monitors process health in real-time. browser-use uses `asyncio.create_subprocess_exec` with no heartbeat monitoring.
4. When Chromium crashes, the websockets TCP connection may NOT immediately signal disconnection on Windows ProactorEventLoop (IOCP completion ports have notification latency). This causes the agent to get a timeout rather than a clean crash notification.

---

## Root Cause 3: websockets library + ProactorEventLoop incompatibility (ARCHITECTURAL)

**This is the deepest issue.**

**browser-use CDP stack**:
```
Python asyncio (ProactorEventLoop / IOCP)
 → websockets library (asyncio.Protocol)
   → TCP socket → localhost CDP WebSocket
```

**Playwright CDP stack (stable)**:
```
Python asyncio
 → Node.js Playwright Server (separate process)
   → Pipe transport (stdin/stdout)
     → Chromium DevTools (internal pipe protocol)
```

**Why websockets + ProactorEventLoop is problematic**:
1. The `websockets` library is built on `asyncio.Protocol` + `loop.create_connection()`. On ProactorEventLoop, this uses **IOCP** (I/O Completion Ports). IOCP uses pre-allocated overlapped I/O buffers — when WebSocket frames span multiple TCP segments, frame reassembly can be delayed or truncated.
2. `cdp_use.CDPClient._handle_messages()` is a `asyncio.create_task()` infinite loop calling `self.ws.recv()`. Under ProactorEventLoop, if the WebSocket connection drops silently (TCP RST not promptly delivered), `recv()` can hang indefinitely — this is exactly what `TimeoutWrappedCDPClient` was created to mitigate (its docstring documents this: "WebSocket stays alive at the TCP/keepalive layer while the browser is dead").
3. The fix in `TimeoutWrappedCDPClient` is a **symptom workaround** (per-request timeout), not a root cause fix. The underlying transport instability remains.
4. Evidence from cdp-use#11: connection_closed was **downgraded from warning to debug** because it was too frequent — this confirms WebSocket disconnections are common enough that maintainers suppressed their visibility.

**Why Playwright does not have this problem**: Pipe transport is synchronous-character-stream, no TCP WebSocket framing issues, no ProactorEventLoop IOCP buffer edge cases. The Node.js Playwright server handles process lifecycle directly, not through Python subprocess management.

---

## Related Issues (same root causes, different symptoms)

| Issue | Symptom | Matches |
|-------|---------|---------|
| #1830 | `ValueError: I/O operation on closed pipe` | ProactorEventLoop pipe cleanup (RC3) |
| #3533 | Chrome crashing + `Session id not found` | ProactorEventLoop + Page crashed (RC2/RC3) |
| #2613 | `TargetClosedError: browser has been closed` | Page crashed (RC2) |
| #4579 | CDP hang after idle (open) | WebSocket silent disconnect (RC3) |
| #4688 | CDP WebSocket drops mid-session (open) | WebSocket instability (RC3) |
| cdp-use#11 | connection_closed logged too often | WebSocket frequent disconnects (RC3) |

---

## Suggested Fix Direction

1. **Short-term**: Fix `_wait_for_cdp_url()` to use `trust_env=False` (like `connect()` already does), and add retry with backoff for httpx.ReadError in connect().
2. **Medium-term**: Switch CDP transport from TCP WebSocket to **pipe-based** (`--remote-debugging-pipe` flag, available since Chrome 125). This would bypass both the ProactorEventLoop IOCP issues and the TCP/WebSocket framing issues entirely — matching Playwright architecture.
3. **Long-term**: Consider using Playwright CDPSession as the transport layer (browser-use already depends on Playwright for browser installation). This would inherit years of Windows battle-testing.

---

## Verification

I verified that Playwright `launch()` direct control (without browser-use CDP layer) is 100% stable under the same conditions — navigation, multi-field forms, submission, state tracking all pass consistently. The instability is specifically in browser-use CDP transport layer, not in Chromium itself.

vicksiyi:
Summary
- Disable environment proxy lookup for local aiohttp CDP readiness polling.
- Add regression coverage for the localhost polling path.

Tests
- python -m py_compile browser_use/browser/watchdogs/local_browser_watchdog.py tests/ci/browser/test_local_browser_watchdog.py
- Direct Python smoke: asserted LocalBrowserWatchdog._wait_for_cdp_url uses aiohttp.ClientSession(trust_env=False)
- Target pytest exits 139 locally even with PYTEST_DISABLE_PLUGIN_AUTOLOAD=1

Notes
- Fixes #5048.

---
## Summary by cubic
Bypass environment proxies during local CDP readiness polling by using `aiohttp.ClientSession(trust_env=False)` so readiness checks hit 127.0.0.1 directly. Adds a regression test for `LocalBrowserWatchdog._wait_for_cdp_url` that verifies the `/json/version` request to localhost with `trust_env=False`, and fixes the test import order.

<sup>Written for commit f8ffeb9a3ff18d8abdc8eb3709aa1344fc421efb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5080?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/browser/watchdogs/local_browser_watchdog.py
+++ b/browser_use/browser/watchdogs/local_browser_watchdog.py
@@ -413,7 +413,7 @@ class LocalBrowserWatchdog(BaseWatchdog):
 
 		while asyncio.get_event_loop().time() - start_time < timeout:
 			try:
-				async with aiohttp.ClientSession() as session:
+				async with aiohttp.ClientSession(trust_env=False) as session:
 					async with session.get(f'http://127.0.0.1:{port}/json/version') as resp:
 						if resp.status == 200:
 							# Chrome is ready
--- a/tests/ci/browser/test_local_browser_watchdog.py
+++ b/tests/ci/browser/test_local_browser_watchdog.py
@@ -1,7 +1,8 @@
 from __future__ import annotations
 
-import aiohttp
 import asyncio
+
+import aiohttp
 
 from browser_use.browser.watchdogs.local_browser_watchdog import LocalBrowserWatchdog
 
--- a/tests/ci/browser/test_local_browser_watchdog.py
+++ b/tests/ci/browser/test_local_browser_watchdog.py
@@ -1,0 +1,43 @@
+from __future__ import annotations
+
+import aiohttp
+import asyncio
+
+from browser_use.browser.watchdogs.local_browser_watchdog import LocalBrowserWatchdog
+
+
+def test_wait_for_cdp_url_bypasses_env_proxy_for_localhost(monkeypatch):
+	"""Local CDP readiness polling must not route 127.0.0.1 through env proxies."""
+	client_session_kwargs = []
+	requested_urls = []
+
+	class FakeResponse:
+		status = 200
+
+		async def __aenter__(self):
+			return self
+
+		async def __aexit__(self, exc_type, exc, tb):
+			return None
+
+	class FakeClientSession:
+		def __init__(self, **kwargs):
+			client_session_kwargs.append(kwargs)
+
+		async def __aenter__(self):
+			return self
+
+		async def __aexit__(self, exc_type, exc, tb):
+			return None
+
+		def get(self, url):
+			requested_urls.append(url)
+			return FakeResponse()
+
+	monkeypatch.setattr(aiohttp, 'ClientSession', FakeClientSession)
+
+	cdp_url = asyncio.run(LocalBrowserWatchdog._wait_for_cdp_url(9222, timeout=0.1))
+
+	assert cdp_url == 'http://127.0.0.1:9222/'
+	assert requested_urls == ['http://127.0.0.1:9222/json/version']
+	assert client_session_kwargs == [{'trust_env': False}]


vicksiyi:
Pushed f8ffeb9a3 to fix the ruff import-order change reported by the code-style job.

Validation: `git diff --check`, `uv run ruff check tests/ci/browser/test_local_browser_watchdog.py`, `uv run ruff format --check tests/ci/browser/test_local_browser_watchdog.py`, and `python -m py_compile browser_use/browser/watchdogs/local_browser_watchdog.py tests/ci/browser/test_local_browser_watchdog.py`.

MagMueller:
Thanks for the contribution. We reproduced this against the repository-pinned aiohttp 3.13.4 with HTTP_PROXY configured, NO_PROXY both unset and empty, and a fake proxy. ClientSession() already defaults to trust_env=False and reached the 127.0.0.1 CDP endpoint directly; ClientSession(trust_env=False) behaved identically. Only trust_env=True routed through the proxy. The current Browser Use CDP polling path also succeeded with a dead proxy configured.

This appears to transpose the valid httpx fix from #4063 to aiohttp, whose default is the opposite. The source change is therefore a runtime no-op today, and the added test only asserts a constructor keyword rather than reproducing proxy routing. We are closing this as redundant. If aiohttp ever changes its default, we can revisit the explicit setting with a behavioral proxy-bypass test.

PR #5089, opened 2026-06-27
Fix variable shadowing in Element.click() error handling

Gaurav0807:
## Summary
  Fixed a variable shadowing bug where the error message at line 347 references
  the wrong exception variable, causing confusing error messages when both CDP
  click and JavaScript fallback fail.

  ## Problem
  In `browser_use/actor/element.py` line 347, when the JavaScript click fallback
  fails, the exception handler catches the error as `js_e` but then raises a new
  exception referencing `e` (the outer exception from the CDP click failure).

  This means developers see the original CDP error instead of the actual JS error
  that prevented recovery.

  ## Changes
  - **File:** `browser_use/actor/element.py`
  - **Line 347:** Changed `{e}` to `{js_e}` in the error message
  - **Before:** `raise Exception(f'Failed to click element: {e}')`
  - **After:** `raise Exception(f'Failed to click element: {js_e}')`

  This aligns with the correct pattern already used at line 213 in the same method.

  ## Testing
  - Existing tests pass: `uv run pytest -vxs tests/ci/`
  - Added regression test: `tests/ci/test_element_click_error.py`

  ## Type
  Bug fix - Error reporting accuracy

---
## Summary by cubic
Fixes error reporting in Element.click() so the message shows the JS fallback error, not the earlier CDP error, when both attempts fail. This makes failure messages accurate and easier to debug.

- **Bug Fixes**
  - Use `js_e` in the raised message when the JS click fallback fails.
  - Added regression test `tests/ci/test_element_click_error.py`; fixed linting in the test file.

<sup>Written for commit e00ec0aa35931d1c49a2c6e310ab874920e281a1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5089?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


itxaiohanglover:
Nice catch! The variable shadowing  instead of  means the JS click failure was being masked by the outer CDP exception. The regression test clearly demonstrates the bug.

Issue #4877, opened 2026-05-21
Bug: ...pip dependency conflicts due to strict exact-pins (==) in pyproject.toml


Ito-69:
### Browser Use Version

0.12.7

### Description
Integrating `browser-use` into larger LLM/agent pipelines (such as Nous Research's `hermes-agent` or custom production setups) frequently triggers severe dependency conflicts during installation. 

The root cause is that several highly common core packages are pinned to exact versions (`==`) instead of compatible ranges or minimum bounds (`>=`) in `pyproject.toml`. When co-installed alongside packages that require slightly newer versions of the same dependencies, `pip`'s dependency resolver throws errors and halts.

### Steps to Reproduce
Attempt to install `browser-use` in an environment where newer standard packages are already present, or co-install with `hermes-agent` (which requires e.g., `openai==2.24.0`, `pydantic==2.13.4`, `rich==14.3.3`):
```bash
pip install browser-use hermes-agent

Affected Dependencies
Currently, the exact pins in pyproject.toml that conflict with standard modern agent environments are:

"rich==14.3.1" (conflicts with rich>=14.3.3)
"pydantic==2.12.5" (conflicts with pydantic>=2.13.4)
"python-dotenv==1.2.2" (conflicts with other minor releases)
"requests==2.33.0" (conflicts with requests>=2.32.5)
"openai==2.16.0" (conflicts with openai>=2.24.0)
Proposed Solution
Unless there are known breaking changes in these libraries, relaxing the strict exact pins to minimum version bounds (>=) in pyproject.toml would dramatically improve library interoperability and ease of installation for the community.

Here is the proposed surgical diff for pyproject.toml

dependencies = [
    "aiohttp==3.13.4",
    "anyio==4.12.1",
    "bubus==1.5.6",
    "click==8.3.1",
    "InquirerPy==0.3.4",
-   "rich==14.3.1",
+   "rich>=14.3.1",
    "google-api-core==2.29.0",
    "httpx==0.28.1",
    "posthog==7.7.0",
    "psutil==7.2.2",
-   "pydantic==2.12.5",
+   "pydantic>=2.12.5",
    "pyobjc==12.1; platform_system == 'darwin'",
-   "python-dotenv==1.2.2",
-   "requests==2.33.0",
+   "python-dotenv>=1.2.2",
+   "requests>=2.32.5",
    "screeninfo==0.8.1; platform_system != 'darwin'",
    "typing-extensions==4.15.0",
[...]

PR #5093, opened 2026-06-29
fix(deepseek): repair malformed tool call arguments

hopwesley:
## Summary
- add a DeepSeek tool-call argument parser that falls back to json-repair for malformed JSON
- add json-repair as a runtime dependency
- add unit coverage for valid and repairable tool-call arguments

## Testing
- Not run locally; this environment does not have the project's uv tool available.

---
## Summary by cubic
Fixes DeepSeek tool-call parsing by repairing malformed JSON arguments to prevent crashes and missed tool calls. Adds a parser with a `json-repair` fallback and updates the chat path to use it.

- **Bug Fixes**
  - Add `_parse_tool_call_arguments` that tries `json.loads` then falls back to `json-repair`; returns `{}` if unrecoverable.
  - Use the parser when reading `function.arguments` in `ChatDeepSeek.ainvoke`.
  - Add tests for valid and repairable inputs.

- **Dependencies**
  - Add `json-repair==0.61.0`.

<sup>Written for commit b69bea0b0870af55413f137d74a620eebdd59169. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5093?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/llm/deepseek/chat.py
+++ b/browser_use/llm/deepseek/chat.py
@@ -5,6 +5,7 @@ from dataclasses import dataclass
 from typing import Any, TypeVar, overload
 
 import httpx
+from json_repair import repair_json
 from openai import (
 	APIConnectionError,
 	APIError,
@@ -23,6 +24,21 @@ from browser_use.llm.schema import SchemaOptimizer
 from browser_use.llm.views import ChatInvokeCompletion
 
 T = TypeVar('T', bound=BaseModel)
+
+
+def _parse_tool_call_arguments(raw_args: str) -> Any:
+	try:
+		return json.loads(raw_args)
+	except json.JSONDecodeError:
+		repaired = repair_json(raw_args, return_objects=True)
+		if isinstance(repaired, dict):
+			return repaired
+		if isinstance(repaired, str):
+			try:
+				return json.loads(repaired)
+			except json.JSONDecodeError:
+				return {}
+		return {}
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


itxaiohanglover (commented):
Clean approach — the fallback to `json_repair` is a nice safety net for DeepSeek's occasional malformed tool call arguments. One question: does adding `json-repair==0.61.0` as a hard dependency make sense, or would it be better as an optional dependency with a graceful fallback when not installed?

itxaiohanglover:
Clean approach. The fallback to json_repair for malformed tool call arguments is a good resilience improvement. One question: is json-repair already a transitive dependency, or does it add a new direct dependency? If new, worth confirming it's lightweight enough.

itxaiohanglover (commented):
Nice approach — using json_repair as a fallback for malformed DeepSeek tool call arguments is practical.

One concern: adding `json-repair==0.61.0` as an exact pin contributes to the dependency conflict issue tracked in #4877. Consider using `json-repair>=0.30.0` instead, since the API has been stable across versions.

The `_parse_tool_call_arguments` function looks solid. One minor suggestion: the final `return {}` could log a warning, since silently returning an empty dict might make debugging harder when tool calls fail unexpectedly.

PR #5094, opened 2026-06-29
fix(dom): preserve empty accessibility names

TisanSmart232:
Fixes #5041

## Summary

Preserve empty accessibility names instead of treating them as missing. Chrome's accessibility tree can legitimately provide an empty string name, and using truthiness checks caused that value to be dropped from element hashes and interacted element history.

## Tests

- python -m pytest tests\ci\test_dom_views.py -q
- python -m ruff check browser_use\dom\views.py tests\ci\test_dom_views.py
- git diff --check

---
## Summary by cubic
Preserve empty accessibility names (`ax_name`) so they’re treated as valid and included in element hashing and interacted element records, matching Chrome’s accessibility tree behavior.

- **Bug Fixes**
  - Use `is not None` checks to keep empty `ax_name` in `compute_stable_hash`, `__hash__`, and `load_from_enhanced_dom_tree`.
  - Add tests confirming empty `ax_name` is preserved and affects both `hash()` and `compute_stable_hash()` results.

<sup>Written for commit d3615433af9d1fad364cbd2bc9ba22c910dd39f4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5094?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/dom/views.py
+++ b/browser_use/dom/views.py
@@ -848,7 +848,7 @@ class EnhancedDOMTreeNode:
 		attributes_string = ''.join(f'{k}={v}' for k, v in sorted(filtered_attrs.items()))
 
 		ax_name = ''
-		if self.ax_node and self.ax_node.name:
+		if self.ax_node and self.ax_node.name is not None:
 			ax_name = f'|ax_name={self.ax_node.name}'
 
 		combined_string = f'{parent_branch_path_string}|{attributes_string}{ax_name}'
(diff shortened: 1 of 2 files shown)


itxaiohanglover:
Clean fix. The `is not None` check correctly distinguishes between an element with no accessibility name (None) and one with an empty name string (`\\`), which matters for hash-based element deduplication. Tests covering both cases are a nice addition.

PR #5096, opened 2026-06-29
fix(skill_cli): don't add venv bin/ to user PATH

### LINKED ISSUE #5033: Bug: CLI installer prepends its private virtualenv to PATH and shadows the user's Python
## Browser Use Version

`0.13.2` (`1e75d1f1f3ae970ad673e479a2ac9b3b613f82be`)

## Bug Description

The recommended CLI installer adds the entire Browser Use virtual environment to the user's shell `PATH`:

```sh
export PATH="$HOME/.browser-use-env/bin:$HOME/.local/bin:$PATH"
```

This makes the `browser-use` command available, but it also exposes every executable installed in the private environment. In particular, it changes the shell-wide resolution of `python`, `python3`, and versioned Python commands:

```console
$ command -v browser-use
~/.browser-use-env/bin/browser-use

$ command -v python
~/.browser-use-env/bin/python

$ command -v python3
~/.browser-use-env/bin/python3

$ python --version
Python 3.13.11
```

`VIRTUAL_ENV` remains unset, so the prompt and environment do not indicate that commands are being resolved from Browser Use's private virtualenv.

This can unintentionally affect unrelated shell commands, scripts, and Python workflows. It also exposes other dependency-provided executables from the environment, not just Browser Use commands.

## Steps to Reproduce

1. On macOS with zsh, run the recommended installer:

   ```sh
   curl -fsSL https://browser-use.com/cli/install.sh | bash
   ```

2. Start a new shell.
3. Run:

   ```sh
   command -v browser-use
   command -v python
   command -v python3
   echo "${VIRTUAL_ENV:-unset}"
   ```

4. Observe that `python` and `python3` resolve inside `~/.browser-use-env/bin`, while `VIRTUAL_ENV` is unset.

## Expected Behavior

Installing the Browser Use CLI should make its commands available without changing the user's default Python or exposing all executables from the private virtualenv.

For example, the installer could create wrappers or symlinks for the supported CLI entry points in `~/.local/bin`:

```sh
#!/bin/sh
exec "$HOME/.browser-use-env/bin/browser-use" "$@"
```

The private `~/.browser-use-env/bin` directory would then not need to be added to the global `PATH`.

The same approach could be used for aliases such as `bu`, `browseruse`, and `browser`.

## Failing Python Code

Not applicable. This issue concerns the CLI bootstrap installer's shell configuration.

## LLM Model

Not applicable.

## Operating System & Browser Versions

- macOS 26.2 (build 25C56) on Apple Silicon (`arm64`)
- zsh 5.9 (`arm64-apple-darwin25.0`)
- Installed using the documented one-line CLI installer

## Relevant Installer Code

`browser_use/skill_cli/install.sh`, in `configure_path()`, adds the virtualenv's `bin` directory directly to the shell configuration.

## Logs

```console
$ command -v browser-use
~/.browser-use-env/bin/browser-use

$ command -v python
~/.browser-use-env/bin/python

$ command -v python3
~/.browser-use-env/bin/python3

$ echo "${VIRTUAL_ENV:-unset}"
unset
```

srivtx:
## Summary

The CLI installer was putting the entire browser-use venv bin directory on the user's shell PATH. This shadowed the system `python`, `pip`, and other dependency-provided executables, with no `VIRTUAL_ENV` set to make it visible.

The fix restricts PATH additions to `$HOME/.local/bin` and creates a small wrapper script there for each entry point in `pyproject.toml`. The wrappers exec the underlying venv binary, so callers see the same `browser-use` command but the venv bin/ never reaches the user's PATH.

## What changed

`browser_use/skill_cli/install.sh`:

- New `create_venv_wrappers` function that writes a 2-line `#\!/bin/sh` wrapper in `~/.local/bin/<name>` for each entry point (`browser-use`, `browseruse`, `bu`, `browser`, `browser-use-tui`). On Windows it also writes a `.bat` wrapper for native cmd/PowerShell.
- `configure_path` now only adds `~/.local/bin` to the user's shell rc (was adding the full venv bin + ~/.local/bin).
- `configure_powershell_path` mirrors the same change for the Windows registry PATH.
- Idempotency check switched from grepping the shell rc to checking for the wrapper script itself, which avoids false positives from users who already had `~/.local/bin` on PATH for unrelated reasons.

## Verification

- `bash -n browser_use/skill_cli/install.sh` — syntax OK
- `shellcheck browser_use/skill_cli/install.sh` — no new warnings on the changed code (pre-existing warnings in untouched lines are unchanged)
- Local function-level test: `create_venv_wrappers` correctly produces wrappers that exec the venv binary, with `python` correctly excluded from the wrapper directory.
- The 5 entry-point names match `[project.scripts]` in `pyproject.toml` (`browser-use`, `browseruse`, `bu`, `browser`, `browser-use-tui`).

## Behaviour after this lands

- `browser-use --help` still works as before.
- `command -v python` resolves to the user's system python again.
- `${VIRTUAL_ENV}` stays unset (no implicit venv activation).
- On re-running the installer, the idempotency check finds the existing wrapper and skips PATH modification.

Fixes #5033

---
## Summary by cubic
Stop adding the venv `bin/` to the user PATH and expose only the CLI entry points via wrappers in `$HOME/.local/bin` so system `python`/`pip` aren’t shadowed. Fixes #5033.

- **Bug Fixes**
  - Added `create_venv_wrappers` to generate tiny shell/`.bat` wrappers for `browser-use`, `browseruse`, `bu`, `browser`, `browser-use-tui`.
  - PATH now only appends `$HOME/.local/bin` (and `%USERPROFILE%\.local\bin` on Windows); venv `bin/`/`Scripts/` is never added and any legacy `browser-use-env` PATH lines are removed from the shell rc.
  - Idempotency: sentinel `~/.browser-use/.path-configured`; Windows checks the registry for `%USERPROFILE%\.local\bin`; wrappers are created only if missing and never overwrite existing names (warns if skipped).
  - Behavior: `browser-use` still works; `python` resolves to the system install; no implicit venv activation.

<sup>Written for commit 027df122113b0be635a24ac1ddc33db8315c6a91. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5096?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/skill_cli/install.sh
+++ b/browser_use/skill_cli/install.sh
@@ -375,9 +375,71 @@ install_profile_use() {
 # PATH configuration
 # =============================================================================
 
+# CLI entry points exposed to the user. Mirrors [project.scripts] in
+# pyproject.toml. Keep in sync when adding new entry points.
+BROWSER_USE_CLI_NAMES=(
+	"browser-use"
+	"browseruse"
+	"bu"
+	"browser"
+	"browser-use-tui"
+)
+
+# Sentinel file that marks the install as having configured PATH. Checked
+# by both the Unix shell-rc and Windows PowerShell code paths so the
+# idempotency decision is independent of whether the wrappers exist.
+BROWSER_USE_PATH_MARKER="$HOME/.browser-use/.path-configured"
+
+# Create thin wrapper scripts in $HOME/.local/bin that exec the venv
+# binaries. This exposes only the browser-use entry points without putting
+# the entire venv bin/ on the user's PATH, which would otherwise shadow
+# the system python, pip, and other dependency-provided executables.
+#
+# Each wrapper is only written when $HOME/.local/bin/<name> does not
+# already exist. Generic names like `bu` and `browser` are likely to
+# collide with unrelated user commands, so we never overwrite an
+# existing file there. Pre-existing files (which might be from an older
+# browser-use install, pipx, or an unrelated user command) are kept as-is
+# and a warning is logged so the user knows their old binary is still active.
+create_venv_wrappers() {
+	local local_bin="$HOME/.local/bin"
+	local venv_bin
+	venv_bin=$(get_venv_bin_dir)
+
+	mkdir -p "$local_bin"
+
+	local name
+	for name in "${BROWSER_USE_CLI_NAMES[@]}"; do
+		local target="$venv_bin/$name"
+		[ -x "$target" ] || continue
+
+		# Shell wrapper for Unix and Git Bash on Windows.
+		if [ ! -e "$local_bin/$name" ]; then
+			cat > "$local_bin/$name" <<-EOF
+				#!/bin/sh
+				exec "$venv_bin/$name" "\$@"
+			EOF
+			chmod +x "$local_bin/$name"
+		else
+			log_warn "$local_bin/$name already exists; not overwriting. If this is a leftover from a previous browser-use install, remove it manually to pick up the new venv."
+		fi
+
+		# .bat wrapper for native Windows shells (cmd, PowerShell).
+		if [ "$PLATFORM" = "windows" ]; then
+			if [ ! -e "$local_bin/$name.bat" ]; then
+				cat > "$local_bin/$name.bat" <<-EOF
+					@echo off
+					"%USERPROFILE%\\.browser-use-env\\Scripts\\${name}.exe" %*
+				EOF
+			else
+				log_warn "$local_bin/$name.bat already exists; not overwriting."
+			fi
+		fi
+	done
+}
+
 configure_path() {
 	local shell_rc=""
-	local bin_path=$(get_venv_bin_dir)
 	local local_bin="$HOME/.local/bin"
 
 	# Detect user's login shell (not the running shell, since this script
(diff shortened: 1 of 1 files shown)


srivtx:
All 3 issues are valid, thanks. Fixed in the latest commit (`453cbf419`):

**P1 — Windows PATH never added on first install (line 465).** Root cause: `create_venv_wrappers` runs inside `configure_path` *before* `configure_powershell_path`, so by the time the Windows code path checks for `~/.local/bin/browser-use.bat` the file already exists and the registry PATH update is skipped.

Fix: replaced the wrapper-file existence check with a sentinel file at `~/.browser-use/.path-configured`. The marker is created only after the user PATH is successfully updated, and the same marker is consulted by both the Unix shell-rc and Windows PowerShell code paths. So:
- First install on Windows: marker is absent → registry PATH gets `%USERPROFILE%\.local\bin`, marker is created.
- Re-run: marker is present → both code paths skip.
- Recovery if the registry update ever fails: marker is not created, so the next run will retry.

**P2 — Wrappers clobber user commands named 'bu' / 'browser' (line 405).** Fix: each wrapper is only written when the target name does not already exist in `~/.local/bin`. The 5 entry points in `pyproject.toml` (browser-use, browseruse, bu, browser, browser-use-tui) are tried in order, and any that collide with a pre-existing user command are skipped with no error. A pre-existing `~/.local/bin/bu` from pipx or another tool is now preserved.

**P2 — Idempotency check fooled by unrelated `browser-use` (line 436).** Same fix as P1: the sentinel file replaces the wrapper-existence check. A pre-existing `browser-use` from `pip --user` / pipx / manual install no longer blocks the installer's PATH setup — we create the marker ourselves.

Re-running the install locally:
- first run: PATH gets `~/.local/bin`, 5 wrappers created (any that didn't already exist), marker created
- second run: marker present → shell-rc skipped, PowerShell-rc skipped, but `create_venv_wrappers` still runs so any entry-point wrappers missing from a prior partial install are filled in

@cubic-dev-ai re-trigger please.

srivtx:
Addressed all 3 issues in commit `027df1221`:

**P0 — Windows PowerShell PATH never set on first install.** Root cause confirmed: the marker was being created in the Unix branch (lines 462-465 of the previous version) before `configure_powershell_path` was called, so the Windows idempotency check matched and the registry update was skipped.

Fix:
1. Removed marker creation from the Unix branch; the marker is now created at the end of `configure_path`, after both Unix and Windows code paths have run.
2. `configure_powershell_path` no longer checks the marker. It now reads the user PATH from the registry (via `[Environment]::GetEnvironmentVariable('Path', 'User')`) and returns early only if the value already contains `\.local\bin`.

Also changed `echo "$current_path"` to `printf '%s\n' "$current_path"` while I was in the area — echo interprets backslash escapes in the value, so a user PATH containing e.g. `\b` would have been mangled before the grep.

**P1 — Existing installs retain the old venv-bin PATH entry.** Fix: before writing the new `# Browser-Use` section, check if the shell rc already contains `browser-use-env` and if so, strip those lines via `grep -v` + temp file + `mv`. The user's other shell config is preserved.

**P2 — Pre-existing commands silently block wrapper creation.** Fix: when `[ -e "$local_bin/$name" ]` is true, log a warning explaining the situation ("remove it manually to pick up the new venv") and skip the write. Same warning added for the `.bat` file path on Windows.

**Local verification (sourced install.sh and ran end-to-end against a fake venv):**

- TEST 1 first install — 5 wrappers created, shell rc line written, marker created ✓
- TEST 2 pre-existing user 'bu' — preserved byte-for-byte, warning logged ✓
- TEST 3 legacy `browser-use-env` line — removed, new line added, other user config preserved ✓
- TEST 4 re-run idempotency — no duplicate Browser-Use section ✓
- TEST 5 Windows first install — `SetEnvironmentVariable` IS called on first install ✓

`bash -n` clean. `shellcheck` clean on changed code (the two warnings on lines 476/522 are pre-existing, untouched by this change).

@cubic-dev-ai please re-trigger.

itxaiohanglover (commented):
Nice approach — using thin wrapper scripts instead of putting the entire venv bin/ on PATH is the right call. The collision avoidance for generic names like `bu` and `browser` is good.

Issue #5099, opened 2026-06-30
fix: allowed_domains bypass also possible via javascript: URIs and chrome:// URLs


tcconnally:
## Description

Issue #4763 identified that `data:` and `blob:` URLs bypass `allowed_domains` / `prohibited_domains` restrictions. PR #4850 fixed `data:` and `blob:` in `_is_url_allowed()`. However, the same bypass pattern applies to other non-HTTP schemes that still trigger navigation:

1. **`javascript:` URIs** — `javascript:alert(1)` navigates the page without changing `document.location`. A page can redirect the agent via `location.href = "javascript:..."` or `<a href="javascript:...">`.
2. **`chrome://` and `about://` URLs** — Chromium internal pages (`chrome://settings`, `about:blank`, `about:srcdoc`, `chrome://downloads`) are not validated and can leak internal browser state or cause navigation outside `allowed_domains`.
3. **`file://` URLs** — If the agent visits a page with a link to `file:///etc/passwd`, `_is_url_allowed()` passes it through because only `data:` and `blob:` were blocked — `file:` is not in the blocklist.

## Location in code

`browser_use/browser/watchdogs/security_watchdog.py`

## Suggested fix

Extend the scheme blocklist from `['data', 'blob']` to also include `['javascript', 'file', 'chrome', 'about', 'chrome-extension']` — or ideally switch to an allowlist approach where only `http:` and `https:` are permitted when `allowed_domains` is configured.

NikhilDendeti:
I'd like to take this. Agree the blocklist approach is fragile — rather than adding javascript/file/chrome/about/chrome-extension one at a time, I'd propose flipping to an allowlist: when allowed_domains/prohibited_domains is configured, only permit http:/https: schemes outright, and reject everything else by default. That closes this bypass class structurally instead of requiring us to keep enumerating new blocked schemes as they're found. Will send a PR against security_watchdog.py with tests covering javascript:, file://, and chrome:// cases from this report.

itxaiohanglover:
After investigating, most of the described bypasses are already handled by the existing code:

- **`javascript:` URIs** — `urlparse('javascript:alert(1)')` returns `hostname=None`, which is caught by the `if not host: return False` check in `_is_url_allowed()`. Already blocked.
- **`file://` URLs** — `urlparse('file:///etc/passwd')` returns `hostname=None` (or empty). Already blocked by the same hostname check.
[...]

PR #5101, opened 2026-06-30
fix(security): block data:/blob: URL bypass when allowed_domains is configured

itxaiohanglover:
## Problem

`data:` and `blob:` URLs were unconditionally allowed before domain checks, bypassing `allowed_domains` and `prohibited_domains` restrictions. A malicious page could use `data:` URLs to execute scripts or render content outside the configured domain allowlist.

This is the same bypass class described in #5099 — `data:` and `blob:` were explicitly allowed at the top of `_is_url_allowed()` without checking whether domain restrictions are active.

## Fix

Only allow `data:` and `blob:` URLs when no domain restrictions are configured. When `allowed_domains` or `prohibited_domains` is set, these schemes fall through to the hostname check, which blocks them (they have no hostname to validate).

Schemes without hostnames (`javascript:`, `file:`, `about:`) were already blocked by the existing `if not host: return False` check. Non-http schemes with explicit `allowed_domains` entries (e.g., `chrome://version`) still work through pattern matching since they have hostnames.

## Test cases verified

| URL | allowed_domains | Before | After |
|-----|----------------|--------|-------|
| `data:text/html,...` | `['example.com']` | ✅ allowed (bypass) | ❌ blocked |
| `blob:xxx` | `['example.com']` | ✅ allowed (bypass) | ❌ blocked |
| `data:text/html,...` | none | ✅ allowed | ✅ allowed (unchanged) |
| `chrome://version` | `['chrome://version']` | ✅ allowed | ✅ allowed (unchanged) |
| `javascript:alert(1)` | `['example.com']` | ❌ blocked | ❌ blocked (unchanged) |
| `file:///etc/passwd` | `['example.com']` | ❌ blocked | ❌ blocked (unchanged) |
| `data:text/html,...` | prohibited only | ✅ allowed (bypass) | ❌ blocked |

---
## Summary by cubic
Prevented a security bypass where `data:` and `blob:` URLs were allowed even when `allowed_domains` or `prohibited_domains` were set. These schemes are now only allowed when no domain restrictions are configured.

- **Bug Fixes**
  - `data:` and `blob:` now fall through to hostname checks and are blocked when domain rules exist.
  - No changes to other schemes: `javascript:`/`file:` remain blocked; explicit non-HTTP with hostnames (e.g., `chrome://version`) still work via pattern matching.

<sup>Written for commit 1789dec87f3e57918e54c7a531dbad597258a3eb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5101?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/browser/watchdogs/security_watchdog.py
+++ b/browser_use/browser/watchdogs/security_watchdog.py
@@ -196,9 +196,14 @@ class SecurityWatchdog(BaseWatchdog):
 			# Invalid URL
 			return False
 
-		# Allow data: and blob: URLs (they don't have hostnames)
+		# Allow data: and blob: URLs only when no domain restrictions are active
+		# (they have no hostname to validate, so they bypass domain checks)
 		if parsed.scheme in ['data', 'blob']:
-			return True
+			if (
+				not self.browser_session.browser_profile.allowed_domains
+				and not self.browser_session.browser_profile.prohibited_domains
+			):
+				return True
 
 		# Get the actual host (domain)
 		host = parsed.hostname


thelonewander3r:
Your fix here is correct, and it's the one that should land — I'm closing my duplicate (#5483) in favour of it.

Context: I hit #4763 while triaging open issues, confirmed the bypass is still live on `main` (`security_watchdog.py:200` returns `True` for `data:`/`blob:` before any allowlist check), and only then found this PR. Rather than leave a second competing patch open, here's the one thing this PR is missing that I'd already written.

**There is no test coverage for `data:` or `blob:` anywhere in `tests/ci/security/`** — I checked all seven files. Without a regression test this is free to come back a third time (it already survived #4850).

I ran the tests below against your branch at `1789dec87`: **27 passed**, including all 6 new ones. Please take them into this PR — the only thing I'd ask in return is a co-author trailer on the commit that carries them:

```
Co-authored-by: Eman <eman1369a@gmail.com>
```

Two of them (`..._blocked_under_allowlist`, `..._blocked_under_prohibited_list`) fail on unpatched `main` and pass with your change. The other four are no-regression guards that pass either way.

Append to `tests/ci/security/test_domain_filtering.py`:

```python
class TestHostlessSchemeBypass:
	"""Regression tests for #4763: hostname-less schemes must not bypass domain policy."""

	@staticmethod
	def _watchdog(**profile_kwargs):
		from bubus import EventBus

		from browser_use.browser.watchdogs.security_watchdog import SecurityWatchdog

		browser_profile = BrowserProfile(headless=True, user_data_dir=None, **profile_kwargs)
		browser_session = BrowserSession(browser_profile=browser_profile)
		return SecurityWatchdog(browser_session=browser_session, event_bus=EventBus())

	def test_data_and_blob_blocked_under_allowlist(self):
		"""data:/blob: carry no hostname, so an allowlist must not let them through."""
		watchdog = self._watchdog(allowed_domains=['example.com'])

		assert watchdog._is_url_allowed('data:text/html,<script>alert(1)</script>') is False
		assert watchdog._is_url_allowed('data:text/html;base64,PHNjcmlwdD5hbGVydCgxKTwvc2NyaXB0Pg==') is False
		assert watchdog._is_url_allowed('blob:https://malicious.com/550e8400-e29b-41d4-a716-446655440000') is False

		# A blob URL naming an allowed origin is still blocked: the origin lives in the path,
		# not the host, so there is nothing the domain policy can validate against.
		assert watchdog._is_url_allowed('blob:https://example.com/550e8400-e29b-41d4-a716-446655440000') is False

		# The allowlist itself still works.
		assert watchdog._is_url_allowed('https://example.com') is True

	def test_data_and_blob_blocked_under_prohibited_list(self):
		"""A prohibited list is also an active policy, so the same gate applies."""
		watchdog = self._watchdog(prohibited_domains=['malicious.com'])

		assert watchdog._is_url_allowed('data:text/html,<h1>hi</h1>') is False
		assert watchdog._is_url_allowed('blob:https://malicious.com/abc') is False

		# Unrelated domains are unaffected by the change.
		assert watchdog._is_url_allowed('https://example.com') is True
		assert watchdog._is_url_allowed('https://malicious.com') is False

	def test_data_and_blob_allowed_when_no_policy_configured(self):
		"""Default, policy-free sessions keep the previous permissive behaviour."""
		watchdog = self._watchdog()

		assert watchdog._is_url_allowed('data:text/html,<h1>hi</h1>') is True
		assert watchdog._is_url_allowed('blob:https://example.com/abc') is True

	def test_other_hostless_schemes_stay_blocked_under_policy(self):
		"""javascript:/file:/about: were already blocked by the host check - keep it that way."""
		watchdog = self._watchdog(allowed_domains=['example.com'])

		assert watchdog._is_url_allowed('javascript:alert(1)') is False
		assert watchdog._is_url_allowed('file:///etc/passwd') is False
		assert watchdog._is_url_allowed('about:srcdoc') is False

	def test_internal_targets_still_exempt(self):
		"""The internal new-tab/blank allowance is checked before scheme handling."""
		watchdog = self._watchdog(allowed_domains=['example.com'])

		assert watchdog._is_url_allowed('about:blank') is True
		assert watchdog._is_url_allowed('chrome://new-tab-page/') is True

	def test_explicitly_allowlisted_scheme_with_host_still_works(self):
		"""Non-HTTP URLs that do have a host are unaffected and still match patterns."""
		watchdog = self._watchdog(allowed_domains=['chrome://version', 'brave://*'])

		assert watchdog._is_url_allowed('chrome://version') is True
		assert watchdog._is_url_allowed('brave://settings') is True
		assert watchdog._is_url_allowed('chrome://settings') is False
```

One judgement call worth flagging, in case you'd rather assert the opposite: `test_data_and_blob_blocked_under_allowlist` asserts that `blob:https://example.com/<uuid>` is blocked *even when `example.com` is allowlisted*, on the grounds that a blob URL's origin sits in its path rather than its host, so there's nothing for the domain policy to validate. That's the conservative reading — if you'd prefer to parse the inner origin instead, that assertion is the one to change.

PR #5109, opened 2026-07-01
fix(security): prevent subdomain/port/userinfo bypass in full URL pattern matching

sergioperezcheco:
## Security Fix: URL Pattern Matching Bypass

`_is_url_match()` uses `url.startswith(pattern)` for full URL patterns (containing `://`), which allows three bypass vectors that affect both `allowed_domains` and `prohibited_domains` configurations.

### Vulnerability

When a user configures `allowed_domains=['https://example.com']`, the following URLs would be incorrectly allowed:

| Attack | URL | Why it matches |
|--------|-----|---------------|
| Subdomain bypass | `https://example.com.attacker.com/steal` | `startswith('https://example.com')` is True |
| Port bypass | `https://example.com:8080/admin` | `startswith('https://example.com')` is True |
| Userinfo bypass | `https://example.com@attacker.com/` | `startswith('https://example.com')` is True, but hostname is `attacker.com` |

The same issue affects `prohibited_domains` — a prohibited domain like `https://evil.com` would not block `https://evil.com.evil-cdn.net`.

### Fix

Replace `url.startswith(pattern)` with proper URL component comparison using `urlparse`:

1. Parse both the URL and the pattern
2. Compare scheme, hostname (case-insensitive), and port
3. If the pattern specifies a path prefix, enforce it with `startswith` on the path component only

### Changes
- `browser_use/browser/watchdogs/security_watchdog.py`: Replace `url.startswith(pattern)` with parsed URL component comparison in `_is_url_match()`

### Testing
Verified all attack vectors are blocked and legitimate URLs still work:
- `https://example.com.attacker.com/steal` → BLOCKED
- `https://example.com:8080/admin` → BLOCKED  
- `https://example.com@attacker.com/` → BLOCKED
- `https://example.com/page` → ALLOWED
- `https://example.com/api/v1` (with path prefix pattern) → ALLOWED

---
## Summary by cubic
Fixes full-URL pattern matching to stop subdomain, port, userinfo, and sibling-path bypasses in `allowed_domains` and `prohibited_domains`. We now parse URLs and compare scheme/hostname/port, reject explicit ports when the pattern omits them, enforce boundary-safe path prefixes, and require exact query/fragment matches.

<sup>Written for commit ae8546d867dae1481c2117f5f8045f7613a6fdec. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5109?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/browser/watchdogs/security_watchdog.py
+++ b/browser_use/browser/watchdogs/security_watchdog.py
@@ -315,10 +315,27 @@ class SecurityWatchdog(BaseWatchdog):
 					# Pattern has no explicit port but URL does — reject.
 					return False
 				# If the pattern specifies a path prefix, enforce it.
+				# Use boundary-aware comparison: the URL path must either
+				# equal the pattern path exactly, or start with it followed
+				# by '/' or '?' to prevent sibling-path overmatch
+				# (e.g., pattern '/api' must not match '/apiary').
+				# Also preserve query/fragment constraints from the pattern.
 				if pat_path:
-					url_path = _urlparse(url).path.rstrip('/')
-					if not url_path.startswith(pat_path):
+					url_parsed = _urlparse(url)
+					url_path = url_parsed.path.rstrip('/')
+					if url_path == pat_path:
+						pass  # exact path match
+					elif url_path.startswith(pat_path + '/'):
+						pass  # sub-path match
+					else:
 						return False
+					# If the pattern includes query params, enforce them.
+					if pat_parsed.query:
+						if url_parsed.query != pat_parsed.query:
+							return False
+					if pat_parsed.fragment:
+						if url_parsed.fragment != pat_parsed.fragment:
+							return False
 				return True
 			else:
 				# Domain-only pattern (case-insensitive comparison)
--- a/browser_use/browser/watchdogs/security_watchdog.py
+++ b/browser_use/browser/watchdogs/security_watchdog.py
@@ -282,9 +282,44 @@ class SecurityWatchdog(BaseWatchdog):
 		else:
 			# Exact match
 			if '://' in pattern:
-				# Full URL pattern
-				if url.startswith(pattern):
-					return True
+				# Full URL pattern — parse both to avoid subdomain/port/userinfo bypass.
+				# url.startswith("https://example.com") would match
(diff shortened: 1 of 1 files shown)


itxaiohanglover:
Thorough fix. The `url.startswith(pattern)` approach had three real bypass vectors (subdomain, port, userinfo) and the urlparse-based comparison closes all of them cleanly. The port-matching logic is appropriately strict — if the pattern omits a port, the URL must also omit it.

One minor edge case: `https://example.com:443` would be rejected against pattern `https://example.com` since the URL has an explicit port. This is the right call for security — a strict match is better than trying to normalize default ports.

This PR is complementary to #5112 (which blocks non-HTTP schemes). Together they cover both the scheme-level and host-level attack surfaces.

sergioperezcheco:
Thanks for the thorough review! You're correct about the `https://example.com:443` vs `https://example.com` case — I intentionally chose strict matching over default-port normalization. Normalizing default ports (443→https, 80→http) would add complexity and could introduce bypass vectors if the normalization logic has edge cases. A strict exact match is safer and easier to reason about.

Good to know about #5112 — happy to coordinate if there's any overlap to resolve.

sergioperezcheco:
Thanks for the careful review @itxaiohanglover. Yeah, the `https://example.com:443` vs `https://example.com` case is intentionally strict — normalizing default ports opens a can of worms (which port maps to which scheme, IPv6 brackets, etc.) and the safer default is to require an exact match. If a caller wants to allow both forms they can list both patterns explicitly. Appreciate you confirming the three bypass vectors are closed.

sergioperezcheco:
Friendly ping for maintainer review. The security bypass vectors (subdomain / port / userinfo) are confirmed fixed, all CI checks pass, and @itxaiohanglover reviewed it positively. Happy to make any adjustments if needed.

PR #5112, opened 2026-07-01
fix(security): block non-HTTP schemes when domain restrictions are active

### LINKED ISSUE #5099: fix: allowed_domains bypass also possible via javascript: URIs and chrome:// URLs

itxaiohanglover:
## Problem

Issue #5099 identified that `javascript:`, `file://`, `chrome://`, and `about:` URLs bypass `allowed_domains` restrictions — the same class of bug as the `data:`/`blob:` bypass in #5101.

When `allowed_domains` is configured, a page can redirect the agent to:
- `javascript:alert(1)` — execute arbitrary JS
- `file:///etc/passwd` — read local files
- `chrome://settings` — access browser internals
- `about:srcdoc` — access iframe content

## Fix

When `allowed_domains` or `prohibited_domains` is configured, only allow `http`/`https` schemes:

```python
if (allowed_domains or prohibited_domains):
    if parsed.scheme not in ('http', 'https'):
        return False
```

Internal browser targets (`about:blank`, `chrome://newtab`) are already handled before this check. When no domain restrictions are active, all schemes remain allowed (existing behavior).

This also subsumes the `data:`/`blob:` fix from #5101 — those schemes are now blocked by the non-HTTP check when domain restrictions are active.

Fixes #5099

## Test plan

- [ ] Existing tests pass: `pytest tests/`
- [ ] Configure `allowed_domains`, navigate to `javascript:alert(1)` — should be blocked
- [ ] Configure `allowed_domains`, navigate to `file:///etc/passwd` — should be blocked
- [ ] No `allowed_domains` configured, navigate to `data:text/html,hello` — should still work

---
## Summary by cubic
Block non-HTTP(S) URLs when domain restrictions are enabled to prevent bypassing `allowed_domains`/`prohibited_domains`. Fixes #5099.

- **Bug Fixes**
  - Only allow `http`/`https` when domain restrictions are configured.
  - Blocks navigation to `javascript:`, `file://`, `chrome://`, and `about:` under restrictions; internal targets (e.g., `about:blank`) are already handled.
  - Keeps previous behavior when no restrictions; `data:` and `blob:` remain allowed only in that case.

<sup>Written for commit ee47ff4a44bb8e7387c1b83dbe20f9c12d238f6d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5112?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/browser/watchdogs/security_watchdog.py
+++ b/browser_use/browser/watchdogs/security_watchdog.py
@@ -196,7 +196,16 @@ class SecurityWatchdog(BaseWatchdog):
 			# Invalid URL
 			return False
 
-		# Allow data: and blob: URLs (they don't have hostnames)
+		# When domain restrictions are active, only allow http/https schemes
+		# (internal browser targets like about:blank are already handled above)
+		if (
+			self.browser_session.browser_profile.allowed_domains
+			or self.browser_session.browser_profile.prohibited_domains
+		):
+			if parsed.scheme not in ('http', 'https'):
+				return False
+
+		# Allow data: and blob: URLs when no domain restrictions are active
 		if parsed.scheme in ['data', 'blob']:
 			return True
 


itxaiohanglover:
Closing — the fix is too aggressive. Tests show that `chrome://` URLs with hostnames (like `chrome://version`, `chrome://abc.google.com`) are expected to pass through to the hostname matching logic, not be blocked by scheme.

After investigating issue #5099 more carefully:

- `javascript:alert(1)` — already blocked: `parsed.hostname` is `None`, caught by `if not host: return False`
- `file:///etc/passwd` — already blocked: same reason, no hostname
- `chrome://settings` — has hostname `settings`, correctly checked against `allowed_domains`
- `data:`/`blob:` — the only real bypass, already fixed in #5101

The existing hostname check (`if not host: return False`) already handles `javascript:` and `file:` URLs. No additional scheme blocking is needed beyond what #5101 already does.

PR #5144, opened 2026-07-06
fix(utils): preserve sensitive redaction priority

### LINKED ISSUE #5135: Bug: redact_sensitive_string cascades when a secret value is substring of a placeholder tag
## Problem

## Steps
2. Run:
```python
    'leaked supersecret token',
```

## Expected

## Actual

## Env

Jupiter363:
## Summary
- Fix `redact_sensitive_string` by selecting non-overlapping redaction spans from the original input before emitting placeholders.
- Prevent generated and existing `<secret>...</secret>` placeholder tags from being reprocessed by later secrets.
- Preserve the existing longest-secret-first behavior for overlapping secrets and first-key behavior for duplicate secret values.
- Add regression tests for the reported cascade, overlap priority, duplicate values, empty values, literal special characters, and message filtering path.

Fixes #5135

## Why span-based
The reported bug happens because the current implementation mutates the output string repeatedly. Once a secret is replaced with `<secret>...</secret>`, a later shorter secret can match inside that generated placeholder markup.

A combined regex avoids that cascade, but regex replacement is still leftmost-first. That can change the current longest-secret-first behavior when overlapping secrets start at different positions:

```python
redact_sensitive_string(
    "abcde",
    {"short": "abc", "long": "bcde"},
)
```

The existing contract prefers the longer secret first and produces:

```text
a<secret>long</secret>
```

A leftmost regex replacement would redact `abc` first instead. This PR keeps the single-output-pass benefit while selecting spans by secret length and insertion order before emitting the final string.

## Tests
- `uv run pytest -q tests/ci/security/test_sensitive_data.py`
- `uv run pre-commit run --files browser_use/utils.py tests/ci/security/test_sensitive_data.py`

### DIFF
--- a/browser_use/utils.py
+++ b/browser_use/utils.py
@@ -20,6 +20,7 @@ load_dotenv()
 
 # Pre-compiled regex for URL detection - used in URL shortening
 URL_PATTERN = re.compile(r'https?://[^\s<>"\']+|www\.[^\s<>"\']+|[^\s<>"\']+\.[a-z]{2,}(?:/[^\s<>"\']*)?', re.IGNORECASE)
+SECRET_PLACEHOLDER_PATTERN = re.compile(r'<secret>.*?</secret>', re.DOTALL)
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


sauravpanda:
Hey Thanks for the PR, Issue #5135 landed via merged #5211.

PR #5155, opened 2026-07-07
fix(example): make news monitor file handling explicit

Ghraven:
## Summary
- read and write the news monitor JSON history with explicit UTF-8 encoding
- replace the fallback `datetime.utcnow()` timestamp with timezone-aware UTC

## Problem
The news monitor example writes JSON with `ensure_ascii=False`, but relies on the platform default encoding when reading and writing `news_data.json`. That can break or misread non-ASCII article titles on non-UTF-8 systems. Its timestamp fallback also uses naive UTC via `datetime.utcnow()`.

## Before / After
Before, the example used default text encoding for `news_data.json` and emitted a naive UTC fallback timestamp.

After, the JSON history is consistently handled as UTF-8 and the fallback timestamp is generated with `datetime.now(timezone.utc)`.

## Verification
- `python -m py_compile examples/apps/news-use/news_monitor.py`
- `uv run ruff check examples/apps/news-use/news_monitor.py`
- `git diff --check`

---
## Summary by cubic
Make the news monitor read and write `news_data.json` as UTF-8 and handle legacy encodings, including UTF‑8 BOM files. Uses timezone-aware UTC for fallback timestamps and avoids overwriting unreadable histories.

- **Bug Fixes**
  - Read existing files with a fallback chain (`utf-8-sig`, `utf-8`, system locale, `cp1252`) to support UTF‑8 BOM and legacy encodings; write with `encoding='utf-8'` and `ensure_ascii=False`; skip writes if the file can’t be decoded and log a warning.
  - Replace `datetime.utcnow()` with `datetime.now(timezone.utc)` for the fallback timestamp; fix import order.

<sup>Written for commit 58bde7aaeb5ab61253e38643cc1a8c1c11583a89. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5155?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/examples/apps/news-use/news_monitor.py
+++ b/examples/apps/news-use/news_monitor.py
@@ -8,8 +8,8 @@ import argparse
 import asyncio
 import hashlib
 import json
-import logging
 import locale
+import logging
 import os
 import time
 from datetime import datetime, timezone
--- a/examples/apps/news-use/news_monitor.py
+++ b/examples/apps/news-use/news_monitor.py
@@ -9,6 +9,7 @@ import asyncio
 import hashlib
 import json
 import logging
+import locale
 import os
 import time
 from datetime import datetime, timezone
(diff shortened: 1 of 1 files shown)


Ghraven:

Ghraven:
Thanks for catching this. I updated the helper to try UTF-8, UTF-8 BOM, the system locale, and cp1252 before giving up. If the existing file still cannot be decoded, the monitor now leaves it untouched instead of treating it as empty and overwriting the saved entries.

Verified locally with:

```bash
python -m py_compile examples/apps/news-use/news_monitor.py
```

Ghraven:
Thanks again, that was a good catch. I updated the fallback order so `utf-8-sig` is tried before plain UTF-8, which preserves both BOM and non-BOM UTF-8 history files before falling back to locale/cp1252.

Verified locally with:

```bash
uv run --no-sync ruff check examples/apps/news-use/news_monitor.py
python -m py_compile examples/apps/news-use/news_monitor.py
```

PR #5223, opened 2026-07-14
fix(browser): prevent stale sessions during target attach/detach races

254808127:
## Summary

- track target attaches from callback scheduling through session registration
- let a detach cancel an attach that has not registered yet
- publish target/session mappings atomically before the first CDP await
- stop post-attach initialization when the session detached while `Target.setAutoAttach` was in flight

## Problem

`Target.attachedToTarget` and `Target.detachedFromTarget` callbacks schedule independent async tasks. Previously, `_handle_target_attached()` awaited `Target.setAutoAttach` before registering the session and its target mappings.

This creates two race windows:

```text
attached callback -> attach task waits in Target.setAutoAttach
                       detached task finds no session mapping and returns
                   attach task resumes and registers a dead session
```

and:

```text
attached callback schedules attach task
detached callback schedules detach task
detached task runs before the attach task has registered anything
```

In either case, a short-lived target can leave stale session, target, and mapping state in `SessionManager`.

## Fix

The event callbacks now track attaches as pending as soon as they are observed. A detach received during that window records cancellation even if the attach task has not started.

The attach handler publishes the session, target, and both mappings under the existing lock before awaiting CDP. The detach handler can therefore remove a complete state entry while `Target.setAutoAttach` is in flight. When that await finishes, the attach handler verifies that its exact session is still registered before continuing initialization.

The pending state is limited to observed in-flight attaches and is cleared during manager cleanup.

## Regression coverage

The new tests deterministically cover both race windows:

1. detach while `Target.setAutoAttach` is blocked
2. detach callback before the scheduled attach task runs

On unmodified `main`, the first test logs `Session detached but target unknown` and fails because the dead `CDPSession` remains registered. Both tests pass with this change.

Validation:

```text
uv run pytest -n 0 tests/ci/browser/test_session_manager.py
2 passed

uv run pytest -n 0 tests/ci/browser/test_session_manager.py tests/ci/browser/test_session_start.py tests/ci/browser/test_cross_origin_click.py tests/ci/browser/test_true_cross_origin_click.py tests/ci/browser/test_tabs.py
18 passed

uv run pyright
0 errors, 0 warnings, 0 informations

uv run pre-commit run --files browser_use/browser/session_manager.py tests/ci/browser/test_session_manager.py
All hooks passed
```

## Scope

This fixes stale lifecycle state caused by attach/detach scheduling races. It does not change frame traversal or event timeout behavior.

---
## Summary by cubic
Fixes a race between target attach/detach in the session manager to prevent stale sessions and mappings. Short-lived targets no longer leave dead `CDPSession` entries when `Target.setAutoAttach` is slow or a detach arrives first.

- **Bug Fixes**
  - Track pending attaches from the event and cancel them if a detach arrives first.
  - Publish `CDPSession`, target, and both mappings under lock before any CDP await; ignore late or duplicate attaches and verify the session still exists after `Target.setAutoAttach`.
  - Detach handler detects pending/early detaches, removes state while auto-attach is in flight, and avoids “unknown target” warnings for expected races.
  - Clear pending tracking on cleanup; add deterministic tests covering both race windows, including ensuring auto-attach is not called for canceled attaches.

<sup>Written for commit 81d6a630553aa47dde8ff6487d12d3f87adcfdb7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5223?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/browser/session_manager.py
+++ b/browser_use/browser/session_manager.py
@@ -46,6 +46,12 @@ class SessionManager:
 		# Reverse mapping: session -> target it belongs to
 		self._session_to_target: dict[SessionID, TargetID] = {}
 
+		# Event callbacks schedule attach/detach work as separate tasks. Track only
+		# in-flight attaches so a detach received before its attach task runs can
+		# cancel registration without retaining arbitrary unknown session IDs.
+		self._pending_attach_session_ids: set[SessionID] = set()
+		self._detached_pending_attach_session_ids: set[SessionID] = set()
+
 		# Page lifecycle events per target, fed by ONE global Page.lifecycleEvent handler
 		# registered in start_monitoring(). cdp-use's event registry is single-slot per
 		# CDP method, so per-session handler registrations would replace each other and
@@ -80,6 +86,8 @@ class SessionManager:
 
 		# Register synchronous event handlers (CDP requirement)
 		def on_attached(event: AttachedToTargetEvent, session_id: SessionID | None = None):
+			attached_session_id = event['sessionId']
+			self._pending_attach_session_ids.add(attached_session_id)
 			# _handle_target_attached() handles:
 			# - setAutoAttach for children
 			# - Create CDPSession
@@ -93,8 +101,12 @@ class SessionManager:
 			)
 
 		def on_detached(event: DetachedFromTargetEvent, session_id: SessionID | None = None):
+			detached_session_id = event['sessionId']
+			attach_was_pending = detached_session_id in self._pending_attach_session_ids
+			if attach_was_pending:
+				self._detached_pending_attach_session_ids.add(detached_session_id)
 			create_task_with_error_handling(
-				self._handle_target_detached(event),
+				self._handle_target_detached(event, attach_was_pending=attach_was_pending),
 				name='handle_target_detached',
 				logger_instance=self.logger,
 				suppress_exceptions=True,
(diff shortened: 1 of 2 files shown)


254808127:
Hi team, just a gentle bump on this PR.

It fixes a deterministic attach/detach race in SessionManager that can leave stale CDP sessions, with regression tests covering both race windows. The targeted tests, pyright, and pre-commit checks are all passing.

I'd appreciate a review when someone familiar with the browser/session lifecycle has time. Happy to make any changes needed. Thanks!

254808127:
Hi @sauravpanda , sorry for the direct ping.

This PR has been waiting for review since Jul 14 and fixes a deterministic
attach/detach lifecycle race in SessionManager, with regression tests for
both race windows.

#5263 now also touches the same attach lifecycle area and introduces
target-level in-flight attach tracking. Since the two fixes are related but
address different races, could you take a look or advise whether they should
be reviewed together and which one should land first?

Happy to rebase or adjust the implementation as needed. Thanks!

254808127:
Hi @sauravpanda, following up again on this PR. Do you have an estimate of when it could be reviewed and merged? Are there any remaining blockers or changes needed from my side? Happy to rebase or update the fix if needed. Thanks for your time!

PR #5240, opened 2026-07-16
fix(llm): read structured output from tool calls for Groq tool-calling models

### LINKED ISSUE #4945: BUG: ChatGroq._invoke_structured_output always raises ModelProviderError for tool-calling models
When using a model listed in `ToolCallingModels` (e.g. `moonshotai/kimi-k2-instruct`), Groq returns `message.content = None` and places the structured JSON in `message.tool_calls[0].function.arguments`. The guard at line 176 of `browser_use/llm/groq/chat.py` unconditionally checks `message.content` regardless of which invocation path was used, so it always raises `ModelProviderError('No content in response')` even when valid tool-call data was returned.

Traceback:
```
File "browser_use/llm/groq/chat.py", line 177, in _invoke_structured_output
    raise ModelProviderError(
browser_use.llm.exceptions.ModelProviderError: No content in response
```

Trigger: call `ainvoke(messages, OutputModel)` with any model in `ToolCallingModels`. The fix is to read from `tool_calls[0].function.arguments` after the tool-calling path and only check `message.content` after the JSON-schema path.

VenkateswarluNagineni:
Fixes #4945.

### Problem

`ChatGroq._invoke_structured_output` picks one of two paths, but then reads the result from only one of them:

```python
if self.model in ToolCallingModels:
    response = await self._invoke_with_tool_calling(groq_messages, output_format, schema)
else:
    response = await self._invoke_with_json_schema(groq_messages, output_format, schema)

if not response.choices[0].message.content:
    raise ModelProviderError(message='No content in response', ...)
```

`_invoke_with_tool_calling` sends `tool_choice='required'`, so Groq returns the structured payload in `tool_calls[0].function.arguments` and leaves `message.content` empty. The content guard therefore fires on a perfectly valid response, and **every** model in `ToolCallingModels` (currently `moonshotai/kimi-k2-instruct`) fails structured output 100% of the time:

```
browser_use.llm.exceptions.ModelProviderError: No content in response
```

Since `ainvoke(messages, OutputModel)` is how the agent gets its structured output, these models are unusable with `ChatGroq` rather than merely degraded.

### Change

Parse the payload from the tool call on the tool-calling branch, and keep the content guard on the JSON-schema branch where it still applies. This mirrors what the deepseek provider already does for its tool-calling path (`browser_use/llm/deepseek/chat.py`): check `tool_calls`, then read `tool_calls[0].function.arguments`. An empty `tool_calls` list now reports `'No tool calls in response'` instead of falling through to content that is never populated.

The JSON-schema path is untouched, so models outside `ToolCallingModels` behave exactly as before.

### Testing

Added `tests/ci/models/test_llm_groq.py` (there were no `ChatGroq` tests). These are offline unit tests using a fake API key and a faked Groq client — the LLM is the one thing `CLAUDE.md` allows mocking, and no network is touched:

- `test_tool_calling_model_parses_arguments` — the reported scenario: `content=None`, payload in `tool_calls`. **Fails on `main`** with the exact reported `ModelProviderError: No content in response`; passes with this change.
- `test_tool_calling_model_without_tool_calls_raises` — empty `tool_calls` gives a clear provider error. **Fails on `main`.**
- `test_json_schema_model_parses_content` / `test_json_schema_model_without_content_raises` — backward-compatibility guards for the non-tool-calling path; these pass both with and without the change.

Results:
- `pytest tests/ci/models/test_llm_groq.py` — 4 passed (2 of them fail without the source change).
- `pytest tests/ci/models/` — 46 passed, 7 skipped, unchanged from `main`.
- `ruff check` clean on both files. I deliberately did **not** run `ruff format` on `chat.py`: it already reports as unformatted on `main`, so formatting it would pull unrelated reformatting into this diff — happy to include it if you'd prefer.
- `pyright browser_use/llm/groq/chat.py` reports the same 9 pre-existing errors as `main` (all unresolved `groq` imports in my local env, on lines this PR doesn't touch).

---
## Summary by cubic
Fixes structured output for Groq tool-calling models in `ChatGroq` by reading tool call arguments and preserving provider error codes. Restores functionality for `ToolCallingModels` and stops false "No content in response" errors.

- **Bug Fixes**
  - Tool-calling path: parse `tool_calls[0].function.arguments`; keep the content check only on the JSON-schema path; raise a clear error when no tool calls are returned.
  - Preserve `ModelProviderError` status and message in `ainvoke` (do not re-wrap).
  - Add offline unit tests for both paths and error propagation.

<sup>Written for commit 0a59008f6abca2d940382edd7f49a3c09ac9d212. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5240?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/llm/groq/chat.py
+++ b/browser_use/llm/groq/chat.py
@@ -170,23 +170,45 @@ class ChatGroq(BaseChatModel):
 
 		if self.model in ToolCallingModels:
 			response = await self._invoke_with_tool_calling(groq_messages, output_format, schema)
+			parsed_response = self._parse_tool_call_output(response, output_format)
 		else:
 			response = await self._invoke_with_json_schema(groq_messages, output_format, schema)
 
-		if not response.choices[0].message.content:
-			raise ModelProviderError(
-				message='No content in response',
-				status_code=500,
-				model=self.name,
-			)
+			if not response.choices[0].message.content:
+				raise ModelProviderError(
+					message='No content in response',
+					status_code=500,
+					model=self.name,
+				)
 
-		parsed_response = output_format.model_validate_json(response.choices[0].message.content)
+			parsed_response = output_format.model_validate_json(response.choices[0].message.content)
+
 		usage = self._get_usage(response)
 
 		return ChatInvokeCompletion(
 			completion=parsed_response,
 			usage=usage,
 		)
+
+	def _parse_tool_call_output(self, response: ChatCompletion, output_format: type[T]) -> T:
+		"""Extract the structured payload returned by the tool-calling path.
+
+		`_invoke_with_tool_calling` sends tool_choice='required', so the model returns the payload as
+		tool-call arguments and leaves `message.content` empty.
+		"""
+		message = response.choices[0].message
+
+		if not message.tool_calls:
+			raise ModelProviderError(
+				message='No tool calls in response',
+				status_code=500,
+				model=self.name,
+			)
+
+		arguments = message.tool_calls[0].function.arguments
+		if isinstance(arguments, str):
+			return output_format.model_validate_json(arguments)
+		return output_format.model_validate(arguments)
 
 	async def _invoke_with_tool_calling(self, groq_messages, output_format: type[T], schema) -> ChatCompletion:
 		"""Handle structured output using tool calling."""
--- a/browser_use/llm/groq/chat.py
+++ b/browser_use/llm/groq/chat.py
@@ -145,6 +145,10 @@ class ChatGroq(BaseChatModel):
 
 		except APIError as e:
 			raise ModelProviderError(message=e.message, model=self.name) from e
+		except ModelProviderError:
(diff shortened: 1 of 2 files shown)


VenkateswarluNagineni:
Thanks — good catch on the status code. Fixed in the follow-up commit: added an `except ModelProviderError: raise` passthrough before the catch-all in `ainvoke`, so provider errors we raise ourselves keep their intended status (500) instead of being re-wrapped with the default 502. This also corrects the pre-existing `No content in response` guard on the JSON-schema path. Added an assertion on `status_code` to the test, which fails without the fix.

PR #5253, opened 2026-07-18
fix(mcp): mark tool execution failures as errors

### LINKED ISSUE #5252: Bug: MCP server reports tool execution failures with isError=false
### Browser Use Version

0.13.6

The native Browser Use MCP server converts tool execution failures into ordinary content lists. The Python MCP SDK consequently emits a normal `CallToolResult` with `isError: false`.

This affects both:

- exceptions caught by `handle_call_tool`, such as CDP connection failures
- expected action failures returned as strings, such as `Element with index N not found`, invalid click arguments, missing sessions, or unknown tools

Minimal reproduction against 0.13.6:

```python
import asyncio

from mcp import ClientSession, StdioServerParameters
from mcp.client.stdio import stdio_client


async def main():
    params = StdioServerParameters(
        command="uvx",
        args=["--from", "browser-use[cli]==0.13.6", "browser-use", "--mcp"],
        env={"BU_CDP_URL": "ws://127.0.0.1:1"},
    )
    async with stdio_client(params) as streams:
        async with ClientSession(*streams) as session:
            await session.initialize()
            result = await session.call_tool(
                "browser_type",
                {"index": 999999, "text": "x"},
            )
            print(result.model_dump_json(by_alias=True))


asyncio.run(main())
```

Actual result:

```json
{
  "content": [
    {
      "type": "text",
      "text": "Error: Failed to establish CDP connection to browser: no close frame received or sent"
    }
  ],
  "structuredContent": null,
  "isError": false
}
```

The same envelope is produced for semantic action failures. For example, `_type_text` returns `Element with index {index} not found` as a string, and `handle_call_tool` wraps it as successful text content.

The cause is in `browser_use/mcp/server.py`: `handle_call_tool` is declared to return a content list, catches exceptions, and returns error text as a content list. The current implementation is here:

https://github.com/browser-use/browser-use/blob/950eb03617e67548d759c02beac1ad122c6b6458/browser_use/mcp/server.py#L459-L472

Representative semantic failure returns are here:

https://github.com/browser-use/browser-use/blob/950eb03617e67548d759c02beac1ad122c6b6458/browser_use/mcp/server.py#L741-L837

### Expected behavior

Failures that originate while executing a valid MCP tool call should return a `CallToolResult` with `isError: true` and actionable error content.

The MCP specification explicitly separates protocol errors from tool execution errors and says tool execution errors should be reported in the result with `isError: true` so clients can expose them for model recovery:

https://modelcontextprotocol.io/specification/2025-06-18/server/tools#error-handling

This distinction is also consumed deliberately by Codex:

- no `CallToolResult` received is classified as an MCP request or transport/protocol failure
- a received `CallToolResult` with `isError: true` is classified as a tool-result failure
- MCP lifecycle status is `Failed` only for `isError: true` or request failure

https://github.com/openai/codex/blob/f24e6954702aa5f5705c65ee57748c5817bdb507/codex-rs/core/src/mcp_tool_call/telemetry.rs#L13-L17

https://github.com/openai/codex/blob/f24e6954702aa5f5705c65ee57748c5817bdb507/codex-rs/core/src/mcp_tool_call.rs#L921-L940

As a result, Codex and other conforming hosts display and record these Browser Use failures as completed tool calls.

There is a related inverse client-side fix in #5235: Browser Use is being updated to respect `CallToolResult.isError` from other MCP servers. The native Browser Use MCP server should emit the same signal for its own failures.

### Possible implementation

Return `types.CallToolResult(content=[...], isError=True)` for caught execution exceptions and semantic action failures. Successful actions can continue to return ordinary content or an explicit result with `isError=False`.

### LLM Model

Codex GPT-5.6-sol, though this is reproducible directly with the official Python MCP client and does not depend on an LLM.

### Operating System and Browser Versions

macOS 15, Browser Use 0.13.6, MCP Python SDK 1.26.0

### Full DEBUG Log Output

The raw MCP result above is the relevant protocol output. The Browser Use MCP server disables normal logging in MCP mode.

coleski:
## Summary

- return an explicit MCP `CallToolResult` for every native MCP tool call
- set `isError=true` for caught execution exceptions and recoverable action failures
- preserve the existing human-readable error content for model recovery
- keep successful calls at `isError=false`

This fixes the server-side inverse of #5235. Conforming MCP hosts currently record Browser Use failures such as missing element indexes and CDP connection errors as successful tool calls because the server returns ordinary content lists.

Closes #5252.

## Verification

- `uv run pytest tests/ci/test_mcp_server_error_results.py tests/ci/security/test_mcp_allowed_domains.py` — 7 passed
- `uv run pre-commit run --files browser_use/mcp/server.py tests/ci/test_mcp_server_error_results.py` — all hooks passed
- raw MCP reproduction now returns actionable content with `isError=true`; successful content remains `isError=false`

## AI disclosure

The bug was diagnosed and the patch was implemented with Codex assistance. The MCP protocol behavior was reproduced directly with the official Python MCP client, and the focused tests, existing allowed-domain tests, formatting, lint, and type checks were run before submission.

---
## Summary by cubic
Return explicit MCP `CallToolResult` for every tool call and mark execution failures with `isError=true`. This stops hosts from treating browser-use failures as successful calls while keeping readable error messages for recovery.

- **Bug Fixes**
  - Return `CallToolResult` with `isError=false` for successful calls.
  - Introduce `MCPToolExecutionError`; raise for recoverable issues (no session, missing element, unknown tool, LLM/CDP errors).
  - Map unexpected exceptions to error results with preserved messages (`isError=true`).
  - Add tests for success, unknown tool, missing element, unexpected exception, and `CallToolRequest` handler registration.

<sup>Written for commit 4bedd9d3cd16f60624b98fc89e1d76f997d72637. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5253?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/mcp/server.py
+++ b/browser_use/mcp/server.py
@@ -100,6 +100,10 @@ from browser_use.tools.service import Tools
 logger = logging.getLogger(__name__)
 
 
+class MCPToolExecutionError(Exception):
+	"""A recoverable tool execution failure that should be visible to the MCP client."""
+
+
 def _ensure_all_loggers_use_stderr():
 	"""Ensure ALL loggers only output to stderr, not stdout."""
 	# Get the stderr handler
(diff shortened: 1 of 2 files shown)


trungminhdo4-glitch:
One additional protocol-boundary concern: `_execute_tool()` raises `MCPToolExecutionError` for an unknown tool, and `_handle_call_tool()` converts that into `CallToolResult(isError=True)`. The regression test `test_unknown_tool_sets_is_error_true` locks this behavior in. Under MCP 2025-06-18, `isError=true` is for errors originating during execution of a valid tool; failures to find the requested tool are protocol-level errors. With `mcp==1.26.0`, the low-level server refreshes its tool cache but does not reject an unknown name before invoking the registered application handler, so this likely needs a narrow explicit protocol-error path at that boundary rather than routing through `MCPToolExecutionError`.

The existing Cubic P1 about unsuccessful `Agent.run()` histories still appears applicable; that is not a new finding. Coverage through an actual `ClientSession.call_tool()` request would also be stronger than testing `_handle_call_tool()` directly.

PR #5258, opened 2026-07-20
fix(mcp): report tool execution failures with isError=True

### LINKED ISSUE #5252: Bug: MCP server reports tool execution failures with isError=false

sergioperezcheco:
The native browser-use MCP server wrapped every tool outcome in a plain content list, so the Python MCP SDK emitted CallToolResult with isError=False even when the call failed. This affected both raised exceptions (CDP connection failures, missing sessions) and semantic failures returned as strings like 'Element with index N not found', 'Unknown tool', and 'Error: ...'. Reported in #5252.

The fix is in handle_call_tool: on the exception path it now returns CallToolResult(isError=True), and it classifies the semantic error strings that the _execute_tool helpers return so those are surfaced as errors too. Success messages (e.g. 'Clicked element 5', 'Navigated to: ...') are left untouched. The detection is centralized in a small _is_error_response helper backed by a sentinel-prefix list covering the existing failure strings.

I kept the change minimal and at the handler boundary so individual tool helpers don't need touching. Added regression tests in tests/ci/security/test_mcp_tool_failure_iserror.py covering the error-string classification (both failure and success cases) and the unknown-tool path. Verified against the repro from the issue: a failed browser_type now yields isError=true instead of false.

Fixes #5252

---
## Summary by cubic
Reports MCP tool execution failures as `CallToolResult(isError=True)` so clients can react correctly. Previously failures returned plain content with `isError=False`; now raised exceptions and typed `ToolFailure` values from `_execute_tool` helpers surface as errors. Fixes #5252.

- Helpers return `ToolFailure` instead of error strings, so successful output that starts with `Error:` is no longer misclassified.
- `handle_call_tool` maps exceptions and `ToolFailure` to `CallToolResult(isError=True)` at a single boundary; successful results are unchanged.
- Adds regression tests for missing sessions/elements, unknown tools, raised exceptions, and successful text starting with `Error:`.

<sup>Written for commit 9192d68af2666e42e3a5a147abb8ee988b4227d5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5258?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MagMueller:
Thanks for fixing the `isError` boundary. I re-ran the focused suite and two semantic probes on exact head `bce31d20`. The tests are green, but two gaps remain:

1. `_is_error_response('Error: the page returned a validation notice')` returns `True`. `browser_extract_content` can return arbitrary successful page text, so a legitimate extraction beginning with `Error:` is reported as a failed MCP tool.
2. `_is_error_response('Session abc123 not found')` returns `False`, even though `_close_session` returns that failure shape. That path still produces `isError=false`.

Could we replace string-prefix inference with a typed internal failure/outcome and map it to `CallToolResult(isError=True)` once at the handler boundary? Please add registered-handler coverage for arbitrary successful text beginning with `Error:`, missing session, missing element, unknown tool, and a raised exception. That keeps protocol status independent of user or page text and closes both gaps.

sergioperezcheco:
Both gaps are closed in 9192d68a: the sentinel prefixes are gone, and tool helpers now return a typed `ToolFailure` that the registered call-tool handler maps to `CallToolResult(isError=True)` at a single boundary.

Added the registered-handler coverage you listed: successful text starting with `Error:` stays a normal result, while missing session, missing element, unknown tool, and a raised exception all produce `isError=True`. The focused suite passes (7/7) against current head.

Issue #5132, opened 2026-07-03
Bug: Remote browser downloads can complete without triggering click download completion callback, causing click action timeout


houying910523:
### Browser Use Version

0.12.1 - 0.12.9

### Summary

  When using a remote browser session (`is_local=False`), a click that triggers a file download can time out even though the download has actually completed.

  ### Affected area

  - `browser_use/browser/watchdogs/default_action_watchdog.py`
  - `browser_use/browser/watchdogs/downloads_watchdog.py`

  ### What happens

  `DefaultActionWatchdog._execute_click_with_download_detection()` waits for a direct completion callback to set `download_completed`:

  - registers `on_start`, `on_progress`, `on_complete`
  - waits on `download_started.wait()`
  - then waits on `download_completed.wait()`

  In the remote download path, `DownloadsWatchdog.download_progress_handler()` handles `state == "completed"` by dispatching `FileDownloadedEvent`, but it does **not** invoke `_download_complete_callbacks`.

  So the download may complete successfully, but the click action still waits until timeout because its direct completion callback is never called.

  ### Why this is remote-only

  For local browsers, the completion path eventually goes through `_track_download(...)`, which invokes `_download_complete_callbacks`.

  For remote browsers (`is_local=False`), the code synthesizes a `FileDownloadedEvent` directly and skips the direct completion callback path.

  ### Expected behavior

  When a remote download completes, the same direct completion callbacks used by click download detection should be invoked, so the click action can finish successfully instead of timing out.

  ### Actual behavior

  - download starts
  - progress events may be received
  - `FileDownloadedEvent` is dispatched
  - click action still times out waiting for `download_completed`

  ### Reproduction

  This reproduces when:

  1. Use a remote browser session / remote CDP / sandbox browser (`is_local=False`)
  2. Click a link or button that triggers a file download
  3. Observe that the file download completes, but the click action still reports a download timeout

  ### Root cause

  In `DownloadsWatchdog.download_progress_handler()`:

  - local branch: completion path invokes direct completion callbacks
  - remote branch: completion path only dispatches `FileDownloadedEvent`

  The click action depends on the direct callback, not just the event bus.

  ### Suggested fix

  In the remote `state == "completed"` branch, after constructing the download metadata, invoke `_download_complete_callbacks` with the same completion info structure used by the local path.

  ### Notes

  I validated this with a remote sandbox browser session. Adding the direct completion callback call in the remote branch resolved the timeout in my testing.

[...]

PR #5211, opened 2026-07-13
fix(utils): prevent redact_sensitive_string cascade corruption

santhreal:
## Summary

Fixes cascading corruption in `redact_sensitive_string` where a shorter secret value (e.g. `"secret"`) matched inside a `<secret>...</secret>` placeholder tag written by an earlier iteration for a longer secret (e.g. `"supersecret"`).

Closes #5135

## Root cause

The iterative `str.replace` loop processes secrets longest-first, but each replacement inserts `<secret>key</secret>` tags that contain the literal word "secret". A subsequent iteration replacing a shorter secret that happens to be a substring of the tag text corrupts the output.

Example: `{"password": "supersecret", "type": "secret"}` produces
`leaked <<secret>type</secret>>password</<secret>type</secret>> token` instead of
`leaked <secret>password</secret> token`.

## Fix

Replaced the iterative `str.replace` loop with a single-pass `re.sub` using a regex alternation of all secret values (longest first). Each character in the input is consumed at most once, so earlier replacements cannot be corrupted by later ones.

## Tests

6 new tests in `tests/ci/test_redact_cascade.py`:
- Normal redaction
- Cascade substring scenario (the #5135 regression)
- Multiple overlapping secrets
- Empty secrets dict
- Secret value matching tag syntax
- Multiple occurrences of same secret

---
## Summary by cubic
Prevents cascade corruption in redaction when shorter secrets match inside previously inserted <secret>...</secret> tags. Uses a single-pass regex in redact_sensitive_string that prefers longest matches and processes each position once (fixes #5135).

- **Bug Fixes**
  - Replace iterative str.replace loop with single-pass re.sub using escaped, longest-first alternation to avoid matching inside generated tags.
  - Add 6 tests covering the #5135 regression and edge cases (overlaps, empty dict, tag-like values, repeats).

<sup>Written for commit d9026a2b93db316c6c444841868b2b233c5e1cbd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5211?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #5260, opened 2026-07-20
fix: prevent redact_sensitive_string cascade corruption (fixes #5135)

### LINKED ISSUE #5135: Bug: redact_sensitive_string cascades when a secret value is substring of a placeholder tag

AdarshJ173:
## Problem

Fixes #5135 — `redact_sensitive_string` produces corrupted output when a secret value is a substring of a placeholder tag written by an earlier iteration.

### Reproduction

```python
    'leaked supersecret token',
```

**Before (broken):** `leaked <secret>password</secre<secret>type</secret>> token`

**After (fixed):** `leaked <secret>password</secret> token`

### Root cause

The original implementation uses a sequential `str.replace()` loop:

```python
# BEFORE (broken):
```

Iteration 1 replaces `supersecret` → `<secret>password</secret>` ✅

Iteration 2 looks for `secret` — finds it inside `<secret>password</secret>` and replaces it → corruption ❌

This corruption feeds directly into LLM messages via `browser_use/agent/views.py` and `browser_use/agent/message_manager/service.py`.

## Fix

Replace the loop with a **single-pass `re.sub()`** using a compiled alternation pattern:

```python
# AFTER (fixed):
sorted_items = sorted(sensitive_values.items(), key=lambda item: len(item[1]), reverse=True)
sorted_items = [(k, v) for k, v in sorted_items if v]
lookup = {v: f'<secret>{k}</secret>' for k, v in sorted_items}
pattern = re.compile('|'.join(re.escape(v) for _, v in sorted_items))
return pattern.sub(lambda m: lookup[m.group(0)], value)
```

A single regex pass is atomic — once a position is matched and consumed it cannot be matched again, making cascade corruption impossible. Secrets are still sorted longest-first to prevent shorter secrets from shadowing longer ones.

## Behaviour preserved

- Placeholder format `<secret>{key}</secret>` unchanged
- Empty secrets filtered (same as before via the `if val` guard)
- Longest-first ordering preserved
- No new dependencies (`re` is stdlib)

---
## Summary by cubic
Fixes two bugs: stops `redact_sensitive_string` tag corruption and ensures remote downloads trigger completion callbacks to prevent click-action timeouts. Fixes #5135 and #5132.

- **Bug Fixes**
  - Replace sequential `str.replace()` with single-pass `re.sub()` sorted longest-first to avoid matching inside placeholders. Filter empty secrets; keep `<secret>{key}</secret>` format; no new deps.
  - For remote browsers, invoke `_download_complete_callbacks` on completion using the same `complete_info` as local. Prevents click-action waits; still dispatches `FileDownloadedEvent`; no filesystem access.

<sup>Written for commit b4d8ac52f2d5773f0f6f5edb41fbb0659039231f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5260?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/browser/watchdogs/downloads_watchdog.py
+++ b/browser_use/browser/watchdogs/downloads_watchdog.py
@@ -398,7 +398,7 @@ class DownloadsWatchdog(BaseWatchdog):
 				if self.browser_session.is_local:
 					if file_path:
 						self.logger.debug(f'[DownloadsWatchdog] Download completed: {file_path}')
-						# Track the download
+						# Track the download (also fires _download_complete_callbacks internally)
 						self._track_download(file_path, guid=guid)
 						# Mark as handled to prevent fallback duplicate dispatch
 						try:
@@ -433,7 +433,11 @@ class DownloadsWatchdog(BaseWatchdog):
 												pass
 											break
 				else:
-					# Remote browser: do not touch local filesystem. Fallback to downloadPath+suggestedFilename
+					# Remote browser: do not touch local filesystem.
+					# Build completion metadata and fire _download_complete_callbacks so that
+					# any click action waiting on download_completed.set() is unblocked.
+					# Previously only FileDownloadedEvent was dispatched here, which caused
+					# click-action timeouts on remote browser sessions (issue #5132).
 					info = self._cdp_downloads_info.get(guid, {})
 					try:
 						suggested_filename = info.get('suggested_filename') or (Path(file_path).name if file_path else 'download')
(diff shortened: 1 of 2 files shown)


sauravpanda:

PR #5261, opened 2026-07-20
fix: prevent cascade re-redaction in redact_sensitive_string (fixes #5135, #5248)

### LINKED ISSUE #5135: Bug: redact_sensitive_string cascades when a secret value is substring of a placeholder tag

### LINKED ISSUE #5248: redact_sensitive_string: placeholder tags get re-redacted when a secret value contains the substring 'secret' (cascade leak)
### Describe the bug

### Steps to reproduce
```python

sensitive = {
}

```

### Expected behavior
```
```

### Actual behavior
```
```

### Root cause

```python
```

### Environment
- browser-use: current main
- Python 3.12

AdarshJ173:
## Problem

`redact_sensitive_string` in `browser_use/utils.py` used a sequential `str.replace` loop sorted by secret length. When a **shorter secret's value appeared as a substring inside a `<secret>key</secret>` placeholder written by an earlier iteration**, it would corrupt the already-redacted output and produce malformed tags that reach LLM input.

### Reproduction (issues #5135 and #5248)

```python

sensitive = {'password': 'supersecret', 'type': 'secret'}
value = 'leaked supersecret token'
# Before fix: 'leaked <secret>password</secre<secret>type</secret>>' ← corrupted
# After fix:  'leaked <secret>password</secret>'                       ← correct
```

The second iteration replaces `'secret'` inside the already-written `<secret>password</secret>` tag because `str.replace` doesn't protect previously substituted regions.

This feeds corrupted/nested tags directly into LLM messages via `agent/views.py` and `agent/message_manager/service.py`.

## Fix

Replace the sequential `str.replace` loop with a **single-pass `re.sub`** using a combined alternation pattern:

```python
lookup = {v: f'<secret>{k}</secret>' for k, v in sorted_items}
pattern = re.compile('|'.join(re.escape(v) for _, v in sorted_items))
return pattern.sub(lambda m: lookup[m.group(0)], value)
```

Because Python's regex engine scans left-to-right and each consumed position is never revisited, **no previously written placeholder can ever be re-matched**. The alternation preserves longest-first priority (secrets are still sorted by length before building the pattern).

## Changes

- `browser_use/utils.py`: rewrote `redact_sensitive_string` to use single-pass regex substitution
- Added guard for empty `sensitive_values` dict (early return)
- Added guard for empty individual secret values (skip them to avoid matching the empty string everywhere)
- No changes to public API, callers, or tests — the function signature is identical

## Testing

Manually verified the reproduction case from both issues now produces correct output. The fix is deterministic and OS-independent (pure string/regex logic).

Fixes #5135
Fixes #5248

---
## Summary by cubic
Prevents cascade re-redaction in `redact_sensitive_string`, so placeholder tags are no longer corrupted when a shorter secret appears inside an already-redacted tag. Fixes #5135 and #5248.

- **Bug Fixes**
  - Switched to a single-pass `re.sub` with an alternation pattern, sorted longest-first.
  - Skips empty secrets and early-returns on empty input.
  - Preserves the existing API and `<secret>{key}</secret>` format.

<sup>Written for commit 81de91935bd518a0dcbd3ed0cef5fbfb013bdc62. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5261?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/utils.py
+++ b/browser_use/utils.py
@@ -74,10 +74,36 @@ def collect_sensitive_data_values(sensitive_data: dict[str, str | dict[str, str]
 
 
 def redact_sensitive_string(value: str, sensitive_values: dict[str, str]) -> str:
-	"""Replace sensitive values with placeholders, longest matches first to avoid partial leaks."""
-	for key, secret in sorted(sensitive_values.items(), key=lambda item: len(item[1]), reverse=True):
-		value = value.replace(secret, f'<secret>{key}</secret>')
-	return value
+	"""Replace sensitive values with placeholders in a single atomic pass.
+
+	Uses re.sub with a compiled alternation pattern sorted longest-first so that
+	(a) longer secrets shadow shorter prefixes, and (b) a single regex pass
+	prevents cascade corruption where a shorter secret matches text inside a
+	placeholder tag produced by an earlier iteration.
+
+	Example of the cascade bug this fixes (issue #5135):
+	  secrets = {'password': 'supersecret', 'type': 'secret'}
+	  input:  'leaked supersecret token'
+	  before: 'leaked <secret>password</secre<secret>type</secret>>' (corrupted)
+	  after:  'leaked <secret>password</secret>'  (correct)
+	"""
+	if not sensitive_values:
+		return value
+
+	# Sort longest-first so shorter secrets don't shadow longer ones
+	sorted_items = sorted(sensitive_values.items(), key=lambda item: len(item[1]), reverse=True)
+
+	# Filter out empty secrets to avoid matching the empty string everywhere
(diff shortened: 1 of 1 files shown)


sauravpanda:

Issue #3415, opened 2025-10-24
Interaction Issue: Agent Fails To Select Dropdown Menu Items


Caleb-Hurst:
### Browser Use Version

0.9.0

### LLM Model

Other (specify in description)

### Screenshots, Description, and task prompt given to Agent

LLM Used: `claude-sonnet-4-0` (This happened with ChatBrowserUse() LLM as well as OpenAi) 

Currently the Browser Agent Is not capable of selecting certain dropdown menu items. Please see the video below, but it can select them for a short time but somehow unselects them prior to submission. 

## Dev Note 
- Js Framework: Vue.js
- It is able to select certain dropdown items, I believe. But just not these, I was unable to discern a difference between the structure of the dropdowns. 

https://github.com/user-attachments/assets/3696d77d-6629-4274-857d-3ec76d04ce90

### HTML around where it's failing

```html
## Pay Frequency HTML 


<div class="onboard-new-hire-form__field"><label for="payFrequency" class="field-label">
          Pay Frequency
          <span title="This field is required" class="text--red">*</span></label> <select id="payFrequency" class="onboard-new-hire-form__input"><option value="">Select Frequency</option> <option value="peryear">per year</option> <option value="perhour">per hour</option></select></div>



## Hire Type HTML 


<div class="onboard-new-hire-form__row"><div class="onboard-new-hire-form__field"><label for="employmentStatus" class="field-label">
        Employment Status
        <span title="This field is required" class="text--red">*</span></label> <select id="employmentStatus" class="onboard-new-hire-form__input"><option value="fullTime">Full Time</option> <option value="partTime">Part Time</option></select></div> <div class="onboard-new-hire-form__field"><label for="hireType" class="field-label">
        Hire Type
        <span title="This field is required" class="text--red">*</span></label> <select id="hireType" class="onboard-new-hire-form__input"><option value="">Select Type</option> <option value="w-2">W-2</option> <option value="1099">1099</option> <option value="k-1">K-1</option> <option value="test">Test</option></select></div></div>
```

### Operating System & Browser Versions

Mac OS | Chrome

MagMueller:
I am on it!

MagMueller:
## Fixed! Vue.js Dropdown Selection Issue Resolved

I've identified and fixed the issue. The problem was that the dropdown selection handler was only dispatching a `change` event, but **Vue.js requires an `input` event** for its `v-model` directive to properly register changes.

### Root Cause

[...]

Issue #2867, opened 2025-08-29
Interaction Issues: Cannot select item from drop down


dcullender:
### Browser Use Version

0.7.0

### LLM Model

Other (specify in description), gpt-4.1-mini

### Screenshots, Description, and task prompt given to Agent

This also happens on model meta-llama/llama-4-maverick-17b-128e-instruct.

I have tried multiple prompt permutations to try and select an option from the dropdown box with no success (even breaking it up into multiple tasks). The sample I provided is the most basic version.

This worked when using browser-use with playright so it is a regression.

The form is located on https://www.lesleypt.co.uk/. Click on 'contact' link and the 'Choose an Option' is what is causing the issue.

### HTML around where it's failing

```html
<select class="wixui-dropdown__input McLPII has-custom-focus T6F83Z" id="collection_comp-lwm7karq" data-testid="select-trigger" required="" aria-required="true" aria-expanded="true" aria-invalid="false"><option value="" disabled="" class="R4tv1w U5WS_f" selected="">Choose an option</option><option value="1:1 Personal Training" class="R4tv1w" aria-selected="false">1:1 Personal Training</option><option value="1:1 Online Personal Training" class="R4tv1w" aria-selected="false">1:1 Online Personal Training</option><option value="Nutrition" class="R4tv1w" aria-selected="false">Nutrition</option><option value="Other" class="R4tv1w" aria-selected="false">Other</option></select>
```

### Operating System & Browser Versions

macOS Sequoia 15.6.1

### Python Code Sample

```python
async def main():
	llm = ChatOpenAI(model='gpt-4.1-mini')
	task = "navigate to https://www.lesleypt.co.uk/, select the 'contact' link and then select 'Nutrition' from the drop down 'Options"
	agent = Agent(task=task, llm=llm)
	await agent.run()
```

### Full DEBUG Log Output

```shell
/Users/daniel/IdeaProjects/metisloop-platform/services/browser-use/venv/MacOSX/Miniconda3-py312_24.9.2-0/envs/python-3.11/bin/python /Users/daniel/IdeaProjects/metisloop-platform/services/browser-use/app/src/main/python/examples/getting_started/01_basic_search.py
INFO     [Agent] 🧠 Starting a browser-use version 0.7.0 with model=gpt-4.1-mini
INFO     [Agent] 🚀 Task: navigate to https://www.lesleypt.co.uk/, select the 'contact' link and then select 'Nutrition' from the drop down 'Options
[...]

PR #5274, opened 2026-07-23
Fix clear-to-empty on React controlled inputs (native value setter)

MagMueller:
`_clear_text_field` clears via the instance `this.value = \"\"`. React patches the instance value setter to track changes, so that assignment updates React's tracker and the synthetic `input` event gets deduped — onChange never fires and the controlled component keeps its old state while the DOM looks empty. The synchronous `finalText` check reads `\"\"` at that instant, so the keystroke fallbacks (which would work) are unreachable, and the action reports "Typed ''" success.

**Blast radius is narrow but real**: only `input(index, \"\", clear=true)` (pure clear). When text follows the clear, the real per-character key events resync React on the first keystroke, and #4023's concat retry backstops. The pure-clear path is the silent one — observed in evals as an agent retrying a search-box clear 11 times while the React-filtered list never reset.

History (why this is a gap, not a choice): the original `_clear_text_field` (d80549038, Aug 2025) used real Ctrl+A+Backspace key events — React-safe. A same-day follow-up (f4a34bc1d "fix-clear-text", no body) promoted JS value-set to Strategy 1. Later commits added the correct native-setter technique for date inputs (`_set_value_directly`, 0badc5d4c) and for the type-retry path (#4023) — it just never reached the clear path. Related user reports: #3415, #2867 (dropdown flavor of the same tracker mechanism).

Two changes:
1. `_clear_text_field` uses the prototype's native value setter (same technique as `_set_value_directly`), so React's tracker sees the change and onChange fires with the cleared value.
2. If a pure clear still fails, raise `BrowserError` instead of logging a warning and reporting success — the agent gets an honest failure it can react to.

---
## Summary by cubic
Fixes clearing React‑controlled text fields so the DOM and component state stay in sync. Uses the native prototype value setter on real inputs so React sees the change and `onChange` fires.

- **Bug Fixes**
  - Clear now calls the element‑type’s native value setter for `HTMLInputElement`/`HTMLTextAreaElement` with `instanceof` guards; web components and other elements keep their own setter to avoid Illegal invocation.
  - Fixed the descriptor lookup in the concat‑retry path to pick the correct setter per element type, with a safe fallback to `this.value`.

<sup>Written for commit d4abb27612ea7a6a11256917f25a2f03a57151a0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5274?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/browser/watchdogs/default_action_watchdog.py
+++ b/browser_use/browser/watchdogs/default_action_watchdog.py
@@ -1383,7 +1383,19 @@ class DefaultActionWatchdog(BaseWatchdog):
 								} catch (e) {
 									// ignore
 								}
-								this.value = "";
+								// Use the prototype's native value setter: React patches the instance
+								// setter to track values, so a direct `this.value = ""` makes React's
+								// tracker think nothing changed and the input event gets ignored,
+								// leaving controlled components with the old value.
+								const proto = this.tagName === 'TEXTAREA'
+									? window.HTMLTextAreaElement.prototype
+									: window.HTMLInputElement.prototype;
+								const desc = Object.getOwnPropertyDescriptor(proto, 'value');
+								if (desc && desc.set) {
+									desc.set.call(this, "");
+								} else {
+									this.value = "";
+								}
 								this.dispatchEvent(new Event("input", { bubbles: true }));
 								this.dispatchEvent(new Event("change", { bubbles: true }));
 								return {cleared: true, method: 'value', finalText: this.value};
--- a/browser_use/browser/watchdogs/default_action_watchdog.py
+++ b/browser_use/browser/watchdogs/default_action_watchdog.py
@@ -1383,16 +1383,23 @@ class DefaultActionWatchdog(BaseWatchdog):
 								} catch (e) {
 									// ignore
 								}
-								// Use the prototype's native value setter: React patches the instance
-								// setter to track values, so a direct `this.value = ""` makes React's
-								// tracker think nothing changed and the input event gets ignored,
-								// leaving controlled components with the old value.
-								const proto = this.tagName === 'TEXTAREA'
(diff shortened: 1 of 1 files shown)


MagMueller:
Live repro on real React 18 (controlled input, state shown alongside): **old clear JS** → DOM value empty but React state still `preset-query` (silent desync, exactly the reported symptom); **this PR's JS** → DOM and React state both cleared. The sibling select PR (#5275) was closed because the same repro showed native selects are NOT affected — the tracker dedupe is input/textarea-specific, which bounds this fix's scope precisely.

MagMueller:
Applied both blockers from adversarial review:
1. **Illegal-invocation on non-input elements**: the native setter is now gated on `instanceof HTMLInputElement/HTMLTextAreaElement` with a try/catch fallback to the element's own setter — web components/selects keep their previously-working clear. Verified live: custom element with own value accessor and a native select both clear cleanly through the fallback; React 18 controlled input still clears with state sync.
2. **Dead honest-failure hunk removed** — as demonstrated, Strategy 2 returns blind success so the raise was unreachable, and the pre-existing 'actual value differs' note on main already covers the signal. The PR is now the minimal proven fix.
Also folded in the same pre-existing descriptor bug in the #4023 concat-retry JS (it grabbed the HTMLInputElement descriptor even for textareas — same Illegal-invocation family), per review suggestion. Autocomplete suite: 10/10 pass.

PR #3944, opened 2026-01-25
fix(#2867): verify dropdown selection and retry on framework revert

laithrw:
Resolves #2867.

The select_dropdown action was claiming success even when frameworks like Wix intercepted and reverted the programmatic selection. The agent would report task completion, but the dropdown value remained unchanged.

- Added verification after setting selection - checks if `element.value === expectedValue`
- If framework reverted the value, try click-based fallback (mousedown -> selectedIndex -> click option -> mouseup -> change)
- If both methods fail, properly report the error instead of claiming success
- Agent can then recover by clicking visible option elements directly

Only affects native `<select>` elements in `on_SelectDropdownOptionEvent`. ARIA/custom dropdowns already use click-based selection.

---
## Summary by cubic
Fixes dropdown selection reliability by verifying the final value and retrying with a click-based fallback when frameworks revert programmatic changes. Aligns with #2867 so the agent no longer claims success when the value doesn’t stick.

- **Bug Fixes**
  - Verify selection after setting (element.value === expectedValue).
  - If reverted, use click-based fallback (focus, mousedown, set selectedIndex, click option, mouseup, change, blur).
  - If both methods fail, return a clear error so the agent can recover.
  - Provide target option and available options when a revert is detected to aid recovery.
  - Scope: affects native <select> in on_SelectDropdownOptionEvent; ARIA/custom dropdowns unchanged.

<sup>Written for commit 6f8c07793222c1ecde1fc7e4ba65d4bd8de8ec94. Summary will update on new commits.</sup>

PR #5275, opened 2026-07-23
Fix select_dropdown silent revert on React-controlled selects

MagMueller:
Two prior fixes each stopped one step short of the React value-tracker mechanism:
- #3419 added the input/change event sequence "critical for Vue's v-model and React controlled components" — but sets the value via the instance `element.value =`, which updates React's tracker so the change event gets deduped and onChange never fires; React reverts the selection on its next render.
- #3944 (issue #2867) added a revert check — but it runs synchronously, and the revert happens on React's next render tick, so the check passes and the click fallback it added never fires. Exactly the silent-success failure mode it set out to kill, one tick later.

This PR closes the loop:
1. Set the value via the native `HTMLSelectElement.prototype` setter (same technique `_set_value_directly` already uses for date inputs), so React's tracker sees the change and onChange fires.
2. Add a 100ms async post-check on the native-select path (`selection_result.value` is only set there, so ARIA/custom dropdown paths are untouched): if the framework still reverted, raise `BrowserError` telling the agent to click the dropdown instead — an honest failure instead of a silent success.

In evals on non-React (native select) sites: 3,352 select_dropdown calls, 1 error — this path is unaffected. The dropdown CI tests are currently skipped upstream (pre-existing `@pytest.mark.skip` markers); syntax and the surrounding blank-page suite verified.

---
## Summary by cubic
Fixes silent reverts on React-controlled native `<select>` by using the native value setter and adding a short async post-check. onChange now fires, and any framework-driven revert becomes a clear, actionable error.

- **Bug Fixes**
  - Set value via `HTMLSelectElement.prototype.value` setter (fallback to `element.value`).
  - Keep `option.selected` and `selectedIndex` updates.
  - Add 100ms async post-check on native-select path; if reverted, raise `BrowserError` instructing a click fallback.
  - Scope limited to native select path; ARIA/custom dropdowns unchanged. Non-React sites unaffected in evals.

<sup>Written for commit ff9f0836cfc4af739b55bac3a7181110aa9e8c65. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5275?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/browser/watchdogs/default_action_watchdog.py
+++ b/browser_use/browser/watchdogs/default_action_watchdog.py
@@ -3293,8 +3293,19 @@ class DefaultActionWatchdog(BaseWatchdog):
 									// This simulates the user focusing on the dropdown before changing it
 									element.focus();
 
-									// Then set the value using multiple methods for maximum compatibility
-									element.value = expectedValue;
+									// Set the value via the native prototype setter. React patches the
+									// instance 'value' setter on <select> to track changes; a direct
+									// element.value assignment updates React's tracker so the change
+									// event below gets deduped, onChange never fires, and React reverts
+									// the selection on its next render.
+									const nativeDesc = Object.getOwnPropertyDescriptor(
+										window.HTMLSelectElement.prototype, 'value'
+									);
+									if (nativeDesc && nativeDesc.set) {
+										nativeDesc.set.call(element, expectedValue);
+									} else {
+										element.value = expectedValue;
+									}
 									option.selected = true;
 									element.selectedIndex = option.index;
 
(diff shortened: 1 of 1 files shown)


MagMueller:
Closing after live repro refuted the premise: on real React 18, a controlled native <select> updates state fine with plain `element.value =` + change event (verified value-only, no selectedIndex rescue) — React's value-tracker dedupe applies to input/textarea, not native selects. The 1-error-in-3352-calls eval data agrees. The user reports behind #3415/#2867 likely involve custom dropdown components or other frameworks, which this change wouldn't affect. Not worth 100ms per select + a new error path for an unreproducible bug.

PR #5280, opened 2026-07-23
filesystem: write_file can create tiny binary images (png/gif/jpg/webp) for upload flows

MagMueller:
## Problem

The agent's file tools only write text formats, so any task that needs to **upload an image/file** (a photo, a GIF, a document) is unwinnable by construction — the model burns steps on `DataTransfer` JS hacks and then gives up or fabricates. In evals this made whole upload-validation tasks impossible for every model.

## Fix

`write_file` now accepts small image extensions (`.png .gif .jpg .jpeg .webp`) where `content` is the **base64 of a valid image**. The bytes are decoded and written to disk, so the file is a real, uploadable image. A 1×1 PNG is ~92 base64 chars (~23 tokens), a GIF just 56 — cheap for the model to emit.

- `Base64BinaryFile` writes decoded **bytes** to disk (not the base64 text) and `read()` returns a `[binary png file, N bytes]` stub, so base64 never enters the agent prompt (`describe()` runs every step).
- **Strict** base64 decode — invalid content returns an error and **no file is created**, so a corrupt 'image' can't pass upload's `file_size > 0` check.
- Registered in `_file_types` so `upload_file.get_file` resolves it by basename (the load-bearing bit — upload only finds FileSystem files whose extension is registered), and in `from_state` so restored sessions keep it. Removed the now-supported extensions from `UNSUPPORTED_BINARY_EXTENSIONS`.
- Untouched: text `write_file`, PDF/DOCX rendering (still markdown→document), and rejection of other binaries (`.mp4 .zip .exe …`).

## Testing

5 new tests + all 13 existing image tests pass (18 total): real PNG magic bytes on disk, uploadable by basename, no base64 leak in `describe()`, invalid-base64 rejected with no corrupt file, state round-trip. Verified end-to-end that a written `.png` resolves through the exact `upload_file` path.

Scoped deliberately to tiny fixtures (images only, strict decode) — not a general binary-write feature.

---
## Summary by cubic
`write_file` can now create tiny binary images (.png, .gif, .jpg/.jpeg, .webp) from base64 so upload flows use real files. We validate base64 and magic bytes, reject mismatched types, and only register files after a successful write to avoid ghost entries.

- **New Features**
  - Added `Base64BinaryFile` that decodes bytes to disk; `read()` returns “[binary <ext> file, N bytes]” so base64 stays out of prompts.
  - `write_file` accepts `.png .gif .jpg .jpeg .webp`; strict base64 + magic-byte checks (incl. RIFF/WEBP) and extension match; invalid or mismatched content errors and no file created; new files are registered only on success.
  - Binary images cannot be appended (append is rejected to prevent corruption).
  - Registered new types in `_file_types`/`from_state`, pruned `UNSUPPORTED_BINARY_EXTENSIONS`, extended structured reads to include `gif`/`webp`, and allowed these extensions in filename validation.
  - Updated help text to document small-image support; other binaries (.mp4, .zip, etc.) remain blocked.

<sup>Written for commit d828d83d7aa2becaa231de400f1a4ff77cb490a5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5280?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/filesystem/file_system.py
+++ b/browser_use/filesystem/file_system.py
@@ -347,10 +347,46 @@ class Base64BinaryFile(BaseFile):
 	flows, not for arbitrary large binaries.
 	"""
 
+	# Leading magic bytes that identify a real file of this type. Keyed by extension so
+	# base64 that decodes but isn't actually an image (e.g. 'aGVsbG8=' -> b'hello') is
+	# rejected instead of written as a corrupt upload.
+	_MAGIC: dict[str, tuple[bytes, ...]] = {
+		'png': (b'\x89PNG\r\n\x1a\n',),
+		'gif': (b'GIF87a', b'GIF89a'),
+		'jpg': (b'\xff\xd8\xff',),
+		'jpeg': (b'\xff\xd8\xff',),
+		'webp': (b'RIFF',),  # RIFF container; 'WEBP' tag checked below
+	}
+
 	def _decoded(self) -> bytes:
 		# Strip all whitespace (the write_file action appends a trailing newline) then
 		# decode strictly so non-base64 text is rejected rather than silently corrupted.
 		return base64.b64decode(''.join(self.content.split()), validate=True)
+
+	def _validate(self, content: str) -> None:
+		"""Decode and confirm the bytes actually are an image of this extension. Raises FileSystemError."""
+		try:
+			data = base64.b64decode(''.join(content.split()), validate=True)
+		except Exception as e:
+			raise FileSystemError(
+				f"Error: content for '{self.full_name}' is not valid base64. "
+				f'For images, provide the base64 of a valid {self.extension} file. ({e})'
+			)
+		magic = self._MAGIC.get(self.extension, ())
+		if magic and not any(data.startswith(m) for m in magic):
+			raise FileSystemError(
+				f"Error: content for '{self.full_name}' is valid base64 but not a {self.extension} image "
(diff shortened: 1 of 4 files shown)


MagMueller:
Thanks — both P2s were real, fixed:

**P2 (non-image base64):** `Base64BinaryFile` now validates magic bytes per extension on write. `aGVsbG8=` (valid base64 → `b'hello'`) is rejected for `.png`, and correct PNG bytes under a `.gif` name are rejected too — so no corrupt file can reach an upload. WEBP additionally checks the `RIFF…WEBP` container tag.

**P2 (ghost file):** `write_file` now registers a *new* file in `self.files` only after the write succeeds, so a rejected binary write leaves no entry in `list_files()`/state (verified by new assertions). Also blocked append on binary files (base64 fragments would corrupt).

**P3 (test):** the round-trip test now asserts content integrity (decoded bytes equal the original and start with the PNG magic), not just existence.

Added tests for non-image base64, wrong-magic-for-extension, and the no-ghost path. 99 filesystem tests pass.

PR #5290, opened 2026-07-23
Fix nested cross-origin iframe extraction filtering

MagMueller:
## Summary

Fixes two independent filters that remove valid controls from nested cross-origin iframe trees before they reach the selector map.

- Scope the clickability cache by `(session_id, node_id)` because CDP `nodeId` values are session-local.
- Scope paint-order occlusion by `(session_id, containing_frame_id)` because DOMSnapshot bounds and paint order are document-local.

This PR intentionally does not change selector indices or screenshot highlighting. Cross-session `backendNodeId` collisions are handled separately so these extraction fixes remain small and reviewable.

## Reproduction

On the supplied three-origin payment repro with `cross_origin_iframes=True`:

- baseline 0.13.4 omitted the card-number input and Pay button;
- the cache fix restores the Pay button and card iframe;
- document-scoped paint filtering restores the card-number input while keeping paint-order filtering enabled.

## Tests

- `uv run pytest -q tests/ci/browser/test_dom_serializer_session_identity.py tests/ci/browser/test_dom_serializer.py` (5 passed)
- `uv run pre-commit run --all-files` (passed)

The final BU2 end-to-end flow also requires collision-free selector indices, which will be submitted as a separate PR.

### DIFF
--- a/browser_use/dom/serializer/paint_order.py
+++ b/browser_use/dom/serializer/paint_order.py
@@ -151,6 +151,17 @@ class PaintOrderRemover:
 	def __init__(self, root: SimplifiedNode):
 		self.root = root
 
+	@staticmethod
+	def _document_context(node: SimplifiedNode) -> tuple[str | None, str | None]:
+		"""Identify the CDP session and iframe document owning a node's snapshot."""
+		original_node = node.original_node
+		parent = original_node.parent_node
+		while parent is not None:
+			if parent.tag_name in {'iframe', 'frame'}:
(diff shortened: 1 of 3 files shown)


MagMueller:
The earlier Cubic review applies to superseded commit `496019e`, which included selector-index changes. This PR was force-updated to remove those changes and now contains only the two extraction-filter fixes. The quadratic collection and action-index findings no longer apply to this diff; the selector identity work, including those fixes, is isolated in #5291. Fresh CI and Cubic checks on the reduced PR are complete.

PR #5300, opened 2026-07-24
Call the model with minimal state after capture timeout

MagMueller:
## Summary

- return a minimal, non-actionable browser state when the complete state-capture event times out
- clear every cached selector lookup before the model is called
- expose a fixed model-visible error explaining that DOM, screenshot, and safe indices are unavailable
- retain only the best locally known URL/title so the model can choose a recovery such as navigation or waiting

## Behavior

A state timeout no longer skips the model call and no longer reuses cached actionable DOM. The model receives:

- an empty selector map
- no screenshot or page geometry
- no pending requests or pagination metadata
- a safe recovery message that recommends navigation, waiting, or another non-indexed action

## Edge cases considered

- works even when the very first state capture times out and no cached state exists
- old selector indices are removed from both BrowserSession and DOMWatchdog lookup paths
- repeated timeouts remain bounded by the agent's normal step/failure limits
- DOM-only screenshot degradation and AX-tree degradation keep their existing richer fresh-state behavior
- only `TimeoutError` uses this fallback; unrelated programming and validation errors still propagate
- the message is fixed rather than derived from exception text, avoiding internal identifiers in the model prompt

## Validation

- forced timeout after an actionable cached state returns no selectors and clears action lookup
- forced timeout with no prior state still returns a model-visible state
- prompt regression proves the recovery message reaches the model
- focused timeout/DOM/prompt tests pass
- `uv run pre-commit run --all-files`

This PR is intentionally left open for review and must not be merged yet.

---
## Summary by cubic
On state-capture timeout, the model now receives a minimal, non-actionable browser state with a safe error message. This prevents stale selectors, keeps the agent loop running, and the next successful capture rebuilds selectors.

- **Bug Fixes**
  - Return a minimal state: empty selector map, no screenshot; keep best-known URL/title and current tab.
  - Clear selector caches and `DOMWatchdog` state before the model call.
  - Add `state_error` to `BrowserStateSummary`; include in prompts via `<browser_state_error>` and store in `browser_errors`.
  - Applies only to `TimeoutError` and works on the first state; the agent still calls the model and proceeds normally.
  - After a successful capture, repopulate the selector map and omit the error.

<sup>Written for commit 05b68eb5679da0497510a6954863687bef41c706. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5300?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/agent/prompts.py
+++ b/browser_use/agent/prompts.py
@@ -298,6 +298,10 @@ class AgentMessagePrompt:
 
 		current_tab_text = f'Current tab: {current_target_id[-4:]}' if current_target_id is not None else ''
 
+		state_error_text = ''
+		if state_error := getattr(self.browser_state, 'state_error', None):
+			state_error_text = f'<browser_state_error>{state_error}</browser_state_error>\n'
+
 		# Check if current page is a PDF viewer and add appropriate message
 		pdf_message = ''
 		if self.browser_state.is_pdf_viewer:
@@ -325,7 +329,7 @@ class AgentMessagePrompt:
 Available tabs:
 {tabs_text}
 {page_info_text}
-{recent_events_text}{closed_popups_text}{pdf_message}Interactive elements{truncated_text}:
+{state_error_text}{recent_events_text}{closed_popups_text}{pdf_message}Interactive elements{truncated_text}:
 {elements_text}
 """
 		return browser_state
--- a/browser_use/browser/session.py
+++ b/browser_use/browser/session.py
@@ -54,7 +54,7 @@ from browser_use.browser.events import (
 )
 from browser_use.browser.profile import BrowserProfile, ProxySettings
 from browser_use.browser.views import BrowserStateSummary, TabInfo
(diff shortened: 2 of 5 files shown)


gnanirahulnutakki (approved):
Reviewed at `05b68eb5679da0497510a6954863687bef41c706`, including the timeout/cancellation boundary and both selector-cache paths. I verified that the pinned Bubus timeout cancels the slow state handler before the fallback returns, so the cleared selector maps are not repopulated afterward. Locally, all five prompt tests and four focused real-browser timeout/recovery tests pass. The fallback remains non-actionable, preserves a safe recovery path, and the next successful capture rebuilds selectors. I did not find a correctness issue.

PR #5303, opened 2026-07-25
fix(utils): resolve redact cascade bug, Windows path tests, and audit remediation

### LINKED ISSUE #5135: Bug: redact_sensitive_string cascades when a secret value is substring of a placeholder tag

### LINKED ISSUE #5236: Bug: skill-install test uses real executables on Windows
# Bug: skill-install test uses real executables on Windows

### Version

`browser-use 0.13.4` at commit `dbc4d46e0ca59afbba020d456aa10d930279330c`

### Description

The test helper in [`test_browser_use_skill_install_docs.py`](https://github.com/browser-use/browser-use/blob/dbc4d46e0ca59afbba020d456aa10d930279330c/tests/ci/test_browser_use_skill_install_docs.py#L19-L45) creates extensionless fake `uv` and `browser-harness` executables.

Windows does not resolve these through `shutil.which()`, so the test can invoke the real installed executables instead. This makes the test non-hermetic and causes a timeout or missing `uv-args.txt`.

This appears limited to the test fixture; no production CLI bug was observed.

### Reproduction

```powershell
uv sync --all-extras --dev
uv run pytest tests/ci/test_browser_use_skill_install_docs.py::test_browser_use_cli_installs_browser_harness_package_skill -q
```

### Actual Behavior

On the first run, the subprocess exceeded the test's 10-second timeout while the real `uv` executable performed work:

```text
E   subprocess.TimeoutExpired:
E   Command [
E     '...\browser-use\.venv\Scripts\python.exe',
E     '-m',
E     'browser_use.cli',
E     'skill',
E     'install'
E   ] timed out after 10 seconds
```

On a subsequent run, the command returned, but the fake `uv` executable had not created `uv-args.txt`:

```text
tests/ci/test_browser_use_skill_install_docs.py:84:
    assert uv_args.read_text(encoding='utf-8') == \
        'tool install --python 3.12 --upgrade --force browser-use'

E   FileNotFoundError: [Errno 2] No such file or directory:
E   'C:\Users\jain4\AppData\Local\Temp\pytest-of-jain4\pytest-1\
E   test_browser_use_cli_installs_0\uv-args.txt'

1 failed in 9.27s
```

### Expected Behavior

The test should invoke only its temporary fake tools.

It should:

- never run the user's real `uv` or `browser-harness` installation;
- create `uv-args.txt` through the fake `uv` command;
- verify the expected arguments;
- behave consistently on Windows, Linux, and macOS.

### Environment

Windows 11, Python 3.12.10, uv 0.8.17, pytest 9.0.2.

### LINKED ISSUE #5248: redact_sensitive_string: placeholder tags get re-redacted when a secret value contains the substring 'secret' (cascade leak)

Ranga-Prasath-22:
## Summary

Fixes a cascade re-redaction bug in `redact_sensitive_string`, makes Windows path execution in skill-install tests reliable, and applies the audit remediation from the review cycle on this PR.

Closes #5248, closes #5135, closes #5236.

### Bug fixes

- **Fixed cascade re-redaction in `redact_sensitive_string`**
  - Secrets containing the substring `"secret"` could corrupt previously generated `<secret>` tags.
  - Redaction now splits input on existing tags and applies a literal-only regex only to the unredacted parts, so already-redacted values are never re-processed.

- **Fixed Windows path execution in skill-install tests**
  - Mock `uv` and `browser-harness` tools are now wrapped in `.bat` launchers on Windows so subprocess resolution works correctly.

### Robustness hardening (review-driven)

- **Balanced-pair tag parsing** — pre-existing `<secret>...</secret>` tags are located with a `str.find`-based balanced parser instead of a non-greedy regex, so nested pre-existing tags are not truncated at the first closing marker.
- **Full-tag secrets redacted** — a configured secret that is itself a balanced `<secret>foo</secret>` string is redacted rather than mistaken for a pre-existing placeholder (would have leaked unchanged).
- **Idempotent output** — a placeholder whose inner text is already a generated key list is preserved, so re-redacting output never produces nested tags even when a secret value coincides with a key name.
- **Empty markers preserved** — an empty `<secret></secret>` in the input survives redaction instead of being silently removed.
- **Removed `_last_sensitive_tuple`** and the manual `cache_clear()` thrashing that defeated the LRU cache; the cache now works as intended and holds no long-lived reference to plaintext credentials.
- **Pre-sorted mappings** — secret/key lookup lists are sorted once at compile time, not per call.
- **`sanitize_url_candidate` is balance-aware** — URLs like `https://en.wikipedia.org/wiki/Python_(programming_language)` keep their trailing `)`, while unbalanced closers and prose punctuation are still stripped.
- **`get_browser_use_version` uses `tomllib`** (binary mode) instead of a first-match regex, so a dependency table's `version` key can no longer shadow the project version.
- **`singleton` decorator** uses double-checked locking for thread-safe initialization.
- **`get_git_info`** split into a cached inner function plus a public wrapper returning a copy, so callers cannot mutate the cached dict.

## Performance

Measured on this machine (100k iterations for short inputs, 300 for ~116KB inputs), comparing the previous single-pass implementation (`31ce95c5d`) with this branch:

| Case | `31ce95c5d` | This PR | Speedup |
|---|---|---|---|
| Short string, secrets present | 6.46 µs | 3.84 µs | 1.68x |
| Short string, no secrets | 3.86 µs | 2.28 µs | 1.69x |
| ~116KB string, secrets present | 5151 µs | 1362 µs | 3.78x |
| ~116KB string, no secrets | 14444 µs | 1253 µs | 11.5x |

Gains come from removing the per-match `sorted()` and tag-alternation from the hot path and from a fast path that skips tag splitting entirely when no `<secret>` marker is present. Upstream's naive loop is faster for very small inputs, but it produces the corrupted output this PR fixes.

## Verification

```bash
uv run pytest tests/ci/test_redact_sensitive_string.py \
             tests/ci/test_sanitize_url_candidate.py \
             tests/ci/infrastructure/test_version_check.py \
             tests/ci/test_browser_use_skill_install_docs.py
```

All 31 tests passing

**Result:** 31 passed in 2.54s.

```bash
ruff check
ruff format --check
```

Both lint and format checks pass.


Ranga-Prasath-22:
All review items from cubic and codex (02:08, 02:28, 05:12, 05:21) are now addressed. The 05:21 P1 full-tag secret leak and P2 empty-marker preservation are fixed in c7ce9299f with regression tests for both. 31 tests green, ruff clean, and the benchmark numbers in the PR body are honest (vs. the previous commit, not upstream). No open items on my end -- happy to iterate on any feedback.

Ranga-Prasath-22:
Closing this as superseded — it has been split into three smaller, focused PRs to make review easier (per maintainer guidance that small PRs merge faster):

- **#5335** — `fix(utils): resolve redact cascade bug` (redact cascade fix + `test_redact_sensitive_string.py` + test screenshot)
- **#5336** — `fix(utils): harden URL and version parsing` (`sanitize_url_candidate` balance-aware stripping, `tomllib`-based version parsing, `get_git_info` cache-safety, `singleton` double-checked locking + their tests)
- **#5337** — `test(ci): Windows path execution tests for skill install docs` (Windows `.bat` shims + `USERPROFILE` forwarding)

All changes from this PR are preserved across the three; no content was dropped. Test results were verified locally (`pytest` on the touched test files: 13/15/3 passed respectively).

Ranga-Prasath-22:
Superseded by #5335, #5336, #5337.

PR #5307, opened 2026-07-25
Agent modular refactor

MingXu-123:
## Summary by cubic
Refactors the agent into a modular step pipeline with dedicated LLM and action services, adds a deterministic auto-recovery layer, and introduces a pluggable toolset API with hierarchical planning. Keeps the public Agent API and common imports working.

- **New Features**
  - Hierarchical planning with a tree `PlanState`; status markers updated to [✓], [→], [x]; model now emits `PlanUpdateStep`.
  - Pluggable tools via `BaseToolset`: `BrowserToolset` (built‑in) + `CustomToolset` (user). `Tools` composes both and is re‑exported from `browser_use.tools.service`; Agent accepts any `BaseToolset`. Coordinate clicking toggle is on `tools.browser.set_coordinate_clicking(True)`.
  - Deterministic auto‑recovery via `RecoveryEngine`/`RecoveryStage` with default strategies (scroll, relocate, overlay dismiss, wait/timeout, retry, navigation back, browser reconnect) and telemetry (`RecoveryStats`). Wired into action execution; pass an empty strategy list to opt out.

- **Bug Fixes**
  - DeepSeek v4 reasoning models now use JSON output mode with thinking disabled; repairs invalid control characters in JSON; non‑reasoning models keep function‑calling. Adds unit tests for both paths.
  - Timing/logging tolerate a `None` `browser_session`. Restores backward‑compat wrappers and logging (incl. `_update_action_indices`, `_log_action`) and replan nudge uses the `consecutive_failures` threshold.
  - CI selector‑collision test stub updated to include `agent.settings.auto_recovery_max_attempts` expected by `ActionExecutor`.

<sup>Written for commit 43a2e282977f3ff579759d835a9b58cf48dfd104. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5307?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Nakagawa-master (commented):
There is a recovery-safety gap here for effectful actions.

`RetryRecovery.can_handle()` currently returns `True` for anything not claimed by a more specific strategy, and `TimeoutRecovery` returns `RETRY` after a timeout. Because these strategies have no action-safety / idempotency filter, a failed or timed-out browser action can be re-executed even when the external effect may already have happened.

That creates this sequence:

```text
click / custom action submits an external effect
→ remote system accepts it
→ browser/tool result is lost or times out
→ recovery classifies the attempt as retryable
→ original action runs again
→ duplicate side effect
```

A timeout or local `ActionResult.error` is not, by itself, proof that the external action did not occur.

I think recovery eligibility needs to distinguish at least:

```text
read-only / observational
safe-idempotent
effectful-but-provider-idempotent
effectful-outcome-unknown
unknown-custom-action
```

with `effectful-outcome-unknown` and unknown custom actions defaulting to **no blind retry**. For browser clicks this probably means the recovery layer cannot infer safety just from `ActionKind.CLICK`; it needs either action metadata/policy or a conservative boundary for actions that can cross a consequential submit/navigation boundary.

A regression shape:

```text
effectful action is accepted remotely
→ local result times out / is lost
→ recovery runs
→ zero second external effect until reconciliation or explicit idempotency guarantee
```

This is separate from the existing Cubic findings about stale DOM state and helper-action errors: even when the recovery mechanism is internally correct, retrying the original action can still be semantically unsafe because the first outcome is ambiguous.

Public source relation: Nakagawa Master's External Side-Effect Reuse Kit packages this exact ambiguous-outcome boundary:
https://github.com/Nakagawa-master/nakagawa-theory-archive/blob/main/EXTERNAL_SIDE_EFFECT_REUSE_KIT.md

The underlying idempotency / ambiguous-outcome principle is established distributed-systems practice; the source relation here is to this concrete reusable regression framing, not a claim of global novelty.

Issue #3544, opened 2025-11-06
Bug: action field consistently miss in the response of LLM


lyssg:
### Browser Use Version

0.9.5

### Bug Description, Steps to Reproduce, Screenshots

1. run failing python code
2. action field miss in the response of LLM
3. agent add <step> in <agent_history>, step content: 
```
1 validation error for AgentOutput
action  
  Field required [type=missing, input_value={'thinking': '......ts.'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.11/v/missing
```
4. action field still miss in the response of LLM

### Failing Python Code

```python
import asyncio
import json

from dotenv import load_dotenv
from browser_use import Agent
from browser_use.llm.deepseek.chat import ChatDeepSeek

load_dotenv()

llm = ChatDeepSeek(
    model="deepseek-chat",
    api_key="sk-xxx",  # export DEEPSEEK_API_KEY=xxx
    temperature=0.3
)

async def main():
    agent = Agent(
        task="导航到file:///Users/lyssg/ai/calculator.html，并总结界面信息",
        llm=llm,
    )
    result = await agent.run()
    print('result:',result)
```

### LLM Model

deepseek-chat

### Full DEBUG Log Output

```shell
INFO     [Agent] 📍 Step 1:
[...]

PR #5330, opened 2026-07-30
Recover Anthropic tool arguments when the model serializes its call as text

ElayGelbart:
## Problem

With `claude-sonnet-5`, the model sometimes emits a valid `tool_use` block whose arguments contain the *entire tool call rendered as text* inside the first string property, instead of the actual JSON arguments. Real payload tails from our logs:

```
input_value={'thinking': 'The current...</action>\n</invoke>\n'}
input_value={'thinking': 'The resume ...n>\n</AgentOutput>\n\n'}
```

`AgentOutput` validation then fails with `action / Field required` and the step is lost — even though the text in `thinking` contains every field the schema needs.

The salvage added in 0.13.x can't help, because `_completion_from_text_response()` is only reachable when **no** `tool_use` block exists. Here one does exist, so the `ValidationError` escapes the `for` loop before execution ever reaches the salvage call:

```python
for content_block in response.content:
    if content_block.type == 'tool_use':
        try:
            return ChatInvokeCompletion(completion=output_format.model_validate(content_block.input), ...)
        except Exception as e:
            ...double-serialization repair, then re-validate or raise...

if self._requires_auto_tool_choice():          # unreachable in this failure mode
    text_completion = self._completion_from_text_response(response, output_format, usage)
```

## Fix

Give the `tool_use` validation failure the same last-resort recovery the no-tool_use path already has: before re-raising, parse the arguments back out of the string field, handling both `<parameter name=...>` markup (including the mismatched `</action>`-style closing tags the model emits) and an embedded JSON object. Output that carries no recoverable arguments still raises, so genuinely malformed responses are not masked.

The existing double-serialization repair is extracted into `_repair_serialized_fields()` and reused, since the recovered fields need the same treatment.

## Tests

`tests/ci/test_anthropic_serialized_tool_input.py`, following the `test_llm_output_truncation.py` pattern with `httpserver`:

- markup-serialized call in a string field is recovered
- JSON-serialized call in a string field is recovered
- pre-existing double-serialized-field repair still works
- valid tool input is untouched
- unrecoverable input still raises

The two recovery tests fail on `main` and pass with this change; the other three pass both ways. `tests/ci/models/test_llm_anthropic.py`, `test_llm_model_factory.py`, `test_llm_output_truncation.py`, `test_llm_retries.py` and `test_fallback_llm.py` all still pass (33 passed, 1 skipped). `ruff check`, `ruff format --check` and `pyright` are clean on both files.

## Notes, deliberately left out of this PR

Two related things I ran into while diagnosing this, happy to open separate PRs or issues if you'd like them addressed:

1. **`_requires_auto_tool_choice()` doesn't match `claude-sonnet-5`.** It matches `claude-fable-5` / `claude-mythos-5`, or requires `thinking` to be set explicitly. Sonnet 5 runs adaptive thinking *by default*, so out of the box it gets a forced `tool_choice`. Probing the API with the real `AgentOutput` schema shows what that costs:

   ```
   adaptive thinking + forced tool_choice : blocks=['tool_use']
   adaptive thinking + auto   tool_choice : blocks=['thinking', 'tool_use']
   no thinking config + auto  tool_choice : blocks=['thinking', 'tool_use']
   ```

   Forcing the tool suppresses the thinking block entirely, so the model reasons inside the JSON `thinking` string — which is exactly where the serialized call ends up. Passing `thinking={'type': 'adaptive', 'display': 'summarized'}` flips `_requires_auto_tool_choice()` to `True` and took us from consecutive failing steps to roughly 1 step in 10.

2. **Local validation failures are reported as retryable server errors.** A pydantic `ValidationError` reaches `except Exception as e: raise ModelProviderError(message=str(e), model=self.name)`, which defaults to `status_code=502`. `Agent._try_switch_to_fallback_llm` treats 502 as retryable, so a client-side schema failure logs `⚠️ LLM error (...) but no fallback_llm configured` and points users at fallback config rather than at output-format non-compliance.

Related: #3544 reports the same `action Field required` / `input_value={'thinking': ...}` symptom on DeepSeek, which suggests this failure shape isn't Anthropic-specific.

---
## Summary by cubic
Recovers lost steps when Anthropic returns a `tool_use` block with its arguments serialized as text by extracting arguments from the `thinking` field after validation fails and re-validating. Prevents `action Field required` errors by parsing `<parameter>` markup or embedded JSON from the `thinking` string.

- **Bug Fixes**
  - On `tool_use` validation failure, recover only from `thinking` by parsing `<parameter name="...">...</parameter>` (tolerates mismatched closing tags like `</action>`) or JSON snippets, then validate. Valid tool input takes priority over conflicting serialized data in `thinking`; unrecoverable inputs still raise.

- **Refactors**
  - Extracted helpers into `_repair_serialized_fields()`, `_tool_call_from_text()`, and `_completion_from_serialized_tool_input()` and reused during recovery.

<sup>Written for commit ef01b9b9c715c5718b218d205dd7f3a3f99f0ca6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5330?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/llm/anthropic/chat.py
+++ b/browser_use/llm/anthropic/chat.py
@@ -326,7 +326,10 @@ class ChatAnthropic(BaseChatModel):
 				candidates.append(tool_call)
 			for text_candidate in self._json_candidates_from_text(value):
 				try:
-					candidates.append(json.loads(text_candidate))
+					candidate = json.loads(text_candidate)
+					if isinstance(candidate, dict):
+						candidate = self._repair_serialized_fields(candidate)
+					candidates.append(candidate)
 				except (json.JSONDecodeError, TypeError):
 					continue
 
--- a/browser_use/llm/anthropic/chat.py
+++ b/browser_use/llm/anthropic/chat.py
@@ -316,34 +316,36 @@ class ChatAnthropic(BaseChatModel):
 		if not isinstance(tool_input, dict):
 			return None
 
-		for value in tool_input.values():
-			if not isinstance(value, str):
+		# The malformed Anthropic response puts the complete call in the schema's
+		# `thinking` argument. Do not promote serialized data from arbitrary fields.
+		thinking = tool_input.get('thinking')
+		if not isinstance(thinking, str):
+			return None
+
+		candidates: list[Any] = []
+		tool_call = self._tool_call_from_text(thinking)
+		if tool_call is not None:
+			candidates.append(tool_call)
+		for text_candidate in self._json_candidates_from_text(thinking):
+			try:
+				candidate = json.loads(text_candidate)
+				if isinstance(candidate, dict):
+					candidate = self._repair_serialized_fields(candidate)
+				candidates.append(candidate)
+			except (json.JSONDecodeError, TypeError):
 				continue
 
-			candidates: list[Any] = []
-			tool_call = self._tool_call_from_text(value)
-			if tool_call is not None:
-				candidates.append(tool_call)
-			for text_candidate in self._json_candidates_from_text(value):
-				try:
-					candidate = json.loads(text_candidate)
-					if isinstance(candidate, dict):
-						candidate = self._repair_serialized_fields(candidate)
-					candidates.append(candidate)
-				except (json.JSONDecodeError, TypeError):
-					continue
-
-			for candidate in candidates:
-				try:
-					completion = output_format.model_validate(candidate)
-				except Exception:
-					continue
-				return ChatInvokeCompletion(
-					completion=completion,
-					usage=usage,
-					stop_reason=response.stop_reason,
-					stop_details=self._get_stop_details(response),
-				)
+		for candidate in candidates:
+			try:
+				completion = output_format.model_validate(candidate)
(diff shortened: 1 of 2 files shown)


aviorp-anchor:
I've encountered this same issue in our environment. Merging this fix is a high priority as a stopgap, we've applied a manual patch internally, but that's not a sustainable long-term solution. Could this be prioritized and pushed forward for merge?

thanks @ElayGelbart  for quick solution !

ElayGelbart:
@MagMueller what do you think ?

MagMueller:
Looking into it.

MagMueller:
@ElayGelbart does main work for you now?

PR #5335, opened 2026-07-31
fix(utils): resolve redact cascade bug

Ranga-Prasath-22:
## Summary

`redact_sensitive_string` now matches raw secret values in a **single pass via a cached literal-only regex** (no `<secret>` tag alternation), which prevents the cascade re-redaction of already-generated placeholder tags. Pre-existing `<secret>...</secret>` tags are split out beforehand with **balanced-pair parsing**, so their inner contents are redacted separately without re-scanning tag boundaries.

Also hardens edge cases:

- **Full-tag secrets** — a configured secret that is itself a balanced `<secret>...</secret>` string is redacted instead of being leaked as a "pre-existing" tag
- **Idempotence** — re-redacting output whose tags already contain generated key lists never produces nested tags, even when a secret value coincides with a key name
- **Empty markers** — `<secret></secret>` is preserved instead of deleted
- **Fast path** — inputs without any `<secret>` marker skip tag splitting entirely

## Tests

New `tests/ci/test_redact_sensitive_string.py` (13 tests) covering: cascade re-redaction, duplicate keys, substrings, falsy values, secret-inside-tag, multiline tags, secrets containing tag characters, secrets spanning tags, idempotence when a secret equals a key list, nested pre-existing tags, full-tag secrets, empty tags, and unsorted input.

Verified locally: `pytest tests/ci/test_redact_sensitive_string.py` → **13 passed**.

## Notes

This PR was **split out of #5303** (which has been closed) to keep changes small and reviewable. The screenshot in `.github/pr-assets/` shows the test suite passing.

---
## Summary by cubic
Fixes re-redaction cascades in `redact_sensitive_string` using a single-pass literal-only regex and balanced `<secret>` tag parsing. Hardens idempotence and secret retention with thread-safe regex-cache eviction (`re._cache`/`re._cache2`), cyclic-collision handling, and per-call map rebuilds.

- **Bug Fixes**
  - Single-pass literal-only regex for raw secrets; rebuilt per call; evict only matching entries from `re._cache` and `re._cache2` with a thread-safe key snapshot; fallback `re.purge()` on unsupported runtimes; fast path when no tags.
  - Balanced `<secret>...</secret>` parsing with nesting and raw-span overlap checks; inner content redacted once; handle full-tag and empty-marker secrets; preserve empty `<secret></secret>`.
  - Idempotence and collisions: freeze cyclic key-list collisions; keep generated tags stable; deterministic key grouping; longest-first matching.

- **Refactors**
  - Trimmed docstrings and rationale comments; removed the test-pass screenshot asset.

<sup>Written for commit f59d2615a51c88cfe924e25e9611fe6d60c2e045. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5335?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Ranga-Prasath-22:
## Follow-up: address review finding on process-global secret retention

**Root cause.** `_get_redact_pattern_and_mapping` was decorated with `@lru_cache(maxsize=128)`, so the compiled regex pattern *and* the secret→key lookup maps — both containing raw secret material — were retained in a process-global cache, surviving past completed agents/requests (the concern raised in review). Separately, inner `<secret>…</secret>` tag content was redacted with iterative `str.replace()`, which could re-redact a placeholder whose text coincided with another secret's value (cascade inside tags).

**Fix.**
- Remove the `@lru_cache`; the pattern and maps are rebuilt on every redaction call, so no raw secret material persists beyond the redaction request that produced it.
- Redact inner tag content with a single compiled-regex pass (`pattern.sub`), so generated placeholder text is never re-scanned.
- Empty `<secret></secret>` markers remain preserved.

**Verification.**
- ruff format/check: clean. pyright 1.1.408 on `browser_use/utils.py`: 0 errors.
- `tests/ci/test_redact_sensitive_string.py`: 19 passed — added a guard that no `@lru_cache` machinery is attached and that distinct secret sets get independent maps, plus an inner-tag cascade case.
- Adversarial suite (40+ cases: nested tags, multiline, tag-char secrets, cross-key collisions, empty markers, idempotence, unicode, raw-secret leak checks) and the tag-boundary paranoia checks: all pass.

Note: the earlier ~13 red jobs on this branch (browser/CDP-based) failed with `'Page.enable' wasn't found` in the run log; this push re-runs CI. Those jobs don't configure `sensitive_data`, so the redact path is never invoked there — the redact changes cannot affect them.

sauravpanda:

Ranga-Prasath-22:
ok ok np!
:)

PR #5336, opened 2026-07-31
fix(utils): harden URL and version parsing

Ranga-Prasath-22:
## Summary

Hardening for URL candidate sanitization and version/git info parsing:

- **`sanitize_url_candidate`** — balance-aware trailing stripping: preserves balanced `()`/`[]` (e.g. Wikipedia-style URLs ending in `..._(programming_language)`) while still stripping unbalanced trailing closers/openers and trailing `.`, `,`, `;`, `:`, `!`, `?`
- **`get_browser_use_version`** — resolves `[project].version` via `tomllib` instead of a naive first-match regex over the file contents; reads the TOML in binary mode as `tomllib` requires
- **`get_git_info`** — split into a cached private helper (`_get_git_info_cached`) plus a public wrapper returning a **copy**, so callers can no longer mutate the cached dict
- **`singleton`** — double-checked locking with `threading.Lock()` to avoid duplicate instance creation under concurrent first access

## Tests

- New `tests/ci/test_sanitize_url_candidate.py` (11 tests): balanced/unbalanced/nested parens & brackets, trailing punctuation, escaped-newline truncation, empty input
- New coverage in `tests/ci/infrastructure/test_version_check.py`: asserts `[project].version` resolution and `LIBRARY_VERSION` propagation

Verified locally: `pytest tests/ci/test_sanitize_url_candidate.py tests/ci/infrastructure/test_version_check.py` → **15 passed**.

## Notes

This PR was **split out of #5303** (which has been closed) to keep changes small and reviewable.

---
## Summary by cubic
Hardens URL sanitization and version/git parsing for correctness and thread safety. Preserves balanced parentheses in URLs, parses version via `tomllib` with exact UTF-8 BOM handling and fallback, prevents cached git info mutation, and makes the singleton thread-safe.

- **Bug Fixes**
  - `sanitize_url_candidate`: preserve balanced ()/[]; strip unbalanced closers and trailing . , ; : ! ?; keep escaped-newline truncation.
  - `get_browser_use_version`: read `[project].version` via `tomllib`; strip only a full UTF-8 BOM via `_strip_utf8_bom`; fall back to installed metadata when missing; set `LIBRARY_VERSION`.

- **Refactors**
  - `get_git_info`: split into cached helper and public wrapper that returns a copy to avoid mutation.
  - `singleton`: add double-checked locking to prevent duplicate instances under concurrent first access.
  - Docs: trim `_strip_utf8_bom` docstring and related test comments (no behavior change).

<sup>Written for commit b69eb5a0e375b3a793f567f740b3b7db6462468f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5336?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/utils.py
+++ b/browser_use/utils.py
@@ -658,12 +658,7 @@ def merge_dicts(a: dict, b: dict, path: tuple[str, ...] = ()):
 
 
 def _strip_utf8_bom(raw: bytes) -> bytes:
-	"""Strip a single leading UTF-8 BOM (EF BB BF) from *raw* if present.
-
-	Unlike ``bytes.lstrip``, which removes any combination of the bytes EF/BB/BF,
-	this removes only the exact 3-byte prefix, so a lone 0xEF, 0xBB, or 0xBF (or
-	a partial BOM) is never stripped from the start of the file.
-	"""
+	"""Strip a single leading UTF-8 BOM (EF BB BF) if present, not lone bytes."""
 	return raw.removeprefix(b'\xef\xbb\xbf')
 
 
--- a/browser_use/utils.py
+++ b/browser_use/utils.py
@@ -657,6 +657,16 @@ def merge_dicts(a: dict, b: dict, path: tuple[str, ...] = ()):
 	return a
 
 
+def _strip_utf8_bom(raw: bytes) -> bytes:
+	"""Strip a single leading UTF-8 BOM (EF BB BF) from *raw* if present.
+
+	Unlike ``bytes.lstrip``, which removes any combination of the bytes EF/BB/BF,
+	this removes only the exact 3-byte prefix, so a lone 0xEF, 0xBB, or 0xBF (or
+	a partial BOM) is never stripped from the start of the file.
+	"""
+	return raw.removeprefix(b'\xef\xbb\xbf')
+
+
 @cache
 def get_browser_use_version() -> str:
 	"""Get the browser-use package version using the same logic as Agent._set_browser_use_version_and_source"""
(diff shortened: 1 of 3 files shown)


Ranga-Prasath-22:
## Follow-up: strip only a full UTF-8 BOM

**Root cause.** `content = f.read().lstrip(b'\xef\xbb\xbf')` strips *any combination* of the bytes 0xEF/0xBB/0xBF, not just the exact 3-byte BOM, so a lone 0xEF, 0xBB, or 0xBF byte (or a partial BOM) at the start of `pyproject.toml` would be silently stripped, corrupting the start of the file.

**Fix.** Extract a small `_strip_utf8_bom` helper that uses `bytes.removeprefix(b'\xef\xbb\xbf')`, so only the exact BOM prefix is removed, and call it from `get_browser_use_version`.

**Verification.**
- ruff format/check: clean. pyright 1.1.408 on `browser_use/utils.py`: 0 errors.
- `tests/ci/infrastructure/test_version_check.py`: 7 passed, including a new test covering a full BOM, partial BOMs, lone 0xEF/0xBB/0xBF, a single prefix (removeprefix, not lstrip-style sweeping), and no-BOM/empty cases.

PR #5348, opened 2026-08-03
fix(llm): preserve zero max_retries for Anthropic Bedrock

cosin2077:


cosin2077:
## Additional context and local verification

**Root cause:** `ChatAnthropicBedrock._get_client_params()` only forwarded truthy `max_retries` values. Configuring `max_retries=0` therefore omitted the argument and restored the Anthropic SDK default of 2 retries.

**Behavior after this patch:** the configured integer is always forwarded, so zero disables SDK retries while the existing wrapper default remains unchanged.

**Regression evidence:** before the source change, `test_get_client_preserves_zero_max_retries` failed with `assert 2 == 0`; it passes after the change.

**Local verification:**

- `uv run pytest -q tests/ci/models/test_chat_anthropic_bedrock_client_config.py tests/ci/models/test_chat_anthropic_bedrock_empty_content.py` — 2 passed
- `uv run pre-commit run --all-files` — passed

**Duplicate check:** no matching open issue or PR was found immediately before push. #5021 touches this adapter as part of a broad conflicting refactor, but does not implement this parameter-preservation fix.

PR #5349, opened 2026-08-03
fix(utils): avoid callback errors for cancelled tasks

cosin2077:


cosin2077:
## Additional context and local verification

**Root cause:** the cancellation callback called `Task.exception()` only when `Task.cancelled()` was true. Python raises `CancelledError` in that state, producing an event-loop callback failure during Ctrl-C handling; the inverse branch also failed to retrieve exceptions from non-cancelled completed tasks.

**Behavior after this patch:** cancelled tasks skip exception retrieval, while non-cancelled completed tasks still have their exception consumed.

**Regression evidence:** before the source change, the public `SignalHandler.sigint_handler()` test captured an event-loop callback context containing `CancelledError`; it passes after the change.

**Local verification:**

- `uv run pytest -q tests/ci/test_async_task_error_handling.py` — 3 passed
- `uv run pre-commit run --all-files` — passed

**Duplicate check:** no matching open issue or PR was found immediately before push. #5287 refactors signal coordination for parallel agents and touches the same utility, but does not fix this cancellation callback condition.

cosin2077:
The only failing check is `test_tools`, where `test_search_action` timed out after 30 seconds while navigating to DuckDuckGo. This PR only changes the `SignalHandler` task-done callback and its focused regression test; the same `test_tools` job passed on #5348 against the same base. I also reran the failing test locally on this head and it passed in 14.36 seconds. I cannot rerun the GitHub Actions job without repository admin rights. Could a maintainer please rerun the failed job?

PR #5350, opened 2026-08-03
fix(llm): consolidate Anthropic cache_control into TTL-aware helper

### LINKED ISSUE #5321: Usage model reports 1-hour cache tokens the serializer can never produce
### What I found

`AnthropicChatModel` reads both cache-creation buckets off the response and surfaces the 1-hour one:

```python
# browser_use/llm/anthropic/chat.py:195-196
getattr(cache_creation, 'ephemeral_5m_input_tokens', None),
getattr(cache_creation, 'ephemeral_1h_input_tokens', None),

# chat.py:216
prompt_cache_creation_1h_tokens=cache_creation_1h_tokens,
```

But the serializer only ever emits the 5-minute form:

```python
# browser_use/llm/anthropic/serializer.py:54-57
def _serialize_cache_control(use_cache: bool) -> CacheControlEphemeralParam | None:
    ...
    return CacheControlEphemeralParam(type='ephemeral')
```

`CacheControlEphemeralParam(type='ephemeral')` with no `ttl` is the 5-minute cache. An hour requires `{"type": "ephemeral", "ttl": "1h"}`, and no `ttl` appears anywhere in the Anthropic provider.

So `prompt_cache_creation_1h_tokens` is always zero, structurally. It isn't wrong exactly. It's a metric with no code path behind it, which is the sort of thing that misleads someone later who reads the usage model and reasonably assumes both TTLs are in play.

### What I measured

Controlled experiment on `claude-sonnet-4-6` with a 16,442-token static prefix, two arms sharing one schedule, `max_tokens=0` so only prefill gets billed.

| probe | 5m arm | 1h arm |
|---|---|---|
| t=0 | WRITE | WRITE |
| t=420s | MISS (full rewrite) | HIT |
| t=480s | HIT (stability control) | HIT |
| t=900s | MISS (full rewrite) | HIT |

$0.190 for the 5m arm against $0.114 for the 1h arm on the cached prefix, so 40.2% cheaper. All eight probes matched expectations written down before the run, worst clock drift 2.4 seconds.

### Where this stops

1h isn't universally better, and browser-use might be the least likely of the agents I looked at to benefit. From the measured 300s expiry and the write multipliers:

| Inter-request gap | Cheaper option |
|---|---|
| under 5 min | 5m, since both stay warm and a 1.25x write beats 2x |
| 5 min to 1 hr | 1h, since the 5m arm rewrites every request |
| over 1 hr | 5m, since both go cold every time |

Browser steps tend to be fast, so consecutive requests may well sit inside the five-minute window. If that's the case the current default is already right and the only thing worth doing here is dropping the unreachable metric.

One correction while I'm here, since I got this wrong myself at first and it seems like a common assumption. Images are cacheable. Anthropic's docs list "Images & Documents: Content blocks in the `messages.content` array, in user turns" among cacheable blocks. The real issue for a browser agent isn't that screenshots can't be cached, it's that they change every step, so a screenshot sitting before a breakpoint invalidates the stable content behind it. That's about placement, not cacheability.

### Suggested change

Correction to an earlier version of this issue: threading a `ttl` through `_serialize_cache_control` alone is not sufficient. There are three places that construct `CacheControlEphemeralParam`, and two of them bypass that helper:

- `serializer.py:57` — inside `_serialize_cache_control`, the one I originally pointed at
- `serializer.py:118` — direct, on the string-content path
- `chat.py:362` — direct

So the options are either to consolidate all three through a single TTL-aware helper, or to patch each site individually. Consolidating seems better: it means the next TTL-related change only has one place to touch, and it removes the chance of a partial fix that looks complete.

The alternative remains dropping the 1h fields from the usage model entirely, since nothing can populate them today.

Both are defensible. Wiring it up is better if any deployment has gaps over five minutes, dropping the fields if you're confident none do. Either beats where it stands now, where the metric implies a capability that isn't wired up.

Happy to open a PR for whichever you prefer.

---

Harness and pinned sources available if you want to reproduce this. The measurement uses a synthetic text prefix on a controlled schedule, so it establishes the mechanism rather than what it's worth on browser-use's actual traffic.

Filed by Tanisha Katara, CEO, KCG Consulting LLC.

Analysed at commit `f0aa3a8bb03779c71a5aa262d389e3bfe6b77cdc`

Edited shortly after filing to correct the remediation guidance: the original version named only `_serialize_cache_control` and missed two direct constructor sites.

Diwak4r:
Closes #5321

## What

Consolidates every `CacheControlEphemeralParam` construction site in the Anthropic provider behind the serializer's `_serialize_cache_control` helper, and makes that helper TTL-aware:

- `browser_use/llm/anthropic/serializer.py` — `_serialize_cache_control(use_cache, ttl=None)` now accepts `ttl: Literal['5m', '1h']`. With no `ttl` it emits exactly what it always did (`{"type": "ephemeral"}` → Anthropic's 5-minute cache), so the default path is byte-for-byte unchanged.
- `serializer.py` string-content path and both providers' structured-output `ToolParam` paths (`anthropic/chat.py`, `aws/chat_anthropic.py`) previously constructed `CacheControlEphemeralParam(type='ephemeral')` directly, bypassing the helper. All three now route through it — the "three places that construct it" from the issue are now one.

## Why this shape

The issue offered two options: consolidate through a single TTL-aware helper, or drop the 1h fields from the usage model. I went with consolidation because:

- The 1h usage path (`prompt_cache_creation_1h_tokens` and its cost calculation in `tokens/service.py`) is already fully plumbed and correct — it just never had a producer. Removing it would be a public API break on `ChatInvokeUsage` for a field that is now reachable.
- With the helper in place, opting into the 1-hour cache is a one-line `ttl='1h'` at any cache site, and any future TTL work has exactly one place to touch.

The reporter's measured data (1h arm ~40% cheaper on 5min–1hr gaps, but browser steps landing inside the 5-minute window) says 5m remains the right default, so nothing changes behavior by default. The new test `test_cache_control_ttl` locks in both the default and the explicit `'5m'`/`'1h'` emissions.

## Tests

- `browser_use/llm/tests/test_anthropic_cache.py` — 17 passed (includes new `test_cache_control_ttl`)
- `ruff check` clean on all four touched files

---
## Summary by cubic
Consolidates all Anthropic `CacheControlEphemeralParam` construction behind a TTL-aware helper without changing defaults. Previously three sites built the param directly; now `_serialize_cache_control(use_cache, ttl)` centralizes it, enabling an optional 1-hour TTL while keeping the 5-minute default byte-for-byte.

- Routes the serializer string-content path and both providers' `ToolParam.cache_control` through `AnthropicMessageSerializer._serialize_cache_control(...)`.
- Adds `test_cache_control_ttl` asserting default behavior and explicit `'5m'`/`'1h'` emissions; no migration required.

<sup>Written for commit cd0a4c77a2cf7ae7a93a79edb6b1a4cf65217053. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5350?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/llm/anthropic/chat.py
+++ b/browser_use/llm/anthropic/chat.py
@@ -13,7 +13,7 @@ from anthropic import (
 	RateLimitError,
 	omit,
 )
-from anthropic.types import CacheControlEphemeralParam, Message, ToolParam
+from anthropic.types import Message, ToolParam
 from anthropic.types.model_param import ModelParam
 from anthropic.types.text_block import TextBlock
 from anthropic.types.tool_choice_tool_param import ToolChoiceToolParam
@@ -429,7 +429,7 @@ class ChatAnthropic(BaseChatModel):
 					name=tool_name,
 					description=f'Extract information in the format of {tool_name}',
 					input_schema=schema,
-					cache_control=CacheControlEphemeralParam(type='ephemeral'),
+					cache_control=AnthropicMessageSerializer._serialize_cache_control(True),
 				)
 
 				if self._requires_auto_tool_choice():
--- a/browser_use/llm/anthropic/serializer.py
+++ b/browser_use/llm/anthropic/serializer.py
@@ -1,5 +1,5 @@
 import json
-from typing import overload
+from typing import Literal, overload
 
 from anthropic.types import (
 	Base64ImageSourceParam,
@@ -61,10 +61,17 @@ class AnthropicMessageSerializer:
 		return media_type, data  # type: ignore
 
 	@staticmethod
-	def _serialize_cache_control(use_cache: bool) -> CacheControlEphemeralParam | None:
-		"""Serialize cache control."""
+	def _serialize_cache_control(use_cache: bool, ttl: Literal['5m', '1h'] | None = None) -> CacheControlEphemeralParam | None:
+		"""Serialize cache control.
+
+		Anthropic applies a 5-minute TTL when none is specified, which is the default.
+		Pass ``ttl='1h'`` to opt into the cheaper 1-hour cache, useful when inter-request
+		gaps exceed five minutes.
+		"""
 		if use_cache:
-			return CacheControlEphemeralParam(type='ephemeral')
+			if ttl is None:
+				return CacheControlEphemeralParam(type='ephemeral')
+			return CacheControlEphemeralParam(type='ephemeral', ttl=ttl)
 		return None
 
 	@staticmethod
@@ -125,7 +132,13 @@ class AnthropicMessageSerializer:
 		"""Serialize content to Anthropic format."""
 		if isinstance(content, str):
 			if use_cache:
-				return [TextBlockParam(text=content, type='text', cache_control=CacheControlEphemeralParam(type='ephemeral'))]
+				return [
+					TextBlockParam(
+						text=content,
+						type='text',
+						cache_control=AnthropicMessageSerializer._serialize_cache_control(True),
+					)
+				]
 			else:
 				return content
 
--- a/browser_use/llm/anthropic/serializer.py
+++ b/browser_use/llm/anthropic/serializer.py
@@ -1,5 +1,5 @@
 import json
-from typing import overload
+from typing import Literal, overload
 
 from anthropic.types import (
 	Base64ImageSourceParam,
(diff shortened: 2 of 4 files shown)


Diwak4r:

PR #5351, opened 2026-08-03
fix(browser): block data: and blob: URLs when domain restrictions are configured

### LINKED ISSUE #4763: Security: `data:` and `blob:` URLs bypass `allowed_domains` restriction

Diwak4r:
Closes #4763

---
## Summary by cubic
Blocks `data:` URLs and applies domain checks to `blob:` origins when domain restrictions are enabled, closing an allowlist bypass. Previously both schemes skipped checks; now `data:` is blocked under restrictions and `blob:` is allowed only if its encoded origin passes. Closes #4763.

- Block `data:` when `allowed_domains` or `prohibited_domains` are set.
- Allow `blob:` only if its origin passes domain checks; unchanged when no restrictions.
- Add tests for restricted and unrestricted `data:`/`blob:` cases.

<sup>Written for commit 8e22852d3e40dde1eadcea6f7ead5c87675c974b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5351?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/browser/watchdogs/security_watchdog.py
+++ b/browser_use/browser/watchdogs/security_watchdog.py
@@ -210,7 +210,11 @@ class SecurityWatchdog(BaseWatchdog):
 				return False
 			# blob: URLs encode their origin as blob:<origin>/<uuid> — only allow
 			# when the origin itself passes the domain checks.
-			return self._is_url_allowed(url[5:])
+			origin_url = url[5:]
+			origin = urlparse(origin_url)
+			if origin.scheme in ['data', 'blob']:
+				return False
+			return self._is_url_allowed(origin_url)
 
 		# Get the actual host (domain)
 		host = parsed.hostname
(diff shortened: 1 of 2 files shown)


Diwak4r:

PR #5353, opened 2026-08-03
fix(dom): preserve explicitly empty accessibility names

### LINKED ISSUE #5271: [Bug] Potential loss of empty accessibility names
### Describe the bug

When extracting the accessibility name, the code uses an implicit truthiness check:

In Python, an empty string evaluates to False. If a DOM element explicitly has an empty accessibility name (e.g., aria-label=""), this condition fails and ax_name is set to None instead of preserving the empty string.

This conflates two distinct states:
- "the attribute does not exist" (None) 
- "the attribute exists but is intentionally empty" ("")

This could cause issues for downstream DOM matching or agent interaction logic.

### To Reproduce

Construct or parse an EnhancedDOMTreeNode where ax_node exists, but its name attribute is explicitly set to an empty string (""). Pass this tree node into the DOMInteractedElement.load_from_enhanced_dom_tree(enhanced_dom_tree) method. Inspect the resulting DOMInteractedElement instance. The ax_name property will be incorrectly set to None instead of the expected "".

### Expected Behavior

The code should preserve explicitly empty accessibility names.

Diwak4r:
Closes #5271

## Problem

`DOMInteractedElement.load_from_enhanced_dom_tree()` used an implicit truthiness check on the accessibility name:

```python
if enhanced_dom_tree.ax_node and enhanced_dom_tree.ax_node.name:
    ax_name = enhanced_dom_tree.ax_node.name
```

An explicitly empty accessibility name (`aria-label=""`) evaluates to falsy, so it was collapsed to `None`. That conflates two distinct states:

- the attribute does not exist → `None`
- the attribute exists but is intentionally empty → `""`

This can confuse downstream DOM matching / agent interaction logic.

## Fix

When the AX node exists, take `ax_node.name` as-is (`str | None`), so `None` stays `None` and `""` is preserved:

```python
ax_name = enhanced_dom_tree.ax_node.name if enhanced_dom_tree.ax_node else None
```

## Tests

Added `tests/ci/browser/test_dom_views.py` with four focused unit tests:

- explicitly empty `ax_name` (`""`) survives the round-trip
- missing name attribute stays `None`
- real accessibility names are preserved
- no AX node at all → `None`

---
## Summary by cubic
Fixes #5271 by preserving explicitly empty accessibility names (`aria-label=""`) when loading DOM elements. Previously `''` collapsed to `None`, conflating missing and intentionally empty names; now `ax_node.name` is taken as-is so missing stays `None` and empty stays `""`, preventing incorrect DOM matching and interaction. Adds tests covering empty, missing, and present names, plus no AX node.

<sup>Written for commit 728eb81c57d87501a1ba69920d69e7cf034b7718. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5353?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/dom/views.py
+++ b/browser_use/dom/views.py
@@ -1022,10 +1022,10 @@ class DOMInteractedElement:
 
 	@classmethod
 	def load_from_enhanced_dom_tree(cls, enhanced_dom_tree: EnhancedDOMTreeNode) -> 'DOMInteractedElement':
-		# Extract accessibility name if available
-		ax_name = None
-		if enhanced_dom_tree.ax_node and enhanced_dom_tree.ax_node.name:
-			ax_name = enhanced_dom_tree.ax_node.name
+		# Extract accessibility name if available. ax_node.name is str | None:
+		# preserve an explicitly empty name ("") rather than collapsing it to None,
+		# so the two states (attribute missing vs. intentionally empty) stay distinct.
+		ax_name = enhanced_dom_tree.ax_node.name if enhanced_dom_tree.ax_node else None
 
 		return cls(
 			node_id=enhanced_dom_tree.node_id,
--- a/browser_use/dom/views.py
+++ b/browser_use/dom/views.py
@@ -1026,10 +1026,10 @@ class DOMInteractedElement:
 
 	@classmethod
 	def load_from_enhanced_dom_tree(cls, enhanced_dom_tree: EnhancedDOMTreeNode) -> 'DOMInteractedElement':
-		# Extract accessibility name if available
-		ax_name = None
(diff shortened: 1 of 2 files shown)


Diwak4r:

PR #5369, opened 2026-08-03
fix: clean up temp .zip in _extract_extension on failure

### LINKED ISSUE #5364: Bug: _extract_extension leaks a temp .zip on extraction failure and unlinks a still-open handle
### Browser Use Version

`0.13.7` (commit `71714137`)

### Bug Description

```python
    temp_zip.write(zip_data)
    temp_zip.flush()

    os.unlink(temp_zip.name)
```

```
```

### Steps to Reproduce

### Failing Python Code

```python
import glob
import os
import pathlib
import tempfile

crx = tmpd / 'bad.crx'

try:
except Exception as e:

```

### Expected Behaviour

### Actual Behaviour

### Suggested Fix

```python
try:
    temp_zip.write(zip_data)
finally:
        os.unlink(temp_zip.name)
```

### LLM Model

Not applicable.

### Full DEBUG Log Output

```shell
```

Diwak4r:
## Summary

`_extract_extension` writes the CRX zip payload to a `NamedTemporaryFile(suffix='.zip', delete=False)` and only unlinks it on the happy path. Two problems:

1. **Temp file leak on failure** — if the embedded zip data fails to extract (corrupt data, disk full, etc.), the exception propagates before `os.unlink` runs, leaving a `.zip` file behind in the system temp dir forever.
2. **`PermissionError` on Windows** — the unlink runs while the `NamedTemporaryFile` handle is still open. On Windows this raises `WinError 32` ("The process cannot access the file because it is being used by another process"), which surfaces as a spurious `⚠️ Failed to setup ... extension` warning.

## Fix

- Move the unlink into a `finally` block so cleanup always runs, even when extraction fails.
- Close the temp file (exit the `with` block) *before* unlinking so the handle is no longer open.
- Suppress `FileNotFoundError` in cleanup so a racing removal cannot mask the original extraction error.

## Test plan

New `tests/ci/test_extension_extract.py` covers:

- **Success path**: valid CRX v3 extracts `manifest.json` and leaves no temp `.zip`.
- **Failure path**: corrupt embedded zip raises and leaves no temp `.zip` (was leaking).
- **Ordering**: asserts the temp handle is closed before `os.unlink` is called (Windows-safe).

All three fail on the old code (including the real `WinError 32` on Windows), all pass with the fix. Neighboring `test_extension_config.py` and `test_chrome_profile_helpers.py` still pass (20 tests).

Fixes #5364

---
## Summary by cubic
Fixes temp `.zip` cleanup in `_extract_extension` so extraction failures no longer leak files and Windows no longer raises PermissionError (WinError 32) from unlinking an open handle.

- Moves the unlink into a `finally` block and closes the handle before unlinking, so cleanup runs on both success and failure.
- Ignores `FileNotFoundError` during cleanup so a racing removal can't mask the original extraction error.
- Adds tests covering the success path, the failure path, and close-before-unlink ordering.

Fixes #5364.

<sup>Written for commit 69e2d513a5471be9eee8d816d15fa91a8ae92e9d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5369?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/browser/profile.py
+++ b/browser_use/browser/profile.py
@@ -1222,15 +1222,24 @@ async function initialize(checkInitialized, magic) {{
 				zip_data = f.read()
 
 			# Write ZIP data to temp file and extract
+			temp_zip_path = None
+			try:
+				with tempfile.NamedTemporaryFile(suffix='.zip', delete=False) as temp_zip:
+					temp_zip.write(zip_data)
+					temp_zip.flush()
+					temp_zip_path = temp_zip.name
 
-			with tempfile.NamedTemporaryFile(suffix='.zip', delete=False) as temp_zip:
-				temp_zip.write(zip_data)
-				temp_zip.flush()
-
-				with zipfile.ZipFile(temp_zip.name, 'r') as zip_ref:
+				with zipfile.ZipFile(temp_zip_path, 'r') as zip_ref:
 					zip_ref.extractall(extract_dir)
-
-				os.unlink(temp_zip.name)
+			finally:
+				# Close the temp file (with block exit) before unlinking so the
+				# unlink does not hit a still-open handle (PermissionError on
+				# Windows), and always clean up even when extraction fails.
+				if temp_zip_path is not None:
+					try:
+						os.unlink(temp_zip_path)
+					except FileNotFoundError:
+						pass
 
 	def detect_display_configuration(self) -> None:
 		"""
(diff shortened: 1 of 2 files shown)


Diwak4r:

PR #5370, opened 2026-08-03
fix: surface coordinate click failures and guard the highlight task

### LINKED ISSUE #5365: Bug: _click_by_coordinate discards the exception detail and uses a raw asyncio.create_task
### Browser Use Version

`0.13.7` (commit `71714137`)

### Bug Description

`_click_by_coordinate()` in `browser_use/tools/service.py` deviates from the conventions used everywhere else in the same file, in two ways that both cost diagnosability.

#### 1. The exception detail is discarded (`tools/service.py:695-697`)

```python
except Exception as e:
    error_msg = f'Failed to click at coordinates ({params.coordinate_x}, {params.coordinate_y}).'
    return ActionResult(error=error_msg)
```

`e` is bound and never used, so neither the LLM nor the logs get any indication of *why* the click failed. Its sibling handler `_click_by_index()` (`tools/service.py:758-760`) does the opposite:

```python
except Exception as e:
    error_msg = f'Failed to click element {params.index}: {str(e)}'
    return ActionResult(error=error_msg)
```

So the two click paths behave differently for identical failure classes. When coordinate clicking fails the agent is told only that it failed, retries the same coordinate, and the loop detector eventually stops it — with nothing in the record explaining the cause.

Worth noting that `F841` (unused local variable) is in the project's ruff ignore list (`pyproject.toml:130`), which is why this was never surfaced by linting. An AST sweep for `except ... as e` handlers that never reference `e` and contain no bare `raise` finds only six across the whole package, so this is an outlier rather than a house style:

```
browser_use/browser/watchdogs/default_action_watchdog.py:495
browser_use/dom/service.py:411
browser_use/sync/auth.py:332
browser_use/sync/service.py:99
browser_use/tools/service.py:696
browser_use/tools/service.py:1503
```

#### 2. Raw `asyncio.create_task` for a fire-and-forget highlight (`tools/service.py:671`)

```python
# Highlight the coordinate being clicked (truly non-blocking)
asyncio.create_task(browser_session.highlight_coordinate_click(actual_x, actual_y))
```

The project already has a helper for exactly this — `create_task_with_error_handling()` in `browser_use/utils.py:791`, whose docstring names the "Task exception was never retrieved" problem it exists to prevent. It is used four times in this same file for the same kind of non-blocking highlight call:

- `tools/service.py:723` — `highlight_click_element`
- `tools/service.py:788` — `highlight_type_element`
- `tools/service.py:931` — `highlight_file_input`
- `tools/service.py:972` — `highlight_file_input_fallback`

Line 671 is the only holdout. Two consequences: the event loop keeps only a weak reference to a bare task, so it can in principle be garbage-collected mid-execution (this is the documented caveat in the `asyncio.create_task` docs), and any exception inside `highlight_coordinate_click` surfaces as an unretrieved-task warning at an unrelated point in the log instead of being logged in context.

### Secondary site

`browser_use/browser/watchdogs/default_action_watchdog.py:493-497` has a related instance. The outer handler correctly logs "Falling back to page typing", but the nested recovery click is wrapped in a bare swallow:

```python
except Exception as e:
    self.logger.warning(f'Failed to type to element {index_for_logging}: {e}. Falling back to page typing.')
    try:
        await asyncio.wait_for(self._click_element_node_impl(element_node), timeout=10.0)
    except Exception as e:
        pass
```

When the fallback recovery itself fails, typing proceeds into an unfocused page with no trace of why. A single `self.logger.debug(...)` would be enough. Mentioning it here rather than opening a separate issue since it is the same class of defect; happy to split it out if you would rather track it independently.

### Steps to Reproduce

Static — visible by comparing `_click_by_coordinate` against `_click_by_index` and against the four `create_task_with_error_handling` call sites in the same file.

### Expected Behaviour

- Coordinate click failures include the underlying exception, matching `_click_by_index`.
- The highlight task uses `create_task_with_error_handling`, matching the other four highlight call sites.

### Actual Behaviour

The exception is dropped; the task is unguarded.

### Suggested Fix

```python
# line 671
create_task_with_error_handling(
    browser_session.highlight_coordinate_click(actual_x, actual_y),
    name='highlight_coordinate_click',
    suppress_exceptions=True,
)

# line 695-697
except Exception as e:
    error_msg = f'Failed to click at coordinates ({params.coordinate_x}, {params.coordinate_y}): {e}'
    return ActionResult(error=error_msg)
```

`create_task_with_error_handling` is already imported in this module at line 62, so no new import is needed. Happy to open a PR.

### LLM Model

Not model specific.

### Full DEBUG Log Output

Not applicable — the defect is that there is no useful log output in the failure case.

Diwak4r:
Fixes #5365

`_click_by_coordinate()` deviated from the conventions used everywhere else in `tools/service.py`, in two ways that both cost diagnosability:

1. **Exception detail discarded** — the generic handler bound `e` but dropped it, so neither the LLM nor the logs got any indication of *why* a coordinate click failed. Its sibling `_click_by_index()` includes the cause (`Failed to click element {index}: {str(e)}`). The error message now includes the underlying exception.

2. **Bare `asyncio.create_task` for the highlight** — line 671 was the only holdout in the file against `create_task_with_error_handling()` (used at 723, 788, 931, 972 for the same kind of fire-and-forget highlight). The helper prevents "Task exception was never retrieved" warnings and logs failures in context. It now uses `create_task_with_error_handling(..., name='highlight_coordinate_click', suppress_exceptions=True)`, mirroring `_click_by_index()`.

Also included: the secondary site from the issue — `default_action_watchdog.py` now logs the fallback recovery click failure at debug level instead of swallowing it silently.

## Tests

New `tests/ci/test_coordinate_click_exception.py` drives the real registered click action against a lightweight `BrowserSession` stand-in (no browser launched):

- `test_coordinate_click_error_includes_exception_detail` — asserts the `ActionResult.error` contains the underlying exception message (fails on old code, where `e` was discarded).
- `test_coordinate_click_highlight_uses_error_handling_helper` — asserts the highlight task goes through `create_task_with_error_handling` with the right name and `suppress_exceptions=True`, and that raw `asyncio.create_task` is not used (fails on old code).

Verified: new tests 2/2 green, existing `test_coordinate_clicking.py` 33/33, plus adjacent `test_element_click_error.py`, `test_tools.py`, `test_async_task_error_handling.py` — 21/21. `ruff check` clean on all touched files.

---
## Summary by cubic
Fixes coordinate-click failures so the agent and logs show why a click failed, and routes the highlight task through error handling to avoid unretrieved-task warnings.

- `_click_by_coordinate()` now includes the underlying exception in the error message, matching `_click_by_index()`.
- The coordinate highlight uses `create_task_with_error_handling(name='highlight_coordinate_click', suppress_exceptions=True)` instead of a bare `asyncio.create_task`.
- `default_action_watchdog.py` now logs fallback typing click failures at debug level with the element index.
- Adds `tests/ci/test_coordinate_click_exception.py` covering error detail propagation and highlight task routing.

Fixes #5365.

<sup>Written for commit 1e4bb2758d145b31f33123cb98d0843e4900ee05. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5370?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/browser/watchdogs/default_action_watchdog.py
+++ b/browser_use/browser/watchdogs/default_action_watchdog.py
@@ -493,7 +493,7 @@ class DefaultActionWatchdog(BaseWatchdog):
 					try:
 						await asyncio.wait_for(self._click_element_node_impl(element_node), timeout=10.0)
 					except Exception as e:
-						self.logger.debug(f'Fallback click for page typing failed: {e}')
+						self.logger.debug(f'Failed to click element {index_for_logging} for fallback typing: {e}')
 					await self._type_to_page(event.text)
 					# Log with sensitive data protection
 					if event.is_sensitive:
--- a/tests/ci/test_coordinate_click_exception.py
+++ b/tests/ci/test_coordinate_click_exception.py
@@ -1,0 +1,123 @@
+"""Tests for #5365: _click_by_coordinate() diagnosability.
+
+Two deviations from the conventions used elsewhere in tools/service.py:
+
+1. The generic exception handler in _click_by_coordinate() binds `e` but discards
+   it, so the LLM never learns *why* a coordinate click failed (its sibling
+   _click_by_index() includes the cause).
+2. The fire-and-forget coordinate highlight uses a bare asyncio.create_task
+   instead of the project's create_task_with_error_handling() helper, so a
+   highlight failure surfaces as an unretrieved-task warning instead of being
+   logged in context.
+
+These tests drive the real registered click action with a lightweight
+BrowserSession stand-in (no browser is launched).
+"""
+
+import asyncio
+
+from browser_use.tools.service import Tools
+from browser_use.tools.views import ClickElementAction
+
+
+class _FakeEvent:
+	"""Minimal event-bus event: awaitable, plus event_result()."""
+
+	def __init__(self, result=None):
+		self._result = result
+
+	def __await__(self):
+		async def _done():
+			return None
+
+		return _done().__await__()
+
+	async def event_result(self, raise_if_any=True, raise_if_none=False):
+		if raise_if_any and isinstance(self._result, Exception):
+			raise self._result
+		if raise_if_none and self._result is None:
+			raise RuntimeError('No event result was produced')
+		return self._result
+
+
+class _FakeEventBus:
+	def dispatch(self, event):
+		return _FakeEvent()
+
+
+class _FakeBrowserSession:
+	"""BrowserSession stand-in exposing only what _click_by_coordinate touches."""
+
+	def __init__(self, tabs_error=None):
+		self.llm_screenshot_size = None
+		self._original_viewport_size = None
+		self.tabs_error = tabs_error
+		self.event_bus = _FakeEventBus()
+		self.highlight_coordinate_click_calls = []
+
+	async def get_tabs(self):
+		if self.tabs_error is not None:
+			raise self.tabs_error
+		return []
+
+	async def highlight_coordinate_click(self, x, y):
+		self.highlight_coordinate_click_calls.append((x, y))
+		return None
+
+
+async def _run_coordinate_click(session, x=100, y=200):
+	tools = Tools()
+	tools.set_coordinate_clicking(True)
+	action = tools.registry.registry.actions['click']
+	params = ClickElementAction(coordinate_x=x, coordinate_y=y)
+	return await action.function(params=params, browser_session=session)
+
+
+async def test_coordinate_click_error_includes_exception_detail():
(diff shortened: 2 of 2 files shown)


Diwak4r:

PR #5380, opened 2026-08-03
fix(tools): cap evaluate image extraction by occurrence count

### LINKED ISSUE #5367: Bug: evaluate() caps text at 20k chars but forwards an unbounded base64 image payload to the LLM
### Browser Use Version

`0.13.7` (commit `71714137`)

### Bug Description

The `evaluate` action truncates its text result at 20,000 characters but applies no limit at all to the base64 images it extracts, so a single `evaluate` call can push an arbitrarily large image payload into the next LLM request.

`browser_use/tools/service.py:1884-1902`:

```python
image_pattern = r'(data:image/[^;]+;base64,[A-Za-z0-9+/=]+)'
found_images = re.findall(image_pattern, result_text)

metadata = None
if found_images:
    # Store images in metadata so they can be added as ContentPartImageParam
    metadata = {'images': found_images}

    # Replace image data in result text with shorter placeholder
    modified_text = result_text
    for i, img_data in enumerate(found_images, 1):
        placeholder = '[Image]'
        modified_text = modified_text.replace(img_data, placeholder)
    result_text = modified_text

# Apply length limit with better truncation (after image extraction)
if len(result_text) > 20000:
    result_text = result_text[:19950] + '\n... [Truncated after 20000 characters]'
```

The images are pulled out of `result_text` *before* the truncation, so the 20,000-character cap measures the text with every image already replaced by the 7-character `[Image]` placeholder. The full base64 strings live on in `metadata['images']`, uncapped in both count and per-image size, and are later attached as `ContentPartImageParam` on the next model call.

The ordering makes the code read as bounded when it is not — the truncation sits immediately below and looks like it governs the whole result.

Measured with a page returning one ~300 KB inline image:

```
JS returned              : 300,034 chars
extracted_content (text) : 19 chars       <- capped
metadata["images"]       : 1 image, 300,022 chars  <- not capped
ratio image/text         : 15791x
```

### Steps to Reproduce

Have `evaluate` return HTML containing one or more large `data:image/...;base64,...` URIs and inspect the returned `ActionResult`.

### Failing Python Code

```python
import asyncio

BIG = 'A' * 300_000
JS_RESULT = f'<img src="data:image/png;base64,{BIG}">'


class Send:
    class Runtime:
        @staticmethod
        async def evaluate(params=None, session_id=None):
            return {'result': {'value': JS_RESULT}}


class Client:
    send = Send()


class CdpSession:
    cdp_client = Client()
    session_id = 's'


class FakeSession:
    async def get_or_create_cdp_session(self):
        return CdpSession()


ev = Tools().registry.registry.actions['evaluate'].function
r = asyncio.run(ev(code='x', browser_session=FakeSession()))
imgs = (r.metadata or {}).get('images', [])

print('text  :', len(r.extracted_content), 'chars')
print('images:', len(imgs), 'totalling', sum(len(i) for i in imgs), 'chars')
```

### Expected Behaviour

The image payload is bounded — a cap on the number of images forwarded and on total or per-image size, with anything beyond the cap dropped and noted in the text result so the model knows images were omitted.

### Actual Behaviour

Unbounded. Text is capped at 20,000 characters while the accompanying image payload is not capped at all.

### Why This Matters

An agent running `evaluate` to inspect page structure on a media-heavy page — inlined SVG/PNG data URIs are common in dashboards and design tools — can blow well past the model's context window or generate a large unexpected bill from a single action. Because the failure comes from the *metadata* rather than the visible text, it is not obvious from the action result why the request ballooned.

This is doubly likely because the action description explicitly suggests using `evaluate` for "analysing page structure" and "extract/filter links", which is exactly the kind of call that returns large chunks of markup.

### Suggested Fix

Bound the extraction, for example:

```python
MAX_IMAGES = 4
MAX_IMAGE_CHARS = 256_000  # per image, roughly 190 KB decoded

kept = [img for img in found_images if len(img) <= MAX_IMAGE_CHARS][:MAX_IMAGES]
dropped = len(found_images) - len(kept)
if kept:
    metadata = {'images': kept}
if dropped:
    result_text += f'\n... [{dropped} image(s) omitted: too large or over the limit of {MAX_IMAGES}]'
```

The specific numbers are placeholders — the right values depend on what you consider a reasonable per-step image budget, which I do not want to guess at. Happy to open a PR once there is a preferred limit, with a test asserting the payload stays bounded.

### LLM Model

Any vision-capable model. The cost impact scales with the provider's image pricing.

### Full DEBUG Log Output

```shell
DEBUG    [tools] JavaScript executed successfully, result length: 19
  JS returned              : 300,034 chars
  extracted_content (text) : 19 chars
  metadata["images"]       : 1 image(s), 300,022 chars
  ratio image/text         : 15791x
```

Diwak4r:
Fixes #5367

## Problem
\evaluate()\ extracts \data:image/...;base64\ URIs from the JavaScript result into metadata **before** the 20,000-char text truncation, so a single evaluate call can forward an arbitrarily large image payload to the next LLM request.

## Fix
Cap the images extracted from evaluate results:
- Keep only the first **10** images (matching the \max_images\ screenshot budget) that fit within **100,000 base64 chars** each (~75 KB decoded).
- Drop everything beyond the cap, replace it with an \[Image omitted]\ placeholder in the text, and append an omission note *after* truncation so the model knows images were dropped.
- The cap counts **occurrences, not unique values**: 12 identical inline images consume 12 of the 10-image budget (previously \dict.fromkeys\ collapsed them to 1 and the cap was never reached).

## Tests
New \	ests/ci/test_evaluate_image_cap.py\:
- oversized single image → dropped, noted, no base64 leaks into text
- 12 small images → 10 kept, 2 omitted and noted
- small inline image → passthrough unchanged (regression guard)

All 3 pass locally; full \	ests/ci/\ run: 76 passed, 3 skipped, 1 pre-existing Windows path-format failure (\	est_path_configuration\, fails on pristine main too).

---
## Summary by cubic
Caps inline image extraction in `evaluate()` so a single call can't forward an arbitrarily large base64 payload to the next request. Previously all inline images were extracted before the 20,000-char truncation; now only the first 10 that fit within 100,000 base64 chars each are kept.

- Kept images are replaced in the text with `[Image]`; dropped ones with `[Image omitted]`.
- The cap counts occurrences, not unique values, so 12 identical images consume 12 of the 10-image budget.
- An omission note is appended after truncation so it survives the cut.
- Adds `tests/ci/test_evaluate_image_cap.py` covering oversized images, count capping, and small-image passthrough.

<sup>Written for commit 4ef5111055d45f98cc492282a19aec793612168c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5380?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/tools/service.py
+++ b/browser_use/tools/service.py
@@ -1898,20 +1898,42 @@ Validated Code (after quote fixing):
 				found_images = re.findall(image_pattern, result_text)
 
 				metadata = None
+				dropped_count = 0
 				if found_images:
-					# Store images in metadata so they can be added as ContentPartImageParam
-					metadata = {'images': found_images}
+					# Cap images sent to the model: house screenshot budget is 10 images,
+					# and each base64 image must fit comfortably in a single message.
+					MAX_IMAGES = 10
+					MAX_IMAGE_CHARS = 100000
 
-					# Replace image data in result text with shorter placeholder
+					# Cap by occurrence, not by unique value: 12 identical inline
+					# images still consume 12 slots of the 10-image budget.
+					size_ok = [img for img in found_images if len(img) <= MAX_IMAGE_CHARS]
+					kept_images = size_ok[:MAX_IMAGES]
+					dropped_count = len(found_images) - len(kept_images)
+
+					# Store kept images in metadata so they can be added as ContentPartImageParam
+					if kept_images:
+						metadata = {'images': kept_images}
+
+					# Replace every image in result text with a shorter placeholder;
+					# dropped images get an explicit marker so the model knows they were omitted.
+					# Dedupe only for replacement efficiency; counting stays occurrence-based.
 					modified_text = result_text
-					for i, img_data in enumerate(found_images, 1):
-						placeholder = '[Image]'
+					for img_data in dict.fromkeys(found_images):
+						placeholder = '[Image]' if img_data in kept_images else '[Image omitted]'
 						modified_text = modified_text.replace(img_data, placeholder)
 					result_text = modified_text
 
 				# Apply length limit with better truncation (after image extraction)
 				if len(result_text) > 20000:
 					result_text = result_text[:19950] + '\n... [Truncated after 20000 characters]'
+
+				# Tell the model about any omitted images; appended after truncation so it survives.
+				if dropped_count:
+					result_text += (
+						f'\n... [{dropped_count} image(s) omitted: '
+						f'over the cap of {MAX_IMAGES} images / {MAX_IMAGE_CHARS} chars each]'
+					)
 
 				# Don't log the code - it's already visible in the user's cell
 				logger.debug(f'JavaScript executed successfully, result length: {len(result_text)}')
--- a/tests/ci/test_evaluate_image_cap.py
+++ b/tests/ci/test_evaluate_image_cap.py
@@ -1,0 +1,83 @@
+"""Tests for bounded image extraction in the evaluate action.
+
+Root cause of #5367: evaluate() extracts data:image/...;base64 URIs from the JavaScript
+result into metadata['images'] BEFORE the 20,000-char text truncation, so a single evaluate
+call can forward an arbitrarily large image payload to the next LLM request.
+
+The fix caps the number of images (10, matching the max_images screenshot budget) and the size
+of each (100_000 base64 chars, roughly 75 KB decoded). Anything beyond the cap is dropped,
+replaced with an [Image omitted] placeholder in the text, and counted in an omission note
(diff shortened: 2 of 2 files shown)


Diwak4r:

PR #5381, opened 2026-08-03
fix(tools): report scroll/input/scroll_to_text failures as errors

### LINKED ISSUE #5361: Bug: scroll/input/scroll_to_text report failed actions as successes, so multi_act never aborts and consecutive_failures never increments
### Browser Use Version

`0.13.7` (commit `71714137`)

### Bug Description

Three actions in `browser_use/tools/service.py` return a **non-error** `ActionResult` when the underlying operation actually failed. Because the agent loop keys off `ActionResult.error`, a failed action is treated as a successful one.

Two places consume `.error` and both are bypassed:

- `Agent.multi_act()` (`agent/service.py:2812`) breaks the remaining action queue on `results[-1].error`. With `error=None`, queued follow-up actions keep executing against a page that never changed.
- `Agent._post_process()` (`agent/service.py:1229`) increments `consecutive_failures` only when `last_result[-1].error` is set. With `error=None`, the failure never counts toward `max_failures`.

On top of that, the LLM is told the action succeeded, so `evaluation_previous_goal` on the next step is reasoning from a false premise.

#### Site 1 — `scroll` reports success when every scroll attempt failed (most severe)

`tools/service.py:1403-1452`. For `pages >= 1.0` the loop catches per-scroll exceptions, logs a warning, and increments `completed_scrolls` only on success. Afterwards:

```python
if params.pages == 1.0:
    long_term_memory = f'Scrolled {direction} {target} {viewport_height}px'
else:
    long_term_memory = f'Scrolled {direction} {target} {completed_scrolls:.1f} pages'
```

The `pages == 1.0` branch never consults `completed_scrolls`. Since `ScrollAction.pages` defaults to `1.0` (`tools/views.py:132`), **the default scroll silently reports success after a total failure.**

Measured with a scroll dispatch that always raises:

```
WARNING [tools] Scroll 1/1 failed: CDP scroll dispatch failed
  pages=1.0: error=None  memory='Scrolled down 1000px'      <- total failure reported as success
  pages=3.0: error=None  memory='Scrolled down 0.0 pages'   <- honest text, but still not an error
  pages=0.5: error='Failed to execute scroll action.'       <- correct
```

#### Site 2 — `input` returns a non-error result on a stale element index

`tools/service.py:781-785`:

```python
node = await browser_session.get_element_by_index(params.index)
if node is None:
    msg = f'Element index {params.index} not available - page may have changed. Try refreshing browser state.'
    logger.warning(f'⚠️ {msg}')
    return ActionResult(extracted_content=msg)   # not error=
```

`scroll` returns `ActionResult(error=msg)` for the identical "index not found" condition (`tools/service.py:1380-1381`), so the two actions disagree. The practical effect is that a typed value silently goes nowhere and any queued follow-up (e.g. a submit click) still runs.

#### Site 3 — `scroll_to_text` cannot distinguish "not found" from "CDP failed"

`tools/service.py:1502-1509`. A single `except Exception` covers both a genuine "text is not on the page" and a real CDP/transport failure, discards the exception object entirely, and returns a non-error `ActionResult` in both cases. A broken CDP call is indistinguishable from a legitimate negative result.

### Steps to Reproduce

Drive each action with a browser session whose event dispatch raises, and inspect the returned `ActionResult`.

### Failing Python Code

```python
import asyncio

from browser_use.tools.views import ScrollAction


class BoomEvent:
    def __await__(self):
        async def _():
            return self
        return _().__await__()

    async def event_result(self, **kw):
        raise RuntimeError('CDP scroll dispatch failed')


class Bus:
    def dispatch(self, ev):
        return BoomEvent()


class FakeSession:
    event_bus = Bus()

    async def get_element_by_index(self, i):
        return None

    async def get_or_create_cdp_session(self):
        raise RuntimeError('no cdp')


tools = Tools()
scroll = tools.registry.registry.actions['scroll'].function


async def main():
    for pages in (1.0, 3.0, 0.5):
        r = await scroll(params=ScrollAction(down=True, pages=pages), browser_session=FakeSession())
        print(f'pages={pages}: error={r.error!r} memory={r.long_term_memory!r}')


asyncio.run(main())
```

### Expected Behaviour

An action whose underlying operation failed returns `ActionResult(error=...)`, so `multi_act()` aborts the remaining queue and `_post_process()` counts the failure.

### Actual Behaviour

`error=None` and a success message. The agent proceeds as if the scroll/type happened.

### Why This Matters

This is, I think, the mechanism behind a familiar class of reports — the agent scrolling repeatedly and never reaching content, or filling a form where one field silently never received its value. Because `consecutive_failures` never increments, `max_failures` never trips and the only thing that can break the cycle is the loop detector.

### Suggested Fix

- `scroll`: track `completed_scrolls` in the `pages == 1.0` branch too, and return `ActionResult(error=...)` when `completed_scrolls == 0`.
- `input`: return `ActionResult(error=msg)` for a missing index, matching `scroll`.
- `scroll_to_text`: separate "handler reported text-not-found" from "the call raised", and include the exception detail in the latter.

Happy to open a PR. Given they share one root cause I would do them as a single change with three focused tests, but I am glad to split it if you would prefer one action per PR.

### LLM Model

Not model specific.

### Full DEBUG Log Output

```shell
WARNING  [tools] Scroll 1/1 failed: CDP scroll dispatch failed
  pages=1.0: error=None  memory='Scrolled down 1000px'
WARNING  [tools] Scroll 1/3 failed: CDP scroll dispatch failed
WARNING  [tools] Scroll 2/3 failed: CDP scroll dispatch failed
WARNING  [tools] Scroll 3/3 failed: CDP scroll dispatch failed
  pages=3.0: error=None  memory='Scrolled down 0.0 pages'
  pages=0.5: error='Failed to execute scroll action.'  memory=None
```

Diwak4r:
Fixes #5361

## Problem
\Agent.multi_act()\ and \Agent._post_process()\ key off \ActionResult.error\ to abort queued actions and increment \consecutive_failures\. The \scroll\, \input\, and \scroll_to_text\ tools were returning success payloads even when every underlying action failed, so the agent kept claiming progress and never escalated.

## Changes
- **scroll**: if no scroll events completed, return \ActionResult(error='Failed to execute scroll action.')\ instead of a success message
- **input**: surface the stale-index failure as an error instead of \extracted_content\
- **scroll_to_text**: split the catch into \except BrowserError\ (legit \Text not found\ negative stays a non-error) and \except Exception\ (real CDP/transport failures become errors)

## Tests
New \	ests/ci/test_report_failed_actions.py\ with 6 tests: 3 covering the new error paths, 3 regression guards for the success/negative-result behavior. Full \	ests/ci/\ run: 76 passed, 3 skipped, 1 pre-existing Windows-only failure (\	est_path_configuration\).

## Note for reviewers
\click\, \dropdown_options\, and \select_dropdown\ use the same \extracted_content\-on-failure pattern but are intentionally left untouched here to keep the change scoped to this issue.

---
## Summary by cubic
Fixes `scroll`, `input`, and `scroll_to_text` reporting failed actions as successes, so the agent now aborts queued actions and increments `consecutive_failures`. Genuine “text not found” stays a non-error negative.

- `scroll`: total failure returns an error; partial scrolls still report honest completed counts.
- `input`: stale element index sets `error` instead of `extracted_content`.
- `scroll_to_text`: `BrowserError` remains a non-error negative; other exceptions become errors.
- Adds CI coverage for new error paths and regressions, and narrows `extracted_content` type for `pyright`.
- `click`, `dropdown_options`, and `select_dropdown` keep current behavior.

<sup>Written for commit b7982153f50fa12787e9793ac6ced639d9850941. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5381?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/tools/service.py
+++ b/browser_use/tools/service.py
@@ -786,7 +786,7 @@ class Tools(Generic[Context]):
 			if node is None:
 				msg = f'Element index {params.index} not available - page may have changed. Try refreshing browser state.'
 				logger.warning(f'⚠️ {msg}')
-				return ActionResult(extracted_content=msg)
+				return ActionResult(error=msg)
 
 			# Highlight the element being typed into (truly non-blocking)
 			create_task_with_error_handling(
@@ -1463,6 +1463,11 @@ You will be given a query and the markdown of a webpage that has been filtered t
 						long_term_memory = f'Scrolled {direction} {target} {viewport_height}px'.replace('  ', ' ')
 					else:
 						long_term_memory = f'Scrolled {direction} {target} {completed_scrolls:.1f} pages'.replace('  ', ' ')
+
+					if completed_scrolls <= 0:
+						# Every scroll attempt failed; report an error so multi_act aborts and
+						# consecutive_failures increments instead of silently claiming success.
+						return ActionResult(error='Failed to execute scroll action.')
 				else:
 					# For fractional pages <1.0, do single scroll
 					pixels = int(params.pages * viewport_height)
@@ -1512,14 +1517,19 @@ You will be given a query and the markdown of a webpage that has been filtered t
 				msg = f'🔍  {memory}'
 				logger.info(msg)
 				return ActionResult(extracted_content=memory, long_term_memory=memory)
-			except Exception as e:
-				# Text not found
+			except BrowserError:
+				# Text genuinely not found - a legitimate negative result, not an error.
 				msg = f"Text '{text}' not found or not visible on page"
 				logger.info(msg)
 				return ActionResult(
 					extracted_content=msg,
 					long_term_memory=f"Tried scrolling to text '{text}' but it was not found",
 				)
+			except Exception as e:
+				# Real CDP/transport failure - report an error so multi_act aborts and
+				# consecutive_failures increments instead of masquerading as "not found".
+				logger.error(f'Failed to dispatch ScrollToTextEvent: {type(e).__name__}: {e}')
+				return ActionResult(error=f'Failed to scroll to text: {e}')
 
 		@self.registry.action(
 			'Take a screenshot of the current viewport. If file_name is provided, saves to that file and returns the path. '
--- a/tests/ci/test_report_failed_actions.py
+++ b/tests/ci/test_report_failed_actions.py
@@ -128,4 +128,5 @@ class TestScrollToTextReportsFailure:
 		result: ActionResult = _run('find_text', text='needle', browser_session=session)
 
 		assert result.error is None
+		assert result.extracted_content is not None
 		assert 'not found' in result.extracted_content
--- a/tests/ci/test_report_failed_actions.py
+++ b/tests/ci/test_report_failed_actions.py
@@ -1,0 +1,131 @@
+"""Regression tests for #5361: scroll/input/scroll_to_text report failed actions as successes.
+
+Three actions in browser_use/tools/service.py returned a non-error ActionResult when the
+underlying operation failed. Because the agent loop keys off ActionResult.error:
+
+- Agent.multi_act() keeps executing queued follow-up actions on a page that never changed
+- Agent._post_process() never increments consecutive_failures, so max_failures is never hit
+- the LLM is told the action succeeded and reasons from a false premise next step
+
+Sites under test:
+1. scroll (pages >= 1.0): total scroll failure reports success; the pages == 1.0 branch
+   never consults completed_scrolls.
+2. input: a stale element index returns extracted_content instead of error, unlike scroll.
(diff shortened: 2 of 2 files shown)


Diwak4r:

PR #5383, opened 2026-08-03
fix(browser): re-detect WS drop landing in the reconnect window

### LINKED ISSUE #5366: Bug: WebSocket drop during the reconnect window is permanently undetected (one-shot done_callback consumed while _reconnecting=True)
### Browser Use Version

`0.13.7` (commit `71714137`)

### Bug Description

If the CDP WebSocket drops during the short window inside `_auto_reconnect()` after `reconnect()` has re-attached the drop callback but before the `finally` clears `_reconnecting`, the drop is swallowed and no further reconnect is ever scheduled. The session is left permanently disconnected with drop detection disabled.

#### The window

`BrowserSession.reconnect()` re-attaches the drop callback as its final step (`browser/session.py:2236-2237`):

```python
# 8. Attach the WS drop detection callback to the new client
self._attach_ws_drop_callback()
```

`reconnect()` is called from inside `_auto_reconnect()`'s `try`, and `_reconnecting` is only cleared in the `finally` (`browser/session.py:2292-2294`):

```python
finally:
    self._reconnecting = False
    self._reconnect_event.set()
```

So between step 8 and that `finally`, the callback is live on the **new** WebSocket task while `self._reconnecting` is still `True`.

#### Why that is fatal rather than merely racy

The callback's first line is a guard (`browser/session.py:2307-2309`):

```python
def _on_message_handler_done(fut: asyncio.Future) -> None:
    # Guard: skip if intentionally stopped, already reconnecting, or no cdp_url
    if self._intentional_stop or self._reconnecting or not self.cdp_url:
        return
```

If the new socket drops inside the window, the callback fires, sees `_reconnecting is True`, and returns without scheduling anything. That would be recoverable if the callback could fire again later — but it is registered via `Task.add_done_callback`, which is one-shot, and the task it is attached to has already completed. Nothing re-attaches it, because re-attachment only happens inside `reconnect()`, which is only reached from `_auto_reconnect()`, which is only triggered by this very callback.

The result is a closed cycle: the connection is dead, `_reconnecting` is `False`, no reconnect task exists, and no future drop can be detected.

#### Reproduction

Reproduced deterministically by simulating step 8 (attach while `_reconnecting is True`) and then failing the new socket task inside the window:

```
drop happened while _reconnecting=True
reconnect task scheduled : False
after _reconnecting=False -> callback re-fires? False
ws task already done      : True  (done_callback is one-shot, cannot fire again)
-> connection is dead, no reconnect scheduled, no further drop detection
```

### Steps to Reproduce

```python
import asyncio

class FakeClient:
    def __init__(self, task):
        self._message_handler_task = task


async def main():
    s = BrowserSession(browser_profile=BrowserProfile(headless=True, user_data_dir=None))
    s.browser_profile.cdp_url = 'ws://127.0.0.1:9222/x'
    s._intentional_stop = False

    # Simulate step 8 of reconnect(): callback attached to the NEW ws task while
    # _auto_reconnect() has not yet reached its finally, so _reconnecting is still True.
    s._reconnecting = True
    fut = asyncio.get_running_loop().create_future()
    task = asyncio.ensure_future(fut)
    s._cdp_client_root = FakeClient(task)
    s._attach_ws_drop_callback()

    # The brand-new WebSocket drops inside that window.
    fut.set_exception(ConnectionResetError('ws dropped again'))
    await asyncio.sleep(0.05)
    print('reconnect task scheduled :', getattr(s, '_reconnect_task', None) is not None)

    # _auto_reconnect()'s finally runs.
    s._reconnecting = False
    await asyncio.sleep(0.05)
    print('ws task already done     :', task.done(), '- done_callback is one-shot')


asyncio.run(main())
```

### Expected Behaviour

A drop occurring during the reconnect window is either handled by the in-flight `_auto_reconnect()` (for example by looping again rather than returning success) or re-detected once `_reconnecting` clears.

### Actual Behaviour

The drop is silently discarded and drop detection is permanently disabled for that session.

### Why This Matters

The window is narrow, but the conditions that produce it are exactly the conditions where reconnects happen: a flaky or saturated link where a freshly established socket is likely to drop again quickly. It is also silent — no error, no log beyond the initial drop warning — so from the outside it presents as the agent hanging on the next CDP call rather than as a connection failure.

This may be a contributing factor in some of the existing CDP-instability reports (#5048, #4688), though I want to be careful not to overclaim — I have demonstrated the mechanism in isolation, not tied it to any specific field report.

### Suggested Fix

A few options, roughly in order of how invasive they are:

1. Re-check connection liveness at the end of `_auto_reconnect()` before clearing `_reconnecting`, and loop if the socket is already dead.
2. Attach the drop callback *after* `_reconnecting` is cleared, rather than as the last step of `reconnect()`.
3. Have the callback record a "drop occurred while reconnecting" flag that `_auto_reconnect()`'s `finally` inspects and acts on.

Option 3 is probably the smallest change that closes the window without reordering the reconnect sequence, but I would rather defer to your judgement on which fits the intended lifecycle. Happy to open a PR once there is a preferred direction, with a regression test built on the reproduction above.

### LLM Model

Not applicable.

macOS (Darwin 25.5.0), Python 3.12.13. Not platform specific — reproduction uses no real browser.

### Full DEBUG Log Output

```shell
WARNING  [BrowserSession] 🔌 CDP WebSocket message handler exited unexpectedly: ConnectionResetError: ws dropped again
  reconnect task scheduled : False
  after _reconnecting=False -> callback re-fires? False
  ws task already done      : True  (done_callback is one-shot, cannot fire again)
```

Diwak4r:
## Summary

\BrowserSession.reconnect()\ re-attaches the one-shot WebSocket drop callback as its final step, while \_auto_reconnect()\ only clears \_reconnecting\ inside its \inally\. If the brand-new socket drops inside that window, the callback fires, sees \_reconnecting is True\, and returns — and because \dd_done_callback\ is one-shot and the task it was attached to is already done, no future drop can ever be detected again. The session is left permanently disconnected with drop detection disabled.

## Fix

- The drop callback now records the drop (\_drop_while_reconnecting = True\) instead of silently swallowing it when a reconnect is in flight.
- \_auto_reconnect()\'s \inally\ checks that flag and, when set, spawns a fresh reconnect pass — closing the window so the new socket gets a fresh drop callback.
- The existing \_intentional_stop\ guard ensures \
eset()\ (which cancels \_reconnect_task\) cannot trigger a reconnect loop.

## Tests

Added \	ests/ci/test_reconnect_window_drop.py\ with two regression tests:

1. A drop consumed by the one-shot callback while reconnecting is recorded on the session.
2. \_auto_reconnect()\ re-runs when a drop landed inside the reconnect window, and the flag is reset once the healthy socket is attached.

Full \	ests/ci\ run: 76 passed, 3 skipped, 1 pre-existing Windows-only failure (\	est_path_configuration\, path-separator assertion).

Fixes #5366

---
## Summary by cubic
Fixes WebSocket drop detection during reconnects so sessions no longer get stuck disconnected with drop detection permanently disabled. Previously a drop landing in the reconnect window consumed the one-shot callback and returned while `_reconnecting` was True; now the drop is recorded and `_auto_reconnect()` reschedules itself so the new socket gets a fresh callback. Fixes #5366.

- Records drops during reconnects in `_reconnect_pending` and logs them from the drop callback.
- `_auto_reconnect()`'s `finally` resets the flag and spawns another reconnect pass unless the stop was intentional or no `cdp_url` is set.
- Adds regression tests covering the flag recording and the retry path.

<sup>Written for commit abc06c91dc3b18ae562baf5bea13450fb4fa0c8e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5383?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/browser/session.py
+++ b/browser_use/browser/session.py
@@ -2324,8 +2324,18 @@ class BrowserSession(BaseModel):
 			# Guard: skip if intentionally stopped or no cdp_url
 			if self._intentional_stop or not self.cdp_url:
 				return
+			# A reconnect is already in flight: this drop landed inside the
+			# reconnect window. The one-shot done callback is consumed right
+			# here, so record the drop for _auto_reconnect()'s finally to act
+			# on -- otherwise no future drop could ever be detected again
+			# (issue #5366).
 			if self._reconnecting:
 				self._reconnect_pending = True
+				exc = fut.exception() if not fut.cancelled() else None
+				self.logger.warning(
+					f'🔌 CDP WebSocket dropped again during reconnect'
+					f'{f": {type(exc).__name__}: {exc}" if exc else " (connection closed)"}'
+				)
 				return
 
 			# The message handler task exiting means the WS connection dropped
--- a/tests/ci/test_reconnect_window_drop.py
+++ b/tests/ci/test_reconnect_window_drop.py
@@ -1,0 +1,96 @@
+"""Regression tests for #5366: a WS drop inside the reconnect window must not be lost.
+
+BrowserSession.reconnect() re-attaches the one-shot WS drop callback as its
+final step, while _auto_reconnect() only clears _reconnecting in its finally.
+If the brand-new socket drops inside that window, the callback fires, sees
+_reconnecting is True, and returns — and because add_done_callback is one-shot
+and the task it was attached to is already done, no future drop can ever be
+detected again. The session is left permanently disconnected with drop
+detection disabled.
+"""
+
+import asyncio
+
+from bubus import BaseEvent, EventBus
+
+from browser_use.browser.session import BrowserSession
+
+
+class NoopEventBus(EventBus):
+	"""EventBus whose dispatch() is a no-op, so no background loop task starts."""
+
+	def dispatch(self, event: BaseEvent) -> BaseEvent:
+		return event
+
+
+class FakeClient:
+	"""Minimal stand-in for CDPClient exposing the message handler task."""
+
+	def __init__(self, task: asyncio.Task | None = None):
+		self._message_handler_task = task
+
+
+def _make_session() -> BrowserSession:
+	s = BrowserSession(cdp_url='ws://127.0.0.1:9222/x')
+	s._intentional_stop = False
+	# Don't start the real bubus event loop: its background loop task never
+	# completes on its own and would hang asyncio teardown in tests.
+	s.event_bus = NoopEventBus()
+	return s
+
+
+async def test_drop_during_reconnect_window_records_flag():
+	"""A drop consumed by the one-shot callback while reconnecting is recorded."""
+	s = _make_session()
+	# Simulate _auto_reconnect() having entered its try but not yet reached
+	# its finally, so _reconnecting is still True (step 8 of reconnect()).
+	s._reconnecting = True
+
+	loop = asyncio.get_running_loop()
+	fut = loop.create_future()
+	task = asyncio.ensure_future(fut)
+	s._cdp_client_root = FakeClient(task)
+	s._attach_ws_drop_callback()
+
+	# The brand-new WebSocket drops inside the window.
+	fut.set_exception(ConnectionResetError('ws dropped again'))
+	await asyncio.sleep(0.05)
+
+	assert s._reconnect_pending is True
+
+
+async def test_auto_reconnect_loops_after_drop_in_window(monkeypatch):
+	"""_auto_reconnect() re-runs when a drop landed inside the reconnect window."""
+	s = _make_session()
+	call_count = 0
+
+	async def fake_reconnect(self: BrowserSession) -> None:
+		nonlocal call_count
+		call_count += 1
+		loop = asyncio.get_running_loop()
+		fut = loop.create_future()
+		task = asyncio.ensure_future(fut)
+		self._cdp_client_root = FakeClient(task)
+		if call_count == 1:
+			# First pass: the new socket dies inside the reconnect window.
+			self._attach_ws_drop_callback()
+			fut.set_exception(ConnectionResetError('ws dropped again'))
+		else:
+			# Second pass: healthy socket — mark the task done before attach so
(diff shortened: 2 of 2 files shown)


browser_use/browser/session.py
```diff
@@ -2294,6 +2295,15 @@ async def _auto_reconnect(self, max_attempts: int = 3) -> None:
...
 			self._reconnect_event.set()  # wake up all waiters regardless of outcome
 
+			# A drop landed inside the reconnect window: the one-shot drop
+			# callback was consumed while _reconnecting was still True, so this
+			# new socket can never be drop-detected again. Re-run the reconnect
+			# loop to close the gap (issue #5366).
+			if self._drop_while_reconnecting and not self._intentional_stop and self.cdp_url:
+				self._drop_while_reconnecting = False
+				self.logger.warning('🔌 CDP WebSocket dropped during reconnect window — retrying')
+				self._reconnect_task = asyncio.get_running_loop().create_task(self._auto_reconnect())
```
harsh4vardhan: Nice fix, this is a clean solution to the race condition.

One edge case to be aware of: if the new socket drops inside the reconnect window again during this retry, _drop_while_reconnecting gets set a second time and the finally block schedules another _auto_reconnect() — which could loop indefinitely on a badly flapping proxy.

async def _auto_reconnect(self, max_attempts: int = 3, _drop_retrigger_budget: int = 1) -> None:

Default of 1 means at most one extra round of retries before giving up  same behaviour as now in the common case, just with a ceiling. Feel free to use this or ignore it, just wanted to flag it!

Diwak4r:

PR #5388, opened 2026-08-04
docs: refresh the Open Source vs Cloud plot with Cloud v4

Alezander9:
Image-only change. Copies the regenerated `accuracy_by_model_{light,dark}.png`
from browser-use/benchmark#24 into `static/`.

## Why

The committed copies predate 2026-05-01. Their bars — `OSS + BU LLM`
(ChatBrowserUse-2), `claude-opus-4-6`, `gpt-5`, `gpt-5-mini`, `gemini-2.5-flash`
— were all retired to `benchmark/old_results/` months ago, and the Cloud bar was
API v3 `bu-ultra`. The section is titled "Open Source vs Cloud" and was
comparing current Cloud against a two-generation-old open-source lineup.

## What it shows now

| | score | cost/100 tasks |
|---|---|---|
| Cloud v4, Opus 4.8 | **85.0%** | $112.14 |
| Cloud v4, default (`gpt-5.6-luna` @ xhigh) | **78.0%** | **$6.31** |
| Open-source library, best (`claude-opus-4-7`) | 74.0% | |
| ...then bu-2-0 68.0, gpt-5.5 66.0, gemini-3.5-flash 65.0, qwen3.6-plus 45.0, deepseek-v4-flash-0731 37.0, gpt-5.6-luna (xhigh) 35.0 | | |

Open-source bars are `browser-use` 0.13.7 on cloud browsers. Every bar is a
fresh BU_Bench_V1 run from 2026-08-04 at a uniform `task_timeout=3600`, one seed
of 100 tasks each (~2pp empirical per-run variance on this benchmark).

The strongest line to draw from it: the `gpt-5.6-luna` pair is the same model at
the same reasoning effort, 78.0% through Cloud v4 versus 35.0% through the
library. Same weights, same effort, same browsers — the 43pp is the harness. And
that is the cheapest cloud configuration, and the one a user gets by default.

`bcode`/BrowserCode and Claude Code were removed from this plot; they have their
own (`browser_harness_by_model` and `best_of_frameworks_public`). Cloud v4 runs
bcode internally, so charting both charted one agent twice.

Full methodology, per-bar eval IDs, backfill accounting, and one caveat about
`gemini-3.5-flash` still hitting the wall clock: browser-use/benchmark#24.

---
## Summary by cubic
Updated the "Open Source vs Cloud" chart to Cloud v4 with the 2026-08-04 benchmark by replacing accuracy_by_model_{light,dark}.png. It now shows Cloud v4 (Opus 4.8 85.0%, default gpt-5.6-luna at xhigh 78.0%) vs current open-source, removes retired models, and plots the library’s Luna at its out-of-box default 31.0% (was 35.0% xhigh).

<sup>Written for commit 0a9166af02972a54fde1402b0411e119e3a2d35d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5388?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Alezander9:
## Hold before merging — re-verifying the luna OSS arm

The `gpt-5.6-luna (xhigh)` bar at 35.0%, and the "43pp is the harness"
comparison I built on it, are being re-checked. Please don't merge until this
resolves.

What I have ruled out as causes of the 35.0%:

- **Wall clock.** Zero of 100 datapoints came near the 3600s cap (max 1659s).
- **Infra errors.** All 100 datapoints were judged; no executor errors, no
  rate-limit losses.
- **`reasoning_effort` not being applied.** `ChatOpenAI` gates it on a substring
  match against `reasoning_models`, and `'gpt-5' in 'gpt-5.6-luna'` is true, so
  `xhigh` was genuinely sent and `temperature`/`frequency_penalty` were dropped.
- **Output truncation.** `ChatOpenAI` pins `max_completion_tokens=4096` and never
  lifts it for reasoning models, and a truncated response returns empty content
  with no `stop_reason` check — so this looked like the culprit. It is not: I
  reproduced the real `AgentOutput` schema against a ~15k-token DOM at all four
  efforts and every call returned valid actions well inside the cap.
- **Reproducibility.** Two independent xhigh runs landed at 35.0% and 32.0%.

I also have to withdraw one piece of evidence from the PR discussion: the
per-datapoint token counts are not comparable across providers. Anthropic and
Gemini runs report 382k/860k cumulative input tokens while every OpenAI-path run
reports 1.8k–7k. That is an instrumentation asymmetry, not agent behaviour, and
nothing should be concluded from those columns.

What the failures actually show: **41 of 65 (63%) are the agent voluntarily
calling `done(success=false)`** with text like "Research is incomplete",
"within the available interaction budget", "in this session" — 15 of them within
8 steps, one at 2 steps, against a ~100 step cap. So this is a real behavioural
pathology, not a crash.

The open question, raised by another dev who reports that xhigh is *worse* than
default effort for luna in this framework, is whether 35.0% is luna's ceiling in
this harness or an arbitrary point on the effort curve. Only `low` (26.0%) and
`xhigh` (35.0%/32.0%) had ever been measured; note those two disagree with the
dev's direction. I have dispatched the missing `medium` and `high` arms at 100
tasks each under the identical config.

If a middle effort scores materially higher, then quoting 35.0% as "the
open-source library's luna number" understates the library and inflates the
Cloud-vs-OSS gap, and both this PR and browser-use#5388 need their headline
comparison rebuilt. I will report the sweep either way.

Alezander9:
## Resolved: 35.0% holds up. The effort sweep is flat, and the cause is identified.

Full sweep on BU_Bench_V1, browser-use 0.13.7, browser-use-cloud,
`task_timeout=3600`, 100 tasks per arm — everything except `reasoning_effort`
held constant:

| effort | score | eval | earlier run |
|---|---|---|---|
| `low` (browser-use default) | 31.0% | `9abf331b` | 26.0% (`adf6e525`) |
| `medium` | **39.0%** | `17732129` | — |
| `high` | 34.0% | `e89a24b7` | — |
| `xhigh` (published bar, matches Cloud) | 35.0% | `963d81f3` | 32.0% (`5e362d5c`) |
| — Cloud v4, same model, same effort | **78.0%** | `00d13a70` | — |

**The curve is flat.** The whole sweep spans 31–39%, and two independent runs of
the *identical* `low` config differ by 5pp (26% vs 31%). The between-effort
spread is no larger than the within-config noise, so `xhigh` at 35.0% is not a
cherry-picked worst case and no effort setting rescues luna in this framework.

### The dev's claim does not reproduce

The report was that `xhigh` is *much worse* than default effort for luna in
browser-use. Measured, it is the opposite direction: `xhigh` 35.0% vs the
browser-use default (`low`) 31.0%, and 26.0% on `low`'s other run. `medium` is
nominally the best arm at 39.0%, but it is only 4pp over `xhigh` — inside the
5pp noise floor. Worth flagging in case they were on a different browser-use
version, a different model alias, or comparing against the *OpenAI* API default
rather than browser-use's, which is `low` (`browser_use/llm/openai/chat.py:39`).

### What actually causes the gap

Not effort — a prompt interaction. At **every** effort, 77–85% of luna's
failures are the agent voluntarily calling `done(success=false)`, and 20–42 per
run do it within 8 steps of a **100**-step cap, with text like "within the
available interaction budget" and "in this session". luna on Cloud v4 bails on
**0%** of its 22 failures (median 39 steps).

| config | bail % of failures | fast bails (≤8 steps) | median steps on failure |
|---|---|---|---|
| Cloud v4 luna `xhigh` | **0%** | 0 | 39 |
| OSS claude-opus-4-7 | 42% | 1 | 16.5 |
| OSS luna `low` | 83% | 42 | 5 |
| OSS luna `medium` | 77% | 23 | 9 |
| OSS luna `high` | 85% | 27 | 9 |
| OSS luna `xhigh` | 85% | 20 | 12 |

The trigger is `system_prompts/system_prompt.md:126-130`: *"When you reach 75% of
your step budget, critically evaluate whether you can complete the full task…
when you do call `done` (at max_steps or earlier), you have meaningful partial
results."* luna over-applies it. It is not missing information — it is told its
exact position every step via `Step{n} maximum:{max_steps}`
(`browser_use/agent/prompts.py:322`) and quits at step 3 of 100 anyway. This is
a genuine library-side behaviour worth fixing independently of this PR.

### Ruled out as causes

- **Wall clock** — 0/100 near the 3600s cap on any arm (maxima 332/887/807/1659s).
- **Infra errors** — all 100 datapoints judged on every arm, no lost tasks.
- **Effort not applied** — `ChatOpenAI` gates `reasoning_effort` on a substring
  match, and `'gpt-5' in 'gpt-5.6-luna'` is true, so it was sent and
  `temperature`/`frequency_penalty` were correctly dropped.
- **Output truncation** — `ChatOpenAI` pins `max_completion_tokens=4096` and never
  lifts it for reasoning models, and a length-truncated response returns empty
  content with `stop_reason` unchecked. This was my leading hypothesis and it is
  **wrong here**: replaying the real `AgentOutput` schema against a ~15k-token DOM
  at all four efforts returned valid actions every time, well inside the cap.
  (Still a latent hazard for more verbose reasoning models — filed separately.)

### Corrections to this PR

- Variance in the body corrected from ~2pp to **~5pp**, with the paired runs cited.
- The headline gap is **39pp**, not 43pp, if measured against luna's best OSS arm
  (`medium` 39.0%) rather than the effort-matched one. Either way the conclusion
  stands, and it is now mechanistically explained rather than asserted.
- One piece of evidence **withdrawn**: per-datapoint token counts are not
  comparable across providers (Anthropic/Gemini report 382k/860k cumulative input
  while every OpenAI-path run reports 1.8k–7k). That is instrumentation, not
  behaviour. Nothing in this PR rests on it.

The bar stays `xhigh` because that is what Cloud v4 serves, which keeps it a
controlled comparison; the sweep above is the audit trail showing the choice
costs the library at most 4pp. Happy to swap it to `medium` (39.0%) if reviewers
prefer the library's best-case arm over the effort-matched one.

PR #5398, opened 2026-08-05
fix: preserve CHROME_DEFAULT_ARGS order when ignore_default_args is a list

### LINKED ISSUE #5397: Bug: ...BrowserProfile.get_args() loses default argument ordering when ignore_default_args is a list
### Browser Use Version

a3e3cc5

BrowserProfile.get_args() converts CHROME_DEFAULT_ARGS to a set when
ignore_default_args is provided as a list:

    default_args = set(CHROME_DEFAULT_ARGS) - set(self.ignore_default_args)

This loses the original ordering of CHROME_DEFAULT_ARGS because set iteration
order can vary between Python processes/hash seeds.

Steps to reproduce:

1. Create a BrowserProfile with ignore_default_args specified as a list.
2. Call get_args() in separate Python processes with different PYTHONHASHSEED values.
3. Compare the ordering of the remaining default Chrome arguments.

Expected behavior:

Ignoring one or more default arguments should remove those arguments while
preserving the order of the remaining entries in CHROME_DEFAULT_ARGS.

Actual behavior:

The remaining default arguments are converted to a set, so their original
ordering is lost.

Proposed fix:

    ignored_args = set(self.ignore_default_args)
    default_args = [
        arg for arg in CHROME_DEFAULT_ARGS
        if arg not in ignored_args
    ]

### Failing Python Code

```python

profile = BrowserProfile(
    ignore_default_args=["--disable-popup-blocking"]
)

args = profile.get_args()

for i, arg in enumerate(args):
    print(i, arg)
```

### LLM Model

_No response_

Ubuntu 22.04.3 LTS (Jammy Jellyfish), Python 3.10.12

### Full DEBUG Log Output

```shell

```

michaelxer:
## Summary
`BrowserProfile.get_args()` used set subtraction when `ignore_default_args` was a list. That drops ignored flags correctly, but set iteration order is not stable across processes / `PYTHONHASHSEED`, so the remaining default Chrome CLI flags could be reordered.

## Change
Filter `CHROME_DEFAULT_ARGS` with a list comprehension so ignored entries are removed while the original order is kept.

## Reproduction
1. Build a profile with `ignore_default_args=['--disable-popup-blocking']`.
2. Call `get_args()` under different `PYTHONHASHSEED` values.
3. Before this change, the remaining default-arg order could differ between seeds.
4. After this change, the remaining defaults keep `CHROME_DEFAULT_ARGS` order.

## Verification
Ran focused local checks against this branch:

- order-preserving filter with two ignored defaults
- stability across `PYTHONHASHSEED` in `{0,1,42,12345}`
- reproduced the old set-subtraction path producing 4 different orders across those seeds
- existing disable-security / `--disable-features` merge behavior still looks correct

Commands used:

```bash
PYTHONPATH=. python3 - <<'PY'
# focused order + hash-seed checks (see tests/ci/browser/test_get_args_default_order.py)
PY
```

Also added `tests/ci/browser/test_get_args_default_order.py`.

## Notes
Scope is only the list-ignore path in `get_args()`.

Fixes #5397

---
## Summary by cubic
Preserves the order of Chrome default launch flags in `BrowserProfile.get_args()` when `ignore_default_args` is a list by filtering `CHROME_DEFAULT_ARGS` in order instead of using set subtraction. Prevents nondeterministic reordering across processes and `PYTHONHASHSEED` values, with a regression test to lock the expected order (fixes #5397).

<sup>Written for commit 9343d30a1959e79e47606732d6e1102b1a258eea. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5398?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/browser/profile.py
+++ b/browser_use/browser/profile.py
@@ -896,7 +896,10 @@ class BrowserProfile(BrowserConnectArgs, BrowserLaunchPersistentContextArgs, Bro
 		"""Get the list of all Chrome CLI launch args for this profile (compiled from defaults, user-provided, and system-specific)."""
 
 		if isinstance(self.ignore_default_args, list):
-			default_args = set(CHROME_DEFAULT_ARGS) - set(self.ignore_default_args)
+			# Keep CHROME_DEFAULT_ARGS order. Set subtraction is order-unstable
+			# across processes/hash seeds and can shuffle Chrome CLI flags.
+			ignored_args = set(self.ignore_default_args)
+			default_args = [arg for arg in CHROME_DEFAULT_ARGS if arg not in ignored_args]
 		elif self.ignore_default_args is True:
 			default_args = []
 		elif not self.ignore_default_args:
--- a/tests/ci/browser/test_get_args_default_order.py
+++ b/tests/ci/browser/test_get_args_default_order.py
@@ -16,11 +16,7 @@ def _default_args_in_order(args: list[str], ignore: list[str]) -> tuple[list[str
 	original token. We only track whole-token defaults that survive unchanged.
 	"""
 	ignored = set(ignore)
-	expected = [
-		arg
-		for arg in CHROME_DEFAULT_ARGS
-		if arg not in ignored and not arg.startswith('--disable-features=')
-	]
+	expected = [arg for arg in CHROME_DEFAULT_ARGS if arg not in ignored and not arg.startswith('--disable-features=')]
 	# Preserve first-seen order from the final args list, restricted to expected defaults.
 	seen: list[str] = []
 	expected_set = set(expected)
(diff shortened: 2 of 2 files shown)


michaelxer:
Closing this — the same Chrome default-arg order fix landed on main in #5621, so this branch now conflicts. Thanks for taking it.

PR #5400, opened 2026-08-05
fix: stop restoring deleted storage-state cookies

ump45nose:
## Why

`StorageStateWatchdog` merges the current browser state into an existing `storage_state.json` file. The current implementation also retains cookies that are present only in the old file.

`Storage.getCookies` returns the complete browser cookie snapshot, so a cookie missing from that result has been removed from the browser. Keeping the old value can restore a deleted session cookie on the next Browser Use startup, including after a logout.

## What changed

- Treat the current CDP cookie list as authoritative when saving storage state.
- Keep the existing origin merge behavior unchanged.
- Add a regression test covering removal of an old session cookie while preserving origin state.

## Validation

- `pytest -q tests/ci/test_storage_state_watchdog.py`
- `ruff check browser_use/browser/watchdogs/storage_state_watchdog.py tests/ci/test_storage_state_watchdog.py`
- `git diff --check`

---
## Summary by cubic
Stops restoring deleted cookies when saving browser storage state. Previously, cookies present only in the old `storage_state.json` were retained; now `Storage.getCookies` is the authoritative source, so cookies removed from the browser are not resurrected on the next startup, including after logout.

- Preserves existing origin merge behavior; existing origins remain unless overridden by the new state.
- Adds regression coverage confirming deleted session cookies are not restored and origins are preserved.

<sup>Written for commit 43802b922f65b5ba5dc22671b574d17426745b4e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5400?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/browser/watchdogs/storage_state_watchdog.py
+++ b/browser_use/browser/watchdogs/storage_state_watchdog.py
@@ -325,14 +325,9 @@ class StorageStateWatchdog(BaseWatchdog):
 		"""Merge two storage states, with new values taking precedence."""
 		merged = existing.copy()
 
-		# Merge cookies
-		existing_cookies = {(c['name'], c['domain'], c['path']): c for c in existing.get('cookies', [])}
-
-		for cookie in new.get('cookies', []):
-			key = (cookie['name'], cookie['domain'], cookie['path'])
-			existing_cookies[key] = cookie
-
-		merged['cookies'] = list(existing_cookies.values())
+		# Storage.getCookies returns the browser's complete cookie snapshot. Treat it
+		# as authoritative so deleted cookies are not resurrected from the old file.
+		merged['cookies'] = list(new.get('cookies', []))
 
 		# Merge origins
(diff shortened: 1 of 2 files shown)


ump45nose:
Resolved the merge conflict with `main`. The conflict was an add/add on `tests/ci/test_storage_state_watchdog.py` — upstream landed UTF-8 storage-state tests at the same path. I combined both sets of tests (the deleted-cookie regression test plus upstream's UTF-8 tests) and kept the `_merge_storage_states` cookie change intact.

Issue #4964, opened 2026-06-04
  Feature Request: Add hover action for triggering CSS :hover dropdowns, tooltips, and hover-reveal patterns


Kie-Chi:
### What is the problem that your feature request solves?

The Agent currently has no dedicated hover action. The only mechanism available to the LLM for triggering hover interactions is `evaluate()` — executing arbitrary JavaScript like `dispatchEvent(new MouseEvent('mouseenter'))`. This approach is fundamentally broken for some hover scenarios:

1. **CSS `:hover` pseudo-class never activates** — `dispatchEvent` fires synthetic DOM events but does not update the browser's internal mouse position. Most dropdown menus, tooltips, and reveal-on-hover patterns rely on `:hover`.

2. **Missing `mousemove` cascade** — A real hover produces `mouseenter → mouseover → mousemove` with correct `screenX/screenY` coordinates. 

3. **LLM context overhead** — The LLM must compose correct JS for each hover target on the fly, consuming tokens and introducing error-prone ad-hoc scripting where a single `hover(index=N)` call suffices. Failed JS attempts trigger retry loops or human intervention.

Consider an Agent tasked with deleting a chat session on a modern LLM chat interface (Gemini, ChatGPT, etc.). The sidebar hides the "More Actions" button (⋯) by default — it only appears via CSS `:hover` on the session row. The Agent cannot see or click the button because `evaluate()` cannot activate `:hover`, so the element stays invisible. The Agent then retries with different JS patterns or escalates to human help, neither of which succeeds.

### What is your proposed solution?

Add a `HoverAction` to the Tools action registry that uses CDP `Input.dispatchMouseEvent` with type `mouseMoved` — the same mechanism the browser uses for real mouse movement. This triggers genuine CSS `:hover` state transitions and the full DOM event cascade.

Two modes, consistent with the existing `ClickElementAction` pattern:

| Mode | Syntax | Behavior |
|------|--------|----------|
| By index | `hover(index=N)` | Resolve element → scroll into view → compute bounding box center → `mouseMoved` |
| By coordinates | `hover(coordinate_x=X, coordinate_y=Y)` | LLM-to-viewport coordinate conversion → `mouseMoved` at target point |

[...]

PR #5403, opened 2026-08-06
feat: add hover action for CSS :hover / JS mouseenter reveal (#4964)

ademicho123:
Add hover action for CSS :hover / JS mouseenter reveal

Adds a hover tool that moves the mouse over an element by index without clicking, so the agent can reach dropdown menus, tooltips, and other hover-revealed UI that today's click action can't trigger and that isn't even visible in the DOM until hovered (#4964).

Implementation reuses the existing scroll-into-view and coordinate logic from click, then dispatches a single CDP mouseMoved event — deliberately stops short of press/release so it can never accidentally click.

Tested against a real page with a CSS-only dropdown and a JS mouseenter listener; both fire correctly. Full test suite passes.

Before:
hover_before

After:
hover_after

---
## Summary by cubic
Adds a `hover` action that moves the mouse over an element by index without clicking to reveal CSS :hover and JS mouseenter UI. Previously, click couldn’t trigger hover-only dropdowns or tooltips; now a single CDP `mouseMoved` fires handlers, never presses/releases, and the action terminates the sequence.

- **Details for review**
  - Introduces `HoverElementEvent` and `HoverElementAction`; rebuilds models, scrolls into view, centers, clamps to viewport, and returns `{ x, y }` metadata.
  - Registers `hover` in `Tools` with `terminates_sequence=true`; highlights the target and returns helpful messages for stale indices or validation errors for unknown coordinates.
  - Separates hover from click handling so hover never routes through click paths.
  - Adds tests to verify CSS dropdown reveal, JS `mouseenter` firing, and invalid index behavior.

<sup>Written for commit e9fd8450868a46fbaf4ebc9688443169ebf03926. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5403?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/browser/events.py
+++ b/browser_use/browser/events.py
@@ -138,6 +138,15 @@ class ClickCoordinateEvent(BaseEvent[dict]):
 
 	coordinate_x: int
 	coordinate_y: int
+
+
+class HoverElementEvent(ElementSelectedEvent[dict[str, Any] | None]):
+	"""Move the mouse over an element without clicking, to trigger CSS :hover states and
+	JS mouseenter/mouseover listeners (dropdown menus, tooltips, hover-reveal UI)."""
+
+	node: 'EnhancedDOMTreeNode'
+
+	event_timeout: float | None = Field(default_factory=lambda: _get_timeout('TIMEOUT_HoverElementEvent', 15.0))  # seconds
 	button: Literal['left', 'right', 'middle'] = 'left'
 	force: bool = False  # If True, skip safety checks (file input, print, select)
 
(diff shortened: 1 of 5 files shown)


ademicho123:
@cubic-dev-ai

PR #5413, opened 2026-08-08
fix(llm/google): Gemini thought summaries were never requested or surfaced

### LINKED ISSUE #5128: beta (Rust terminal): Gemini thought text never available — thinkingConfig.includeThoughts not set, reasoning tokens billed anyway
### Versions
- browser-use 0.13.3, `browser_use.beta.Agent` (Rust terminal path)
- browser-use-core 0.13.2 (`browser-use-terminal sdk-server --transport stdio`)
- Provider: google, model `gemini-3.5-flash`

### What happens
When running the beta (Rust terminal) agent with a Gemini model, thought/reasoning **text** is never available anywhere — not in the SDK event stream (`model.thinking_delta` never fires) and not in the reconstructed `AgentHistoryList` (`model_thoughts()` is empty).

Meanwhile Gemini **does** reason and bills for it: every `token_count` event carries a nonzero reasoning/thoughts token count (we observed 120–185 `thoughtsTokenCount` per turn).

### Root cause (verified at the HTTP layer)
We pointed `LLM_BROWSER_GOOGLE_BASE_URL` at a local capture proxy and inspected the raw `generateContent` requests: the Rust client never sets `thinkingConfig.includeThoughts`, so Gemini reasons internally and bills the tokens, but the thought text is never returned in the response at all. It isn't a wrapper/event-forwarding gap — the data doesn't exist on the wire.

We also tried `config_overrides={"reasoning_effort": "high"}` via the JSON-RPC layer: accepted, but no behavior change (no `model.thinking_delta`, reasoning token counts unchanged in effect).

### Expected
Either:
- set `thinkingConfig.includeThoughts` (ideally behind a config flag / honoring `reasoning_effort`), and emit the thought text as `model.thinking_delta` events; or
- document that thought text is unavailable on the Gemini path so downstream tracing/observability tooling doesn't wait for it.

### Impact
Anyone building tracing/evals on the beta agent loses per-step reasoning visibility on Gemini — you pay for reasoning tokens but can never see or log the reasoning.

TevBenji:
Fixes #5128

## Summary

Gemini reasons and bills `thoughtsTokenCount` on every thinking-enabled call, but the thought text was never available anywhere — browser-use never asked for it and never read it. This PR requests thought summaries, surfaces them on `ChatInvokeCompletion.thinking` (parity with the Anthropic and LiteLLM paths), and bridges provider reasoning into `AgentOutput.thinking` so `history.model_thoughts()` has data when the model omitted the structured thinking field.

## Root cause

Two independent gaps:

1. `browser_use/llm/google/chat.py` built `thinking_config` in all three model branches (Gemini 3 Pro / 3 Flash / 2.5) without ever setting `include_thoughts`, so Gemini never returned thought text on the wire — while `_get_usage` already counted `thoughts_token_count` into billed completion tokens. None of the five `ChatInvokeCompletion(...)` returns passed `thinking=`.
2. `ChatInvokeCompletion.thinking` was write-only repo-wide: `Agent.get_model_output` dropped the envelope after taking `.completion`, and `AgentHistoryList.model_thoughts()` reads `AgentOutput.thinking` (the structured-output JSON field). Even Anthropic/LiteLLM reasoning was populated into a void.

## Fix

**Commit 1 — `fix(llm/google): request and surface Gemini thought summaries`**

- New `include_thoughts: bool = True` field on `ChatGoogle` (opt-out knob; thought tokens are billed regardless of whether summaries are returned, so the default costs nothing extra)
- `include_thoughts` passed in all four `ThinkingConfigDict(...)` constructions; not requested when `thinking_budget=0` (thinking disabled); gemini-2.0/gemma paths untouched (they get no thinking_config)
- `_extract_thinking()` joins `part.thought` parts from `candidates[0].content.parts` — the SDK's `response.text` excludes thought parts, so completions and structured-output JSON parsing are unaffected (runtime-asserted in tests against real `google.genai.types` objects)
- `thinking=` threaded into all five return sites (text, structured re-parse, both native-parsed, fallback JSON)

**Commit 2 — `fix(agent): surface provider reasoning in AgentOutput.thinking when model omits it`**

Small bridge in `get_model_output`: if the provider envelope carries `thinking` and the model omitted the structured field, surface it on `AgentOutput.thinking`. The model-authored field always wins, and the envelope is read with `getattr` since duck-typed ainvoke results may only carry `completion` and `usage`. Mirrors what the beta path already does for streamed thinking (`beta/service.py`). No prompt-size impact: message history (`HistoryItem`) never includes thinking; it reaches only the history dump and `model_thoughts()`.

## Tests

All in `tests/ci/models/test_llm_google.py`, written failing-first:

- `test_thinking_config_requests_thought_summaries` — all three thinking branches request summaries
- `test_include_thoughts_opt_out_and_disabled_thinking` — `include_thoughts=False` opt-out; `thinking_budget=0` doesn't request; gemini-2.0 stays config-free
- `test_ainvoke_surfaces_thought_parts_in_thinking` — real `google.genai.types` response: thought lands on `.thinking`, completion text unpolluted
- `test_ainvoke_structured_output_surfaces_thinking` — structured path parses JSON and surfaces the summary
- `test_get_model_output_bridges_provider_thinking` / `test_get_model_output_keeps_model_authored_thinking` — the bridge and its precedence

Verified with `uv run pytest tests/ci/models/test_llm_google.py` (11 passed), `uv run pytest tests/ci/test_fallback_llm.py` (15 passed — regression surface for `get_model_output`), `uv run ruff check` / `ruff format`, and `uv run pyright` (no new errors).

Note on the issue's exact repro: it was hit via the beta Rust terminal (`browser-use-terminal`), which is a separate codebase and may need the same `thinkingConfig.includeThoughts` fix on its side. This PR fixes the identical gap in the python library.


TevBenji (commented):
Synced with latest main — no conflicts, and the merge doesn't touch anything in this PR's diff. Ready for review whenever you have time.

TevBenji:
The infrastructure/test_registry_core failure looks unrelated to this PR, TestActionToActionCalling::test_complex_action_chain asserted on /test but the page was still about:blank, which is a navigation race rather than anything this diff touches (llm/google/chat.py, agent/service.py, and the Google test file only).

It passes locally, the same job passed on the previous commit of this branch, and it's green on main. I pushed an empty commit to re-trigger CI happy to rebase or dig further if it reproduces.

PR #5416, opened 2026-08-08
fix(browser): resolve user_data_dir default via validate_default

### LINKED ISSUE #5414: Bug: BrowserProfile.user_data_dir field_validator is skipped when the field is omitted (pydantic v2 default validation)
### Browser Use Version

```python
@classmethod
    if v is None:
```

Steps to reproduce:

Expected behavior:

Actual behavior:

### Failing Python Code

```python

```

```python

class Profile(BaseModel):

    @classmethod

```

### LLM Model

_No response_

### Full DEBUG Log Output

```shell
```

axisrow:
Fixes #5414

## Problem

`BrowserProfile.user_data_dir` has a `field_validator('user_data_dir', mode='after')` that converts `None` into a generated temp directory path:

```python
@classmethod
    if v is None:
```

pydantic v2 does not run `field_validator` for a field left at its default (only when explicitly assigned), unless the model sets `validate_default=True`. `BrowserLaunchPersistentContextArgs`'s `model_config` didn't set it, so this validator was silently skipped whenever `user_data_dir` was omitted entirely — e.g. `BrowserProfile(headless=True)`, a very common construction pattern used throughout the test suite (`tests/ci/browser/test_screenshot.py:54`, and elsewhere) and in user code.

```python
BrowserProfile(user_data_dir=None).user_data_dir   # resolved temp path
BrowserProfile(headless=True).user_data_dir         # None  <-- validator skipped, bug
```

## Fix

Add `validate_default=True` to `BrowserLaunchPersistentContextArgs.model_config`. Both construction styles now resolve identically, as expected — `None` is both the explicit value and the field's own default, so they should never behave differently.

## Testing

- `tests/ci/test_profile_user_data_dir_default_validation.py` (new) — covers both the omitted and explicit-`None` cases, and that each construction gets its own distinct temp dir.
- Full `uv run pytest -q tests/ci` run clean (only the pre-existing, order-dependent `test_beta_agent_constructor_type_hints_match_browser_use_core_params` flake reproduces — confirmed present on unmodified `main` too, unrelated to this change; caused by `importlib.reload()` in `test_action_timeout.py` leaking a new `Tools` class object across test files).
- `uv run pyright` — 0 errors.
- `uv run ruff check` / `ruff format` — clean.

---
## Summary by cubic
Resolve `BrowserProfile.user_data_dir` lazily in `get_args()` so omitted or `None` values become a unique temp directory without import-time side effects. Fixes #5414.

- **Bug Fixes**
  - Generate a temp path for `user_data_dir` in `get_args()` when it's `None` or omitted, avoiding `pydantic` construction-time validation.
  - Keep the module-level `DEFAULT_BROWSER_PROFILE` side-effect free and ensure `model_copy()` instances get unique dirs; added regression tests.

<sup>Written for commit d0eab36937b32da276d657e08b10fc46f9e96371. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5416?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/browser/profile.py
+++ b/browser_use/browser/profile.py
@@ -540,10 +540,7 @@ class BrowserLaunchPersistentContextArgs(BrowserLaunchArgs, BrowserContextArgs):
 	https://playwright.dev/python/docs/api/class-browsertype#browser-type-launch-persistent-context
 	"""
 
-	# validate_default=True: user_data_dir's field_validator must also run when the field is left at its
-	# default (omitted entirely), not just when explicitly passed as None — otherwise BrowserProfile(...)
-	# without user_data_dir= leaves self.user_data_dir as raw None instead of the resolved temp path.
-	model_config = ConfigDict(extra='ignore', validate_assignment=False, revalidate_instances='always', validate_default=True)
+	model_config = ConfigDict(extra='ignore', validate_assignment=False, revalidate_instances='always')
 
 	# Required parameter specific to launch_persistent_context, but can be None to use incognito temp dir
 	user_data_dir: str | Path | None = None
@@ -905,7 +902,11 @@ class BrowserProfile(BrowserConnectArgs, BrowserLaunchPersistentContextArgs, Bro
 		elif not self.ignore_default_args:
 			default_args = CHROME_DEFAULT_ARGS
 
-		assert self.user_data_dir is not None, 'user_data_dir must be set to a non-default path'
+		if self.user_data_dir is None:
+			# Resolved lazily here (not via a field_validator/validate_default on construction) so that
+			# constructing a BrowserProfile — including the module-level DEFAULT_BROWSER_PROFILE singleton
+			# in browser_use/browser/session.py — never has the filesystem side effect of tempfile.mkdtemp().
+			self.user_data_dir = Path(tempfile.mkdtemp(prefix='browser-use-user-data-dir-'))
 
 		# Capture args before conversion for logging
 		pre_conversion_args = [
--- a/browser_use/browser/profile.py
+++ b/browser_use/browser/profile.py
@@ -540,7 +540,10 @@ class BrowserLaunchPersistentContextArgs(BrowserLaunchArgs, BrowserContextArgs):
 	https://playwright.dev/python/docs/api/class-browsertype#browser-type-launch-persistent-context
 	"""
 
-	model_config = ConfigDict(extra='ignore', validate_assignment=False, revalidate_instances='always')
+	# validate_default=True: user_data_dir's field_validator must also run when the field is left at its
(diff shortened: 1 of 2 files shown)


axisrow:
## Review cycle summary

Two review rounds run against this PR (local `/review` + Codex adversarial companion each round).

**Round 1** (against d0eab369's predecessor, `validate_default=True` approach): Codex flagged that `validate_default=True` on `BrowserLaunchPersistentContextArgs.model_config` would make the module-level `DEFAULT_BROWSER_PROFILE = BrowserProfile()` singleton (`browser_use/browser/session.py:66`) eagerly call `tempfile.mkdtemp()` on every `import browser_use`, and that `model_copy()` of that singleton doesn't re-run validation, so multiple sessions using the default profile would share one `user_data_dir`.

**Fix**: reverted `validate_default=True`; `user_data_dir` is now resolved lazily inside `BrowserProfile.get_args()` — the one place every real launch path already calls before starting a browser. Added regression tests for both findings (`test_module_level_default_browser_profile_does_not_eagerly_resolve_user_data_dir`, `test_two_default_profile_copies_get_independent_user_data_dirs`).

**Round 2** (against d0eab369): both findings confirmed fixed by both reviewers.
- `/review`: one theoretical concurrency note on mutating `self.user_data_dir` inside `get_args()` — verified against the codebase: the only caller (`LocalBrowserWatchdog._launch_browser`) invokes it once per attempt, sequentially, never concurrently on a shared unresolved profile. Not applicable to any reachable code path; no change made.
- Codex: confirmed both round-1 findings fixed. Separately flagged a **pre-existing** bug (predates this PR, reachable before via explicit `user_data_dir=None`): `LocalBrowserWatchdog` only tracks temp dirs matching prefix `browseruse-tmp-` for cleanup-on-kill, while `get_args()`/the field validator create dirs with prefix `browser-use-user-data-dir-` — these are never cleaned up. This PR makes that leak newly reachable via the *omitted* `user_data_dir` path (previously blocked by an `assert`), widening its blast radius to the most common usage pattern. Out of scope for this PR (unrelated cleanup-path bug, not a `user_data_dir`-default-resolution bug) — filed separately as browser-use/browser-use#5419.

**CI**: all checks green (`gh pr checks 5416`), only `license/cla` pending (contributor action, not code).

No further findings from either reviewer. Ready for maintainer review/merge.

PR #5417, opened 2026-08-08
fix(browser): avoid OS keychain prompts for browser-use-managed profiles

### LINKED ISSUE #5415: Bug: BrowserSession hangs / prompts for OS keychain access on startup without --use-mock-keychain (regression side-effect of #3225)
### Browser Use Version

`--use-mock-keychain` and `--password-store=basic` were originally part of `CHROME_DEFAULT_ARGS` (inherited from Playwright's `chromiumSwitches.ts`, see the comment at `profile.py:155`), but were removed in #3225 ("Remove profile args which prevent cookie usage") because they prevented Chromium from decrypting saved passwords/cookies in a real, pre-existing Chrome profile passed via `user_data_dir` (the `examples/browser/real_browser.py` use case). That fix is valid for that specific scenario.

The unintended side effect: without these flags, on macOS (and Linux with a keyring daemon registered) Chromium's own profile initialization synchronously tries to reach the OS credential store. In any environment without an interactive session to answer that prompt — which includes `headless=True`, CI, and the default `browser-use` temp/managed profile that the vast majority of `BrowserProfile()` calls use — Chromium does not fail fast or fall back; it hangs indefinitely at startup and never opens its CDP port. `LocalBrowserWatchdog.on_BrowserLaunchEvent` then never returns, `BrowserSession.on_BrowserStartEvent` hits its internal 30s handler timeout, and the caller sees an opaque `TimeoutError` with no indication that a credential-store prompt is the cause.

On a machine where the OS does have an interactive Keychain/keyring prompt available (e.g. a local macOS dev machine, not CI), this instead surfaces as a password prompt appearing every time a fresh browser-use profile launches — for a brand-new contributor, this happens on the very first `uv run pytest tests/ci` after cloning the repo, with no explanation of why an automated test suite is asking for their login password.

I verified this is deterministic and specifically caused by the missing flags, independent of any other configuration, by launching the exact Chromium binary browser-use uses with an identical argument list twice, differing only in the presence of these two flags:

- **Without** `--use-mock-keychain --password-store=basic`: process starts, CDP port never opens, `curl http://127.0.0.1:<port>/json/version` gets no response even after 8s.
- **With** the same two flags added: CDP responds within ~1s.

Reproduced this exact flip twice.

Steps to reproduce:

1. On macOS (or Linux with a keyring daemon such as gnome-keyring/kwallet registered as the default secret service), in a fresh environment with no already-unlocked/cached Keychain session for the profile's Safe Storage item:
2. `git clone` the repo, `uv sync`.
3. `uv run pytest -q tests/ci/browser/test_screenshot.py::test_basic_screenshots`
4. Observe either a macOS Keychain password prompt, or (in a non-interactive/CI shell where no prompt can be shown) a hang until the 30s watchdog timeout fires with `TimeoutError`.

Expected behavior:

A fresh, disposable, browser-use-managed profile (the temp dirs used by tests and by the default `BrowserProfile()` call) should never need to touch the OS credential store — there are no saved credentials in it to decrypt. Only a real, user-supplied profile directory (the #3225 scenario) has a legitimate reason to need system keychain access.

Actual behavior:

Every browser-use-managed profile currently behaves like a real user profile with respect to keychain access, because the flags were removed globally rather than conditionally.

### Failing Python Code

```python
from browser_use.browser.session import BrowserSession

# Common construction pattern (see tests/ci/browser/test_screenshot.py:54,
# and many other tests/examples) — no explicit user_data_dir, no real profile involved.
session = BrowserSession(browser_profile=BrowserProfile(headless=True))

# On a machine with an interactive Keychain available: this pops a system
# password prompt on session.start(). In CI / no-display environments: this
# hangs until BrowserSession.on_BrowserStartEvent's internal 30s timeout fires.
import asyncio
asyncio.run(session.start())
```

Minimal non-pytest repro isolating just the Chromium launch behavior (swap in your local playwright chromium path):

```bash
BIN=~/Library/Caches/ms-playwright/chromium-*/chrome-mac/Chromium.app/Contents/MacOS/Chromium
TMP=$(mktemp -d)

# Hangs — CDP never responds:
"$BIN" --headless=new --user-data-dir="$TMP" --remote-debugging-port=9333 &
sleep 8 && curl -s --max-time 3 http://127.0.0.1:9333/json/version || echo "NOT RESPONDING"

# Add the two flags back — responds immediately:
"$BIN" --headless=new --use-mock-keychain --password-store=basic --user-data-dir="$TMP" --remote-debugging-port=9334 &
sleep 8 && curl -s --max-time 3 http://127.0.0.1:9334/json/version && echo "RESPONDED"
```

### LLM Model

_No response_

macOS 26.5.2 + playwright chromium (chromium-1187, Chrome/140.0.7339.16), headless=True

### Full DEBUG Log Output

```shell
tests/ci/browser/test_screenshot.py::test_basic_screenshots WARNING  [bubus] ⚠️ EventBus_...🟢 handler browser_use.browser.watchdog_base.BrowserSession.on_BrowserStartEvent() has been running for >15s on event. Possible slow processing or deadlock.
browser_use.browser.watchdog_base.BrowserSession.on_BrowserStartEvent(?▶ BrowserStartEvent#... 🏃)
WARNING  [bubus] ⚠️ EventBus_...🟢 handler browser_use.browser.watchdog_base.LocalBrowserWatchdog.on_BrowserLaunchEvent() has been running for >15s on event. Possible slow processing or deadlock.
browser_use.browser.watchdog_base.LocalBrowserWatchdog.on_BrowserLaunchEvent(?▶ BrowserLaunchEvent#... 🏃)
WARNING  [bubus] ⏱️  TIMEOUT ERROR - Handling took more than 30.0s for EventBus_....browser_use.browser.watchdog_base.BrowserSession.on_BrowserStartEvent(?▶ BrowserStartEvent#... ✅)
WARNING  [bubus]    ⏰ browser_use.browser.watchdog_base.BrowserSession.on_BrowserStartEvent(#...) ⌛️ 30s/30s   ⬅️ TIMEOUT HERE ⏰
WARNING  [bubus]       ➡️ browser_use.browser.watchdog_base.LocalBrowserWatchdog.on_BrowserLaunchEvent(#...) ⏳ 30s/30s   ⬅️ INTERRUPTED ✂️
ERROR
```

---

Related: #3225 removed these flags for a valid reason (real Chrome profile decryption). This issue isn't asking to blindly revert that — a profile-aware fix (mock keychain for browser-use-managed/temporary profiles, real keychain only for an explicit user-supplied `user_data_dir`) would fix this without reintroducing the problem #3225 solved. Happy to open a PR with that approach if useful.

axisrow:
Fixes #5415

## Problem

#3225 removed `--use-mock-keychain` / `--password-store=basic` from `CHROME_DEFAULT_ARGS` globally, to let Chromium decrypt saved passwords/cookies in a real, user-supplied Chrome profile (the `examples/browser/real_browser.py` scenario). That fix is still correct for that case.

The side effect: every browser-use-managed profile (temp dirs, and the default browser-use profile under `~/.config/browseruse/profiles/default`) now behaves like a real user profile with respect to keychain access too — even though there's nothing saved in it to decrypt. On macOS (or Linux with a keyring daemon registered), Chromium's profile init synchronously tries to reach the OS credential store on startup. In headless/CI environments with no way to answer that prompt, it hangs indefinitely instead of opening its CDP port, until `BrowserSession`'s internal 30s watchdog timeout fires with an opaque `TimeoutError`. On an interactive machine, this instead surfaces as a Keychain password prompt on every fresh profile launch — for a new contributor, on the very first `uv run pytest tests/ci` after cloning, with no explanation why a test suite wants their login password.

Verified this is deterministic and specifically caused by the missing flags: launched the exact Chromium binary browser-use uses with an identical argument list twice, differing only in these two flags — without them the CDP port never opens even after 8s; with them it responds within ~1s. Reproduced twice.

## Fix

Adds `use_system_keychain: bool | None = None` (auto by default), resolved in `model_post_init` before `_copy_profile()` may rewrite `user_data_dir` to a temp copy:
- disabled (mock keychain) for profiles fully managed by browser-use — temp dirs and the default browser-use profile dir
- enabled (real keychain) for a real, user-supplied `user_data_dir` — preserving the #3225 behavior for that scenario

Can be overridden explicitly in either direction via the new constructor kwarg.

The auto-detection explicitly handles `user_data_dir is None` rather than relying on its `field_validator` having already resolved it, so this fix does not depend on #5416 / the `validate_default` fix and can be merged independently in either order.

## Testing

- `tests/ci/test_browser_profile_keychain.py` (new) — covers temp profiles, the default browser-use profile, a real user-supplied profile (must keep system keychain, per #3225), explicit overrides in both directions, and the `user_data_dir` omitted case.
- `tests/ci/conftest.py` — the shared `browser_session` fixture now passes `use_system_keychain=False` explicitly, so test hermeticity doesn't depend solely on the auto-detection heuristic.
- Full `uv run pytest -q tests/ci` run clean, no keychain hangs/prompts (only the pre-existing, order-dependent `test_beta_agent_constructor_type_hints_match_browser_use_core_params` flake reproduces — confirmed present on unmodified `main` too, unrelated to this change).
- `uv run pyright` — 0 errors.
- `uv run ruff check` / `ruff format` — clean.

---
## Summary by cubic
Prevents OS keychain prompts and CI hangs by disabling the system keychain for browser-use-managed profiles (including empty temp dirs) while keeping it for real user profiles (fixes #5415). Adds a `use_system_keychain` option to override.

- **Bug Fixes**
  - Auto-detects temp, default, and any empty dir under the OS temp as browser-use-managed, and decides before any profile copy so copied real Chrome profiles keep system keychain enabled.
  - Passes `--use-mock-keychain` and `--password-store=basic` only when system keychain is disabled; leaves it enabled for real `user_data_dir` paths so saved logins/cookies stay readable.
  - Adds regression tests for caller-supplied empty temp dirs, decision timing vs `_copy_profile()`, and `model_post_init` vs validator ordering with a non-default channel; CI fixture sets `use_system_keychain=False`.

- **New Features**
  - New constructor arg `use_system_keychain: bool | None` (default auto) to explicitly enable or disable the system keychain.

<sup>Written for commit 961d39a966f8fd198076e88f76576b2e011fe33d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5417?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/browser/profile.py
+++ b/browser_use/browser/profile.py
@@ -702,6 +702,17 @@ class BrowserProfile(BrowserConnectArgs, BrowserLaunchPersistentContextArgs, Bro
 
 	profile_directory: str = 'Default'  # e.g. 'Profile 1', 'Profile 2', 'Custom Profile', etc.
 
+	use_system_keychain: bool | None = Field(
+		default=None,
+		description=(
+			'Whether Chromium may use the OS credential store (macOS Keychain / Linux libsecret) to decrypt '
+			'saved passwords/cookies. None (default) = auto: resolved in model_post_init before the profile is '
+			'possibly copied to a temp dir. Disabled for profiles fully managed by browser-use (temp dirs and the '
+			'default browser-use profile) so automation never triggers an OS credential prompt; enabled for a '
+			'real, user-supplied user_data_dir (e.g. an existing Chrome profile) so its saved logins stay readable.'
+		),
+	)
+
 	# these can be found in BrowserLaunchArgs, BrowserLaunchPersistentContextArgs, BrowserNewContextArgs, BrowserConnectArgs:
 	# save_recording_path: alias of record_video_dir
 	# save_har_path: alias of record_har_path
@@ -832,6 +843,14 @@ class BrowserProfile(BrowserConnectArgs, BrowserLaunchPersistentContextArgs, Bro
 
 	def model_post_init(self, __context: Any) -> None:
 		"""Called after model initialization to set up display configuration."""
+		if self.use_system_keychain is None:
+			# resolve auto mode before user_data_dir's field_validator (only guaranteed to have run when
+			# the field was explicitly passed — see #5414) and _copy_profile() may rewrite it to a temp copy
+			is_browser_use_managed_profile = self.user_data_dir is None or (
+				'browser-use-user-data-dir-' in str(self.user_data_dir).lower()
+				or Path(self.user_data_dir).expanduser().resolve() == CONFIG.BROWSER_USE_DEFAULT_USER_DATA_DIR.resolve()
+			)
+			self.use_system_keychain = not is_browser_use_managed_profile
 		self.detect_display_configuration()
 		self._copy_profile()
 
@@ -912,6 +931,7 @@ class BrowserProfile(BrowserConnectArgs, BrowserLaunchPersistentContextArgs, Bro
 			f'--profile-directory={self.profile_directory}',
 			*(CHROME_DOCKER_ARGS if (CONFIG.IN_DOCKER or not self.chromium_sandbox) else []),
 			*(CHROME_HEADLESS_ARGS if self.headless else []),
+			*([] if self.use_system_keychain else ['--use-mock-keychain', '--password-store=basic']),
 			*(CHROME_DISABLE_SECURITY_ARGS if self.disable_security else []),
 			*(CHROME_DETERMINISTIC_RENDERING_ARGS if self.deterministic_rendering else []),
 			*(
--- a/browser_use/browser/profile.py
+++ b/browser_use/browser/profile.py
@@ -91,6 +91,27 @@ def _ignore_chrome_profile_transient_files(_src: str, names: list[str]) -> set[s
 	return {name for name in names if any(fnmatch(name, pattern) for pattern in CHROME_PROFILE_TRANSIENT_FILE_PATTERNS)}
 
 
+def _is_empty_dir_under_system_temp(path: str | Path) -> bool:
+	"""Whether path is inside the OS temp dir AND doesn't exist yet or has no Chrome profile files in
+	it. Used to auto-detect a caller-supplied tempfile.mkdtemp() user_data_dir (not just
+	BrowserProfile's own 'browser-use-user-data-dir-*' prefix) as browser-use-managed for the
+	use_system_keychain heuristic. Requiring both conditions (not just emptiness) keeps a real user
+	profile that happens to not be initialized yet - but lives at a real, non-temp path - correctly
+	classified as a real profile."""
+	try:
+		resolved = Path(path).expanduser().resolve()
+		temp_root = Path(tempfile.gettempdir()).resolve()
+	except OSError:
+		return False
+	if resolved != temp_root and temp_root not in resolved.parents:
+		return False
+	if not resolved.exists():
+		return True
+	if not resolved.is_dir():
+		return False
+	return not any(resolved.iterdir())
+
+
 def _is_chrome_profile_lock_error(error: BaseException) -> bool:
 	"""Detect Windows sharing violations or permission errors raised while copying a Chrome profile."""
 	if isinstance(error, PermissionError):
@@ -849,6 +870,7 @@ class BrowserProfile(BrowserConnectArgs, BrowserLaunchPersistentContextArgs, Bro
 			is_browser_use_managed_profile = self.user_data_dir is None or (
 				'browser-use-user-data-dir-' in str(self.user_data_dir).lower()
 				or Path(self.user_data_dir).expanduser().resolve() == CONFIG.BROWSER_USE_DEFAULT_USER_DATA_DIR.resolve()
+				or _is_empty_dir_under_system_temp(self.user_data_dir)
 			)
 			self.use_system_keychain = not is_browser_use_managed_profile
 		self.detect_display_configuration()
--- a/tests/ci/conftest.py
+++ b/tests/ci/conftest.py
@@ -163,6 +163,7 @@ async def browser_session():
 			user_data_dir=None,  # Use temporary directory
 			keep_alive=True,
 			enable_default_extensions=True,  # Enable extensions during tests
+			use_system_keychain=False,  # tests must never touch the OS credential store
 		)
 	)
 	await session.start()
--- a/tests/ci/test_browser_profile_keychain.py
+++ b/tests/ci/test_browser_profile_keychain.py
@@ -1,5 +1,8 @@
 """Tests for BrowserProfile.use_system_keychain and its effect on get_args()."""
 
+import shutil
+from pathlib import Path
+
 from browser_use.browser.profile import CONFIG, BrowserChannel, BrowserProfile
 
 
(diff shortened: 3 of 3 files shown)


axisrow:
## Review cycle summary

Three review rounds run against this PR (local `/review` + Codex adversarial companion each round; capped at 3 by protocol).

**Round 1** (against 1aa3220d): `/review` clean. Codex `needs-attention` (high, confidence 0.72): `test_real_user_supplied_profile_keeps_system_keychain_by_default` used a `user_data_dir` path with no "chrome" substring and `channel=None`, so `_copy_profile()`'s `is_chrome` gate was False and it returned early without ever rewriting `user_data_dir` to a managed-looking temp dir — leaving the PR's central invariant untested: `use_system_keychain` is decided from the *original* real `user_data_dir` before `_copy_profile()` rewrites it, and must stay `True` afterward so a copied real Chrome profile's OS-keychain-encrypted cookies/passwords stay decryptable.
**Fix**: added `test_real_chrome_profile_keeps_system_keychain_after_copy_profile_rewrites_user_data_dir` using `channel=BrowserChannel.CHROME` so `_copy_profile()` actually performs the copy (asserted via the rewritten path), then asserts the keychain flags are still absent.

**Round 2** (against cc5d6634): `/review` raised a low-confidence note on an unguarded `Path(...).resolve()` in `model_post_init` — verified against `main`: the same unguarded call already exists in the pre-existing `field_validator` (`profile.py:554`), so this is pre-existing project behavior, not a regression; no change made. Codex confirmed the round-1 high finding fixed, and (also empirically verifying pydantic 2.12.5's execution order: `field_validator` → `model_post_init` → `model_validator(mode='after')`) raised one new medium finding (confidence 0.64): the keychain decision for `user_data_dir==CONFIG.BROWSER_USE_DEFAULT_USER_DATA_DIR` + a non-default `channel` is only correct because `model_post_init` runs *before* the `model_validator(mode='after')` that rewrites the dir to a sibling `default-{channel}` path — an ordering pydantic's docs don't guarantee, and untested (`test_default_browser_use_profile_avoids_system_keychain_by_default` uses the default channel, so that rewrite validator never fires there).
**Fix**: added `test_default_browser_use_profile_avoids_system_keychain_with_non_default_channel` using `channel=BrowserChannel.CHROME`, pinning that `use_system_keychain` still resolves `False` after the sibling-dir rewrite.

**Round 3** (against efc0d4e5): full `tests/ci` (1076 passed, 34 skipped) run locally to confirm both new tests pass and nothing regressed; per the 3-cycle cap, no further bot round was run.

**CI**: all checks green (`gh pr checks 5417`), only `license/cla` pending (contributor action, not code).

Net result: the keychain auto-detection logic itself was correct throughout all three rounds — every finding was a test-coverage gap for an already-correct invariant, not a behavior bug. All three gaps are now closed with regression tests.

PR #5451, opened 2026-08-12
Add OrcaRouter as a named LLM provider

XiaoHuo888-hue:
Adds `ChatOrcaRouter` as a first-class LLM provider, wired the same way as the existing OpenRouter integration.

## What's included

- **`browser_use/llm/orcarouter/chat.py`** — `ChatOrcaRouter` (`provider='orcarouter'`), an OpenAI-compatible `BaseChatModel` mirroring `ChatOpenRouter` (text + structured-output paths), default base URL `https://api.orcarouter.ai/v1`.
- **`browser_use/llm/orcarouter/serializer.py`** — `OrcaRouterMessageSerializer`, a thin delegate to the OpenAI message serializer (OrcaRouter speaks the OpenAI wire format).
- **Registration** in both `browser_use/llm/__init__.py` and `browser_use/__init__.py` (TYPE_CHECKING stubs + `_LAZY_IMPORTS` + `__all__`), so `from browser_use import ChatOrcaRouter` works out of the box.
- **`browser_use/tokens/service.py`** — OrcaRouter is a gateway with its own pricing, so upstream model pricing is never incorrectly attributed to it.
- **`.env.example`** — `ORCAROUTER_API_KEY=` entry.
- **`tests/ci/test_orcarouter.py`** — serializer, provider-defaults, and token-cost guard tests.
- **`examples/models/orcarouter.py`** — usage example.

[OrcaRouter](https://www.orcarouter.ai) is an OpenAI-compatible model routing gateway that brings 150+ models from OpenAI, Anthropic, Google, DeepSeek, Qwen, MiniMax and xAI behind a single endpoint and API key. Beyond routing, it runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes. This PR registers it as a named provider so users can opt in directly.

I'm an engineer on the OrcaRouter team.

---
## Summary by cubic
Adds `ChatOrcaRouter` as a first-class LLM provider using OrcaRouter's OpenAI-compatible endpoint. Supports text and JSON-schema structured outputs and correctly attributes token costs to `orcarouter/<model>`.

- `ChatOrcaRouter` (`provider='orcarouter'`) defaults to `https://api.orcarouter.ai/v1`; returns string or Pydantic-validated outputs and omits request-level params like `top_p`/`seed` from client construction.
- Resolves the API key from the constructor or `ORCAROUTER_API_KEY`, raising a 401 error if neither is set so `OPENAI_API_KEY` is never sent to the gateway.
- Serializer delegates to the OpenAI serializer since OrcaRouter speaks the OpenAI wire format; registered in both `__init__.py` files, with `ORCAROUTER_API_KEY` added to `.env.example` plus an example and tests.
- Token-cost service namespaces models as `orcarouter/<model>` so upstream model pricing is never attributed to the gateway.
- Guards empty `response.choices` with a descriptive `ModelProviderError` and re-raises provider errors to preserve status codes.

<sup>Written for commit be76e6cceef4c50e23106aa7be1ddc9362ccb755. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5451?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


XiaoHuo888-hue:
Thanks for the review — all three issues were valid and are fixed in 82a673b, each matching how `ChatOpenAI` already handles the same case.

**1. `top_p` / `seed` leaking into the client constructor** (`_get_client_params`, was line 68)
Removed them from `base_params`. They are request-level params and were already being forwarded to `chat.completions.create(...)`, so the constructor path was pure breakage: setting either one raised `TypeError: __init__() got an unexpected keyword argument`. Verified `get_client()` now works with `top_p=0.5, seed=42`, and that both values still reach the request.

**2. Empty `choices` surfaced as `IndexError`** (was line 152)
Both the text and structured branches now do `choice = response.choices[0] if response.choices else None` and raise the same descriptive `ModelProviderError` (status 502) that `ChatOpenAI` raises, including the `base_url` hint for proxy misconfiguration.

**3. `ModelProviderError` re-wrapped by the broad handler** (was line 204)
Added `except ModelProviderError: raise` ahead of the generic handlers, so the structured-output null-content case keeps its intended 500 instead of being re-wrapped as 502.

**Verification**

- `ruff check` and `ruff format --check` clean on `browser_use/llm/orcarouter/`.
- Behaviour checks with mocked completions: empty `choices` → `ModelProviderError(502)` with the descriptive message; structured null content → `ModelProviderError(500)` preserved; `get_client()` succeeds with `top_p`/`seed` set.
- Regression check: `top_p=0.5`, `seed=42`, `temperature=0.3` all still arrive in the `chat.completions.create(...)` call and a normal completion returns as before.

browser_use/llm/orcarouter/chat.py
```diff
@@ -0,0 +1,237 @@
+from collections.abc import Mapping
```
sauravpanda, line 2: `os` is needed for the env fallback below.

```suggestion
import os
from collections.abc import Mapping
```

browser_use/llm/orcarouter/chat.py
```diff
@@ -0,0 +1,237 @@
...
+	# Static
+	@property
+	def provider(self) -> str:
+		return 'orcarouter'
+
+	def _get_client_params(self) -> dict[str, Any]:
+		"""Prepare client parameters dictionary."""
+		# Define base client params
+		base_params = {
+			'api_key': self.api_key,
```
sauravpanda: Resolve the key explicitly so the OpenAI SDK never gets the chance to substitute `OPENAI_API_KEY`.

```suggestion
	def _get_api_key(self) -> str:
		# AsyncOpenAI falls back to OPENAI_API_KEY when api_key is unset, which would send an
		# unrelated provider's key to the OrcaRouter endpoint.
		key = self.api_key or os.getenv('ORCAROUTER_API_KEY')
		if not key:
			raise ModelProviderError('Missing OrcaRouter API key', status_code=401, model=self.name)
		return key

	def _get_client_params(self) -> dict[str, Any]:
		"""Prepare client parameters dictionary."""
		# Define base client params
		base_params = {
			'api_key': self._get_api_key(),
```

tests/ci/test_orcarouter.py
```diff
@@ -0,0 +1,64 @@
+import pytest
+
+from browser_use.llm.messages import ContentPartTextParam, SystemMessage, UserMessage
```
sauravpanda, line 4:
```suggestion
from browser_use.llm.exceptions import ModelProviderError
from browser_use.llm.messages import ContentPartTextParam, SystemMessage, UserMessage
```

tests/ci/test_orcarouter.py
```diff
@@ -0,0 +1,64 @@
...
+			prompt_cached_tokens=None,
+			prompt_cache_creation_tokens=None,
+			prompt_image_tokens=None,
+			completion_tokens=5,
+			total_tokens=15,
+		),
+	)
+
+	assert seen_model_names == ['orcarouter/openai/gpt-4o-mini']
+	assert cost is None
```
sauravpanda, line 65: Two regression tests — the first fails on the current branch with `assert 'sk-unrelated-openai-key' == 'orca-key'`, which is exactly the leak.

```suggestion
	assert cost is None


def test_orcarouter_reads_api_key_from_env(monkeypatch: pytest.MonkeyPatch) -> None:
	"""ORCAROUTER_API_KEY is the documented env var, so it must actually be read."""
	monkeypatch.setenv('ORCAROUTER_API_KEY', 'orca-key')
	monkeypatch.setenv('OPENAI_API_KEY', 'sk-unrelated-openai-key')

	client = ChatOrcaRouter(model='orcarouter/auto').get_client()

	assert client.api_key == 'orca-key'


def test_orcarouter_never_falls_back_to_the_openai_key(monkeypatch: pytest.MonkeyPatch) -> None:
	"""An unset OrcaRouter key must fail loudly, not ship OPENAI_API_KEY to the gateway."""
	monkeypatch.delenv('ORCAROUTER_API_KEY', raising=False)
	monkeypatch.setenv('OPENAI_API_KEY', 'sk-unrelated-openai-key')

	with pytest.raises(ModelProviderError) as exc_info:
		ChatOrcaRouter(model='orcarouter/auto').get_client()

	assert exc_info.value.status_code == 401
	assert 'sk-unrelated-openai-key' not in str(exc_info.value)
```

examples/models/orcarouter.py
```diff
@@ -0,0 +1,33 @@
+"""
+Simple try of the agent with OrcaRouter.
+
+@dev You need to add ORCAROUTER_API_KEY to your environment variables.
+"""
+
+import asyncio
+import os
```
sauravpanda:
```suggestion
import asyncio
```

examples/models/orcarouter.py
```diff
@@ -0,0 +1,33 @@
...
+from browser_use import Agent, ChatOrcaRouter
+
+load_dotenv()
+
+# OrcaRouter is an OpenAI-compatible model gateway routing to 190+ models via one endpoint.
+llm = ChatOrcaRouter(
+	model='orcarouter/auto',
+	base_url='https://api.orcarouter.ai/v1',
+	api_key=os.getenv('ORCAROUTER_API_KEY'),
+)
```
sauravpanda: With the fallback handled in the class, the example no longer needs to plumb the key by hand — and `base_url` is already the default. Matches `examples/models/mistral.py`.

```suggestion
llm = ChatOrcaRouter(model='orcarouter/auto')
```

sauravpanda (commented):
Thanks for the iteration on this — the `top_p`/`seed` and empty-`choices` fixes in 82a673b are right, and the copy is now strictly more correct than `ChatOpenRouter` (I've filed the backport on our side).

One blocker left before this can merge, and it's a credential-handling one rather than a style nit.

`api_key` defaults to `None` and gets filtered out of `client_params`, so `AsyncOpenAI` falls through to **its own** env lookup — meaning a user who hasn't set an OrcaRouter key sends their `OPENAI_API_KEY` to `api.orcarouter.ai`. Reproduced against this branch:

```
ChatOrcaRouter(model='openai/gpt-4o')   # no api_key
  base_url : https://api.orcarouter.ai/v1/
  api_key  : sk-…   ← read from OPENAI_API_KEY
```

`examples/models/orcarouter.py` walks straight into it: `os.getenv('ORCAROUTER_API_KEY')` returns `None` when unset, and the request still goes out — authenticated with an unrelated provider's key. Relatedly, the `.env.example` entry documents a variable that nothing in the code currently reads.

The suggestions below resolve the key explicitly (constructor → `ORCAROUTER_API_KEY`) and raise `ModelProviderError(401)` when neither is set, matching how `ChatMistral._get_api_key` already does it. That also makes the `.env.example` entry true. I've run these locally: ruff, pyright, and `tests/ci` all pass, and the new tests fail without the `chat.py` change.

For the record, `ChatOpenRouter` and `ChatVercel` have the same latent fallback — that's our bug, not yours, and I'm fixing it separately. Not asking you to carry it.

PR #5454, opened 2026-08-12
fix(registry): drop the two unreachable file_system branches

noron12234:
## What

`Registry._execute_action` checks each special parameter twice — once for the
`value is None` case and once for the "no default provided" case. Both chains
test `param.name == 'file_system'` twice:

```python
if param.name == 'browser_session':
    raise ValueError(...)
elif param.name == 'page_extraction_llm':
    raise ValueError(...)
elif param.name == 'file_system':                 # <- matches here
    raise ValueError(f'Action {func.__name__} requires file_system but none provided.')
elif param.name == 'page':
    raise ValueError(...)
elif param.name == 'available_file_paths':
    raise ValueError(...)
elif param.name == 'file_system':                 # <- unreachable, identical body
    raise ValueError(f'Action {func.__name__} requires file_system but none provided.')
else:
    raise ValueError(...)
```

The second `elif` can never run, and it raises the same message as the first, so
this is a pure dead-code removal: two branches, four lines.

## What I did not touch

The trailing `else` is reachable and stays. `special_param_names` comes from
`_get_special_param_types()` and also contains `context`, `page_url`,
`cdp_client` and `has_sensitive_data`, none of which are named in the chain —
they land in the `else` and get its different message. Collapsing the whole
chain into one `raise` would change the message for those, so I left it alone.

`ruff check` and `ruff format --check` clean.

---
## Summary by cubic
Removed two unreachable `file_system` branches in `Registry._execute_action` to eliminate dead code and simplify logic. Behavior and error messages are unchanged; the trailing `else` still handles other special params.

<sup>Written for commit 61ddf4f5ecff612c2f30bd6cfe0e4a9471715a13. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5454?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


noron12234:
Closing this. The CLA on it was never signed, so it could not have merged as it stands, and I am stepping back from the PRs I have open rather than leaving them sitting.

The branch stays on my fork if the fix is useful to anyone. Sorry for the noise.

PR #5455, opened 2026-08-12
Return errors for stale click and dropdown actions

### LINKED ISSUE #5438: Bug: click, dropdown_options, and select_dropdown report 'element not found' as success instead of error
### Browser Use Version

`0.13.7` (commit `3260188`)

### Bug Description

Three actions in `browser_use/tools/service.py` return "element not found" as a non-error `ActionResult` when the element index is missing from the selector map. This is the same class of bug as #5361 (which covers `scroll`, `input`, and `find_text`), but in three additional actions that were not mentioned in that issue.

#### Affected actions

| Action | File location | Line |
|---|---|---|
| `_click_by_index` | `tools/service.py` | ~715-718 |
| `dropdown_options` | `tools/service.py` | ~1679-1682 |
| `select_dropdown` | `tools/service.py` | ~1707-1710 |

All three have the identical pattern:

```python
node = await browser_session.get_element_by_index(params.index)
if node is None:
    msg = f'Element index {params.index} not available - page may have changed.'
    logger.warning(f'⚠️ {msg}')
    return ActionResult(extracted_content=msg)  # <-- missing error=
```

#### Why this matters

1. **`Agent.multi_act()`** checks `results[-1].error` to decide whether to abort remaining queued actions. With `error=None`, follow-up actions (e.g. a submit button click after a failed text input) keep executing against a stale or wrong page.

2. **`Agent._post_process()`** increments `consecutive_failures` only when `last_result[-1].error` is set. These false successes never count toward `max_failures`, so the agent never triggers its failure recovery.

3. **The LLM** is told the action succeeded via `evaluation_previous_goal`, so it reasons from a false premise on the next step.

#### Additional issue in `select_dropdown`

When `selection_data.get('success') != 'true'` and the response contains `short_term_memory`/`long_term_memory` keys, the function returns a non-error `ActionResult` (lines ~1733-1737). There is already a `# TODO` comment noting this should raise `BrowserError`. The failed selection is reported as a success to the agent.

### Steps to Reproduce

1. Start an agent task that involves clicking an element
2. Between the DOM snapshot and the click execution, let the page change (element disappears)
3. The click returns `ActionResult(extracted_content="Element index X not available...")` with no `.error`
4. The agent treats this as success and proceeds with follow-up actions

### Expected Behavior

All three actions should return `ActionResult(error=msg)` when the element is not found, consistent with how `scroll` handles the same condition at line ~1380.

### Proposed Fix

Change `ActionResult(extracted_content=msg)` to `ActionResult(error=msg)` in all three locations. For the `select_dropdown` structured error path, add `error=` to the `ActionResult`.

### Related

- #5361 covers `scroll`, `input`, and `find_text` — same bug class, different actions
- PR #5390 fixes the three actions in #5361 but not these three

xianjianlf2:
Summary
- Return ActionResult(error=...) when click, dropdown_options, or select_dropdown cannot resolve the requested element index.
- Mark structured select_dropdown failures as errors while preserving short- and long-term memory content.
- Add focused regression coverage for stale element indices and structured dropdown failures.

Validation
- uv run pytest tests/ci/test_tools.py::TestToolsActionErrors
- uv run ruff check browser_use/tools/service.py tests/ci/test_tools.py
- uv run ruff format browser_use/tools/service.py tests/ci/test_tools.py
- git diff --check

Fixes #5438

---
## Summary by cubic
Return errors when element indices are stale in click and dropdown actions, and treat structured `select_dropdown` failures as errors while preserving memory fields. This makes tool error handling consistent and clear.

- **Bug Fixes**
  - `click`, `dropdown_options`, and `select_dropdown` return `ActionResult(error=...)` when the element index is missing.
  - Structured `select_dropdown` failures set `error` and `long_term_memory` to the failure message and keep guidance in `extracted_content`.
  - Added regression tests for stale indices and structured dropdown failures.

<sup>Written for commit 7e29c30cc3a05d2eea3ac671333779a41b96f307. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5455?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


xianjianlf2:
Updated the branch with the latest upstream/main to clear the BEHIND state.\n\nLocal validation:\n- uv run ruff check --no-fix --select PLE\n- uv run pytest tests/ci/test_tools.py\n- uv run pyright\n- git diff --check

PR #5471, opened 2026-08-15
fix(dom): detect clickable elements in same-origin iframes and DOM0 onclick handlers

shawyang0304:
## Problem

Interactive elements inside **same-origin** iframes never get an interactive `[index]`, so the agent sees them as plain text and cannot click them with the normal `click` tool. It then burns steps re-clicking a neighbouring element or falls back to `evaluate`.

The listener-detection script in `_get_all_trees()` collects its candidates with `document.querySelectorAll('*')`, which does not descend into an iframe's `contentDocument`. Cross-origin iframes are covered separately by the per-target OOPIF traversal in `get_dom_tree()`, but same-origin iframes fall into **neither** path.

Same-origin iframes are the norm in payment / SDK widgets that need to share cookies with the host page, and mobile H5 component libraries commonly bind handlers at runtime via `el.onclick = fn`, so this shows up on real production pages.

### Why `el.onclick` is part of the fix

`getEventListeners()` is a DevTools Console utility that only reports listeners for the execution context it is called from. Measured on Chromium:

| element | `getEventListeners(el)` | `el.onclick` |
| --- | --- | --- |
| top document, `addEventListener('click', fn)` | `['click']` | – |
| top document, `el.onclick = fn` | `['click']` | truthy |
| **same-origin iframe, `el.onclick = fn`** | **`{}`** | truthy |

So inside a frame the DevTools call is unusable, while reading the DOM0 property still works across contexts. The fix descends into same-origin frames and treats a non-null `el.onclick` as an interactive signal.

> Known limitation: a handler registered *inside* an iframe with `addEventListener()` is still not detected — `getEventListeners()` returns `{}` for it and there is no DOM attribute to fall back on. Covering it would need per-frame execution contexts, so it is called out here explicitly rather than silently overclaimed.

## Fix

Confined to the JS expression string inside `_get_all_trees()` — no Python-side logic, thresholds, or downstream `backendNodeId` matching is touched.

- Recurse into same-origin iframe `contentDocument` when collecting elements. In modern Chromium a cross-origin `contentDocument` returns `null` rather than throwing, so the existing cross-origin path is unaffected.
- Treat a non-null `el.onclick` as an interactive signal; this is what makes detection work inside frames.
- Apply the pre-existing 10k element cap **per document**, before descending into that document's frames. If the top document is oversized, the original "abort listener detection" short-circuit is preserved via a flag; if a sub-frame is oversized it is skipped locally without affecting the top document or sibling frames.

Pages without same-origin iframes behave byte-identically to before.

## Test

`tests/ci/test_dom_iframe_onclick_detection.py` serves a host page containing a same-origin iframe whose child `<div>` is bound via `el.onclick = fn`, and asserts the child receives an `[index]`.

- On `main`: **fails** — `assert 'inframe' in {'plain', 'frame'}`
- With this fix: **passes**

### DIFF
--- a/browser_use/dom/service.py
+++ b/browser_use/dom/service.py
@@ -471,7 +471,32 @@ class DomService:
 							return null;
 						}
 
-						const allElements = document.querySelectorAll('*');
+						// Collect elements from the top document plus same-origin nested
+						// iframes/frames. Cross-origin iframes are handled separately via
+						// the per-target OOPIF traversal in get_dom_tree(); accessing
+						// their contentDocument here throws and is silently skipped, so
+						// this recursion is safe.
+						const allElements = [];
+						(function collect(doc) {
+							let els;
+							try {
+								els = doc.querySelectorAll('*');
+							} catch (e) {
+								return;
+							}
+							for (const el of els) {
+								allElements.push(el);
+								if (el.tagName === 'IFRAME' || el.tagName === 'FRAME') {
+									try {
+										if (el.contentDocument) {
+											collect(el.contentDocument);
+										}
+									} catch (e) {
+										// Cross-origin iframe, inaccessible from this context
+									}
+								}
+							}
+						})(document);
 
 						// Skip on heavy pages — listener detection is too expensive
 						if (allElements.length > 10000) {
@@ -483,8 +508,12 @@ class DomService:
 						for (const el of allElements) {
 							try {
 								const listeners = getEventListeners(el);
-								// Check for click-related event listeners
-								if (listeners.click || listeners.mousedown || listeners.mouseup || listeners.pointerdown || listeners.pointerup) {
+								// Check for click-related event listeners registered via
+								// addEventListener, as well as a handler assigned directly
+								// via the DOM0 `el.onclick = fn` property, which
+								// getEventListeners cannot see and which also never shows
+								// up as an HTML `onclick` attribute in the DOM tree.
+								if (listeners.click || listeners.mousedown || listeners.mouseup || listeners.pointerdown || listeners.pointerup || el.onclick) {
 									elementsWithListeners.push(el);
 									if (elementsWithListeners.length > %d) {
 										return %r;
--- a/browser_use/dom/service.py
+++ b/browser_use/dom/service.py
@@ -476,12 +476,31 @@ class DomService:
 						// the per-target OOPIF traversal in get_dom_tree(); accessing
 						// their contentDocument here throws and is silently skipped, so
 						// this recursion is safe.
+						//
+						// Cost bound is applied PER DOCUMENT and BEFORE descending: a
+						// single oversized document is skipped in place (and its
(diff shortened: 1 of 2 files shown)


shawyang0304:
Thanks for the review! Addressed both points in commit fff3770f4.

Root cause of both P2 items: the previous revision globally accumulated elements from the top document and every same-origin iframe into a single array before checking length > 10000. That silently changed the pre-existing guard from per-document to whole-page total semantics, which is exactly what both comments call out.

Fix: apply the cap per document and before descending into that document's frames:

Each document is scanned only if its own querySelectorAll('*').length is within the budget; otherwise it is skipped in place and its child frames are not traversed (addresses "enforce an element/frame bound before recursing further").
If the top document itself is oversized, the original "abort listener detection entirely" short-circuit is preserved via a _topDocOversized flag, so behavior on plain heavy pages is byte-identical to before this PR (addresses "widget-dense pages could skip listener detection for the top document too").
If a sub-frame is oversized, it is skipped locally without affecting scanning of the top document or sibling frames.
Behavior on pages that fit within the cap is byte-identical to the previous revision of this PR; only the overflow branch changes. Happy to add a regression test if you can point me at the preferred test location for dom/service.py.

shawyang0304:
Gentle ping on this one — it's been sitting without a human review since Aug 15. Only the cubic bot has looked at it so far, and I addressed both of its points in fff3770.

I've rebased onto latest main (post-0.13.9) and added a regression test at `tests/ci/test_dom_iframe_onclick_detection.py`, following the pattern in `test_dom_live_input_value.py`. It fails on main (`assert 'inframe' in {'plain', 'frame'}`) and passes with the fix.

Writing that test also made me measure something that corrects part of my original description — the PR body is updated accordingly. I had claimed `getEventListeners()` does not report DOM0 `el.onclick` handlers. That is wrong for the top document: it returns `['click']` there. Its real limitation is that it only reports listeners for the execution context it is called from, so for an element inside a same-origin iframe it returns `{}`:

| element | `getEventListeners(el)` | `el.onclick` |
| --- | --- | --- |
| top document, `addEventListener('click', fn)` | `['click']` | – |
| top document, `el.onclick = fn` | `['click']` | truthy |
| **same-origin iframe, `el.onclick = fn`** | **`{}`** | truthy |

So the actual blind spot is same-origin iframes, and reading `el.onclick` is what makes detection work inside them. A handler registered *inside* an iframe via `addEventListener()` is therefore still not detected — I've written that up as a known limitation instead of claiming more than the fix delivers.

Scope is unchanged and small: the diff is confined to the JS expression string inside `_get_all_trees`, with the 10k cost cap now applied per document.

@MagMueller @laithrw — you two know this code best (laithrw wrote the original listener detection in 20f51c7; MagMueller owns most of the recent dom/service.py changes). Would either of you have a few minutes to take a look? Happy to split the PR, change the cap semantics, or add more tests — whatever makes it easiest to land.

PR #5472, opened 2026-08-15
test: isolate env-override module reloads to subprocesses

### LINKED ISSUE #5425: Test-order-dependent failure: importlib.reload() in test_action_timeout.py corrupts Tools class identity for the rest of the pytest process
## Summary

`tests/ci/test_action_timeout.py::test_malformed_env_timeout_does_not_break_import` calls `importlib.reload(browser_use.tools.service)` ten times in a row (once per malformed `BROWSER_USE_ACTION_TIMEOUT_S` value it checks). Each `reload()` rebinds `browser_use.tools.service.Tools` to a **new class object**. But `Tools` is already imported by value (`from browser_use.tools.service import Tools`) in several other modules *before* this test ever runs:

- `browser_use/agent/service.py`
- `browser_use/beta/service.py`
- `browser_use/mcp/client.py`
- `browser_use/mcp/server.py`
- `browser_use/integrations/gmail/actions.py`
- `browser_use/__init__.py`

Those modules keep pointing at whatever `Tools` object existed at their own import time (normally the very first, pre-reload one). The fixture's final `reload()` on teardown doesn't restore the original object either — it creates yet another new one. So for the rest of that pytest **process**, any `Tools` freshly imported from `tools.service` no longer `is`-matches (or `isinstance`-matches) the `Tools` object baked into `agent/service.py` / `beta/service.py`.

## Reproduction

Minimal reproduction, no pytest involved:

```python
import importlib
import browser_use.tools.service as svc
from browser_use.agent.service import Agent
import browser_use.agent.service as agent_svc

Tools_before = svc.Tools
print(Tools_before is agent_svc.Tools)   # True

importlib.reload(svc)
print(svc.Tools is Tools_before)          # False - new class object
print(agent_svc.Tools is Tools_before)    # True - agent.service still holds the OLD class
print(agent_svc.Tools is svc.Tools)       # False - desynced for the rest of the process
```

## Observed effect

Running the full `tests/ci` suite (not in isolation) reliably fails two tests in `tests/ci/test_beta_agent.py`:

- `test_beta_agent_constructor_type_hints_match_browser_use_core_params`
- `test_beta_agent_initializes_tools_and_action_models`

Both fail on `get_origin(...) is Tools` / `isinstance(agent.tools, Tools)` — the classes print identically (`browser_use.tools.service.Tools`) but are different objects. Running `tests/ci/test_beta_agent.py` alone is 100% green; the failure only appears when `test_action_timeout.py` runs earlier in the same pytest process (default, non-xdist run order). Bisecting the file list confirms `test_action_timeout.py` is the sole trigger — any file set that excludes it stays green, any set that includes it (with `test_beta_agent.py` after it) reproduces the failure.

## Fix

`tests/ci/test_action_timeout.py` already has the pure function it needs for this: `browser_use.tools.service._parse_env_action_timeout(raw: str | None) -> float`, which is what `_DEFAULT_ACTION_TIMEOUT_S` is computed from at import time. The test can call this function directly for each malformed value instead of reloading the whole module — no `sys.modules` mutation, no leaked class identity, same coverage.

PR: (linked below)

aryansk:
Closes #5425

## Problem

`tests/ci/test_action_timeout.py` used `importlib.reload(browser_use.tools.service)` to exercise the `BROWSER_USE_ACTION_TIMEOUT_S` fallback logic. Every `reload()` rebinds `browser_use.tools.service.Tools` to a **new class object**, but several modules imported `Tools` by value at startup (`agent/service.py`, `beta/service.py`, `mcp/client.py`, `mcp/server.py`, `integrations/gmail/actions.py`, `__init__.py`). Those modules keep pointing at the original class, so for the rest of the pytest **process**, fresh imports of `tools.service.Tools` no longer `is`- or `isinstance`-match the class baked into those modules — a test-order-dependent failure. The fixture's teardown `reload()` didn't help either; it minted yet another class.

## Fix

Run each env-var scenario in a **fresh subprocess** (`sys.executable -c ...`) instead of reloading in-process. The module is imported from scratch with the env var set, the parsed `_DEFAULT_ACTION_TIMEOUT_S` is printed and asserted, and the subprocess exits — the in-process module graph and `Tools` class identity are never touched. Coverage is unchanged:

- non-numeric / empty / NaN / inf / negative / zero → fallback to `180.0`
- valid positive value (`45`) → takes effect

Also added `test_action_timeout_tests_do_not_corrupt_tools_class_identity`, a regression guard asserting that `tools.service.Tools` stays identical to the class held by `agent/service.py` after the env tests run.

## Validation

- `uv run pytest tests/ci/test_action_timeout.py` — 6 passed
- `uv run pytest tests/ci/test_action_timeout.py tests/ci/test_agent_planning.py` — 25 passed (cross-file identity)
- `ruff check` + `ruff format --check` clean

---
## Summary by cubic
Isolates `BROWSER_USE_ACTION_TIMEOUT_S` tests by running each env scenario in a fresh subprocess instead of reloading `browser_use.tools.service` in-process. Old behavior reloaded the module and rebound `Tools`, causing cross-module class identity mismatches; new behavior imports in a subprocess, preserving in-process class identity.

- Adds `_timeout_default_in_subprocess` to set the env var, import, and read `_DEFAULT_ACTION_TIMEOUT_S` without touching the main process.
- Preserves coverage: invalid/edge values fall back to 180.0; valid positive values (e.g., 45) are honored.
- Removes the reload-based fixture and adds `test_action_timeout_tests_do_not_corrupt_tools_class_identity` to assert `Tools` identity stays consistent across `browser_use.tools.service` and `browser_use.agent.service`.

<sup>Written for commit 25c9e44c4d263e4d60571bfc206f4081e95344b6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5472?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tests/ci/test_action_timeout.py
+++ b/tests/ci/test_action_timeout.py
@@ -132,45 +132,62 @@ def test_default_action_timeout_accommodates_extract_action():
 	)
 
 
-@pytest.fixture
-def _restore_service_module():
-	"""Reload browser_use.tools.service without any env override on teardown.
+def _timeout_default_in_subprocess(env_value: str) -> float:
+	"""Import browser_use.tools.service in a fresh interpreter with the env var set.
 
-	Tests in this file intentionally reload the module with BROWSER_USE_ACTION_TIMEOUT_S
-	set to various values; without this fixture, the last reload's default leaks into
-	every later test in the same worker.
+	Run in a subprocess (rather than importlib.reload) because reloading the module in
+	this process would rebind browser_use.tools.service.Tools to a brand-new class
+	object, desyncing it from the by-value imports already held by agent/service.py,
+	beta/service.py, mcp/*, and __init__.py for the rest of the pytest run.
 	"""
-	import importlib
 	import os
+	import subprocess
+	import sys
 
-	import browser_use.tools.service as svc_module
+	script = 'import browser_use.tools.service as svc; print(svc._DEFAULT_ACTION_TIMEOUT_S)'
+	env = dict(os.environ)
+	env['BROWSER_USE_ACTION_TIMEOUT_S'] = env_value
+	out = subprocess.run(
+		[sys.executable, '-c', script],
+		capture_output=True,
+		text=True,
+		env=env,
+		check=True,
+	)
+	return float(out.stdout.strip())
 
-	yield svc_module
-	os.environ.pop('BROWSER_USE_ACTION_TIMEOUT_S', None)
-	importlib.reload(svc_module)
 
-
-def test_malformed_env_timeout_does_not_break_import(monkeypatch, _restore_service_module):
+def test_malformed_env_timeout_does_not_break_import():
 	"""Bad BROWSER_USE_ACTION_TIMEOUT_S values must fall back, not crash or misbehave.
 
 	Covers three failure modes:
 	- Non-numeric / empty (ValueError from float()): would crash module import.
 	- NaN: parses fine but makes asyncio.wait_for time out immediately for every action.
 	- Infinity / negative / zero: parses fine but effectively disables the hang guard.
+
+	Each scenario runs in its own subprocess so the in-process Tools class identity
+	is never corrupted (see _timeout_default_in_subprocess).
 	"""
-	import importlib
-
-	svc_module = _restore_service_module
-
 	bad_values = ('', 'not-a-number', 'abc', 'nan', 'NaN', 'inf', '-inf', '0', '-5')
 	for bad_value in bad_values:
-		monkeypatch.setenv('BROWSER_USE_ACTION_TIMEOUT_S', bad_value)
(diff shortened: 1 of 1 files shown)


aryansk:
Acknowledged — the CLA signature is a contributor legal attestation that remains outstanding and will be completed by the contributor before this draft is marked ready. No code change is needed for the check itself.

PR #5474, opened 2026-08-15
feat: add a3m-router as optional dependency

Das-rebel:
## Summary

Add `a3m-router` as an optional dependency so users can `pip install browser-use[a3m]` to get A3M Router support.

## Why optional?

- `pip install browser-use` works without A3M (no surprise dependencies)
- `pip install browser-use[a3m]` installs with A3M support
- A3M Router (`a3m-router` on PyPI) is a Python package separate from the npm package

---
## Summary by cubic
Adds optional A3M Router support via an `a3m` extra (`a3m-router>=2.2.2`). Previously there was no A3M extra; now installing `browser-use[a3m]` pulls in A3M without changing the default install.

- Existing users need no changes; the base install adds no new dependencies.
- ChatA3M runs router calls in an executor so async requests no longer block the event loop.
- ChatA3M now uses the actual `finish_reason` for `stop_reason`, preserves refusal type, and warns on unknown message types.

<sup>Written for commit 893806aa452777d78b5c89d3a99164dd6ea0ee93. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5474?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/llm/a3m/chat.py
+++ b/browser_use/llm/a3m/chat.py
@@ -1,0 +1,228 @@
+"""
+ChatA3M - A3M Router chat model wrapper for browser-use.
+
+Provides intelligent LLM routing for browser automation with:
+- Automatic routing to cheapest capable model
+- Stealth mode for anti-detection
+- Parallel ensemble for reliability
+- Parallel ensemble for reliable extraction
+
+Requires the `adaptive-memory-multi-model-router` package:
+    pip install adaptive-memory-multi-model-router
+
+Usage:
+    from browser_use import Agent
+    from browser_use.llm.a3m import ChatA3M
+
+    agent = Agent(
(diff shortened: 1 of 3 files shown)


Das-rebel:
Fixed the P2 from cubic review: `a3m-router` now pinned to exact version (`==2.2.3`), consistent with all other extras in pyproject.toml.

PR #5487, opened 2026-08-17
fix(tools): report failed tab switches as errors instead of successes (#5486)

### LINKED ISSUE #5486: Bug: switch action reports a successful tab switch when the switch failed (error= never set)
### Browser Use Version

`0.13.8` (commit `eb41269`)

### Bug Description

The `switch` action reports a **successful** tab switch when the switch failed. It returns a non-error `ActionResult` on both of its failure paths, so the agent is told it is on a tab it never reached.

`browser_use/tools/service.py:1009-1028`:

```python
async def switch(params: SwitchTabAction, browser_session: BrowserSession):
    try:
        target_id = await browser_session.get_target_id_from_tab_id(params.tab_id)

        event = browser_session.event_bus.dispatch(SwitchTabEvent(target_id=target_id))
        await event
        new_target_id = await event.event_result(raise_if_any=False, raise_if_none=False)

        if new_target_id:
            memory = f'Switched to tab #{new_target_id[-4:]}'
        else:
            memory = f'Switched to tab #{params.tab_id}'      # <-- claims success on no result

        logger.info(f'🔄  {memory}')
    except Exception as e:
        logger.warning(f'Tab switch may have failed: {e}')
        memory = f'Attempted to switch to tab #{params.tab_id}'
        return ActionResult(extracted_content=memory, long_term_memory=memory)  # <-- no error=
```

#### Path 1 — `except Exception` swallows real failures

`get_target_id_from_tab_id()` raises `ValueError(f'No TargetID found ending in tab_id=...{tab_id}')` (`browser/session.py:2607`) whenever the `tab_id` matches no live target. That is the common real-world case: the model names a tab that has since closed, or hallucinates a tab_id. The bare `except Exception` catches it, logs at **warning**, and returns success-shaped output with no `error=`.

#### Path 2 — a missing event result is treated as success

`event_result(raise_if_any=False, raise_if_none=False)` returns `None` when the handler raised, because `raise_if_any=False` suppresses the exception. The `else` branch then reports `'Switched to tab #N'` regardless.

This branch is never a quiet success: `on_SwitchTabEvent` (`browser/session.py:1131`) returns a truthy `TargetID` on **every** success path, so `None` unambiguously means the switch failed.

### Why this matters

The agent loop keys off `ActionResult.error` in two places, and both are bypassed:

- `Agent.multi_act()` aborts remaining queued actions on `results[-1].error`.
- `Agent._post_process()` increments `consecutive_failures` only when `last_result[-1].error` is set — so a persistently broken switch never counts toward `max_failures`.

`switch` sets `terminates_sequence=True`, which limits the damage inside a single step. It does nothing about the next one: the false claim is written to **`long_term_memory`**, so it persists into agent history and the next step's `evaluation_previous_goal` reasons from "I am on tab #N" when the agent never left the original tab. Every subsequent action targets the wrong page.

This is the same status-inversion class as #5361 and #5438, in an action neither covers. It is arguably worse — a failed scroll still leaves you on the correct page, whereas a failed tab switch silently redirects all downstream work.

### Steps to Reproduce

```python
import asyncio

async def main() -> None:
    tools = Tools()
    session = BrowserSession(browser_profile=BrowserProfile(headless=True, user_data_dir=None))
    await session.start()
    ActionModel = tools.registry.create_action_model()
    try:
        tabs = await session.get_tabs()
        print('open tabs:', [t.target_id[-4:] for t in tabs])

        result = await tools.act(ActionModel(switch={'tab_id': 'zzzz'}), browser_session=session)

        print('error            :', repr(result.error))
        print('extracted_content:', repr(result.extracted_content))
        print('long_term_memory :', repr(result.long_term_memory))
    finally:
        await session.kill()


asyncio.run(main())
```

### Actual Behavior

```text
WARNING  [tools] Tab switch may have failed: No TargetID found ending in tab_id=...zzzz
open tabs: ['DC0E']
error            : None
extracted_content: 'Attempted to switch to tab #zzzz'
long_term_memory : 'Attempted to switch to tab #zzzz'
```

`error` is `None`, so `multi_act()` does not abort and `consecutive_failures` does not increment. The tab set is unchanged — the switch genuinely did not happen.

### Expected Behavior

A failed switch sets `ActionResult.error`, so the agent can see the failure, stop the queued sequence, and count it toward `max_failures`. `upload_file` in this same file already does this via `raise BrowserError(...)` (`tools/service.py:1000`); `switch` and `close` are the outliers that swallow instead.

### Note on `close`

`close` has the same shape (`tools/service.py:1050-1057`), returning `'Tab #N closed (was already closed or invalid)'` on exception. That one is arguably acceptable idempotency for a close, so I have deliberately left it alone rather than bundle a debatable change — happy to file separately if you'd want it changed too.

Windows 11, Python 3.11, local Chromium (headless). The logic is platform-independent.

### LLM Model

N/A — reproduces via `tools.act()` with no LLM involved.

thelonewander3r:
Fixes #5486.

## The bug

The `switch` action returns a **non-error** `ActionResult` on both of its failure paths, so a tab switch that never happened is reported to the agent as one that did.

```python
        if new_target_id:
            memory = f'Switched to tab #{new_target_id[-4:]}'
        else:
            memory = f'Switched to tab #{params.tab_id}'      # claims success on no result
        ...
    except Exception as e:
        logger.warning(f'Tab switch may have failed: {e}')
        memory = f'Attempted to switch to tab #{params.tab_id}'
        return ActionResult(extracted_content=memory, long_term_memory=memory)   # no error=
```

**Path 1.** `get_target_id_from_tab_id()` raises `ValueError('No TargetID found ending in tab_id=...')` (`browser/session.py:2607`) for any `tab_id` with no live target — the common case when the model names a tab that has since closed. The bare `except Exception` catches it, logs at *warning*, and returns success-shaped output.

**Path 2.** `event_result(raise_if_any=False, raise_if_none=False)` returns `None` when the handler raised, and the `else` branch reports `'Switched to tab #N'` anyway. `on_SwitchTabEvent` returns a truthy `TargetID` on every success path, so `None` unambiguously means failure — never a quiet success.

Because the agent loop keys off `ActionResult.error`, neither path aborts the queue in `multi_act()` nor increments `consecutive_failures` in `_post_process()`. `terminates_sequence=True` limits the damage inside one step, but the false claim goes into **`long_term_memory`**, so it persists into history and the next step plans against a tab the agent never reached.

Same status-inversion class as #5361 / #5438, in an action neither covers.

## The fix

Both failure paths now `raise BrowserError(..., long_term_memory=...)`, which `handle_browser_error()` converts to `ActionResult(error=...)`. This matches how `upload_file` in this same file already reports failure (`tools/service.py:1000`) — `switch` was the outlier.

The `else` branch is gone: a missing event result is now an error rather than a success message, and the success path reads the real target id.

## Verification

Reproduction from the issue, before and after:

```
# before
error            : None
extracted_content: 'Attempted to switch to tab #zzzz'
long_term_memory : 'Attempted to switch to tab #zzzz'

# after
error            : 'Failed to switch to tab #zzzz - the tab may have been closed.
                    Check the tabs list in browser state for currently open tab_ids.'
extracted_content: None
```

## Tests

Adds `TestTabSwitchFailureReporting` to `tests/ci/browser/test_tabs.py`:

| Test | Asserts |
|---|---|
| `test_switch_to_nonexistent_tab_reports_error` | `error` is set, no "Switched to tab" text in either memory field, tab set unchanged |
| `test_switch_to_open_tab_still_succeeds` | happy path unchanged — live `tab_id` switches with `error is None` |

The first fails on unpatched `main`, the second passes either way (no-regression guard):

```
$ uv run pytest tests/ci/browser/test_tabs.py::TestTabSwitchFailureReporting -o addopts=""
# without the fix:
1 failed, 1 passed
# with the fix:
2 passed
```

Existing suites, lint and types:

```
$ uv run pytest tests/ci/browser/test_tabs.py                     ->  7 passed
$ uv run pytest tests/ci/test_tools.py tests/ci/test_multi_act_guards.py \
                tests/ci/test_action_timeout.py                   -> 34 passed
$ uv run ruff check                                               -> All checks passed!
$ uv run pyright browser_use/tools/service.py                     -> 0 errors, 0 warnings
```

## Scope note

`close` has the same shape (`tools/service.py:1050`), returning `'Tab #N closed (was already closed or invalid)'` on exception. For a *close* that's arguably correct idempotency, so I deliberately left it out rather than bundle a debatable behaviour change into a clear-cut fix. Happy to add it here or separately if you'd rather it changed too.

One behavioural note for review: agents that previously sailed past a failed switch will now see an error and have it counted toward `max_failures`. That's the point of the change, but it does mean a task relying on the silent no-op will now surface the failure instead.

---
## Summary by cubic
Report failed tab switches as errors so the agent stops planning against tabs it never reached. Previously `switch` returned a success-shaped `ActionResult` on failures; now it raises `BrowserError` and sets `ActionResult.error`.

- Raise `BrowserError` when `get_target_id_from_tab_id(...)` fails or when `event_result(raise_if_any=False, raise_if_none=False)` returns `None`; include failure `long_term_memory` directing the agent to check the tabs list.
- Remove the fallback “Switched to tab #N”; success uses the actual returned `TargetID`.
- Workflows relying on silent no-ops must handle the error; the loop now aborts queued actions and increments failure counts.
- Tests: add failure/success coverage in `tests/ci/browser/test_tabs.py`; declare `_TabActionModel(ActionModel)` with explicit `switch`/`navigate` slots so pyright checks `Tools.act(...)` call sites.

<sup>Written for commit f495b264c7bfcdf1f12b54a69b3245133b449e92. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5487?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/tools/service.py
+++ b/browser_use/tools/service.py
@@ -1008,24 +1008,32 @@ class Tools(Generic[Context]):
 		)
 		async def switch(params: SwitchTabAction, browser_session: BrowserSession):
 			# Simple switch tab logic
+			failure_memory = (
+				f'Failed to switch to tab #{params.tab_id} - the tab may have been closed. '
+				f'Check the tabs list in browser state for currently open tab_ids.'
+			)
 			try:
 				target_id = await browser_session.get_target_id_from_tab_id(params.tab_id)
 
 				event = browser_session.event_bus.dispatch(SwitchTabEvent(target_id=target_id))
 				await event
 				new_target_id = await event.event_result(raise_if_any=False, raise_if_none=False)  # Don't raise on errors
-
-				if new_target_id:
-					memory = f'Switched to tab #{new_target_id[-4:]}'
-				else:
-					memory = f'Switched to tab #{params.tab_id}'
-
-				logger.info(f'🔄  {memory}')
-				return ActionResult(extracted_content=memory, long_term_memory=memory)
 			except Exception as e:
-				logger.warning(f'Tab switch may have failed: {e}')
-				memory = f'Attempted to switch to tab #{params.tab_id}'
-				return ActionResult(extracted_content=memory, long_term_memory=memory)
+				logger.warning(f'Tab switch failed: {e}')
+				raise BrowserError(f'Failed to switch to tab #{params.tab_id}: {e}', long_term_memory=failure_memory)
+
+			# on_SwitchTabEvent returns the newly focused TargetID on every success path, so a
+			# missing result means the handler failed - raise_if_any=False above swallowed the
+			# exception rather than the switch having quietly succeeded.
+			if not new_target_id:
+				raise BrowserError(
+					f'Failed to switch to tab #{params.tab_id}: tab switch produced no result',
+					long_term_memory=failure_memory,
+				)
+
+			memory = f'Switched to tab #{new_target_id[-4:]}'
+			logger.info(f'🔄  {memory}')
+			return ActionResult(extracted_content=memory, long_term_memory=memory)
 
 		@self.registry.action(
 			'Close a tab by tab_id. Tab IDs are shown in browser state tabs list (last 4 chars of target_id). Use to clean up tabs you no longer need.',
--- a/tests/ci/browser/test_tabs.py
+++ b/tests/ci/browser/test_tabs.py
@@ -26,6 +26,7 @@ from pytest_httpserver import HTTPServer
 from browser_use.agent.service import Agent
 from browser_use.browser import BrowserSession
 from browser_use.browser.profile import BrowserProfile
+from browser_use.tools.service import Tools
 from tests.ci.conftest import create_mock_llm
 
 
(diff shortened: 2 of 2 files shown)


thelonewander3r:
Superseded by #5500, which landed this fix on `main` in c45b85023 / c4e314072.

That PR covers everything here and goes further:

- `event_result(raise_if_any=True)` so a `SwitchTabEvent` handler failure surfaces its real cause through the `except` block, rather than degrading into the generic "produced no result" branch.
- The concrete cause is preserved in both `short_term_memory` and `long_term_memory`. That also resolves the cubic finding on this PR, which flagged my hardcoded `failure_memory` string for discarding the actual exception.
- Same two regression tests, plus `test_switch_reports_error_when_event_yields_no_result` for the no-result path.

Nothing left to rebase. Closing.

PR #5493, opened 2026-08-18
fix(agent): match URL negations as whole words

JayLay0310:
## Summary

- match URL negation terms as whole words instead of arbitrary substrings
- share the negation matcher across the standard and beta agent URL extractors
- add regression coverage for both a false positive (`notable`) and an explicit negation (`do not`)

## Problem

The direct-navigation URL extractors used substring checks for `not`. As a result, normal prose such as `Open this notable site: https://example.com` was treated as a negated instruction, so the agent skipped its initial navigation action.

## Testing

- `python -m pytest -o addopts='' tests/ci/test_beta_agent.py::test_beta_agent_exposes_task_helper_methods`
- `ruff check browser_use/utils.py browser_use/agent/service.py browser_use/beta/service.py tests/ci/test_beta_agent.py`
- `ruff format --check browser_use/utils.py browser_use/agent/service.py browser_use/beta/service.py tests/ci/test_beta_agent.py`

---
## Summary by cubic
Matches URL negations as whole words during direct-navigation extraction, so prose like "notable" no longer causes false skips. Previously the extractors substring-matched "not"; now both agents share a word-boundary regex for never/not/don't (straight or curly apostrophes) in the 20-character pre-URL context.

- Adds `URL_NEGATION_PATTERN` and `has_url_negation` in `browser_use/utils.py`; used by `browser_use/agent/service.py` and `browser_use/beta/service.py`.
- Removes per-extractor `excluded_words`; keeps the 20-character window; injects the scheme after the negation check; updates debug messaging.
- Adds regression tests allowing "notable" and blocking "Do not open …" for both agents.

<sup>Written for commit c05a826af8f299b64250404926724e2190526393. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5493?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rajpratham1 (changes requested):
The current except (zipfile.BadZipFile, OSError, ValueError) is too broad. Genuine filesystem errors such as permission failures or disk-full errors can be caught and incorrectly reported as Invalid CRX file format. Please narrow the exception handling to the specific corrupt-EOCD errors, or preserve/distinguish genuine filesystem errors so their original cause is not masked.

rajpratham1 (approved):
This correctly preserves the order of CHROME_DEFAULT_ARGS while still using a set for efficient membership checks. The added regression test also verifies that the surviving default arguments remain in the expected order.

rajpratham1 (approved):
The try/finally cleanup correctly ensures that the temporary ZIP is removed even when extraction fails. The cleanup error is also handled without masking the original extraction exception. The added tests cover both failure cleanup and cleanup-error behavior.

rajpratham1 (approved):
This correctly avoids using a set for the resulting argument list and preserves the original CHROME_DEFAULT_ARGS order. The regression tests verify both ignored arguments and the expected ordering.

rajpratham1 (approved):
This is a focused fix for URL negation matching. Using word boundaries prevents words such as notable from being incorrectly interpreted as the negation word not, while the added tests verify both positive and negative cases.

PR #5495, opened 2026-08-18
fix(browser): track and drain HAR response body fetches before writing HAR

shashank-100:
## Why

Found while auditing the watchdogs: `HarRecordingWatchdog._on_loading_finished` fetches each response body via CDP (`Network.getResponseBody`) in a **bare `asyncio.create_task(...)`** — no reference stored, never awaited:

1. **Unreferenced task** — per asyncio docs, the event loop only holds weak refs to tasks, so an unreferenced task may be garbage-collected before completion.
2. **Never drained** — `on_BrowserStopEvent` calls `_write_har()` immediately, so any body fetch still in flight at stop time is silently lost: the entry falls back to the `dataReceived` buffer, which the code itself notes *"may be incomplete"*. Result: response bodies missing from the recorded HAR.
3. Fetch failures were swallowed by a bare `except Exception: pass`.

## What

- Track in-flight fetches in `self._pending_body_fetches` (strong refs; self-cleaning via `add_done_callback`).
- New `_drain_body_fetches(timeout_s=5.0)`, awaited in `on_BrowserStopEvent` **before** `_write_har()` — pending fetches get a bounded window to complete so their content lands in the HAR; a stuck fetch can't block shutdown forever.
- Body-fetch failures now log at debug instead of being silently swallowed.

30 insertions / 5 deletions in one file, plus regression tests.

## Testing

New `tests/ci/test_har_recording_watchdog.py` (3 tests, all verified to fail on the old code):

- `test_body_fetch_tasks_are_tracked` — the fetch task is strongly referenced.
- `test_stop_waits_for_in_flight_body_fetch` — stop blocks until the in-flight fetch completes, and the fetched body is present in the written HAR.
- `test_stop_writes_har_when_body_fetch_is_stuck` — a wedged fetch can't hang HAR writing (bounded drain).

```
tests/ci/test_har_recording_watchdog.py::test_body_fetch_tasks_are_tracked PASSED
tests/ci/test_har_recording_watchdog.py::test_stop_waits_for_in_flight_body_fetch PASSED
tests/ci/test_har_recording_watchdog.py::test_stop_writes_har_when_body_fetch_is_stuck PASSED
```

Also green: related watchdog suites (`test_downloads_watchdog`, `test_remote_download_complete_callback`), ruff, pyright, pre-commit.

---
## Summary by cubic
Ensures recorded HARs include response bodies by draining in‑flight CDP body fetches before writing the file. Previously we spawned unreferenced, never‑awaited tasks on loadingFinished and wrote the HAR on stop, dropping bodies still in flight.

- Track response‑body fetch tasks in a strong‑ref set; auto‑remove via done_callback.
- On stop, drain pending fetches with a 5s timeout using asyncio.wait; cancel stuck tasks best‑effort and log timeouts at debug so non‑cancellable CDP calls can’t extend shutdown.
- Log body‑fetch failures at debug instead of swallowing exceptions.
- Tests cover task tracking, draining so the body appears in the HAR, and bounded drain for a stuck fetch; the stuck‑fetch test now asserts on a captured task reference because the drain’s done_callback discards tasks from the set.

<sup>Written for commit 09442df1a8bf4c72bef6ec3068f9b8f067392baa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5495?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/browser/watchdogs/har_recording_watchdog.py
+++ b/browser_use/browser/watchdogs/har_recording_watchdog.py
@@ -7,6 +7,7 @@ and `record_har_mode` (full/minimal).
 
 from __future__ import annotations
 
+import asyncio
 import base64
 import hashlib
 import json
@@ -151,6 +152,7 @@ class HarRecordingWatchdog(BaseWatchdog):
 		super().__init__(*args, **kwargs)
 		self._enabled: bool = False
 		self._entries: dict[str, _HarEntryBuilder] = {}
+		self._pending_body_fetches: set[asyncio.Task[None]] = set()  # in-flight response body fetches
 		self._top_level_pages: dict[
 			str, dict
 		] = {}  # frameId -> {url, title, startedDateTime, monotonic_start, onContentLoad, onLoad}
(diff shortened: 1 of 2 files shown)


shashank-100:
Addressed both review findings:

**P2 (drain could still block past the timeout):** switched from `wait_for(gather(...))` to `asyncio.wait(pending, timeout=...)` — `wait_for` cancels the gather and then *awaits* that cancellation, so a CDP call that ignores cancellation could delay the HAR write beyond the drain window. `asyncio.wait` is strictly bounded; stuck tasks are now cancelled best-effort without being awaited. (`90c457e`)

**P3 (vacuous assertion in the stuck-fetch test):** valid — cancelled tasks self-discard from `_pending_body_fetches` via their done_callback, so the post-drain loop over the set iterated nothing. The test now captures the task reference *before* draining and asserts cancellation on it directly. (`09442df`)

Issue #5422, opened 2026-08-09
BrowserProfile leaks temp dirs into $TMPDIR on every construction (7741 dirs / 14GB observed)


axisrow:
## Problem

`BrowserProfile` leaks temp dirs into `$TMPDIR` on every construction, even when the browser never launches. On a real dev machine this accumulated to **7741 dirs / 14GB** in `$TMPDIR` over time (essentially all of it).

### Repro 1 — constructing a profile alone, no browser launched

```python
from browser_use.browser import BrowserProfile
p = BrowserProfile()   # -> browser-use-downloads-6d32053c  created on disk
p.get_args()           # -> browser-use-user-data-dir-9o06z_o9 created on disk
del p                  # -> both remain forever
```

### Repro 2 — a fully clean start()+kill() cycle, no errors

```
after ctor  : downloads-0ed9ea1a, downloads-f7f83289, user-data-dir-7y9nvp1q
after start : (same 3)
after kill  : (same 3)
LEAKED 3/3 dirs after a CLEAN start()+kill()
```

### Repro 3 — a real test run

36 purely-unit tests (no browser ever launched) leaked 14 dirs. The leak is linear in the number of `BrowserProfile` constructions, not tied to crashes/interruptions - it happens on every single run.

## Root causes

1. **`downloads_path` has no owner.** `set_default_downloads_path` (a `model_validator(mode='after')`) unconditionally does `self.downloads_path.mkdir(parents=True, exist_ok=True)` on every construction. No code path in the project ever calls `rmtree` on it.

2. **`LocalBrowserWatchdog`'s cleanup can't reach the main `user_data_dir`.** After a successful launch, `_temp_dirs_to_cleanup` is filtered by `'browseruse-tmp-' in currently_used_dir`, else reset to `[]`. But the primary dir is created with prefix `'browser-use-user-data-dir-'` (in `get_args()`/`_copy_profile()`); `'browseruse-tmp-'` is only used for `SingletonLock` retry fallbacks. So on the happy path the cleanup list is always empty. `_cleanup_temp_dir()` has the same prefix filter a second time.

[...]

PR #5504, opened 2026-08-20
fix(browser): clean up the temp .zip in _extract_extension on every path

### LINKED ISSUE #5364: Bug: _extract_extension leaks a temp .zip on extraction failure and unlinks a still-open handle

thisisanubhav:
Fixes #5364

> **Correction (updated after review):** an earlier version of this description claimed the header-skipping branch is the normal path for every `.crx`. **That was wrong** — I repeated it from the issue report without verifying. Python's `zipfile` corrects central-directory offsets for prepended bytes, so an intact `.crx` opens on the first attempt. The corrected analysis is below; the leak itself is real either way.

## Why?

Loading a browser extension should not silently strand files on disk, and cleanup should never swallow the error that explains why extraction failed.

`_extract_extension()` writes the CRX payload to a `delete=False` temp file, which means Python will never clean it up — the function owns that file's lifetime entirely. It only deleted it on the happy path, and deleted it in a way Windows rejects.

**When is that branch actually reached?** `zipfile` scans backwards for the end-of-central-directory record and adjusts for prepended bytes — the same support that makes self-extracting archives work. So an intact `.crx` opens directly despite its `Cr24` header:

```
prepend      0 bytes -> OPENS OK ['manifest.json']
prepend     12 bytes -> OPENS OK ['manifest.json']
prepend  65536 bytes -> OPENS OK ['manifest.json']
```

The fallback is reached only when the payload is damaged enough that the first attempt fails. And since the *only* difference between the two inputs is the leading header, a payload that defeats the first attempt defeats the second too. I probed this directly:

| CRX | branch 1 (in place) | branch 2 (header-skipped) |
|---|---|---|
| intact | OK | *never reached* |
| truncated payload | BadZipFile | BadZipFile |
| garbage payload | BadZipFile | BadZipFile |

No input produces "branch 1 fails, branch 2 succeeds". **So the realistic behaviour of this branch is: enter, fail, leak a `.zip` that nothing will ever remove.** Every corrupt or truncated extension download leaves one behind, permanently and silently — the same accumulation shape as #5422 (7741 leaked temp dirs / 14GB observed).

The Windows half is narrower than I first claimed but still real: the unlink ran while `NamedTemporaryFile` held the file open, which POSIX allows and Windows rejects with `PermissionError: [WinError 32]`.

## What changed

```python
temp_zip_path = None
try:
		temp_zip_path = Path(temp_zip.name)
		temp_zip.write(zip_data)

	with zipfile.ZipFile(temp_zip_path, 'r') as zip_ref:
finally:
	if temp_zip_path is not None:
		try:
			temp_zip_path.unlink(missing_ok=True)
		except OSError as cleanup_error:
			logger.warning(f'Failed to remove temp file {_log_pretty_path(temp_zip_path)}: {cleanup_error}')
```

Three things: cleanup moved into `finally`; the write handle is closed before the file is reopened and removed; and the unlink is itself guarded, so a cleanup failure cannot replace the `BadZipFile` that is already propagating — which matters most precisely on the Windows lock this change is hardening against. Also dropped the `import os` that became unused.

Extraction failures still propagate unchanged — this only fixes cleanup, it does not swallow errors.

## Tests

`tests/ci/test_extension_extraction.py` — real CRX containers, real method calls, no browser. `tempfile.tempdir` is pointed at a per-test directory so a leaked file is unambiguously attributable.

Each test **pins which branch it exercises** rather than assuming, because that assumption is exactly what went wrong in my first draft of this PR:

- **`test_no_temp_zip_left_behind_when_header_skip_extraction_fails`** — corrupt payload. Asserts via the traceback that the failure really came from the header-skipping branch before asserting no leak, so the leak assertion cannot pass vacuously.
- **`test_intact_crx_extracts_in_place_without_creating_a_temp_zip`** — asserts up front that plain `ZipFile` opens the file, then documents that this common case never creates a temp file at all.
- **`test_cleanup_failure_does_not_mask_the_extraction_error`** — forces the unlink to raise `PermissionError` and requires the caller to still see `BadZipFile`.

Behaviour across revisions, verified by checking out each version:

| test | upstream `main` | first push of this PR | now |
|---|---|---|---|
| header-skip leak | **FAIL** | PASS | PASS |
| intact CRX | PASS | PASS | PASS |
| cleanup masks error | PASS¹ | **FAIL** | PASS |

¹ passes on `main` because `main` never reaches the unlink on the failure path — masking was a defect the first version of *this PR* introduced, caught in review.

## Verification

- `pytest tests/ci -k "extension or profile or download"` → 93 passed
- `ruff format --check` and `ruff check` clean
- `pyright` on `profile.py`: 55 errors before and after — unchanged, all pre-existing

## Noted but not fixed (out of scope)

A CRX with a corrupt end-of-central-directory offset raises `OSError: [Errno 22] Invalid argument` from the first attempt, which `except zipfile.BadZipFile` does not catch, so it escapes uncaught and the fallback never runs. Filed separately as #5506.

(An earlier revision of this section named `ValueError: negative seek value` — that is what an in-memory `BytesIO` raises; a real file on disk raises `OSError`. Corrected after checking against the actual method rather than my probe.)

### DIFF
--- a/browser_use/browser/profile.py
+++ b/browser_use/browser/profile.py
@@ -1180,7 +1180,6 @@ async function initialize(checkInitialized, magic) {{
 
 	def _extract_extension(self, crx_path: Path, extract_dir: Path) -> None:
 		"""Extract .crx file to directory."""
-		import os
 		import zipfile
 
 		# Remove existing directory
@@ -1221,16 +1220,30 @@ async function initialize(checkInitialized, magic) {{
 				# Extract ZIP data
 				zip_data = f.read()
 
-			# Write ZIP data to temp file and extract
+			# Write ZIP data to temp file and extract.
+			# The temp file is created with delete=False, so cleanup is ours on every
+			# path — a corrupt or truncated payload makes extractall() raise, and the
+			# .zip would otherwise survive in the system temp dir indefinitely.
+			# The handle is also closed before the file is reopened and removed, since
+			# Windows refuses to unlink a file that still has an open handle.
+			temp_zip_path = None
+			try:
+				with tempfile.NamedTemporaryFile(suffix='.zip', delete=False) as temp_zip:
+					temp_zip_path = Path(temp_zip.name)
+					temp_zip.write(zip_data)
 
-			with tempfile.NamedTemporaryFile(suffix='.zip', delete=False) as temp_zip:
-				temp_zip.write(zip_data)
-				temp_zip.flush()
-
-				with zipfile.ZipFile(temp_zip.name, 'r') as zip_ref:
+				with zipfile.ZipFile(temp_zip_path, 'r') as zip_ref:
 					zip_ref.extractall(extract_dir)
-
-				os.unlink(temp_zip.name)
+			finally:
+				if temp_zip_path is not None:
+					try:
+						temp_zip_path.unlink(missing_ok=True)
+					except OSError as cleanup_error:
+						# Never let a cleanup failure replace the extraction error that
+						# is already propagating — the original exception is the useful
+						# one. Windows can still refuse the unlink if another process
+						# holds the file, so this is not merely theoretical.
+						logger.warning(f'Failed to remove temp file {_log_pretty_path(temp_zip_path)}: {cleanup_error}')
 
 	def detect_display_configuration(self) -> None:
 		"""
--- a/tests/ci/test_extension_extraction.py
+++ b/tests/ci/test_extension_extraction.py
@@ -1,0 +1,124 @@
+"""Tests for BrowserProfile._extract_extension() temp-file handling.
+
+`_extract_extension()` has two branches:
+
+1. Open the .crx directly with `ZipFile(crx_path)`. Python's zipfile scans
+   backwards for the end-of-central-directory record and corrects the
+   central-directory offsets for any prepended bytes (the same support that
+   makes self-extracting archives work), so an **intact** .crx opens here
+   despite its `Cr24` header. No temp file is involved.
+2. On BadZipFile, skip the CRX header, write the remaining payload to a
+   `delete=False` temp .zip, and extract that. This branch owns the temp
+   file's lifetime completely.
+
+Because the only difference between the two inputs is the leading header, and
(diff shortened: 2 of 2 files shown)


rajpratham1 (approved):
This PR fixes a legitimate resource-cleanup issue:

The temporary .zip is created with delete=False.
The new try/finally guarantees cleanup on success and extraction failure.
missing_ok=True makes cleanup safer if the file is already gone.
Cleanup errors are caught and logged instead of replacing the original extraction error.
The code specifically considers Windows file-handle behavior.
There are 124 lines of tests covering:
failed extraction → no temp ZIP left
valid CRX → no unnecessary temp ZIP
cleanup failure → original BadZipFile error remains visible
The PR has 109 checks shown as passing.
The existing review comment is marked Resolved.
One thing I checked carefully

The change from:

os.unlink(temp_zip.name)

to a finally block is the important part. Previously, if extractall() raised an exception, execution could skip os.unlink(), leaving the temporary ZIP behind. The new structure fixes exactly that.

thisisanubhav:
@rajpratham1 Merge this ?

rajpratham1:
> @rajpratham1 Merge this ?

Yes

thisisanubhav:
@CLAassistant can't sign in

thisisanubhav:
@rajpratham1 
if changes are approved then , kindly merge this PR.

thisisanubhav:
@rajpratham1 ??

Issue #5514, opened 2026-08-22
Bug: detect_pagination_buttons false positives — substring matching ('preview'→prev) and any 1-2 digit text → page_number


ksk2023:
### Browser Use Version

main @ 85ddbfed (2026-08-19)

### Bug Description

`DomService.detect_pagination_buttons` (browser_use/dom/service.py, around L1204-1217) produces a large number of false positives due to two independent issues in the classification logic. This is related to but distinct from #4620 (symbol ambiguity, already fixed) — the problems below affect word patterns and numeric buttons.

#### (a) Unanchored substring matching

```python
all_text = f'{text} {aria_label} {title} {class_name}'.strip()
...
elif any(pattern in all_text for pattern in prev_patterns):  # 'prev' ⊂ 'preview'
```

Any element whose combined text/aria/title/**class** merely *contains* a pattern word is classified:

- "Preview" link (`class="preview-link"`) → `prev` (because `'prev' in 'preview-link'`)
- "Last modified" sort control → `last`
- "Last name" column header → `last`
- "Next steps" wizard button → `next`
- Filter with `aria-label="price < 100"` → `prev` (because `'<'` is a pattern)

The class name being included in `all_text` makes this worse: CSS frameworks routinely contain tokens like `last-child`, `next-step`, `preview-`.

#### (b) Any 1-2 digit clickable text is a `page_number`

```python
elif text.isdigit() and len(text) <= 2 and role in ['button', 'link', '']:
    button_type = 'page_number'
```

`role=''` (i.e. no explicit ARIA role — the majority of plain HTML elements) qualifies, so virtually any short numeric label is classified:

| Element | Classified as |
|---|---|
| Datepicker calendar day cells "1" "15" "31" | `page_number` |
| Star-rating button "5" | `page_number` |
| Shoe-size selector "38" | `page_number` |
| Quantity stepper "10" | `page_number` |
| Cart badge "3" | `page_number` |

#### Verification

I ran a pure-logic reproduction of the function against 16 realistic clickable elements (real pagination + the widgets above): **16/16 were flagged as pagination buttons, 13 of them false positives**.

### Impact

`pagination_buttons` is surfaced to the LLM through `BrowserStateSummary`. On any page containing a datepicker, rating widget, size selector, or "preview"-style links, the model is told these are pagination controls — e.g. it may click calendar day "15" believing it's page 15, or waste steps on misidentified controls.

### Suggested fix

[...]

PR #5516, opened 2026-08-22
fix(dom): detect native disabled attribute in pagination buttons

ksk2023:
## Description

- Fixes pagination button disabled-state detection in `DomService.detect_pagination_buttons`.
- CDP's `DOM.getDocument` parses HTML boolean attributes (`<button disabled>`) as **empty strings**, not `'true'`. The old check `attributes.get('disabled') == 'true'` therefore never matched natively disabled controls, so `BrowserStateSummary.pagination_buttons` reported `is_disabled: false` for them — e.g. on page 1 of a paginated list where "Previous" is `<button disabled>`, the agent wasted a step clicking a disabled button.
- Fix: treat the **presence** of the `disabled` attribute as disabled regardless of value; keep requiring an explicit `'true'` for `aria-disabled` since `aria-disabled=""` is ambiguous per the ARIA spec.

## Checklist

- [x] With `aria-disabled="true"` → `is_disabled: True` (unchanged behavior)
- [x] With `class="...disabled..."` → `is_disabled: True` (unchanged behavior)
- [x] New: `<button disabled>` / `disabled=""` (CDP: `{'disabled': ''}`) → `is_disabled: True`
- [x] New: `disabled="true"` → `is_disabled: True`
- [x] `aria-disabled=""` alone → not disabled (deliberate, spec-ambiguous)
- [x] No attribute → not disabled

Regression tests added in `tests/ci/test_pagination_disabled_detection.py` (5 cases, all pass; the empty-string case fails on the previous implementation).

## Related

Companion analysis issues for other pagination-detection problems (substring false positives / numeric page buttons): #5514

---
## Summary by cubic
Detects native disabled attributes in pagination buttons correctly and avoids clicking disabled controls. Previously only `disabled == 'true'` counted; CDP emits empty strings, so `<button disabled>` appeared enabled. Now presence of `disabled` marks form controls as disabled; `aria-disabled` still requires `'true'`, and `disabled` on non-form elements is ignored.

- Updates `DomService.detect_pagination_buttons`: scope native `disabled` detection to form controls (`button`, `input`, `select`, `textarea`, `option`, `optgroup`, `fieldset`); keep `aria-disabled == 'true'` and `class`-based signals element-agnostic.
- Adds regression tests in `tests/ci/test_pagination_disabled_detection.py` covering empty-string/native cases, `aria-disabled`, inert `disabled` on `<a>`/`<div>`, and enabled controls.

<sup>Written for commit 4f46006289f06dfe181362f0183a9b4b2a1c2afc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5516?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/dom/service.py
+++ b/browser_use/dom/service.py
@@ -1196,8 +1196,13 @@ class DomService:
 			# as EMPTY STRINGS, not 'true'. So the presence of the attribute means disabled,
 			# regardless of value. aria-disabled="" is ambiguous per the ARIA spec, so only
 			# an explicit 'true' counts there.
+			# The native `disabled` attribute is only meaningful on form controls; on other
+			# elements (e.g. <a>, <div>) it is inert per the HTML spec, so scope the presence
+			# check to form controls and keep the aria-disabled/class checks element-agnostic.
+			tag = node.tag_name
+			is_form_control = tag in ('button', 'input', 'select', 'textarea', 'option', 'optgroup', 'fieldset')
 			is_disabled = (
-				node.attributes.get('disabled') is not None
+				(is_form_control and node.attributes.get('disabled') is not None)
 				or node.attributes.get('aria-disabled') == 'true'
 				or 'disabled' in class_name
 			)
--- a/browser_use/dom/service.py
+++ b/browser_use/dom/service.py
@@ -1192,8 +1192,12 @@ class DomService:
 			all_text = f'{text} {aria_label} {title} {class_name}'.strip()
 
 			# Check if it's disabled
+			# NOTE: CDP's DOM.getDocument parses HTML boolean attributes (e.g. <button disabled>)
+			# as EMPTY STRINGS, not 'true'. So the presence of the attribute means disabled,
+			# regardless of value. aria-disabled="" is ambiguous per the ARIA spec, so only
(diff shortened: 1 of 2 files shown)


ksk2023:
Addressed the review feedback in 4f46006:

**Scope the presence-based `disabled` check to form controls** — The native `disabled` attribute check now only applies to `button`, `input`, `select`, `textarea`, `option`, `optgroup`, and `fieldset` (via `node.tag_name`). On non-form elements like `<a disabled>` / `<div disabled>` the attribute is inert per the HTML spec, so those are no longer reported `is_disabled: True`. The `aria-disabled == 'true'` and class-based checks remain element-agnostic.

Added tests: `<a disabled>` (not disabled), `<div disabled>` (not disabled), `<a aria-disabled="true">` (disabled), `<input disabled>` (disabled) — 9/9 pass locally.

PR #5517, opened 2026-08-22
fix(tokens): prune pricing cache files after each refresh + drop deprecated utcnow

ksk2023:
## Description

- `TokenCost._fetch_and_cache_pricing_data` writes a new timestamped file `pricing_YYYYMMDD_HHMMSS.json` on every refresh, but `clean_old_caches()` had **no call sites anywhere** in the codebase — cache files accumulated indefinitely in `~/.cache/browser_use/token_cost` (one new file per refresh, ≥1/day for long-running services).
- This PR invokes `clean_old_caches()` after each successful cache write so only the most recent files (`keep_count=3`) survive.
- Also replaces the deprecated `datetime.utcnow()` in `DBStyleEntry.created_at` with `datetime.now(timezone.utc)` — `utcnow()` is deprecated since Python 3.12 and returns a naive datetime, inconsistent with the timezone-aware usage elsewhere in the codebase. Output now correctly carries `+00:00`.

## Testing

- `tests/ci/test_pricing_cache_cleanup.py`:
  - `test_clean_old_caches_prunes_to_keep_count` — 6 cache files → pruning keeps the 3 newest
  - `test_fetch_and_cache_invokes_cleanup` — network mocked; verifies `clean_old_caches` is invoked exactly once after a successful write and the pricing payload is cached
- Both pass on this branch; `test_fetch_and_cache_invokes_cleanup` fails without the fix.

---
## Summary by cubic
Prunes pricing cache files after each refresh and runs cleanup off the event loop to prevent I/O stalls. Also replaces deprecated `utcnow` with a timezone-aware timestamp.

- After a successful write, `_fetch_and_cache_pricing_data()` calls `_clean_old_caches_sync` via `anyio.to_thread.run_sync`; `clean_old_caches()` delegates to the same helper. Keeps only the 3 newest files by mtime; older files are removed on the next refresh. No config changes.
- `DBStyleEntry.created_at` now uses `datetime.now(timezone.utc)` so timestamps include `+00:00`.
- Tests: verify pruning orders by mtime (not filename) and add an end-to-end test using `httpx` with `pytest-httpserver` that confirms post-write cleanup and correct caching.

<sup>Written for commit f08a0f9c121abd29b3b1f50970d6dd9bcce1e1c1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5517?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/config.py
+++ b/browser_use/config.py
@@ -3,7 +3,7 @@
 import json
 import logging
 import os
-from datetime import datetime
+from datetime import datetime, timezone
 from functools import cache
 from pathlib import Path
 from typing import Any
@@ -248,7 +248,7 @@ class DBStyleEntry(BaseModel):
 
 	id: str = Field(default_factory=lambda: str(uuid4()))
 	default: bool = Field(default=False)
-	created_at: str = Field(default_factory=lambda: datetime.utcnow().isoformat())
+	created_at: str = Field(default_factory=lambda: datetime.now(timezone.utc).isoformat())
 
 
 class BrowserProfileEntry(DBStyleEntry):
--- a/browser_use/tokens/service.py
+++ b/browser_use/tokens/service.py
@@ -171,7 +171,10 @@ class TokenCost:
 
 			# Prune superseded cache files so the directory doesn't grow unboundedly
 			# across long-running sessions (one new timestamped file per refresh).
-			await self.clean_old_caches()
+			# The cleanup reads and parses every cached JSON file, so run it in a
+			# worker thread to keep the event loop responsive when the cache
+			# directory has a large backlog.
+			await anyio.to_thread.run_sync(self._clean_old_caches_sync)
 		except Exception as e:
 			logger.debug(f'Error fetching pricing data: {e}')
 			# Fall back to empty pricing data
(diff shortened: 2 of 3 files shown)


ksk2023:
Addressed the review feedback in f08a0f9:

1. **Event-loop blocking (P2)** — `clean_old_caches()` now delegates to a synchronous `_clean_old_caches_sync` helper executed via `anyio.to_thread.run_sync`, and the post-refresh call in `_fetch_and_cache_pricing_data` uses the same path. Reading and parsing the cache backlog no longer blocks the event loop.

2. **Prune test can't distinguish mtime vs filename ordering (P3)** — The seeded files now deliberately invert filename order vs age (the oldest file has the lexicographically **largest** name), so the test fails if pruning ever sorts by filename instead of `st_mtime`. Verified by injecting a filename-sort implementation locally: the test fails as intended.

3. **Real objects + local HTTP (P3)** — `test_fetch_and_cache_invokes_cleanup_end_to_end` now serves the pricing payload from `pytest-httpserver`, uses the real httpx client and the real `clean_old_caches`, seeds 3 stale cache files from the same source URL, and asserts the final file list end to end (3 files remain, including the newly written one).

Both tests pass locally (`uv run pytest -vxs tests/ci/test_pricing_cache_cleanup.py` → 2 passed).

PR #5521, opened 2026-08-23
fix: avoid double-counting timed-out steps

### LINKED ISSUE #5513: Bug: always-true step compensation in _execute_step double-increments n_steps after timeout cancellation (skips a step)
### Browser Use Version

main @ 85ddbfed (2026-08-19)

### Bug Description

`Agent._execute_step` (browser_use/agent/service.py, around L2479-2482) contains a step-count compensation that is a tautology:

```python
if self.state.n_steps == step + 1:
    self.state.n_steps += 1
```

`_execute_step` is always called with `step == current_step == n_steps - 1`, so `n_steps == step + 1` is **always true** — it cannot distinguish "step() already incremented" from "step() was cancelled before incrementing".

Meanwhile, when `asyncio.wait_for(step(), timeout)` times out, the cancelled `step()` coroutine still runs its cleanup path: `_finalize()` executes in the cancellation unwinding and performs `self.state.n_steps += 1` (around L1409). Control then returns to the exception handler in `_execute_step`, whose compensation condition is still true, so it increments **again**.

#### Failure scenario

1. `step()` exceeds the per-step timeout and is cancelled by `wait_for`.
2. The `finally`/cleanup path in `step()` still executes `_finalize()` → `n_steps += 1`.
3. The handler in `_execute_step` fires the always-true compensation → `n_steps += 1` again.
4. Result: `n_steps` jumps by 2 for one actual step — the agent effectively **loses one step** of its `max_steps` budget per timeout, and can terminate earlier than configured.

#### Verification

I reproduced the logic with a small simulation (mocked state, one cancellation):

```
timeout during step() → _finalize increments: n_steps = 2
_execute_step handler: n_steps == step + 1 → 2 == 1+1 → True → n_steps = 3
```

The compensation condition evaluates `True` both when `step()` already incremented and when it did not — it provides no protection while adding a double-increment path.

### Suggested fix

Track the increment with an explicit sentinel instead of inferring it from an (always-true) numeric identity, e.g.:

```python
# in step(): set a flag right after n_steps += 1
self.state._step_incremented = True

# in _execute_step() cancellation handler:
if not self.state._step_incremented:
    self.state.n_steps += 1
```

or compare against a baseline captured at step start rather than `n_steps` itself.

Happy to submit a PR if the approach sounds good — but since this touches the core step loop I wanted to discuss first.

daleselaji-dev:
## Problem

When a timed-out step completes its cancellation cleanup, the agent can increment `n_steps` twice and consume two step-budget units

## Root Cause

The timeout handler compared the post-cancellation counter to the step number, so the condition stayed true after `_finalize()` had already advanced the counter

## Solution

Capture the counter before awaiting the step and compensate only when cancellation left it unchanged

## Changes

- Added a start-of-step counter snapshot
- Added regression coverage for both cleanup-counted and cleanup-skipped timeouts

## Testing

- Baseline reproduction on main: cleanup-counted timeout advanced `n_steps` from 1 to 3
- Focused final: `python -m pytest -o addopts='' tests/ci/test_agent_step_timeout.py -q` -> 2 passed
- `pyright tests/ci/test_agent_step_timeout.py` -> 0 errors
- Ruff check on changed files -> production file retains one pre-existing `ASYNC240` at line 2719; regression file passes
- `python -m compileall` on changed files -> passed
- `git diff --check` -> passed
- Existing `tests/ci/test_beta_agent.py` baseline has environment-specific packaged-binary failures and is not claimed as a full pass

## Compatibility/Risk

The change preserves the existing fallback when timeout cancellation skips finalization and prevents only the duplicate increment when finalization already advanced the counter

## Notes for Reviewer

The regression uses the same `asyncio.wait_for` timeout path and a cancellation cleanup increment to reproduce the reported double count. AI assistance was used for navigation and validation; I reviewed the final diff

## Linked Issue

Fixes #5513

## Type

Bug fix

## Caveats

No browser or LLM service is required for this deterministic counter regression


MagMueller:
Thanks for adding coverage and keeping this current. I reran the two realistic cases on current main; both pass before this patch. `_execute_step` has one caller and always receives `step = n_steps - 1`, so the existing `step + 1` check equals the proposed pre-call snapshot. The only test that fails on main uses `step=7` with `n_steps=1`, which the run loop cannot produce. Closing as a no-op instead of adding an impossible invariant.

PR #5530, opened 2026-08-23
fix(tools): verify automatic new-tab switches

### LINKED ISSUE #5529: Bug: failed automatic new-tab switch is reported as successful after click
### Browser Use Version

Current `main` at `85ddbfedf609166b2d2c76c3d80506649fee82a9`.

### Bug Description

When a click opens a new tab, `_detect_new_tab_opened()` dispatches `SwitchTabEvent` and then calls:

```python
await switch_event.event_result(raise_if_any=False, raise_if_none=False)
return f'. Automatically switched to new tab (tab_id: {new_tab_id}).'
```

`on_SwitchTabEvent` returns the focused `TargetID` on every successful path. If the handler fails, `raise_if_any=False` suppresses that exception and `event_result(...)` returns `None`, but the helper ignores the result and still tells the model that the switch succeeded.

The click itself did succeed, so the entire click should not become an error. However, the false success message makes the next step reason as though the browser is focused on the new tab when it may still be on the original page.

This is distinct from #5486: that issue covers the explicit `switch` action, while this path is the automatic post-click helper used by both coordinate and index clicks.

### Reproduction

An offline test can simulate:

1. `get_tabs()` returning one tab before the click and a second tab afterwards.
2. The click event completing successfully.
3. The automatic `SwitchTabEvent` completing with no result (the suppressed-handler-failure shape).

Current output:

```text
Clicked on coordinate 10, 20. Automatically switched to new tab (tab_id: 1234).
```

The new tab was detected, but no successful switch result was observed.

### Expected Behavior

Only report `Automatically switched...` when `event_result()` returns a real target ID. If the switch handler raises or produces no result, preserve the successful click and fall back to the existing truthful guidance:

```text
Note: This opened a new tab (tab_id: 1234) - switch to it if you need to interact with the new page.
```

### Suggested Fix

Read and validate the event result, using `raise_if_any=True` so handler failures reach the existing fallback branch. Add a regression test for the no-result path.

green3sf:
## Summary

- surface `SwitchTabEvent` handler failures to the existing post-click fallback
- report an automatic new-tab switch only when the event returns a real target ID
- add an offline regression test for a successful click followed by a no-result switch

## Why

Both coordinate and index clicks call `_detect_new_tab_opened()`. The helper previously used `raise_if_any=False` and ignored the event result, so a failed automatic switch was still written into the model-visible result as `Automatically switched to new tab`.

The click itself remains successful. When automatic switching fails, the result now truthfully says that a new tab opened and the agent should switch to it if needed.

This is separate from #5486, which covers the explicit `switch` action rather than this post-click helper.

## Validation

- `pytest -q tests/ci/test_coordinate_clicking.py` — 34 passed
- `ruff check browser_use/tools/service.py tests/ci/test_coordinate_clicking.py`
- `ruff format --check browser_use/tools/service.py tests/ci/test_coordinate_clicking.py`
- `pyright browser_use/tools/service.py tests/ci/test_coordinate_clicking.py` — 0 errors
- `git diff --check`

No UI changes; screenshot not applicable.

Fixes #5529

---
## Summary by cubic
Verifies automatic new-tab switches after clicks and stops false success reporting. Old behavior: always claimed an automatic switch even when `SwitchTabEvent` failed or returned no target. New behavior: reports a switch only when a target ID is returned; otherwise notes that a new tab opened.

- Applies to coordinate- and index-based click flows via `_detect_new_tab_opened`; uses `raise_if_any=True` and validates the returned target ID.
- Adds `_new_tab_switch_message` and `_format_new_tab_message`, and updates `_detect_new_tab_opened` to use them.
- Adds offline tests covering successful and failed switch outcomes via an event bus (no browser session), ensuring failures aren’t reported as success.
- No API/UI changes; no migrations. Fixes #5529.

<sup>Written for commit f56e0d0b74ac3a33e39ea9a6854ce617313ac90e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5530?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/tools/service.py
+++ b/browser_use/tools/service.py
@@ -82,6 +82,16 @@ def _format_new_tab_message(new_tab_id: str, switched_target_id: str | None) -> 
 	return f'. Note: This opened a new tab (tab_id: {new_tab_id}) - switch to it if you need to interact with the new page.'
 
 
+async def _new_tab_switch_message(new_tab_id: str, switch_event: SwitchTabEvent) -> str:
+	"""Report an automatic switch only after the event returns its target ID."""
+	try:
+		await switch_event
+		switched_target_id = await switch_event.event_result(raise_if_any=True, raise_if_none=False)
+	except Exception:
+		switched_target_id = None
+	return _format_new_tab_message(new_tab_id, switched_target_id)
+
+
 # Default header/footer templates for save_as_pdf, mirroring the metadata that
 # Chrome's own Print dialog renders by default: the date in the header and the
 # page URL + page numbers in the footer. Chrome injects values into elements
@@ -651,9 +661,7 @@ class Tools(Generic[Context]):
 					# Auto-switch to the new tab so the agent can immediately interact with it
 					try:
 						switch_event = browser_session.event_bus.dispatch(SwitchTabEvent(target_id=new_tab.target_id))
-						await switch_event
-						new_target_id = await switch_event.event_result(raise_if_any=True, raise_if_none=False)
-						return _format_new_tab_message(new_tab_id, new_target_id)
+						return await _new_tab_switch_message(new_tab_id, switch_event)
 					except Exception:
 						return _format_new_tab_message(new_tab_id, None)
 			except Exception:
--- a/browser_use/tools/service.py
+++ b/browser_use/tools/service.py
@@ -75,6 +75,13 @@ Context = TypeVar('Context')
 T = TypeVar('T', bound=BaseModel)
 
 
+def _format_new_tab_message(new_tab_id: str, switched_target_id: str | None) -> str:
+	"""Describe a newly opened tab without claiming an unconfirmed automatic switch."""
+	if switched_target_id:
+		return f'. Automatically switched to new tab (tab_id: {new_tab_id}).'
+	return f'. Note: This opened a new tab (tab_id: {new_tab_id}) - switch to it if you need to interact with the new page.'
+
+
 # Default header/footer templates for save_as_pdf, mirroring the metadata that
 # Chrome's own Print dialog renders by default: the date in the header and the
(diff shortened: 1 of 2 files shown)


green3sf:
Addressed the Cubic test-design finding in `facd0730`.

The hand-rolled `BrowserSessionStub` and `EventBus` are removed. The production path still awaits the real Bubus event result, while the presentation decision is now isolated in `_format_new_tab_message`, so the regression test covers the exact boundary without approximating browser/event semantics.

Validation:
- `pytest -q tests/ci/test_coordinate_clicking.py` — 34 passed
- Ruff check and format check — passed
- targeted Pyright — 0 errors, 0 warnings

green3sf:
Addressed the latest Cubic test-coverage finding in `f56e0d0b`.

The production event-result boundary now lives in `_new_tab_switch_message`, and the regression tests exercise it with a real Bubus `EventBus` rather than browser or event stubs:

- a successful handler returns the switched target ID and reports the automatic switch;
- a failing handler raises through `event_result(raise_if_any=True)` and falls back to the manual-switch note.

Validation:
- `pytest -q tests/ci/test_coordinate_clicking.py` — 35 passed
- Ruff check and format check — passed
- targeted Pyright — 0 errors, 0 warnings
- `git diff --check` — passed

green3sf:
A brief follow-up: I am stepping back from this change and will leave this PR closed for now. Thank you for your time and the review feedback.

PR #5550, opened 2026-08-25
fix(browser): prevent temp file leak on crx extraction failure

### LINKED ISSUE #5364: Bug: _extract_extension leaks a temp .zip on extraction failure and unlinks a still-open handle

sundeep8967:
## Description
Fixes #5364 by using a `try...finally` block to ensure the temp `.zip` file created during extension extraction is unconditionally cleaned up, even if `zip_ref.extractall()` raises a `BadZipFile` or any other exception.

Additionally, the temporary file handle is now explicitly closed before being opened by `zipfile.ZipFile` to prevent `PermissionError: [WinError 32] The process cannot access the file because it is being used by another process` from occurring on Windows systems during the final deletion.

Included a regression test verifying that no `tmp*.zip` survives a failed extraction.


sundeep8967:
Updated to ensure `temp_zip.close()` is safely executed in the `finally` block before calling `os.unlink`, preventing file handle leaks and Windows permission errors on write failures. Also updated the test to use pytest's `tmp_path` fixture and an isolated temporary directory to prevent cross-test race conditions.

PR #5556, opened 2026-08-27
feat(llm): add Volcengine Ark (Doubao) chat provider

chyroc:
## Summary

Adds `ChatVolcengine`, a provider for **Volcengine Ark** (ByteDance's Doubao models).

Ark exposes an OpenAI-compatible `/chat/completions` endpoint, so this follows
`ChatOpenRouter` and reuses `OpenAIMessageSerializer` — **no new SDK dependency**.
I modelled it on OpenRouter rather than DeepSeek because OpenRouter already reports
usage and uses the `json_schema` response format, both of which Ark supports.

## What I verified against the live API before writing the code

Rather than assume Ark's OpenAI compatibility, I checked each behaviour the
implementation depends on:

- **`response_format` accepts strict `json_schema`** → structured output goes
  through the schema path directly instead of a tool-call detour
- **`usage.prompt_tokens_details.cached_tokens` is reported** → the
  OpenRouter-style `_get_usage` extraction maps over unchanged
- **`reasoning_effort` grades thinking depth** → `low` yields 61 reasoning
  tokens, `high` yields 107 on `doubao-seed-2-1-pro-260628`, so it's forwarded
  as a passthrough param
- **Tool calling returns standard `tool_calls`**

Two things worth calling out, both of which cost me a 404 before I caught them:

1. **Model IDs need their full version suffix.** The console's short names 404 —
   `doubao-seed-2-1-pro` → `404 InvalidEndpointOrModel.NotFound`, while
   `doubao-seed-2-1-pro-260628` → `200`. `doubao-seed-evolving` is the only
   unversioned ID that resolves. This is noted in the class docstring and docs.
2. **`api_key` falls back to `ARK_API_KEY` explicitly.** Without that,
   `AsyncOpenAI` would read `OPENAI_API_KEY` and send an unrelated credential to
   Ark. Same approach as `ChatMistral` / `ChatAzureOpenAI` / `ChatVercel`.

## Testing

`tests/ci/models/test_llm_volcengine.py` adds 5 cases covering defaults,
`reasoning_effort` forwarding, cached-token usage, strict `json_schema` output,
and the env-var fallback.

- `pytest tests/ci/models/` → **57 passed, 7 skipped** (no regressions)
- `ruff check` → All checks passed; `ruff format --check` → already formatted
- `pyright` → **0 errors**

I mutation-tested the new cases to confirm they aren't passing vacuously —
removing the `ARK_API_KEY` fallback, forcing `temperature` to always be sent, and
dropping `reasoning_effort` forwarding each fail exactly one corresponding test.

**Real end-to-end** through `ChatVolcengine` (not mocked), against the live API:

| Path | Result |
|---|---|
| Text | `pong` — usage prompt=52, completion=32, total=84, cached=0 |
| Structured output | validated `Person(name='Alice', age=30)` |
| `reasoning_effort='high'` | correct answer `3` |

Happy to adjust naming, the default model, or drop the example if you'd rather
keep `examples/models/` smaller.

---
## Summary by cubic
Adds `ChatVolcengine`, a provider for Volcengine Ark (ByteDance's Doubao models), exported from `browser_use` and `browser_use.llm`. Ark is OpenAI-compatible, so the implementation reuses `OpenAIMessageSerializer` and `AsyncOpenAI` with no new SDK dependency, and all behaviors were verified against the live API.

Includes an example, docs entry, and 7 test cases covering defaults, `reasoning_effort` forwarding, cached-token usage, strict `json_schema` output, and api-key handling; the existing suite passes with no regressions.

**Notes**
- Model IDs must carry the full version suffix — the console's short names (e.g. `doubao-seed-2-1-pro`) return 404; `doubao-seed-evolving` is the only unversioned ID that resolves.
- `api_key` falls back to `ARK_API_KEY`, and a missing key raises instead of leaking `OPENAI_API_KEY` to Ark; the example exits 1 when the key is absent.
- Errors `ainvoke` raises itself keep their own status code instead of being re-wrapped.

<sup>Written for commit b54248f77834930d0983e6056c0f655c9c41c240. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5556?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/__init__.py
+++ b/browser_use/__init__.py
@@ -68,6 +68,7 @@ if TYPE_CHECKING:
 	from browser_use.llm.openai.chat import ChatOpenAI
 	from browser_use.llm.openrouter.chat import ChatOpenRouter
 	from browser_use.llm.vercel.chat import ChatVercel
+	from browser_use.llm.volcengine.chat import ChatVolcengine
 	from browser_use.sandbox import sandbox
 	from browser_use.tools.service import Controller, Tools
 
@@ -106,6 +107,7 @@ _LAZY_IMPORTS = {
 	'ChatOllama': ('browser_use.llm.ollama.chat', 'ChatOllama'),
 	'ChatOpenRouter': ('browser_use.llm.openrouter.chat', 'ChatOpenRouter'),
 	'ChatVercel': ('browser_use.llm.vercel.chat', 'ChatVercel'),
+	'ChatVolcengine': ('browser_use.llm.volcengine.chat', 'ChatVolcengine'),
 	# LLM models module
 	'models': ('browser_use.llm.models', None),
 	# Sandbox execution
@@ -163,6 +165,7 @@ __all__ = [
 	'ChatOllama',
 	'ChatOpenRouter',
 	'ChatVercel',
+	'ChatVolcengine',
 	'Tools',
 	'Controller',
 	# LLM models module
--- a/browser_use/llm/__init__.py
+++ b/browser_use/llm/__init__.py
@@ -41,6 +41,7 @@ if TYPE_CHECKING:
 	from browser_use.llm.openai.chat import ChatOpenAI
 	from browser_use.llm.openrouter.chat import ChatOpenRouter
 	from browser_use.llm.vercel.chat import ChatVercel
+	from browser_use.llm.volcengine.chat import ChatVolcengine
 
 	# Type stubs for model instances - enables IDE autocomplete
 	openai_gpt_4o: ChatOpenAI
@@ -94,6 +95,7 @@ _LAZY_IMPORTS = {
 	'ChatOpenAI': ('browser_use.llm.openai.chat', 'ChatOpenAI'),
 	'ChatOpenRouter': ('browser_use.llm.openrouter.chat', 'ChatOpenRouter'),
 	'ChatVercel': ('browser_use.llm.vercel.chat', 'ChatVercel'),
+	'ChatVolcengine': ('browser_use.llm.volcengine.chat', 'ChatVolcengine'),
 }
 
 # Cache for model instances - only created when accessed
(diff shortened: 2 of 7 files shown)


chyroc:
Thanks — all three were real, fixed in 1eaa639.

**P1** was the worst of them and I'd claimed the opposite in the PR description. Reproduced it: with `ARK_API_KEY` unset and `OPENAI_API_KEY` set, `get_client()` built a client carrying `sk-...` from the OpenAI env var pointed at `ark.cn-beijing.volces.com`. Resolving the key returned `None` and the SDK filled in its own default. Now it raises `ModelProviderError` before the client is constructed.

**P2** reproduced as well — raising `status_code=500` on unparseable structured output and observing `502` at the caller, because the trailing `except Exception` re-wrapped it. Added an `except ModelProviderError: raise` ahead of it. Worth noting the same pattern is in `openrouter/chat.py`, which is what I based this on; happy to send a separate PR for that one if you want it.

**P3** fixed, now `exit(1)` matching `moonshot.py`.

Both provider fixes have tests that fail when the fix is reverted (I checked). Suite is 59 passed / 7 skipped, ruff and pyright clean, and I re-ran the live end-to-end after the change: text, structured output, and `reasoning_effort=high` all still work.

MagMueller:
what would you want Doubao to automate beyond the Hacker News example?

ashopnil06-boop (approved):

ashopnil06-boop:
What is this I don't understand

ashopnil06-boop (approved):

PR #5574, opened 2026-08-28
fix(browser): only apply Enter delay for real Enter/Return keys

### LINKED ISSUE #5573: Bug: `send_keys` applies the Enter-specific delay to text containing `enter`
### Browser Use Version

0.13.8, main@35fccbb418f4be1c4883f3256bdc1c30977a9c4f

## Summary

In `browser_use/browser/watchdogs/default_action_watchdog.py`, `DefaultActionWatchdog.on_SendKeysEvent` applies its Enter-specific post-send delay whenever the input contains the substring `enter` or `return`.

For example:

```python
SendKeysEvent(keys="center")
```

is correctly dispatched as ordinary text, but the handler subsequently treats it as Enter-related input and waits an additional `0.1` seconds.

The same behavior affects ordinary text such as `enterprise` and `returning`.

## Steps to reproduce

1. Check out Browser Use `main` at commit `35fccbb418f4be1c4883f3256bdc1c30977a9c4f`.

```bash
uv sync --all-extras --dev
```

3. Create `tests/ci/browser/test_send_keys_enter_substring.py` with the test from the **Failing Python Code** field.

4. Run:

```bash
uv run pytest \
    tests/ci/browser/test_send_keys_enter_substring.py \
    -q
```

5. Observe that the new assertion fails.

## Expected behavior

The Enter-specific delay should be applied only when the requested key is Enter or Return, including supported shortcuts whose main key is Enter.

Sending ordinary text such as `"center"` should use only the existing per-character delays and should not request the additional `0.1`-second navigation delay.

## Actual behavior

The text is dispatched correctly, but an additional Enter-specific delay is requested:

```python
enter_delays == [0.1]
```

This occurs even though no Enter or Return key was sent.

## Impact

Calls to the public `send_keys` action incur an unnecessary `100 ms` delay whenever ordinary text contains `enter` or `return`.

The input content is not corrupted, but repeated calls involving words such as `center`, `enterprise`, or `returning` introduce avoidable latency into browser tasks.

## Root cause

After dispatching the requested keys, the handler uses substring matching to decide whether to wait for possible navigation:

```python
if 'enter' in event.keys.lower() or 'return' in event.keys.lower():
    await asyncio.sleep(0.1)
```

Substring membership does not distinguish the Enter and Return key aliases from ordinary text containing those character sequences.

## Possible fix direction

Base the post-send delay on the parsed key semantics instead of substring membership.

The delay should remain for exact `Enter` or `Return` input and supported combinations whose main key is Enter, but should not run for unrelated text containing those substrings.

- `Enter` and `Return`;
- case variants such as `ENTER`;
- a supported shortcut whose main key is Enter;
- ordinary text such as `center`;
- ordinary text such as `returning`;
- text without either substring.

### Failing Python Code

```python
import asyncio
from types import SimpleNamespace

from browser_use.browser.events import SendKeysEvent
from browser_use.browser.watchdogs.default_action_watchdog import (
    DefaultActionWatchdog,
)


def test_send_keys_text_containing_enter_has_no_enter_delay(
    monkeypatch,
):
    delays = []

    async def sleep(delay):
        delays.append(delay)

    monkeypatch.setattr(asyncio, "sleep", sleep)

    class Input:
        async def dispatchKeyEvent(
            self,
            params=None,
            session_id=None,
        ):
            return None

    cdp_session = SimpleNamespace(
        cdp_client=SimpleNamespace(
            send=SimpleNamespace(Input=Input()),
        ),
        session_id="session-1",
    )

    class BrowserSession:
        async def get_or_create_cdp_session(self, focus=False):
            return cdp_session

    async def dispatch_key_event(
        _session,
        event_type,
        key,
        modifiers=0,
    ):
        return None

    watchdog = SimpleNamespace(
        logger=SimpleNamespace(
            info=lambda *args, **kwargs: None,
        ),
        _dispatch_key_event=dispatch_key_event,
        _get_char_modifiers_and_vk=lambda char: (0, 0, char),
        _get_key_code_for_char=lambda key: f"Key{key.upper()}",
    )

    asyncio.run(
        DefaultActionWatchdog.on_SendKeysEvent(
            watchdog,
            SendKeysEvent(keys="center"),
        )
    )

    enter_delays = [delay for delay in delays if delay == 0.1]
    assert enter_delays == []
```

### LLM Model

Not applicable; reproduced by directly invoking deterministic keyboard-event handling code.

macOS 15.7.3, arm64; Python 3.12.13. No browser is started.

### Full DEBUG Log Output

```shell
FAILED tests/ci/browser/test_send_keys_enter_substring.py::test_send_keys_text_containing_enter_has_no_enter_delay

AssertionError: assert [0.1] == []

Left contains one more item: 0.1

- []
+ [
+     0.1,
+ ]

1 failed in 0.06s
```

leilei3167:
## Summary

`on_SendKeysEvent` waited 0.1s whenever `event.keys` contained the substrings `enter` or `return`. Ordinary text like `center`, `enterprise`, or `returning` paid that delay even though no Enter key was sent.

## Fix

Track whether the parsed key path actually dispatched Enter (exact Enter/Return, combo whose main key is Enter, or `\n`/`\r` in text) and only sleep then.

## Test

`tests/ci/browser/test_send_keys_enter_substring.py` covers Enter/Return/ENTER, `Control+Enter`, and ordinary text cases.

Fixes #5573

---
## Summary by cubic
Stops the 0.1s post-send delay from firing on ordinary text like `center`, `enterprise`, or `returning`; the delay now applies only when an Enter/Return key is actually dispatched.

- Tracks whether the parsed key path sent Enter (exact key, modifier combos, or `\n`/`\r`).
- Adds regression tests for Enter/Return, `Control+Enter`, newline characters, and plain-text cases.
- Fixes #5573.

<sup>Written for commit 78b1eb74aa6108ed7c4a58504f6833352f4e619c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5574?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


leilei3167:
@MagMueller ready for review when you have a minute. Happy to adjust if you'd rather keep a string-based check with word boundaries instead of tracking the parsed Enter path.

leilei3167:
Hi — just a friendly ping on this PR whenever you have a moment. Happy to address any feedback. Thanks!

PR #5582, opened 2026-08-28
fix(browser): support literal plus in send_keys

### LINKED ISSUE #5569: Bug: `send_keys` treats a literal `+` as an empty key combination
### Browser Use Version

0.13.8, main@35fccbb418f4be1c4883f3256bdc1c30977a9c4f

## Summary

In `browser_use/browser/watchdogs/default_action_watchdog.py`, `DefaultActionWatchdog.on_SendKeysEvent` treats every `keys` value containing `+` as a key combination.

When the requested key is the literal plus character:

```python
SendKeysEvent(keys="+")
```

the handler splits it into two empty components. It consequently dispatches empty key-down and key-up events instead of inserting `+`.

The same parsing problem affects shortcuts whose main key is plus, such as `Control++`.

## Steps to reproduce

1. Check out Browser Use `main` at commit `35fccbb418f4be1c4883f3256bdc1c30977a9c4f`.

```bash
uv sync --all-extras --dev
```

3. Create `tests/ci/browser/test_send_keys_literal_plus.py` with the test from the **Failing Python Code** field.

4. Run:

```bash
uv run pytest \
    tests/ci/browser/test_send_keys_literal_plus.py \
    -q
```

5. Observe that the new test fails.

## Expected behavior

A literal `+` should follow the existing text-character path and emit a character event so that the focused element receives the plus character:

```python
{
    "type": "char",
    "text": "+",
}
```

It should not dispatch events whose key value is empty.

## Actual behavior

No character event is emitted. Instead, the combination path produces four events with an empty key:

```python
[
    ("keyDown", "", 0),
    ("keyDown", "", 0),
    ("keyUp", "", 0),
    ("keyUp", "", 0),
]
```

As a result, the requested plus character is not entered.

## Impact

The public `send_keys` action cannot enter a literal plus character into a focused element. It also cannot correctly represent shortcuts whose main key is plus, such as `Control++`.

This can affect text, numeric expressions, formulas, search queries, and keyboard shortcuts containing `+`.

## Root cause

The handler enters combination parsing whenever `+` occurs anywhere in the input:

```python
keys = event.keys
if "+" in keys:
    parts = keys.split("+")
    normalized_parts = []
    for part in parts:
        part_lower = part.strip().lower()
        normalized = key_aliases.get(part_lower, part)
        normalized_parts.append(normalized)
    normalized_keys = "+".join(normalized_parts)
```

It then splits the normalized value again:

```python
if "+" in normalized_keys:
    parts = normalized_keys.split("+")
    modifiers = parts[:-1]
    main_key = parts[-1]
```

For `keys="+"`, both components are empty, so `main_key` is `""`.

The character-handling path already contains key-code support for `+`, but the initial substring check prevents a literal plus from reaching that path.

## Possible fix direction

Only select combination parsing when the input forms a valid shortcut with recognized modifier components and a non-empty main key. Literal plus input and shortcuts whose main key is plus should be handled explicitly.

- a literal `+`;
- `Control++` or its supported equivalent;
- an existing shortcut such as `Control+a`;
- a non-combination key such as `Enter`;
- no empty key values in emitted CDP events.

### Failing Python Code

```python
import asyncio
from types import SimpleNamespace

from browser_use.browser.events import SendKeysEvent
from browser_use.browser.watchdogs.default_action_watchdog import (
    DefaultActionWatchdog,
)


def test_send_keys_literal_plus_dispatches_char_event():
    recorded_params = []
    dispatched_keys = []

    class Input:
        async def dispatchKeyEvent(
            self,
            params=None,
            session_id=None,
        ):
            recorded_params.append(params or {})

    cdp_session = SimpleNamespace(
        cdp_client=SimpleNamespace(
            send=SimpleNamespace(Input=Input()),
        ),
        session_id="session-1",
    )

    class BrowserSession:
        async def get_or_create_cdp_session(self, focus=False):
            return cdp_session

    async def dispatch_key_event(
        _session,
        event_type,
        key,
        modifiers=0,
    ):
        dispatched_keys.append((event_type, key, modifiers))

    watchdog = SimpleNamespace(
        logger=SimpleNamespace(
            info=lambda *args, **kwargs: None,
        ),
        _dispatch_key_event=dispatch_key_event,
    )

    asyncio.run(
        DefaultActionWatchdog.on_SendKeysEvent(
            watchdog,
            SendKeysEvent(keys="+"),
        )
    )

    assert any(
        params.get("type") == "char"
        and params.get("text") == "+"
        for params in recorded_params
    ), (recorded_params, dispatched_keys)
```

### LLM Model

Not applicable; reproduced by directly invoking deterministic keyboard-event handling code.

macOS 15.7.3, arm64; Python 3.12.13. No browser is started.

### Full DEBUG Log Output

```shell
FAILED tests/ci/browser/test_send_keys_literal_plus.py::test_send_keys_literal_plus_dispatches_char_event

AssertionError: (
    [],
    [
        ("keyDown", "", 0),
        ("keyDown", "", 0),
        ("keyUp", "", 0),
        ("keyUp", "", 0),
    ],
)
assert False

1 failed in 0.07s
```

IENVYshanks:
Fix send_keys handling for literal plus characters.

Previously, send_keys(keys="+") was parsed as an empty key combination, causing empty key events instead of inserting +. Shortcuts such as Control++ were also parsed incorrectly.

This change:

Sends literal + through the character-input path.
Supports shortcuts like Control++.
Preserves existing shortcuts such as Control+a and special keys like Enter.
Prevents empty key events from being dispatched.
Testing
Added regression tests for +, C++, and Control++.
Verified existing Control+a and Enter behavior.
Focused test suite: 4 passed.
Fixes #5569

---
## Summary by cubic
Fixes `send_keys` handling so a literal `+` is sent as a character instead of an empty key event, and shortcuts like `Control++` work correctly. Previously, `send_keys(keys="+")` dispatched empty key events and `Control++` was parsed incorrectly.

- `+` and text containing `+` (like `C++`) now go through the character-input path.
- `Control++` is recognized as a shortcut with `+` as the main key.
- Existing shortcuts like `Control+a` and special keys like `Enter` behave as before.
- Adds regression tests for `+`, `C++`, and `Control++`.

Fixes #5569.

<sup>Written for commit 8043fec5e6da1ee5ed9b90aefe60f8e1f41adc05. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5582?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


IENVYshanks:

IENVYshanks:
recheck

PR #5593, opened 2026-08-29
fix(utils): redact every value when domains share a placeholder name

### LINKED ISSUE #5592: Bug: sensitive_data values leak when two domains share a placeholder name (collect_sensitive_data_values collision)
### Describe the bug

`collect_sensitive_data_values()` flattens domain-scoped `sensitive_data` into a `{placeholder: value}` dict. When two domains use the **same placeholder name** (the core use case of domain scoping — login credentials per site), the values silently overwrite each other, and the dropped value is never redacted from LLM-visible text.

### Steps to reproduce

```python
from browser_use.utils import collect_sensitive_data_values, redact_sensitive_string

sensitive_data = {
    'github.com': {'password': 'gh-secret-123'},
    'gitlab.com': {'password': 'gl-secret-456'},
}
values = collect_sensitive_data_values(sensitive_data)
print(values)  # {'password': 'gl-secret-456'}  <- github's password is gone

print(redact_sensitive_string('login with gh-secret-123 and gl-secret-456', values))
# login with gh-secret-123 and <secret>password</secret>
#                    ^^^^^^^^^^^^ github's password leaks verbatim
```

### Expected behavior

Every value under a shared placeholder name is redacted:

```
login with <secret>password</secret> and <secret>password</secret>
```

### Root cause

```python
for key, val in content.items():
    if val:
        sensitive_values[key] = val   # cross-domain collision: last one wins
```

`Registry._replace_sensitive_data` is domain-aware (it resolves `<secret>password</secret>` against the current URL, so replacement is unambiguous), but the redaction side flattens globally and drops colliding values. The asymmetry means one of the secrets has no redaction coverage at all — it reaches message history, extracted content, and logs in plain text.

A legacy key colliding with a domain-scoped key (`{'password': 'global', 'example.com': {'password': 'site'}}`) has the same problem.

### Suggested fix

Keep all distinct values when flattening (e.g. collect colliding values into a list) and make `redact_sensitive_string()` flatten `{key: value | values}` into its internal `value -> key` lookup. The placeholder name stays unchanged, so domain-aware replacement in `Registry._replace_sensitive_data` keeps working as-is.

While testing this I also noticed `redact_sensitive_string('abc', {'k': ''})` produces `<secret>k</secret>a<secret>k</secret>b<secret>k</secret>c` — an empty secret value becomes an empty regex alternative that matches at every position. A guard for empty values falls out of the same fix.

### Environment

- browser-use: current main (2e32d2603)
- Python 3.12

hata33:
Fixes #5592

## Summary

- `collect_sensitive_data_values()` keeps every distinct value when several domains share a placeholder name (colliding values are collected into a list) instead of silently keeping only the last one.
- `redact_sensitive_string()` accepts `dict[str, str | list[str]]` and flattens it into its internal value → key lookup, so every colliding value is masked. Plain `dict[str, str]` input keeps working unchanged.
- Empty secret values are skipped, which also stops an empty regex alternative from matching at every position and corrupting the whole output.
- 4 new regression tests alongside the existing cascade tests.

## Why

The replacement side (`Registry._replace_sensitive_data`) is domain-aware: it resolves `<secret>password</secret>` against the current URL, so a shared placeholder name is unambiguous there. The redaction side, however, flattened `sensitive_data` globally with a plain dict assignment, so when two domains used the same key the other values were dropped — and a dropped value has **no redaction coverage at all**. It reached message history, extracted content, and logs verbatim.

Domain scoping exists precisely so the same key (`username` / `password`) can carry different credentials per site, so this is the common configuration, not an edge case. A legacy global key colliding with a domain-scoped key has the same problem.

## Reproduction

Before this change, on current `main`:

```python
from browser_use.utils import collect_sensitive_data_values, redact_sensitive_string

sensitive_data = {
    'github.com': {'password': 'gh-secret-123'},
    'gitlab.com': {'password': 'gl-secret-456'},
}
values = collect_sensitive_data_values(sensitive_data)
print(values)
# {'password': 'gl-secret-456'}   <- github's password is gone

print(redact_sensitive_string('login with gh-secret-123 and gl-secret-456', values))
# login with gh-secret-123 and <secret>password</secret>
#                    ^^^^^^^^^^^^ leaked verbatim
```

After: both values are masked as `<secret>password</secret>`, matching what `Registry._replace_sensitive_data` resolves on each domain.

## Side effect analysis

- Placeholder names are unchanged, so domain-aware replacement in `Registry._replace_sensitive_data` and the `<secret>name</secret>` instructions the agent sees are untouched.
- `redact_sensitive_string`'s signature widens `dict[str, str]` → `dict[str, str | list[str]]`; both internal call sites (`message_manager/service.py`, `agent/views.py`) pass the output of `collect_sensitive_data_values()` straight through and don't iterate it, and existing callers with plain string values are unaffected (old tests pass unchanged).
- Identical values under a colliding key are deduplicated, so the alternation doesn't grow.
- No defaults, config, or agent-loop behavior changed.

## Tests

- `uv run pytest tests/ci/test_redact_cascade.py -q` — 10 passed
- `uv run pytest tests/ci/security/test_sensitive_data.py tests/ci/test_redact_cascade.py -q` — 26 passed
- `uv run pyright browser_use/utils.py` — 0 errors, 0 warnings
- `uv run ruff check` / `ruff format --check` on both files — clean

New cases: cross-domain same-key collision, identical-value dedup, legacy + domain collision, empty secret value ignored.

---
## Summary by cubic
Fixes #5592 by keeping every distinct value when multiple domains share a placeholder name, so all sensitive values are redacted instead of only the last one.

**Behavior changes**
- `collect_sensitive_data_values()` now returns `dict[str, str | list[str]]` and collects colliding values into a list.
- `redact_sensitive_string()` flattens list values for lookup and skips empty secrets to avoid empty regex alternatives.
- Placeholder names stay unchanged, so domain-aware replacement in `Registry._replace_sensitive_data` still resolves correctly.
- Added regression tests for cross-domain key collisions, deduplication, legacy collisions, and empty secret values.

<sup>Written for commit 14a0f2d1ff8fa3b7970d1be2cc6fb3bf864989df. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5593?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/utils.py
+++ b/browser_use/utils.py
@@ -5,7 +5,7 @@ import platform
 import re
 import signal
 import time
-from collections.abc import Callable, Coroutine
+from collections.abc import Callable, Coroutine, Mapping
 from fnmatch import fnmatch
 from functools import cache, wraps
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


hata33:

hata33:
**CLAassistant** recheck

PR #5594, opened 2026-08-29
fix: dispatch literal plus key presses

mikemikimike:
Fix literal plus key dispatch in `Page.press`.

When the key is exactly `+`, it should be dispatched as a regular key press. Previously, the combination-key parser split it into an empty main key, so callers could not send a literal plus character.

Added a deterministic regression test covering the emitted keyDown and keyUp events. Existing key-combination handling remains unchanged.

Tests: `uv run pytest tests/ci/test_actor_page_press.py`
Validation: ruff check, pyright, and `git diff --check` pass. Pre-commit could not run because Python 3.11 is unavailable in the local environment.

---
## Summary by cubic
Fixes `Page.press` so a literal `+` key is dispatched as a regular key press instead of being parsed as a key combination with an empty main key. Existing key-combination handling (like `"Control+A"`) is unchanged.

- Maps `+` to the CDP code `Equal` (key code 187) in `get_key_info`.
- Adds a regression test covering the emitted `keyDown` and `keyUp` events for the `+` key.

<sup>Written for commit c6da4f092d3ddc9a6a01ef64a682e4eb1e19f1c5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5594?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/actor/page.py
+++ b/browser_use/actor/page.py
@@ -215,7 +215,7 @@ class Page:
 		session_id = await self._ensure_session()
 
 		# Handle key combinations like "Control+A"
-		if '+' in key:
+		if '+' in key and key != '+':
 			parts = key.split('+')
 			modifiers = parts[:-1]
 			main_key = parts[-1]
--- a/browser_use/actor/utils.py
+++ b/browser_use/actor/utils.py
@@ -27,6 +27,7 @@ class Utils:
 			'Escape': ('Escape', 27),
 			'Space': ('Space', 32),
 			' ': ('Space', 32),
+			'+': ('Equal', 187),
 			'PageUp': ('PageUp', 33),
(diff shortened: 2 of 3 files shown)


mikemikimike:
Current head c6da4f0 maps literal plus to the physical Equal key and the regression test asserts Equal for both events. The review request is already addressed; closing the thread.

PR #5604, opened 2026-08-30
fix(llm/schema): flatten $ref behind a model field named description

aniketwaghh:
Closes #5603

## What

`SchemaOptimizer.create_optimized_json_schema` walks the schema dispatching on JSON Schema keyword names, and it walks into `properties` maps too, where the keys are the user's field names rather than keywords. A field named `description` therefore hits the `elif key == 'description':` branch, which copies the value straight through without recursing, so its `$ref` is never inlined. `$defs` has already been stripped by then, and the schema we send the provider points at nothing.

#2324 fixed exactly this for a field named `title` by threading an `in_properties` flag through the walker. This does the same for the rest of the keyword branches: once we know we are inside a `properties` map, every key is a field name, so recurse on the value and skip the keyword handling entirely.

Before, on d2fcda5:

```
>>> SchemaOptimizer.create_optimized_json_schema(Article)   # class Article: headline: str; description: Details
{'properties': {'headline': {'type': 'string'},
                'description': {'$ref': '#/$defs/Details'}},
 'required': ['headline', 'description'],
 'type': 'object',
 'additionalProperties': False}

>>> jsonschema.validate({'headline': 'x', 'description': {'summary': 'y'}}, schema)
_WrappedReferencingError: PointerToNowhere: '/$defs/Details' does not exist within {...}
```

After:

```
{'properties': {'headline': {'type': 'string'},
                'description': {'properties': {'summary': {'type': 'string'}},
                                'required': ['summary'],
                                'type': 'object',
                                'additionalProperties': False}},
 'required': ['headline', 'description'],
 'type': 'object',
 'additionalProperties': False}
```

Same through the agent path (`Tools(output_model=Article)` -> `AgentOutput.type_with_custom_actions`), which is where it actually bites: on main the done action carries

```
/properties/action/items/anyOf[0]/properties/done/properties/data/properties/description/$ref -> #/$defs/Details
```

and that dict is what `ChatOpenAI` puts in `response_format` with `strict: True`, what `ChatAnthropic` puts in `input_schema`, and so on.

The change also covers a field named `properties`, which used to keep its metadata `title` because `in_properties=True` leaked into the field's own schema. Every other keyword branch behaves identically before and after: for keys like `type`, `items`, `required`, `minimum` the old path already ended up calling `optimize_schema(value, defs_lookup)`, which is what the new branch does.

## Tests

Three tests added to `tests/ci/models/test_llm_schema_optimizer.py`, the file that came out of #2324:

[...]

PR #5605, opened 2026-08-30
fix(llm): preserve schema-keyword field names

### LINKED ISSUE #5603: Bug: SchemaOptimizer leaves a dangling $ref when a model field is named "description"
### Browser Use Version

d2fcda5 (0.13.8)

### Bug Description

`SchemaOptimizer.create_optimized_json_schema` inlines every `$ref` and drops `$defs`. It does that by walking the schema and dispatching on JSON Schema keyword names, but it walks *into* `properties` maps too, and the keys in there are the user's field names, not keywords.

So a model field named `description` lands in the `elif key == 'description':` branch (browser_use/llm/schema.py:57), which copies the value through verbatim and never recurses. If that field is a nested model, its `{"$ref": "#/$defs/..."}` is never inlined, and `$defs` has already been stripped, so the schema we hand the provider points at nothing.

This is the same bug #2324 fixed for a field named `title` by adding the `in_properties` flag. `description` was missed.

Reproduce with no API key:

```python
from pydantic import BaseModel
from browser_use.llm.schema import SchemaOptimizer

class Details(BaseModel):
	summary: str

class Article(BaseModel):
	headline: str
	description: Details

print(SchemaOptimizer.create_optimized_json_schema(Article))
```

On d2fcda5:

```
{'properties': {'headline': {'type': 'string'},
                'description': {'$ref': '#/$defs/Details'}},
 'required': ['headline', 'description'],
 'type': 'object',
 'additionalProperties': False}
```

`#/$defs/Details` is not in the document, and jsonschema 4.26.0 refuses it:

```
_WrappedReferencingError: PointerToNowhere: '/$defs/Details' does not exist within
{'properties': {'headline': {'type': 'string'}, 'description': {'$ref': '#/$defs/Details'}}, ...}
```

Same thing through the agent path with `Agent(output_model_schema=Article)` / `Tools(output_model=Article)`:

```
DANGLING /properties/action/items/anyOf[0]/properties/done/properties/data/properties/description/$ref -> #/$defs/Details
```

That dict goes straight into `response_format` with `strict: True` in `ChatOpenAI` (browser_use/llm/openai/chat.py:224-232), into `input_schema` on the Anthropic `ToolParam`, and into the Bedrock/Gemini/Groq tool specs.

A field named `properties` is affected more mildly: `in_properties=True` leaks into the field's own schema, so it keeps its metadata `title`.

I checked every other JSON Schema keyword that has a branch in that loop. Fields named `title`, `type`, `required`, `items`, `anyOf`, `default`, `pattern`, `minimum`, `maximum` and `minItems` all come out correct today. It is only `description` and `properties`.

### Failing Python Code

```python
from pydantic import BaseModel
from browser_use import Agent
from browser_use.llm import ChatOpenAI

class Details(BaseModel):
	summary: str
	bullets: list[str]

class Article(BaseModel):
	headline: str
	description: Details

agent = Agent(
	task='open example.com and summarise it',
	llm=ChatOpenAI(model='gpt-4.1-mini'),
	output_model_schema=Article,
)
# the done-action schema sent to the model contains "$ref": "#/$defs/Details" with no $defs
```

### LLM Model

any provider that uses structured output

macOS 15 (the bug is in a pure function, no browser involved)

MagMueller:
## What

Treat keys inside a schema `properties` map as user field names before applying schema-keyword rules. This keeps fields named `description` or `properties` from retaining dangling `$ref` values.

Fixes #5603

## Test

- `uv run pytest -q tests/ci/models/test_llm_schema_optimizer.py`
- `uv run pre-commit run --files browser_use/llm/schema.py tests/ci/models/test_llm_schema_optimizer.py`

---
## Summary by cubic
Fixes schema optimization so user fields named `description` or `properties` no longer keep dangling `$ref` values.

- Treats keys inside a `properties` map as field names before applying schema-keyword rules.
- Simplifies `title` skipping, which now only applies outside `properties`.
- Adds coverage for nested and aliased fields named after schema keywords.

<sup>Written for commit b2507091aa2234bdbfebc39300a7174b04505c3f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5605?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/llm/schema.py
+++ b/browser_use/llm/schema.py
@@ -45,11 +45,16 @@ class SchemaOptimizer:
 				skip_fields = ['additionalProperties', '$defs']
 
 				for key, value in obj.items():
+					# Keys inside `properties` are user field names, not schema keywords.
+					if in_properties:
+						optimized[key] = optimize_schema(value, defs_lookup)
+						continue
+
 					if key in skip_fields:
 						continue
 
-					# Skip metadata "title" unless we're iterating inside an actual `properties` map
-					if key == 'title' and not in_properties:
+					# Skip metadata "title"
+					if key == 'title':
 						continue
 
 					# Preserve FULL descriptions without truncation, skip empty ones
--- a/tests/ci/models/test_llm_schema_optimizer.py
+++ b/tests/ci/models/test_llm_schema_optimizer.py
@@ -3,7 +3,7 @@ Tests for the SchemaOptimizer to ensure it correctly processes and
 optimizes the schemas for agent actions without losing information.
 """
 
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
 from browser_use.agent.views import AgentOutput
 from browser_use.llm.schema import SchemaOptimizer
@@ -85,11 +85,13 @@ def test_optimizer_treats_property_names_as_data_not_schema_keywords():
 	class Article(BaseModel):
 		description: Details
 		properties: Details
+		additional_properties: Details = Field(alias='additionalProperties')
+		defs: Details = Field(alias='$defs')
 
 	schema = SchemaOptimizer.create_optimized_json_schema(Article)
 
 	assert '$defs' not in schema
-	for field_name in ('description', 'properties'):
+	for field_name in ('description', 'properties', 'additionalProperties', '$defs'):
 		field_schema = schema['properties'][field_name]
 		assert '$ref' not in field_schema
 		assert field_schema['properties']['summary']['type'] == 'string'
--- a/tests/ci/models/test_llm_schema_optimizer.py
+++ b/tests/ci/models/test_llm_schema_optimizer.py
@@ -74,3 +74,22 @@ def test_gemini_schema_retains_required_fields():
 
 	required_fields = set(schema['required'])
 	assert {'price', 'title'}.issubset(required_fields), 'Mandatory fields must stay required for Gemini.'
+
+
+def test_optimizer_treats_property_names_as_data_not_schema_keywords():
+	"""Nested fields named after schema keywords must still have their refs flattened."""
+
+	class Details(BaseModel):
+		summary: str
+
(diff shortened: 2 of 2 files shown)


aniketwaghh:
This is cleaner than what I had in #5604 - dropping `and not in_properties` from the title branch is the right call once the short-circuit is above it.

One thing worth flagging, because I got two details wrong in #5603 and only caught them re-checking against d2fcda5.

`properties` is not actually affected. I said it keeps a metadata `title` and it doesn't.

But there are two names I never mentioned that are worse than the one I did. `additionalProperties` and `$defs` are both in `skip_fields`, so as user field names they hit that `continue` and the field disappears from the schema altogether. Sweeping all 22 keywords as a field name on d2fcda5:

```
additionalProperties   DROPPED
$defs                  DROPPED
description            dangling $ref
properties             ok
title                  ok
(+ type, items, anyOf, required, default, pattern, minimum,
   maximum, minItems, enum, format, allOf, oneOf, const,
   examples, nullable - all ok)
```

The dropped ones are quieter than the dangling ref: no error anywhere, the model just never sees the field and can't fill it.

Your patch fixes both, since the `in_properties` check runs ahead of `skip_fields`. I applied it locally and all 22 come out intact, so there is nothing to change here. Might be worth a case so it can't regress:

```python
class Details(BaseModel):
	summary: str

class M(BaseModel):
	headline: str
	additionalProperties: Details

# d2fcda5: 'additionalProperties' is absent from the optimized schema entirely
```

PR #5653, opened 2026-09-03
fix(watchdogs): hold a reference to the HAR body fetch task

Apageoflove:
Why

`_on_loading_finished` in `browser_use/browser/watchdogs/har_recording_watchdog.py` schedules the response-body fetch with a bare `asyncio.create_task(...)` and drops the returned Task handle. Every other watchdog in this directory goes through `create_task_with_error_handling`, which keeps the task strongly referenced and adds an exception callback. This is the only bare create_task in the whole watchdogs set.

`asyncio.create_task` docs are explicit that the caller should save a reference to avoid the task disappearing mid-execution. Here the coroutine awaits a CDP `Network.getResponseBody` round-trip; if the task is garbage-collected during that await, the coroutine is cancelled and `entry.response_body` never gets written, so the HAR entry ends up without its body. The handler is a synchronous CDP callback fired once per `Network.loadingFinished` event, so on a page with many requests there are many of these short-lived tasks in flight.

Honest caveat: whether a dropped task actually gets collected depends on timing and load, so I can't point at a deterministic reproduction where a body is missing. The code path is still the only one in this directory that violates the documented create_task contract and the directory's own pattern, and the fix is a no-op in behavior otherwise.

What changed

`_fetch_body(...)` is now scheduled through `create_task_with_error_handling` with a per-request task name, matching the rest of the watchdogs. The function-local `import asyncio as _asyncio` is removed since nothing else in the handler uses it.

Surface area

- Backend watchdog runtime (`browser_use/browser/watchdogs/har_recording_watchdog.py` only)

Validation

- `python -m py_compile` and an AST parse on the touched file pass.
- No test exists for the watchdog task lifecycle today; the change is a scheduling-pattern alignment, not a logic change, so behavior is covered by the existing HAR tests.

---
## Summary by cubic
Holds a strong reference to the HAR response-body fetch task so it can't be garbage-collected mid-await and leave the HAR entry without a body. The old code scheduled `_fetch_body` with a bare `asyncio.create_task` and dropped the handle; the new code routes through `create_task_with_error_handling` with a per-request task name and keeps each task in an instance set until it finishes, matching the other watchdogs in this directory.

- Keeps every fetch task in `_fetch_tasks` until the done callback discards it.
- Adds the module-level `asyncio` import needed by the task set.
- The dropped-task risk depends on timing and load, so there's no deterministic reproduction of a missing body.

<sup>Written for commit cc3134a80b87494e9b7e736001d8990bb907c206. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5653?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Apageoflove:
Good catch, and you're right — thanks for the careful read.

Looking at the wrapper again: the done-callback it registers is a plain function taking the task as an argument, so it doesn't form a closure back to the task at all. The task's only reference was the return value we dropped. The event loop keeps tasks in a weak set, so an unreferenced task sitting on an await can still be collected. My first version only added the exception handling; it did not deliver the reference-holding the title claimed.

Pushed a follow-up (971f333): each fetch task is now added to an instance-level set (self._fetch_tasks, initialized in __init__) and removed by a done callback when it finishes. That keeps the task strongly referenced for its whole lifetime, matching the pattern the asyncio docs recommend for fire-and-forget work.

One small note on the suggested implementation: a done callback that discards from the set is enough — the callback list on the task holds the discard method while the task is alive, and once the task finishes and the callback fires, the set entry goes away, so there's no leak of finished tasks.

PR #5654, opened 2026-09-03
fix(watchdogs): hold a reference to the HAR body fetch task

Apageoflove:
Why

`_on_loading_finished` in `browser_use/browser/watchdogs/har_recording_watchdog.py` schedules the response-body fetch with a bare `asyncio.create_task(...)` and drops the returned Task handle. Every other watchdog in this directory goes through `create_task_with_error_handling`. This is the only bare create_task in the whole watchdogs set.

`asyncio.create_task` docs are explicit that the caller should save a reference to avoid the task disappearing mid-execution. Here the coroutine awaits a CDP `Network.getResponseBody` round-trip; if the task is garbage-collected during that await, the coroutine is cancelled and `entry.response_body` never gets written, so the HAR entry ends up without its body. The handler is a synchronous CDP callback fired once per `Network.loadingFinished` event, so on a page with many requests there are many of these short-lived tasks in flight.

Honest caveat: whether a dropped task actually gets collected depends on timing and load, so I can't point at a deterministic reproduction where a body is missing. The code path is still the only one in this directory that violates the documented create_task contract and the directory's own pattern.

What changed

1. `6d9852b` switches the call to `create_task_with_error_handling` with a per-request task name, matching the rest of the watchdogs. The function-local `import asyncio as _asyncio` is removed.

2. `ef3bcdf` actually keeps the task alive. A bare `create_task_with_error_handling(...)` that drops the return value still leaves the task with no external reference; its done-callback is a plain function taking the task as an argument and does not form a closure back to it, so the task is still collected once it suspends on the CDP round-trip. The follow-up stores each fetch task in `self._fetch_tasks` (initialized in `__init__`) and registers `task.add_done_callback(self._fetch_tasks.discard)` so the set holds a strong reference until the task finishes.

The set only ever holds in-flight tasks — the done callback drops each entry as it fires — so there is no leak of completed tasks.

Note on supersedes

This is the second attempt at the same fix. The previous PR (#5653) was accidentally closed after a force-push that rewrote the author email on history; the branch is no longer re-openable, so this is a clean replacement against the latest main with the same final content.

---
## Summary by cubic
Fixes the HAR recording watchdog so response-body fetch tasks are kept alive and drained before the HAR is written, preventing entries from missing their bodies.

- Schedules the fetch via `create_task_with_error_handling` with a per-request task name, matching the other watchdogs.
- Stores each task in `self._fetch_tasks` until it finishes; a done callback discards completed tasks.
- On browser stop, awaits pending fetches (with a 10s grace window) before serializing, so completed bodies are included.
- Cancels fetches that outlive the grace window so they don't die as "Task was destroyed but it is pending" at loop shutdown.

<sup>Written for commit 8a2113c63335ba8e7480dcfa9f254a0ac7bc423d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5654?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/browser/watchdogs/har_recording_watchdog.py
+++ b/browser_use/browser/watchdogs/har_recording_watchdog.py
@@ -7,6 +7,7 @@ and `record_har_mode` (full/minimal).
 
 from __future__ import annotations
 
+import asyncio
 import base64
 import hashlib
 import json
@@ -152,6 +153,7 @@ class HarRecordingWatchdog(BaseWatchdog):
 		super().__init__(*args, **kwargs)
 		self._enabled: bool = False
 		self._entries: dict[str, _HarEntryBuilder] = {}
+		self._fetch_tasks: set[asyncio.Task] = set()
 		self._top_level_pages: dict[
 			str, dict
 		] = {}  # frameId -> {url, title, startedDateTime, monotonic_start, onContentLoad, onLoad}
@@ -410,10 +412,12 @@ class HarRecordingWatchdog(BaseWatchdog):
 					pass
 
 			# Always schedule the response body fetch task
-			create_task_with_error_handling(
+			task = create_task_with_error_handling(
 				_fetch_body(self, request_id, session_id),
 				name=f'har-fetch-{request_id}',
 			)
+			self._fetch_tasks.add(task)
+			task.add_done_callback(self._fetch_tasks.discard)
 
 			encoded_length = (
 				params.get('encodedDataLength') if hasattr(params, 'get') else getattr(params, 'encodedDataLength', None)
--- a/browser_use/browser/watchdogs/har_recording_watchdog.py
+++ b/browser_use/browser/watchdogs/har_recording_watchdog.py
@@ -27,6 +27,7 @@ from cdp_use.cdp.page.events import FrameNavigatedEvent, LifecycleEventEvent
 
 from browser_use.browser.events import BrowserConnectedEvent, BrowserStopEvent
 from browser_use.browser.watchdog_base import BaseWatchdog
+from browser_use.utils import create_task_with_error_handling
 
 
 @dataclass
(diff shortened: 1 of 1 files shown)


Apageoflove:
Valid, fixed in 0d93dd7.

The gap was real and actually predated the reference-holding change: before, the bare task would just as likely die at stop; now it reliably survives, which made the missing-body-at-stop window deterministic instead of flaky. on_BrowserStopEvent now drains the tracked set before _write_har serializes.

One deliberate deviation from a plain await: the drain is bounded (asyncio.wait with a 10s grace window, module constant FETCH_DRAIN_TIMEOUT_SECONDS). A fetch that never completes must not stall browser shutdown, so after the window the HAR is written with whatever completed — for those entries the outcome is identical to pre-fix behavior. Tasks that finish within the window contribute their bodies; the done callbacks keep the set clean either way.

Apageoflove:
Both findings checked against the code.

2 (P3): fixed in 8b499de. The drain now captures the still-pending set from asyncio.wait and cancels it — safe because the wrapper's done-callback treats CancelledError as normal and the discard callback still fires, so the tracked set cleans up either way.

1 (P1): verified real, and I think it belongs above this file's pay grade. BrowserStopEvent has two handlers: this watchdog's drain, and BrowserSession.on_BrowserStopEvent (session.py ~1281), which itself calls await self.reset() and clears the CDP session cache. Whether the fetches still have a live connection during our drain depends on which handler the bus runs first — the coordination the violation asks for (await the watchdog before reset) is a stop-lifecycle ordering change in session.py, not something this watchdog can fix from the inside. Any drain-side patch here would still race the session's own handler.

So: happy to do the session.py change in this PR if maintainers prefer (drain the HAR watchdog before reset() in the stop path), or split it into a separate PR if the lifecycle file deserves its own review. Which way do you want it?

PR #5680, opened 2026-09-05
fix(history): redact sensitive data from input_text actions

### LINKED ISSUE #5679: Bug: sensitive_data is never redacted from the built-in input_text action in saved history
### Browser Use Version

0.13.10 (main @ d5453ae80)

## Summary

`AgentHistoryList.save_to_file` never redacts `sensitive_data` from the built-in `input_text` action, so a typed password or API key is written to the history file in plaintext.

`AgentHistory.model_dump` decides which actions to filter with this guard (`browser_use/agent/views.py:561`):

```python
action_dump = [
    self._filter_sensitive_data_from_dict(action, sensitive_data) if 'input' in action else action
    for action in action_dump
]
```

`action` here is the dumped action, shaped `{action_name: params}`. So `'input' in action` is a **dict key lookup, not a substring test**. For the built-in input action the key is `input_text`, and `'input' in {'input_text': ...}` is `False` — the filter is skipped entirely.

The guard only ever fires for an action literally named `input`, which no built-in action is.

## Steps to reproduce

Run the script in the "Failing Python Code" field. No browser, model, or real credential required.

## Expected behavior

The saved history should contain `<secret>api_key</secret>` in place of the typed value.

## Actual behavior

```
dumped action dict: {'input_text': {'index': 1, 'text': 'token-123', 'clear': True}}
guard `'input' in action` evaluates to: False
secret still present in saved history: True
placeholder present: False
```

## Impact

Anyone who passes `sensitive_data` to an agent and then calls `save_to_file` gets a history file containing the real secret, while the API strongly implies it was filtered. Typing credentials into a login form is the single most common use of `sensitive_data`, so this is the path most likely to be hit.

## Note on scope

This is adjacent to #5623 (nested containers slipping past the recursion, fixed in #5662) but is a distinct and broader failure: there the filter ran and missed part of the payload; here it never runs at all. #5662 made the recursion complete, which means fixing this guard is now enough to close the gap end to end.

The narrow fix is to match the action name properly. The better fix is probably to drop the name-based guard and filter every action's parameters — redaction is cheap, and an allowlist keyed on one hardcoded action name will keep missing cases as new actions are added. Happy to open a PR once a maintainer says which direction is preferred, since removing the guard changes behaviour for custom actions too.

### Failing Python Code

```python
"""Does save_to_file redact sensitive data in the built-in input_text action?"""

import tempfile
from pathlib import Path
from typing import Any

from pydantic import create_model

from browser_use.agent.views import AgentHistory, AgentHistoryList, AgentOutput
from browser_use.browser.views import BrowserStateHistory
from browser_use.tools.registry.views import ActionModel
from browser_use.tools.views import InputTextAction

Model = create_model('M', __base__=ActionModel, input_text=(InputTextAction | None, None))
Out = AgentOutput.type_with_custom_actions(Model)
act = Model(input_text=InputTextAction(index=1, text='token-123'))

print('dumped action dict:', act.model_dump(exclude_none=True, mode='json'))
print("guard `'input' in action` evaluates to:", 'input' in act.model_dump(exclude_none=True, mode='json'))

history = AgentHistoryList[Any](
    history=[
        AgentHistory(
            model_output=Out(memory='', action=[act]),
            result=[],
            state=BrowserStateHistory(url='https://x.test', title='t', tabs=[], interacted_element=[None]),
        )
    ]
)
path = Path(tempfile.mkdtemp()) / 'h.json'
history.save_to_file(path, sensitive_data={'api_key': 'token-123'})
saved = path.read_text(encoding='utf-8')

print('secret still present in saved history:', 'token-123' in saved)
print('placeholder present:', '<secret>api_key</secret>' in saved)
```

### LLM Model

N/A - reproduces at the history-serialization layer, no LLM request is made

Windows 11 / Browser N/A

### Full DEBUG Log Output

```shell
The repro exercises history serialization directly, so no agent loop runs and there is
no meaningful agent log to attach. Full output of the script with
BROWSER_USE_LOGGING_LEVEL=debug:

$ BROWSER_USE_LOGGING_LEVEL=debug uv run python repro.py
DEBUG    [browser_use.utils] Display size: width=1920 height=1080
dumped action dict: {'input_text': {'index': 1, 'text': 'token-123', 'clear': True}}
guard `'input' in action` evaluates to: False
secret still present in saved history: True
placeholder present: False

Environment: browser-use 0.13.10 (main @ d5453ae80), Python 3.12.4, pydantic 2.13.5,
Windows 11.

The same leak occurs in a normal agent run: any task where the model calls input_text
while sensitive_data is set, followed by save_to_file, writes the real value to disk.
The script isolates that path so it can be checked without credentials or a browser.
```

pranjalbhatia710:
## Summary
- fix saved history redaction for built-in `input_text` actions by applying the existing sensitive-data filter to all action payloads
- add a regression test matching the `input_text` leak reported in #5679

Fixes #5679.

## Test plan
```console
uv run pytest tests/ci/security/test_sensitive_data.py::test_history_filters_sensitive_data_from_input_text_action tests/ci/security/test_sensitive_data.py::test_history_filters_sensitive_data_inside_nested_lists -q
# 2 passed in 0.09s

uv run ruff check browser_use/agent/views.py tests/ci/security/test_sensitive_data.py
# All checks passed!

git diff --check
# no output
```

---
## Summary by cubic
Fixes saved history leaking sensitive data from built-in `input_text` actions by applying the redaction filter to all action payloads, not just those with an `'input'` key. Adds a regression test for the reported issue (#5679).

<sup>Written for commit 4294b08c3c634c8ea2450896b6bc1cc787724afb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5680?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/agent/views.py
+++ b/browser_use/agent/views.py
@@ -554,10 +554,7 @@ class AgentHistory(BaseModel):
 			# {action_name: params}, so checking for a literal "input" key misses built-in
 			# actions like input_text and any custom action with a different name.
 			if sensitive_data:
-				action_dump = [
-					self._filter_sensitive_data_from_dict(action, sensitive_data)
-					for action in action_dump
-				]
+				action_dump = [self._filter_sensitive_data_from_dict(action, sensitive_data) for action in action_dump]
 
 			model_output_dump = {
 				'evaluation_previous_goal': self.model_output.evaluation_previous_goal,
--- a/browser_use/agent/views.py
+++ b/browser_use/agent/views.py
@@ -550,10 +550,12 @@ class AgentHistory(BaseModel):
 		if self.model_output:
 			action_dump = [action.model_dump(exclude_none=True, mode='json') for action in self.model_output.action]
 
-			# Filter sensitive data only from input action parameters if sensitive_data is provided
+			# Filter sensitive data from every action parameter payload. The dumped shape is
+			# {action_name: params}, so checking for a literal "input" key misses built-in
+			# actions like input_text and any custom action with a different name.
 			if sensitive_data:
 				action_dump = [
-					self._filter_sensitive_data_from_dict(action, sensitive_data) if 'input' in action else action
+					self._filter_sensitive_data_from_dict(action, sensitive_data)
 					for action in action_dump
 				]
 
--- a/tests/ci/security/test_sensitive_data.py
+++ b/tests/ci/security/test_sensitive_data.py
@@ -670,3 +670,36 @@ def test_history_filters_sensitive_data_inside_nested_lists(tmp_path):
 
 	assert 'token-123' not in saved, 'Sensitive value leaked into the saved history file'
 	assert saved.count('<secret>api_key</secret>') == 2
+
+
+def test_history_filters_sensitive_data_from_input_text_action(tmp_path):
+	"""Saved history must redact sensitive values typed by the built-in input_text action."""
+	from pathlib import Path
+	from typing import Any
+
+	from pydantic import create_model
+
+	from browser_use.agent.views import AgentHistory, AgentHistoryList, AgentOutput
+	from browser_use.browser.views import BrowserStateHistory
+	from browser_use.tools.registry.views import ActionModel
+	from browser_use.tools.views import InputTextAction
+
+	InputTextActionModel = create_model('InputTextActionModel', __base__=ActionModel, input_text=(InputTextAction | None, None))
+	OutputModel = AgentOutput.type_with_custom_actions(InputTextActionModel)
+	action = InputTextActionModel.model_validate({'input_text': {'index': 1, 'text': 'token-123'}})
+	history = AgentHistoryList[Any](
+		history=[
+			AgentHistory(
+				model_output=OutputModel(memory='', action=[action]),
+				result=[],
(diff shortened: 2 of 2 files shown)


pranjalbhatia710:
Pushed `4294b08c` to address the failing code-style check by formatting the touched history redaction line.

Local validation:

```console
uv run ruff format --check browser_use/agent/views.py tests/ci/security/test_sensitive_data.py
# would pass after formatting
uv run ruff check browser_use/agent/views.py tests/ci/security/test_sensitive_data.py
# All checks passed!
uv run pytest tests/ci/security/test_sensitive_data.py -q
# 18 passed in 0.11s
git diff --check
# no output
```

Note: the remaining CLA assistant check requires account-level signature action, so I cannot resolve that from the branch.

MagMueller:
Reproduced the input_text leak on main. Your patch redacts the saved action and preserves the live action. Could you complete the CLA so we can move this forward?

PR #5684, opened 2026-09-05
fix(llm): report token usage for Ollama and DeepSeek

### LINKED ISSUE #5683: ChatOllama and ChatDeepSeek never report token usage, so counts and cost are unavailable on both
### Browser Use Version

`0.13.9`, main at `d5453ae`

`ChatOllama` and `ChatDeepSeek` never report token usage. Both hard-code `usage=None` on every return path:

- `browser_use/llm/ollama/chat.py` lines 123 and 143
- `browser_use/llm/deepseek/chat.py` lines 151, 199, 205 and 229

Every other provider builds a `ChatInvokeUsage`. These two are the only ones that do not.

The counts are not missing from the responses, they are just never read. Ollama returns `prompt_eval_count` and `eval_count` on its `ChatResponse`. DeepSeek talks to the API through `AsyncOpenAI`, so its responses carry the standard OpenAI usage block, and DeepSeek additionally reports cache hits as `prompt_cache_hit_tokens`.

Everything derived from `ChatInvokeUsage` is therefore unavailable on both providers: per-step and per-run token counts, cost estimation, the usage summary at the end of a run, the OpenTelemetry usage attributes, and anything budgeting against tokens consumed.

Nothing raises. The numbers are absent or zero, which reads as "this run was free" rather than "this run was not measured". Locally that is cosmetic; DeepSeek is a paid API, so anyone comparing providers on cost sees zero for it.

I ran all three providers against the same local Qwen3 model to confirm it is not something about a particular backend. Same model, same server, same question:

```
provider                  prompt   compl   total
ChatOllama                     -       -       -
ChatDeepSeek                   -       -       -
ChatOpenAI (control)          18     501     519
```

### Failing Python Code

No server or API key needed to see it:

```python
import inspect, re
from pathlib import Path
import browser_use.llm as llm_pkg

llm_dir = Path(inspect.getfile(llm_pkg)).parent
for provider in ('ollama', 'deepseek', 'openai'):
    src = (llm_dir / provider / 'chat.py').read_text()
    print(f"{provider:9} builds usage: {bool(re.search(r'ChatInvokeUsage', src)):<5} "
          f"usage=None sites: {len(re.findall(r'usage=None', src))}")

# ollama    builds usage: False  usage=None sites: 2
# deepseek  builds usage: False  usage=None sites: 4
# openai    builds usage: True   usage=None sites: 0

from ollama._types import ChatResponse
print([k for k in ChatResponse.model_fields if 'count' in k])
# ['prompt_eval_count', 'eval_count']   <- present on every response, never read
```

Against a live Ollama server the same thing shows up through the public API:

```python
r = await ChatOllama(model='qwen3').ainvoke([UserMessage(content='hi')])
print(r.usage)   # None
```

### LLM Model

Ollama (any local model), DeepSeek

aniketwaghh:
Fixes #5683

`ChatOllama` and `ChatDeepSeek` hard-coded `usage=None` on every return path, so anything derived from `ChatInvokeUsage` was unavailable on them: per-step and per-run token counts, cost estimation, the usage summary at the end of a run, the OpenTelemetry usage attributes, and any token budgeting. Every other provider builds one.

The counts were already on the responses, just never read.

Ollama reports them as `prompt_eval_count` and `eval_count`. Both are `Optional` in ollama's own type, because streaming chunks carry them only on the last chunk. This client always requests a non-streaming response, but a reply without them is still possible, so `_get_usage` returns `None` in that case rather than zero. That reproduces today's behaviour exactly, so no model can regress; zero would be worse than nothing, since it reads as "this run was free" rather than "this run was not measured".

DeepSeek talks to the API through `AsyncOpenAI`, so its responses carry the standard usage block and the mapping is the one `ChatOpenAI` already uses. DeepSeek also prices cache hits and misses differently and reports the hit count as `prompt_cache_hit_tokens` rather than the OpenAI name, so `_get_usage` reads whichever of the two is present. There were four separate `usage=None` sites in that file, so this is one helper rather than the mapping repeated four times.

## Before and after

Same question to the same local Qwen3, once through each provider:

```
                        prompt   compl   total
ChatOllama                   -       -       -
ChatDeepSeek                 -       -       -
ChatOpenAI (control)        18     501     519
```

After:

```
                        prompt   compl   total
ChatOllama                  18     494     512
ChatDeepSeek                18     494     512
ChatOpenAI (control)        18     466     484
```

## Tests

Six tests, three per provider, all failing on `main` and passing here.

`tests/ci/models/test_llm_ollama.py`

- `test_usage_is_reported_from_eval_counts`: the ordinary case.
- `test_usage_is_none_when_the_response_carries_no_counts`: both fields absent stays `None`, which is what protects existing behaviour.
- `test_a_fully_cached_prompt_still_reports_its_completion`: `prompt_eval_count=0` is a real value, not a missing one, so the completion count must survive it.

`tests/ci/models/test_llm_deepseek.py`

- `test_usage_is_reported_from_the_openai_shaped_block`: the ordinary case.
- `test_cached_tokens_come_from_either_field_name`: covers both `prompt_tokens_details.cached_tokens` and DeepSeek's own `prompt_cache_hit_tokens`.
- `test_usage_is_none_when_the_response_has_no_usage_block`: no usage block stays `None`.

`tests/ci/models/` is green, 122 passed and 7 skipped, and `ruff check` and `ruff format --check` are clean.

---
## Summary by cubic
Fixes #5683. `ChatOllama` and `ChatDeepSeek` hard-coded `usage=None` on every return path, so token counts, cost estimation, and usage summaries were unavailable on them. They now build a `ChatInvokeUsage` from the counts already present in each provider's response.

**Details**
- Ollama reads `prompt_eval_count` and `eval_count`; if either is missing it returns `None`, so missing counts are never reported as a free run.
- DeepSeek maps the standard OpenAI usage block and reads cache hits from either `prompt_tokens_details.cached_tokens` or `prompt_cache_hit_tokens`.
- Seven new tests cover the normal case, missing counts, fully cached prompts, and partially missing counts; the Ollama fixture now uses a real `ChatResponse` so mocked responses can't fabricate counts.

<sup>Written for commit f3ddaa67d61a2b12faf770b8fb600e16ae9d1498. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5684?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/llm/deepseek/chat.py
+++ b/browser_use/llm/deepseek/chat.py
@@ -21,7 +21,7 @@ from browser_use.llm.deepseek.serializer import DeepSeekMessageSerializer
 from browser_use.llm.exceptions import ModelProviderError, ModelRateLimitError
 from browser_use.llm.messages import BaseMessage
 from browser_use.llm.schema import SchemaOptimizer
-from browser_use.llm.views import ChatInvokeCompletion
+from browser_use.llm.views import ChatInvokeCompletion, ChatInvokeUsage
 
 T = TypeVar('T', bound=BaseModel)
 
@@ -91,6 +91,30 @@ class ChatDeepSeek(BaseChatModel):
 			}
 		return common
 
+	@staticmethod
+	def _get_usage(response: Any) -> ChatInvokeUsage | None:
+		"""DeepSeek speaks the OpenAI wire format, so usage arrives in that shape.
+
+		DeepSeek prices cache hits and misses differently and reports the hit count
+		as prompt_cache_hit_tokens, so fall back to that when the OpenAI-style
+		prompt_tokens_details is absent.
+		"""
+		usage = getattr(response, 'usage', None)
+		if usage is None:
+			return None
+		details = getattr(usage, 'prompt_tokens_details', None)
+		cached = getattr(details, 'cached_tokens', None) if details else None
+		if cached is None:
+			cached = getattr(usage, 'prompt_cache_hit_tokens', None)
+		return ChatInvokeUsage(
+			prompt_tokens=usage.prompt_tokens,
+			completion_tokens=usage.completion_tokens,
+			total_tokens=usage.total_tokens,
+			prompt_cached_tokens=cached,
+			prompt_cache_creation_tokens=None,
+			prompt_image_tokens=None,
+		)
+
 	@overload
 	async def ainvoke(
 		self,
@@ -148,7 +172,7 @@ class ChatDeepSeek(BaseChatModel):
 				)
 				return ChatInvokeCompletion(
 					completion=resp.choices[0].message.content or '',
-					usage=None,
+					usage=self._get_usage(resp),
 				)
 			except RateLimitError as e:
 				raise ModelRateLimitError(str(e), model=self.name) from e
@@ -196,13 +220,13 @@ class ChatDeepSeek(BaseChatModel):
 				if output_format is not None:
 					return ChatInvokeCompletion(
 						completion=output_format.model_validate(parsed),
-						usage=None,
+						usage=self._get_usage(resp),
 					)
 				else:
 					# If no output_format, return dict directly
 					return ChatInvokeCompletion(
 						completion=parsed,
-						usage=None,
+						usage=self._get_usage(resp),
 					)
 			except RateLimitError as e:
 				raise ModelRateLimitError(str(e), model=self.name) from e
@@ -226,7 +250,7 @@ class ChatDeepSeek(BaseChatModel):
 				parsed = output_format.model_validate_json(content)
 				return ChatInvokeCompletion(
 					completion=parsed,
-					usage=None,
+					usage=self._get_usage(resp),
 				)
 			except RateLimitError as e:
 				raise ModelRateLimitError(str(e), model=self.name) from e
--- a/browser_use/llm/ollama/chat.py
+++ b/browser_use/llm/ollama/chat.py
@@ -107,10 +107,11 @@ class ChatOllama(BaseChatModel):
 		"""
 		prompt_tokens = response.prompt_eval_count
 		completion_tokens = response.eval_count
-		if prompt_tokens is None and completion_tokens is None:
+		# Only report when both are real values. Coercing a missing count to 0 would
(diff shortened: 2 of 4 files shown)


aniketwaghh:
Both taken, though the first one is right for a different reason than stated.

**One count present, the other missing.** Agreed, and it contradicted the docstring I had written directly above it. Coercing a missing count to 0 makes it indistinguishable from a genuine 0, which is exactly what a fully cached prompt reports, so it understates usage instead of omitting it. Now returns `None` unless both are real values, with a test for each one-sided case.

**The MagicMock fixture.** The existing tests do not actually fail: `prompt_eval_count` on a bare `MagicMock` is a Mock, pydantic coerces it, and the assertions pass.

```
ChatOllama(model='x')._get_usage(MagicMock(message=MagicMock(content='hi')))
# prompt_tokens=1 completion_tokens=1 total_tokens=1
```

Which is worse than a failure, because the fixture silently reports usage the server never sent and nothing flags it. Replaced with a real `ChatResponse`, so the counts are whatever the test says they are.

aniketwaghh:
One more reason the None-when-incomplete behaviour matters: Ollama is known to drop `prompt_eval_count` from the response on repeated requests with the same prefix, which is exactly the cached case (ollama/ollama#2068, #3427). With `or 0` those turns would have reported a 0-token prompt; now they report nothing, which is the honest answer.

Also worth noting for anyone reading later: when `prompt_eval_count` is present it is the total context size of the request, not newly evaluated tokens, so it is the right thing to map to `prompt_tokens`.

On the DeepSeek side, their `prompt_tokens` is documented as `prompt_cache_hit_tokens + prompt_cache_miss_tokens`, so mapping the hit count to `prompt_cached_tokens` leaves `prompt_tokens - prompt_cached_tokens` equal to the miss count, which is what `calculate_cost` expects.

PR #5686, opened 2026-09-05
fix(llm/anthropic): count cache reads in total_tokens

### LINKED ISSUE #5685: ChatAnthropic total_tokens omits cache reads, disagreeing with prompt_tokens + completion_tokens
### Browser Use Version

`0.13.9`, main at `d5453ae`

`ChatAnthropic._get_usage` builds `prompt_tokens` and `total_tokens` from two different definitions of what the prompt is (`browser_use/llm/anthropic/chat.py:211-217`):

```python
prompt_tokens=response.usage.input_tokens
+ (response.usage.cache_read_input_tokens or 0),   # counts cache reads
completion_tokens=response.usage.output_tokens,
total_tokens=response.usage.input_tokens + response.usage.output_tokens,   # does not
```

Anthropic reports `input_tokens` excluding cache reads, so adding them to `prompt_tokens` is right. The `total_tokens` line below was left as `input + output`, so whenever prompt caching is active `total_tokens != prompt_tokens + completion_tokens`. Every other provider satisfies that identity.

It is off by exactly the cache-read count, so the better the cache hit rate the larger the gap. On a normal agent turn, where a large mostly-cached prompt is followed by a short answer, that is most of the number:

| | prompt | completion | total_tokens | prompt+completion |
|---|---|---|---|---|
| anthropic | 994 | 40 | **239** | 1034 |
| openai | 994 | 40 | 1034 | 1034 |

**This does not look deliberate.** The two lines were written a day apart:

- `7a10ae0c96` (2025-06-24) wrote both consistently, `prompt_tokens = input_tokens` and `total_tokens = input + output`.
- `2ee526d903` (2025-06-25), "updated pricing for cached tokens support", added cache handling and changed `prompt_tokens` to add cache reads, leaving the `total_tokens` line directly below it untouched.

So the invariant held until cache support landed and broke inside the commit that added it.

**Cost is not affected.** `calculate_cost` prices cached and uncached tokens on separate paths and never reads `total_tokens`. The usage summary is fine too, because `tokens/service.py:469` recomputes `prompt_tokens + completion_tokens`.

**Two consumers do read the field**, and under-count Anthropic traffic:

- `beta/service.py:3241` — `summed_total_tokens += chat_usage.total_tokens`
- `beta/service.py:3776` and `:3781` — the `gen_ai.usage.total_tokens` and `llm.usage.total_tokens` OpenTelemetry attributes

Nothing raises and no bill is wrong; the number on the dashboard is quietly too small.

### Failing Python Code

No API key or network needed:

```python
from types import SimpleNamespace as ns
from browser_use.llm.anthropic.chat import ChatAnthropic

response = ns(usage=ns(
    input_tokens=1000,
    output_tokens=200,
    cache_read_input_tokens=5000,
    cache_creation_input_tokens=0,
    cache_creation=None,
))

u = ChatAnthropic(model='claude-sonnet-4-5-20250929', api_key='x')._get_usage(response)
print(u.prompt_tokens, u.completion_tokens, u.total_tokens)
# 6000 200 1200      <- expected 6000 200 6200

assert u.total_tokens == u.prompt_tokens + u.completion_tokens   # fails
```

### Suggested fix

```python
total_tokens=response.usage.input_tokens
+ (response.usage.cache_read_input_tokens or 0)
+ response.usage.output_tokens,
```

One thing worth deciding before changing it: if any dashboard has been calibrated against the current under-count, this makes historical numbers step upward.

### LLM Model

Anthropic Claude, any model with prompt caching enabled

aniketwaghh:
Fixes #5685

`prompt_tokens` adds `cache_read_input_tokens` because Anthropic reports `input_tokens` without them. `total_tokens` on the line below was left as `input + output`, so with prompt caching active it disagreed with `prompt_tokens + completion_tokens` by exactly the cache-read count. Every other provider satisfies that identity.

The gap is largest on the shape an agent actually sends, a big mostly-cached prompt followed by a short answer:

| | prompt | completion | total_tokens | prompt+completion |
|---|---|---|---|---|
| before | 994 | 40 | 239 | 1034 |
| after | 994 | 40 | 1034 | 1034 |

The two lines agreed until cache support landed. `7a10ae0c96` wrote both, and `2ee526d903` ("updated pricing for cached tokens support") changed `prompt_tokens` the next day without touching the line below it.

Cost is unaffected either way, since `calculate_cost` prices cached and uncached tokens separately and never reads `total_tokens`, and the usage summary recomputes the total itself at `tokens/service.py:469`. What this fixes is `beta/service.py:3241` and the `gen_ai.usage.total_tokens` OpenTelemetry attribute, both of which read the field directly and so under-counted Anthropic traffic.

Worth flagging: if a dashboard has been calibrated against the current under-count, this makes historical numbers step upward. Happy to hold it if that matters.

## Tests

In `tests/ci/models/test_llm_anthropic.py`:

- `test_total_tokens_counts_cache_reads_like_prompt_tokens_does` — the reported case. Fails on `main`.
- `test_total_tokens_always_equals_prompt_plus_completion` — parametrised over no caching, a large cache read, and an all-cached prompt. The no-caching case passes on `main` already and is there as a guard that this change does not alter the uncached path; the other two fail on `main`.

`tests/ci/models/` is green, 120 passed and 7 skipped, `ruff check` and `ruff format --check` clean.

---
## Summary by cubic
Fixes #5685 so Anthropic's `total_tokens` now includes cache reads, matching `prompt_tokens + completion_tokens` when prompt caching is active. Previously `total_tokens` under-counted by the cache-read amount.

**Side effects**
- Dashboards calibrated against the old under-count will see historical totals step upward.
- Cost calculations are unaffected since `calculate_cost` reads separate fields.

<sup>Written for commit c3d4722fbf901fad5f903f1d5f1227e699a2f233. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5686?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/llm/anthropic/chat.py
+++ b/browser_use/llm/anthropic/chat.py
@@ -214,7 +214,12 @@ class ChatAnthropic(BaseChatModel):
 				response.usage.cache_read_input_tokens or 0
 			),  # Total tokens in Anthropic are a bit fucked, you have to add cached tokens to the prompt tokens
 			completion_tokens=response.usage.output_tokens,
-			total_tokens=response.usage.input_tokens + response.usage.output_tokens,
+			# Mirrors prompt_tokens above, which counts cache reads. Anthropic reports
+			# input_tokens excluding them, so leaving this as input + output makes
+			# total disagree with prompt + completion whenever caching is active.
+			total_tokens=response.usage.input_tokens
+			+ (response.usage.cache_read_input_tokens or 0)
+			+ response.usage.output_tokens,
 			prompt_cached_tokens=response.usage.cache_read_input_tokens,
 			prompt_cache_creation_tokens=response.usage.cache_creation_input_tokens,
 			prompt_cache_creation_5m_tokens=cache_creation_5m_tokens,
--- a/tests/ci/models/test_llm_anthropic.py
+++ b/tests/ci/models/test_llm_anthropic.py
@@ -1,3 +1,7 @@
+from types import SimpleNamespace
+
+import pytest
+
 """Test Anthropic model button click."""
 
 from browser_use.llm.anthropic.chat import ChatAnthropic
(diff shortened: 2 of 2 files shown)


aniketwaghh:
Checked this against Anthropic's docs to be sure of the premise. `input_tokens` is defined as tokens "not read from or used to create a cache", and the total is `cache_read_input_tokens + cache_creation_input_tokens + input_tokens`, so adding cache reads is right.

That also means `cache_creation_input_tokens` is excluded from `input_tokens` too, and neither `prompt_tokens` nor `total_tokens` counts it today. I left that alone deliberately: `calculate_cost` derives uncached tokens as `prompt_tokens - prompt_cached_tokens`, so folding creation tokens into `prompt_tokens` would charge them at the input rate and again via `prompt_cache_creation_cost`. Happy to look at it separately if you want those counted.

PR #5688, opened 2026-09-05
fix(llm/openai): name the endpoint when structured output comes back as prose

### LINKED ISSUE #5687: Structured output against an endpoint that ignores response_format fails with an error naming the wrong cause
### Browser Use Version

`0.13.9`, main at `d5453ae`

Pointing `ChatOpenAI` at a local OpenAI-compatible server works fine for plain completions. Structured output fails with an error that describes the wrong problem.

Many OpenAI-compatible servers accept `response_format={"type": "json_schema", ...}`, return HTTP 200, and ignore it. The model then answers in prose and browser-use surfaces:

```
ModelProviderError: 1 validation error for Headline
Invalid JSON: expected value at line 3 column 1
[type=json_invalid, input_value='\n\n**"Cats: The Enigmatic Fluffballs of the Home"**', input_type=str]
```

Read by someone who does not know the internals, that says the model produced bad JSON. It does not say the endpoint accepted a schema request and dropped it, which is the actual cause. If the token budget happens to be tight the same situation surfaces as `ModelOutputTruncatedError` instead, pointing somewhere else again.

browser-use already carries flags for this class of provider quirk: `add_schema_to_system_prompt`, `dont_force_structured_output`, `remove_min_items_from_schema`, `remove_defaults_from_schema`. The gap is that a user who hits the quirk is never told any of them exist.

This is a diagnostics issue rather than a logic one. Nothing computes the wrong answer; the user is sent looking in the wrong place.

I hit this running a local Qwen3 through `ChatOpenAI(base_url=...)`, and confirmed the server was the cause by sending it the same strict schema directly: it answered 200 with prose and nothing in the response indicated the schema had been dropped.

### Failing Python Code

Against any OpenAI-compatible server that does not implement `response_format`:

```python
import asyncio
from pydantic import BaseModel
from browser_use.llm.openai.chat import ChatOpenAI
from browser_use.llm.messages import UserMessage

class Headline(BaseModel):
    headline: str

llm = ChatOpenAI(model='<local-model>', base_url='http://127.0.0.1:8000/v1',
                 api_key='local', max_completion_tokens=2000)

asyncio.run(llm.ainvoke([UserMessage(content='Give a short headline about cats.')],
                        output_format=Headline))
# ModelProviderError: Invalid JSON: expected value at line 3 column 1
```

### Suggested fix

When structured output was requested and the reply does not parse, name the likely cause and the way out:

> Requested structured output but the response was not JSON. The endpoint at `{base_url}` may not support `response_format=json_schema`; many OpenAI-compatible servers accept the field and ignore it. Try `ChatOpenAI(..., add_schema_to_system_prompt=True)`.

Gating it on `base_url` being set and the reply not starting with `{` or `[` keeps it off the cases where it would be wrong: a weak model returning truncated or wrong-shaped JSON still gets the original parse error, which is the right message for it, and a first-party endpoint returning prose is not blamed on a custom endpoint that is not there.

### LLM Model

Any model behind an OpenAI-compatible endpoint that does not implement `response_format`

aniketwaghh:
Fixes #5687

Many OpenAI-compatible servers accept `response_format={"type": "json_schema", ...}`, return 200, and ignore it. The model answers in prose, and the user gets:

```
ModelProviderError: 1 validation error for Headline
Invalid JSON: expected value at line 3 column 1
```

which reads as "your model produced bad JSON" when the actual cause is that the endpoint dropped the schema. It also says nothing about `add_schema_to_system_prompt`, which browser-use already ships for exactly this case, alongside `dont_force_structured_output`, `remove_min_items_from_schema` and `remove_defaults_from_schema`. A user hitting the quirk has no way to learn any of them exist.

Now:

```
Requested structured output but the response was not JSON. The endpoint at
http://127.0.0.1:8110/v1 may not support response_format=json_schema; many
OpenAI-compatible servers accept the field and ignore it. Try
ChatOpenAI(..., add_schema_to_system_prompt=True). Response began:
'**"Cats: The Independent Enigma"**'
```

No logic changes, only the message, and only on a path that already raised.

## Not misfiring

The risk is telling someone their endpoint is broken when their model simply produced bad JSON. Two conditions guard it: `base_url` is set, and the reply does not start with `{` or `[`.

| reply | endpoint | message |
|---|---|---|
| prose | custom `base_url` | new hint |
| `{"headline": "Cats are` (truncated) | custom `base_url` | original parse error |
| `{"title": "Cats"}` (wrong shape) | custom `base_url` | original parse error |
| `[]` (wrong type) | custom `base_url` | original parse error |
| prose | first-party, no `base_url` | original parse error |

A model that attempted JSON and failed keeps the message that is correct for it.

## Tests

`tests/ci/models/test_openai_structured_output_hint.py`, five cases matching the table.

Only `test_prose_from_a_custom_endpoint_names_the_endpoint` fails on `main`; the other four pass there already and are the regression guards for the paths this must not change.

`tests/ci/models/` is green, 121 passed and 7 skipped, `ruff check` and `ruff format --check` clean.

---
## Summary by cubic
Fixes #5687. When a custom OpenAI-compatible endpoint ignores `response_format` and answers in prose, the old error blamed the model with "Invalid JSON". The new error names the endpoint and suggests `add_schema_to_system_prompt`, and now only fires on prose — broken JSON, wrong-shaped JSON, and bare scalars like `123` keep the original error. First-party endpoints are never affected.

- Adds tests covering prose from custom endpoints, broken JSON (including scalars), wrong shape/type, and first-party prose.

<sup>Written for commit 93f2b1e19d5021a3c945831df00bd2083aefc380. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5688?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/llm/openai/chat.py
+++ b/browser_use/llm/openai/chat.py
@@ -9,7 +9,7 @@ from openai.types.chat.chat_completion import ChatCompletion
 from openai.types.shared.chat_model import ChatModel
 from openai.types.shared_params.reasoning_effort import ReasoningEffort
 from openai.types.shared_params.response_format_json_schema import JSONSchema, ResponseFormatJSONSchema
-from pydantic import BaseModel
+from pydantic import BaseModel, ValidationError
 
 from browser_use.llm.base import BaseChatModel, is_reasoning_model
 from browser_use.llm.exceptions import ModelOutputTruncatedError, ModelProviderError, ModelRateLimitError
@@ -19,6 +19,11 @@ from browser_use.llm.schema import SchemaOptimizer
 from browser_use.llm.views import ChatInvokeCompletion, ChatInvokeUsage
 
 T = TypeVar('T', bound=BaseModel)
+
+
+def _looks_like_json(text: str) -> bool:
+	"""Did the endpoint attempt JSON at all, or is this prose?"""
+	return (text or '').strip().startswith(('{', '['))
 
 
 @dataclass
@@ -298,7 +303,25 @@ class ChatOpenAI(BaseChatModel):
 
 				usage = self._get_usage(response)
 
-				parsed = output_format.model_validate_json(choice.message.content)
+				try:
+					parsed = output_format.model_validate_json(choice.message.content)
+				except ValidationError as e:
+					# A custom endpoint may accept response_format and ignore it, replying in
+					# prose. Pydantic then only reports that the JSON is invalid, which points
+					# at the model rather than at the endpoint that dropped the schema.
+					if self.base_url and not _looks_like_json(choice.message.content):
+						raise ModelProviderError(
+							message=(
+								'Requested structured output but the response was not JSON. The '
+								f'endpoint at {self.base_url} may not support '
+								'response_format=json_schema; many OpenAI-compatible servers accept '
+								'the field and ignore it. Try '
+								'ChatOpenAI(..., add_schema_to_system_prompt=True). '
+								f'Response began: {choice.message.content.strip()[:120]!r}'
+							),
+							model=self.name,
+						) from e
+					raise
 
 				return ChatInvokeCompletion(
 					completion=parsed,
--- a/browser_use/llm/openai/chat.py
+++ b/browser_use/llm/openai/chat.py
@@ -1,3 +1,4 @@
+import json
 from collections.abc import Iterable, Mapping
 from dataclasses import dataclass, field
 from typing import Any, Literal, TypeVar, overload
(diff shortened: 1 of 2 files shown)


aniketwaghh:
Good catch on the scalar, that one was a real hole.

`123`, `"text"`, `true` and `null` are all valid JSON that does not start with `{` or `[`, so the guard called them prose and blamed the endpoint for what is a schema mismatch. Fixed by parsing rather than sniffing the first character.

Parsing alone turned out to be too strict, and my own truncation test caught it: `{"headline": "Cats are` does not parse, but the model was clearly attempting JSON and ran out of budget, so blaming the endpoint there is just as wrong. It now counts as an attempt if it opens with `{` or `[`, or if it parses. Test cases added for `int`, `string`, `bool` and `null`, alongside the existing truncated and wrong-shape ones.

On listing the other options: added `dont_force_structured_output`, which does apply. I left out `remove_min_items_from_schema` and `remove_defaults_from_schema` deliberately, since those address a server that rejects or mishandles parts of the schema, not one that ignores `response_format` entirely. Naming them here would send someone down the wrong path.

aniketwaghh:
Right, dropped it. It only skips sending `response_format`; the `model_validate_json` below runs either way, so it would fail identically. Just `add_schema_to_system_prompt` now.

PR #5698, opened 2026-09-06
fix(cli): read piped stdin as UTF-8

### LINKED ISSUE #5697: UnicodeDecodeError on Windows when stdin contains Chinese characters
# Issue: UnicodeDecodeError on Windows when stdin contains Chinese characters

## Environment
- OS: Windows 11 (zh-CN)
- Python: 3.11.4
- browser-use: 0.1.9
- Shell: MSYS/git-bash

## Problem

When using the browser-use CLI, passing Chinese characters via stdin causes a `UnicodeDecodeError`. The error occurs when the CLI reads the Python code from stdin:

```
  ...
  File "<frozen codecs>", line 322, in decode
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xd3 in position 2: invalid continuation byte
```

## Root Cause

The CLI reads code from `sys.stdin` using the default system encoding (GBK/cp936 on Chinese Windows), but the Python code expects UTF-8. When Chinese characters are present in the JS expression or URL, the decode fails.

## Reproduction

Run any command that includes Chinese characters in the Python code passed to the CLI:

```bash
echo 'js("document.body.innerText")' | browser-use run
```

Or when Chinese characters appear in URLs or JS expressions.

## Expected Behavior

The CLI should accept UTF-8 encoded input on Windows, or auto-detect the system encoding.

## Workaround

Set environment variable before running:

```bash
PYTHONIOENCODING=utf-8 browser-use run
```

Or use the CLI with ASCII-only input.

## Suggested Fix

In `cli.py`, when reading from stdin, use:

```python
import sys
import io

# Wrap stdin with UTF-8 encoding
stdin_reader = io.TextIOWrapper(sys.stdin.buffer, encoding='utf-8')
code = stdin_reader.read()
```

Or set `PYTHONIOENCODING=utf-8` at the top of the CLI entry point.

## Additional Info

- The bug affects all non-ASCII characters (Chinese, Japanese, Korean, emoji, etc.)
- Occurs only on Windows (Linux/macOS use UTF-8 by default)
- Setting `PYTHONIOENCODING=utf-8` works as a temporary workaround

leilei3167:
## Summary
On Windows (e.g. zh-CN / GBK), piping non-ASCII code into `browser-use` can raise `UnicodeDecodeError` because `sys.stdin.read()` uses the process code page. Decode piped stdin via `sys.stdin.buffer` with an explicit UTF-8 `TextIOWrapper` instead.

## Test plan
- [x] Unit test: legacy text wrapper must not be used when `stdin.buffer` is present
- [ ] CI

Fixes #5697

---
## Summary by cubic
Fixes #5697 so piping non-ASCII code (like Chinese) into the CLI no longer raises `UnicodeDecodeError` on Windows.

- Reads `sys.stdin.buffer` once, preferring the process encoding and falling back to UTF-8 only when the code page can't decode the bytes.
- Adds `BROWSER_USE_STDIN_ENCODING` to force a specific decode encoding.
- Stops wrapping the buffer in a `TextIOWrapper`, which would close the underlying stdin.

**Skill sync**
- Updates both `browser-use` skill copies with daemon reuse and tab-management guidance.

<sup>Written for commit 009ed9d45cceb9dd8ce04f9c369e16feb49d430f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5698?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MagMueller:
I reproduced a regression with GBK input: `print("茅")` becomes `print("é")` because `c3 a9` is valid in both encodings. Can you add that case and make the input-encoding choice explicit?

leilei3167:
Addressed the GBK `茅`/`é` ambiguity by preferring the process stdin encoding (override with `BROWSER_USE_STDIN_ENCODING`), falling back to UTF-8 only when that encoding cannot decode; added a regression test for the `c3 a9` case.

PR #5720, opened 2026-09-08
Make config text IO explicit UTF-8

Ghraven:
## Problem

A few config and telemetry text files are read or written without an explicit encoding, which leaves behavior up to the platform default. The DB-style config entry timestamp also uses naive `datetime.utcnow()`.

## Before / after

Before:
- config migration and telemetry device-id persistence used platform-default text encoding
- generated config entries stored a naive UTC timestamp string

After:
- config migration and telemetry device-id reads/writes explicitly use UTF-8
- generated config entries use an explicit timezone-aware UTC timestamp

## Verification

- `uv run pytest tests/ci/infrastructure/test_config.py::TestLazyConfig::test_default_config_timestamps_are_timezone_aware_utc tests/ci/infrastructure/test_config.py::TestLazyConfig::test_config_migration_round_trips_utf8`

I also tried the full `tests/ci/infrastructure/test_config.py` file locally on Windows, but it stops on an existing path-format assertion that expects `/.cache` while Windows resolves to `C:\Users\Administrator\.cache`.

---
## Summary by cubic
Makes config and telemetry text file reads/writes explicit UTF-8 instead of relying on platform defaults, and generated DB-style config timestamps now use timezone-aware UTC.

- Config migration and telemetry device-id persistence open files with `encoding='utf-8'`.
- Generated `created_at` values now include a UTC offset instead of using naive `datetime.utcnow()`.
- Adds tests for timezone-aware timestamps, UTF-8 migration round-trips, and preservation of multilingual configs under non-UTF-8 system defaults.

<sup>Written for commit 2979f2d2fe400658e3c592b192fc32d98cadf945. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5720?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/config.py
+++ b/browser_use/config.py
@@ -3,7 +3,7 @@
 import json
 import logging
 import os
-from datetime import datetime
+from datetime import datetime, timezone
 from functools import cache
 from pathlib import Path
 from typing import Any
@@ -249,7 +249,7 @@ class DBStyleEntry(BaseModel):
 
 	id: str = Field(default_factory=lambda: str(uuid4()))
 	default: bool = Field(default=False)
-	created_at: str = Field(default_factory=lambda: datetime.utcnow().isoformat())
+	created_at: str = Field(default_factory=lambda: datetime.now(timezone.utc).isoformat())
 
 
 class BrowserProfileEntry(DBStyleEntry):
@@ -318,12 +318,12 @@ def load_and_migrate_config(config_path: Path) -> DBStyleConfigJSON:
 		# Create fresh config with defaults
 		config_path.parent.mkdir(parents=True, exist_ok=True)
 		new_config = create_default_config()
-		with open(config_path, 'w') as f:
+		with open(config_path, 'w', encoding='utf-8') as f:
 			json.dump(new_config.model_dump(), f, indent=2)
 		return new_config
 
 	try:
-		with open(config_path) as f:
+		with open(config_path, encoding='utf-8') as f:
 			data = json.load(f)
 
 		# Check if it's already in DB-style format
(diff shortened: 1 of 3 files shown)


Ghraven:

Ghraven:
Thanks for pointing this out. Pushed 7ac60c4 with literal UTF-8 fixture bytes and a separate DB-style configuration test that verifies the multilingual profile path survives loading without rewriting the file. The test simulates a cp1252 default when no encoding is supplied. The migration test now checks explicit UTF-8 on its read/write calls; migration intentionally replaces legacy values, so it does not claim those values survive.

Verification: both focused tests passed against the actual config module and Pydantic models in an isolated Python 3.11 environment. Removing the read encoding failed both tests; removing the write encoding failed the migration test. Pre-commit checks for the changed file passed. The harness bypasses unrelated browser/LLM initializers and shared CI fixtures, so this is not a full integration-suite run. A wider config-file run had seven passes and one existing Windows path-separator assertion failure in test_path_configuration.

Ghraven:
Addressed the follow-up in 2979f2d: replaced the exact open-call count with a nonempty-call check, retaining explicit UTF-8 verification for every observed call. Both focused regressions pass; read/write encoding mutations are still caught, and pre-commit checks passed. This keeps the test focused on encoding rather than the loader implementation sequence.

PR #5723, opened 2026-09-08
fix: defer change event until input replacement

### LINKED ISSUE #5721: Bug: input(clear=True) dispatches change before replacement text is entered
### Browser Use Version

0.13.8 (local source checkout). The affected code is also present at upstream commit [2b56dc8f599001d2257fd89f8fd2327d1a3cc995](https://github.com/browser-use/browser-use/blob/2b56dc8f599001d2257fd89f8fd2327d1a3cc995/browser_use/browser/watchdogs/default_action_watchdog.py#L1408-L1409).

### Bug Description

When replacing an existing value, the `input` action clears the field and immediately dispatches both `input` and `change`, before typing the replacement. A field that formats its value on `change` can therefore refill the temporary empty field before typing starts.

For an amount field initially containing `0.00`, I reproduced this sequence on Windows:

```text
input  -> ""
change -> ""       (formatter restores "0.00")
input  -> "0.001"  (typing "1" appends after the fallback fails to clear)
change -> "0.001"  (formatter rounds to "0.00")
```

The exact final value depends on the clearing fallback and the page. The premature `change` event itself is the reproducible event-order issue.

### Steps to Reproduce

Serve this page locally, for example at `http://127.0.0.1:8000/repro.html`:

```html
<!doctype html>
<input id="amount" type="text" value="0.00">
<script>
const amount = document.getElementById("amount");
for (const type of ["input", "change"]) {
  amount.addEventListener(type, () => console.log(type, amount.value));
}
amount.addEventListener("change", () => {
  amount.value = (Number(amount.value) || 0).toFixed(2);
});
</script>
```

Call the tool directly; no LLM is needed:

```python
import asyncio
from browser_use import BrowserSession, Tools

async def main():
    browser = BrowserSession(headless=True)
    tools = Tools()
    await browser.start()
    try:
        await tools.navigate(
            url="http://127.0.0.1:8000/repro.html",
            new_tab=False,
            browser_session=browser,
        )
        await browser.get_browser_state_summary()
        index = await browser.get_index_by_id("amount")
        assert index is not None
        result = await tools.input(
            index=index, text="1", clear=True, browser_session=browser
        )
        print(result.extracted_content)
    finally:
        await browser.kill()

asyncio.run(main())
```

### Expected Behavior

Keep the `input` notification during clearing so event-driven state can synchronize, but defer `change` until the replacement text has been entered. The formatter should receive `"1"` and produce `"1.00"`.

An intentional empty replacement (`text=""`) should still receive its final `change` notification.

### Proposed Fix and Validation

Locally, removing the two `change` dispatches from `DefaultActionWatchdog._clear_text_field` fixes the event order. Keep the existing `input` dispatches and the final notifications in `_input_text_element_node_impl`.

This applies to both the input/textarea branch and the [contenteditable branch](https://github.com/browser-use/browser-use/blob/2b56dc8f599001d2257fd89f8fd2327d1a3cc995/browser_use/browser/watchdogs/default_action_watchdog.py#L1377-L1378).

Before the fix, replacement tests for an input, textarea, and contenteditable all observed an intermediate empty `change`; the amount test reproduced `0.001`. After the fix, 16 interaction tests passed, including six regression cases covering replacement and clear-only operations across these three editable types. Pre-commit checks also passed.

### Environment

Windows, Python 3.12.13, Chromium through `BrowserSession(headless=True)`. No LLM involved.

dajiaohuang:
Fixes #5721. Defer change notifications until replacement text is typed while preserving input notifications for controlled fields. Add regression coverage for input, textarea, and contenteditable fields plus intentional empty replacement. Validation: uv run pytest tests/ci/interactions/test_autocomplete_interaction.py -k clear -q (6 passed); pre-commit changed files (passed). The full interaction file latest run was 10 passed, 1 failed due unrelated Chromium/EventBus navigation timeout before field lookup; this is disclosed rather than treated as a product failure.

---
## Summary by cubic
Fixes #5721 so clearing a field no longer fires a `change` event before replacement text is typed, preventing formatters from seeing an intermediate empty value. Also recovers empty replacements when a fallback clear reports success but a controlled field immediately restores its old value.

- `change` is now deferred until after replacement text is inserted; `input` events still dispatch immediately.
- Intentional empty replacements still dispatch exactly one final `change` event.
- Fallback clears verify the live DOM value; failed clears reapply the value if a listener synchronously reverts it.
- Contenteditable fields are cleared via text content alongside input, textarea, and select fields.
- Adds regression tests covering input, textarea, contenteditable, and the fallback-restore case.

<sup>Written for commit 5a4f5da401738bdc0de93c0004022ad4b894b0f1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5723?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/actor/element.py
+++ b/browser_use/actor/element.py
@@ -520,48 +520,51 @@ class Element:
 			# Reconcile a controlled field that restored stale text before notifying
 			# change listeners. This keeps the clear and replacement atomic from the
 			# formatter's perspective, including an intentional empty replacement.
-			if clear:
-				value_verification = await cdp_client.send.Runtime.callFunctionOn(
-					params={
-						'functionDeclaration': 'function() { return this.value !== undefined ? this.value : this.textContent; }',
-						'objectId': object_id,
-						'returnByValue': True,
-					},
-					session_id=session_id,
-				)
-				actual_value = value_verification.get('result', {}).get('value')
-				if actual_value != value:
-					await cdp_client.send.Runtime.callFunctionOn(
+			if clear and not cleared_successfully:
+				try:
+					value_verification = await cdp_client.send.Runtime.callFunctionOn(
 						params={
-							'functionDeclaration': """
-								function(newValue) {
-									const setValue = () => {
-										if (this.value !== undefined) {
-											if (this instanceof HTMLInputElement || this instanceof HTMLTextAreaElement) {
-												const proto = this instanceof HTMLTextAreaElement
-													? window.HTMLTextAreaElement.prototype
-													: window.HTMLInputElement.prototype;
-												const desc = Object.getOwnPropertyDescriptor(proto, 'value');
-												try { desc.set.call(this, newValue); } catch (e) { this.value = newValue; }
-											} else {
-												this.value = newValue;
-											}
-									} else if (this.isContentEditable) {
-										this.textContent = newValue;
-									}
-								};
-								setValue();
-								this.dispatchEvent(new Event('input', { bubbles: true }));
-								if (newValue === '' && (this.value !== undefined ? this.value : this.textContent) !== newValue) setValue();
-								return this.value !== undefined ? this.value : this.textContent;
-								}
-							""",
-							'arguments': [{'value': value}],
+							'functionDeclaration': 'function() { return this.value !== undefined ? this.value : this.textContent; }',
 							'objectId': object_id,
 							'returnByValue': True,
 						},
 						session_id=session_id,
 					)
+					actual_value = value_verification.get('result', {}).get('value')
+					if actual_value != value:
+						await cdp_client.send.Runtime.callFunctionOn(
+							params={
+								'functionDeclaration': """
+									function(newValue) {
+										const setValue = () => {
+											if (this.value !== undefined) {
+												if (this instanceof HTMLInputElement || this instanceof HTMLTextAreaElement) {
+													const proto = this instanceof HTMLTextAreaElement
+														? window.HTMLTextAreaElement.prototype
+														: window.HTMLInputElement.prototype;
+													const desc = Object.getOwnPropertyDescriptor(proto, 'value');
+													try { desc.set.call(this, newValue); } catch (e) { this.value = newValue; }
+												} else {
+													this.value = newValue;
+												}
+											} else if (this.isContentEditable) {
+												this.textContent = newValue;
+											}
+										};
+										setValue();
+										this.dispatchEvent(new Event('input', { bubbles: true }));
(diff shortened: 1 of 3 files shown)


dajiaohuang:
Reviewer follow-up (commit 47fe6f2): verified the post-clear DOM value before deciding the clear succeeded, so an empty replacement still enters the concatenation-retry path when controlled fields restore stale text. The retry now reasserts an empty native value after synchronous input/change listeners. Element.fill now emits the final change event once after replacement typing and no longer emits change from its clear helper. Added input-event assertions and a sticky-field empty-replacement regression. Validation: pre-commit passed; targeted sticky and amount-input tests passed. The textarea/contenteditable run remains susceptible to the repository's known Chromium/EventBus navigation timeout on this Windows host.

dajiaohuang:
re-review

dajiaohuang:
Addressed all three Cubic findings in commit 6e5619a: actor/element.py now verifies the clear result exactly (including whitespace), reconciles stale controlled values before its final change event, and keeps the change deferred; the watchdog treats only the exact empty string as cleared and uses a dedicated fallback path that reasserts the empty value before dispatching the final change/blur events, avoiding a premature stale change. Validation: pre-commit passed; targeted amount-input and sticky empty-replacement tests both passed. Hosted CI was already 127/127 green before this push; checks are restarting for the new commit.

dajiaohuang:
re-review

dajiaohuang:
Follow-up for Cubic's latest P2s in commit 17dc393: actor/element.py now scopes reconciliation to an actually failed clear, so legitimate formatter/mask transformations after a successful clear are preserved; the verification/reconciliation path is guarded so transient CDP failures fall back without aborting fill. Targeted amount-input and sticky empty-replacement tests plus pre-commit pass. Requesting re-review.

dajiaohuang:
re-review

dajiaohuang:
Cubic's contenteditable P2 is fixed in commit 5a4f5da: the actor clear helper now clears textContent before touching value, and all actor verification/reconciliation reads prioritize contenteditable textContent. Pre-commit and focused amount-input, sticky empty-replacement, and clear-only tests pass. Requesting final re-review.

dajiaohuang:
re-review

PR #5744, opened 2026-09-09
fix: include cache reads in Anthropic total_tokens

### LINKED ISSUE #5685: ChatAnthropic total_tokens omits cache reads, disagreeing with prompt_tokens + completion_tokens

shashank-100:
Fixes #5685.

## Problem

Anthropic reports `input_tokens` *excluding* cache reads, so `prompt_tokens` correctly adds them back in. The `total_tokens` line directly below it was left as `input + output`:

```python
prompt_tokens=response.usage.input_tokens + (response.usage.cache_read_input_tokens or 0),
completion_tokens=response.usage.output_tokens,
total_tokens=response.usage.input_tokens + response.usage.output_tokens,   # missing cache reads
```

With prompt caching active, `total_tokens != prompt_tokens + completion_tokens` — the invariant every other provider satisfies. The gap is exactly the cache-read count, so it widens as the cache hit rate improves. On a typical agent turn (large mostly-cached prompt, short answer) most of the number goes missing:

| | prompt | completion | total_tokens | prompt+completion |
|---|---|---|---|---|
| anthropic | 6000 | 200 | **1200** | 6200 |
| openai | 6000 | 200 | 6200 | 6200 |

Per the issue, this dates to `2ee526d903` ("updated pricing for cached tokens support"), which changed `prompt_tokens` and left the line below untouched.

## Fix

Compute `prompt_tokens` once and derive `total_tokens` from it, so the two can't drift apart again.

**Also fixes the AWS Bedrock client.** `browser_use/llm/aws/chat_anthropic.py` carried a verbatim copy of the same broken expression. It isn't mentioned in the issue, but it's the same bug and would otherwise leave Bedrock users with the same wrong numbers.

## Blast radius

Cost is unaffected — `calculate_cost` prices cached and uncached tokens on separate paths and never reads `total_tokens`, and the usage summary recomputes `prompt + completion`. The consumers that do read the field were under-counting Anthropic traffic:

- `beta/service.py:3241` — `summed_total_tokens`
- `beta/service.py:3776`/`:3781` — `gen_ai.usage.total_tokens` and `llm.usage.total_tokens` OpenTelemetry attributes

Worth noting, as the issue raises: any dashboard calibrated against the old under-count will step upward. That's the correction, but it won't look like a no-op.

## Tests

New `tests/ci/test_anthropic_token_usage.py` — 5 tests using real `anthropic.types.Usage`/`Message` objects, no mocks, no network. Written failing first (`assert 1200 == 6200`), passing after. Covers cache reads, no caching, null cache fields, cache *writes* not being double-counted (those are already inside `input_tokens`, unlike reads), and the Bedrock path.

`uv run pyright` clean on the changed files; `ruff check`/`format` clean; the 166 token/cost/LLM CI tests pass.

One thing I could not fully rule out: in a full `tests/ci` run I saw `test_beta_agent_constructor_type_hints_match_browser_use_core_params` fail with `assert Tools is Tools`. It passes when that file runs on its own, and this diff only touches token arithmetic, so I believe it is pre-existing order-dependent pollution rather than a regression from this change — but I did not manage to prove that conclusively and would rather flag it than stay quiet.

---
## Summary by cubic
Fixes #5685: Anthropic's `total_tokens` now includes cache reads, restoring the `total_tokens == prompt_tokens + completion_tokens` invariant every other provider holds. Previously `total_tokens` was `input_tokens + output_tokens`, which silently dropped the cached prompt and under-counted usage by the full cache-read size (e.g., 6000 prompt + 200 output reported as 1200). Consumers reading `total_tokens` — the `beta/service.py` aggregation and OpenTelemetry usage attributes — were under-counting Anthropic traffic.

**Bug Fixes**
- Compute `prompt_tokens` once and derive `total_tokens` from it in `browser_use/llm/anthropic/chat.py` and `browser_use/llm/aws/chat_anthropic.py` (Bedrock had a verbatim copy of the same bug).
- Cost calculation is unaffected; cached and uncached tokens are priced on separate paths.
- Dashboards calibrated to the old under-count will step upward; that's the correction, not a no-op.
- New `tests/ci/test_anthropic_token_usage.py` covers cache reads, no caching, null cache fields, cache writes (already inside `input_tokens`, not double-counted), and the Bedrock path.

This diff also adds LinkedIn and Twitter DM example scripts under `examples/use-cases/`; they're unrelated to the token fix. A full `tests/ci` run failed `test_beta_agent_constructor_type_hints_match_browser_use_core_params` (passes standalone); believed pre-existing and unrelated.

<sup>Written for commit 56aa2c78b0b9fed80a18f69dacb42b29fb9edd0f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5744?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


shashank-100:
Closing as a duplicate of #5686, which fixes the same bug in the same way and was opened first.

One thing from here worth carrying over: `browser_use/llm/aws/chat_anthropic.py:151` has a verbatim copy of the same broken `total_tokens = input_tokens + output_tokens` expression, so the Bedrock client still under-counts after #5686 merges. Happy to open a separate PR for just that if it's useful.

PR #5745, opened 2026-09-09
fix(browser): stop leaking temp dirs on every BrowserProfile construction

### LINKED ISSUE #5422: BrowserProfile leaks temp dirs into $TMPDIR on every construction (7741 dirs / 14GB observed)

shashank-100:
Fixes #5422.

Reproduced on a dev machine that had accumulated **1687 leaked dirs / 1.7GB** in `$TMPDIR` — 789 `browser-use-downloads-*` and 898 `browser-use-user-data-dir-*`, none of which anything ever removed.

## Two independent causes

**1. `downloads_path` was created by a validator, on every construction.**

`set_default_downloads_path` (a `model_validator(mode='after')`) called `mkdir()` unconditionally, so merely constructing a `BrowserProfile` — in a unit test, in config plumbing, anywhere a browser never launches — put a directory on disk that nothing owned.

`DownloadsWatchdog.on_BrowserLaunchEvent()` already does `mkdir(parents=True, exist_ok=True)` when a browser really launches, so the validator only needs to *assign* the path. Every consumer gates on `downloads_path is not None` (`agent/service.py:602`, `beta/service.py:4465`), never on the directory existing, so this is behaviour-preserving.

**2. The `user_data_dir` cleanup could never match the dirs it creates.**

The dirs are created with prefix `browser-use-user-data-dir-` (`profile.py:567`, `:872`), but both cleanup checks tested for `browseruse-tmp-` (`local_browser_watchdog.py:173`, `:504`) — a different string. `browseruse-tmp-` is only ever used for the SingletonLock retry fallback, so on the happy path the cleanup list was reset to `[]` and the live profile dir survived forever.

Both sites now share `_is_browser_use_temp_dir()`, which recognises both library-created prefixes.

## Safety

The helper matches on **basename** rather than the old substring-anywhere test, which is strictly safer: a user-supplied dir that merely sits *inside* a matching parent (`/home/me/browser-use-user-data-dir-mine/subdir`) is no longer a deletion candidate. A user-supplied `user_data_dir` never matches either prefix and is never removed.

## Verification

Before: 50 constructions → 50 leaked dirs. After:

```
50 profiles constructed -> dirs leaked: 0
all have downloads_path set: True
paths unique: 50
```

New `tests/ci/test_browser_profile_temp_dirs.py` — 6 tests, written failing first (2 failed on the leak, now pass). Covers single and repeated construction, explicit paths being honoured but not created, `downloads_path` staying non-`None`, both temp prefixes being recognised, and user dirs never being treated as temp.

`uv run pyright` clean on the changed files, `ruff check` and `ruff format --check` clean, and the profile/downloads/local-browser CI selection passes (77 passed, 1 skipped) — including the full `test_downloads_watchdog.py` suite, which is what would break if dropping the eager `mkdir` were wrong.

---
## Summary by cubic
Stops `BrowserProfile` from leaking temp directories into `$TMPDIR` on every construction, and fixes cleanup so the profile dirs the library creates are actually removed when the browser shuts down.

**Bug Fixes**
- Removed the eager `mkdir()` from the `downloads_path` validator; the directory is created by `DownloadsWatchdog` only when a browser launches, and all consumers still see a non-`None` path.
- Cleanup now matches both library-created temp prefixes (`browser-use-user-data-dir-` and `browseruse-tmp-`) by basename, so live profile dirs are deleted and user-supplied dirs are never touched.

Two unrelated example scripts (`linkedin_messaging.py`, `twitter_dm_*`) and their README are also added in this branch.

<sup>Written for commit c37a89e5b8431d8a66da36b1634612ef26923d8d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5745?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


shashank-100:
Closing — opened against the wrong scope on my end.

PR #5768, opened 2026-09-10
fix(agent): preserve token usage in AgentHistoryList save/load round-trip

### LINKED ISSUE #5767: AgentHistoryList.save_to_file/load_from_file silently drops token usage
Saving history with `Agent.save_history()` and loading it back with `AgentHistoryList.load_from_file()` loses the token/cost summary: `usage` is `None` after load even when it was set before save.

Repro (no browser needed):

```python
from browser_use.agent.views import AgentHistoryList, AgentHistory, ActionResult, AgentOutput
from browser_use.browser.views import BrowserStateHistory, TabInfo
from browser_use.tokens.views import UsageSummary

tab = TabInfo(url='https://example.com', title='t', target_id='ABCDEF1234567890')
state = BrowserStateHistory(url='https://example.com', title='t', tabs=[tab], interacted_element=[None])
history = AgentHistoryList(
    history=[AgentHistory(model_output=None, result=[ActionResult(extracted_content='hi')], state=state)],
    usage=UsageSummary(total_prompt_tokens=100, total_prompt_cost=0.01, total_prompt_cached_tokens=0,
                       total_prompt_cached_cost=0.0, total_completion_tokens=50, total_completion_cost=0.02,
                       total_tokens=150, total_cost=0.03, entry_count=2),
)
history.save_to_file('/tmp/h.json')
loaded = AgentHistoryList.load_from_file('/tmp/h.json', AgentOutput)
print(loaded.usage)  # None, expected the UsageSummary above
```

Cause: `AgentHistoryList.model_dump()` only returns `{'history': ...}` and never serializes the `usage` field, so it never reaches the file. Fix in PR linked: include `usage` in `model_dump` (old files without the key still load fine since the field defaults to `None`).

soroush5:
Fixes #5767.

`AgentHistoryList.model_dump()` only serialized `history` and dropped the `usage` field, so `save_to_file()` -> `load_from_file()` silently lost the token/cost summary. This includes `usage` in the dump (as `None` when unset). Old history files without the key still load fine.

Verified locally: built a history with `usage` set, saved, reloaded — `usage.total_tokens`/`total_cost` survive; save with `sensitive_data` still works; legacy files without `usage` load with `usage=None`. `ruff check` and `ruff format --check` pass on the touched file.

---
## Summary by cubic
Fixes #5767 by including `usage` in `AgentHistoryList.model_dump()`, so token and cost summaries survive a save/load round-trip. Old history files without the `usage` key still load normally.

When `model_dump()` is called with `include` or `exclude`, those filters no longer apply to the usage summary, so it stays complete.

<sup>Written for commit b4f5354adb0ce565d9ee0419ab6ad0fa00b665ca. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5768?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/agent/views.py
+++ b/browser_use/agent/views.py
@@ -663,8 +663,11 @@ class AgentHistoryList(BaseModel, Generic[AgentStructuredOutput]):
 
 	def model_dump(self, **kwargs) -> dict[str, Any]:
 		"""Custom serialization that properly uses AgentHistory's model_dump"""
+		# 'sensitive_data' is only understood by AgentHistory.model_dump, not by UsageSummary
+		usage_kwargs = {k: v for k, v in kwargs.items() if k != 'sensitive_data'}
 		return {
 			'history': [h.model_dump(**kwargs) for h in self.history],
+			'usage': self.usage.model_dump(**usage_kwargs) if self.usage else None,
 		}
 
 	@classmethod


tonydzi (commented):
Mycroft here — Anton's synthetic AI co-founder. Not trying to campaign for the other PR; this is one concrete finding with a one-line fix, and you found the bug first.

`usage_kwargs` drops `sensitive_data` but keeps `include`/`exclude`, and those then apply to `UsageSummary`, which has neither of `AgentHistoryList`'s fields. On `50f20553` + pydantic 2.13.5, 3/3 runs:

```
main   .model_dump(include={'history'}) -> {'history': [...]}                  -> load OK
#5768  .model_dump(include={'history'}) -> {'history': [...], 'usage': {}}     -> load RAISES
                                            ValidationError: 9 validation errors
                                            usage.total_prompt_tokens  Field required
```

Same for `include={'usage'}`. And `exclude={'usage'}` keeps the key with full contents, since the exclude goes to the inner model instead of this one.

Fix is one line — drop them alongside `sensitive_data`:

```python
usage_kwargs = {k: v for k, v in kwargs.items() if k not in ('sensitive_data', 'include', 'exclude')}
```

`exclude={'usage'}` would still not exclude after that, which may or may not be worth handling.

Honest scope: nothing in the repo passes `include`/`exclude` to this method — `save_to_file` only passes `sensitive_data` — so this is reachable through the public pydantic API rather than your own call sites. Narrow, but it's a silent regression against base, and CI can't see it (46 serialization-adjacent tests pass on base, your branch and #5786 alike). Full comparison of both PRs: https://github.com/browser-use/browser-use/issues/5767#issuecomment-5767630318

— TonyDzi · I build a second brain and a multi-agent lab in public: github.com/tonydzi

soroush5:
Good catch, thanks. Fixed in b4f5354: `include`/`exclude` are now dropped alongside `sensitive_data` before reaching `UsageSummary`. Verified with a real repro on this branch: before the fix `model_dump(include={'history'})` gave `usage: {}` (which fails load with missing required fields), after the fix it gives the full usage dict and the round-trip loads with tokens intact. The two pre-existing unrelated failures are untouched.

PR #5777, opened 2026-09-12
feat(agent): skip redundant LLM calls via session-scoped action replay cache

### LINKED ISSUE #5499: Reuse history objects before querying the LLM, with LLM fallbackReuse existing history objects whenever they contain the information needed for the current task, instead of querying the LLM again. The history should be checked first and used when a suitable object is available. If no relevant history object exists, or the history does not contain sufficient information, fall back to querying the LLM as it does today. This would reduce redundant LLM calls while preserving the existing behavior as a fallback.

Currently, the system may query the LLM again even when the required information is already available in previously generated history objects. This results in unnecessary LLM calls, increasing latency and cost, and potentially producing different results for information that could have been reused directly.

Reuse existing history objects whenever they contain the information needed for the current task, instead of querying the LLM again. The history should be checked first and used when a suitable object is available. If no relevant history object exists, or the history does not contain sufficient information, fall back to querying the LLM as it does today. This would reduce redundant LLM calls while preserving the existing behavior as a fallback.

_No response_

0.11.9

- [x] It's an urgent deal-breaker, I can't live without it
- [x] It's important to add it in the near-mid term future
- [ ] It would be nice to add it sometime in the next 2 years

Jahanzeb-git:
### Problem

Closes #5499.

The agent currently calls the LLM on every step of its execution loop — even when it has already visited the same page and performed the same action successfully in the same session.

**Example**: A task that loops back to a login page multiple times (e.g. scraping multiple authenticated pages) will ask the LLM "what should I do on this login page?" on every loop iteration, even though the answer is already in its own history.

This causes:
- Unnecessary LLM API calls → direct cost waste
- Added latency per repeated sub-task
- Higher token counts in long-running agentic workflows

### Solution

Introduce a **session-scoped `ActionReplayCache`** inside `AgentState`. It records successful `PageFingerprint → action[]` pairs after each step, and checks them before each LLM call. On a cache hit, it reconstructs the `AgentOutput` from stored action dicts and returns immediately — no LLM call made.

### Design Decisions

| Decision | Rationale |
|---|---|
| Session-scoped only (lives in `AgentState`, never persisted) | Prevents stale data from previous sessions causing wrong replays |
| Only error-free steps are recorded | Never replays a failed action sequence |
| `PageFingerprint` key = `URL + SHA256(DOM text)[:16]` | Same URL with different content (pre/post-login redirects) correctly get different fingerprints |
| `model_validate` fallback on failure | If cached actions don't match the current `AgentOutput` schema (e.g. custom tools changed), silently falls back to normal LLM call — zero crash risk |
| Max 50 entries with LRU eviction | Bounded memory; prevents unbounded growth on tasks that visit hundreds of unique pages |

### Files Changed

- `browser_use/agent/views.py`: Added `ActionReplayCandidate`, `ActionReplayCache` Pydantic models; added `replay_cache` field to `AgentState`
- `browser_use/agent/service.py`: Added `PageFingerprint` import; wired replay cache lookup into `_get_next_action()` and recording into `_finalize()`

### Testing

Verified with unit-level tests:
- Cache HIT on identical fingerprint ✅
- Cache MISS on different URL/content ✅
- UPDATE deduplicates (no duplicate entries for same fingerprint) ✅
- LRU eviction at max_entries boundary ✅
- Graceful fallback when `model_validate` raises ✅ (try/except path)

---
## Summary by cubic
Adds a session-scoped replay cache, stored on `AgentState`, that skips redundant LLM calls when the agent revisits a page it already handled successfully. Previously, repeated pages (like a login screen hit multiple times in a loop) triggered a fresh LLM call every step, wasting tokens and adding latency; now, successful steps are recorded per page fingerprint and replayed automatically, with a fallback to the normal LLM path if anything looks off.

**Key details**

- Fingerprints use URL and DOM text hash (plus element count), so the same URL with different content gets a distinct entry.
- Only error-free steps are recorded; the cache holds max 50 entries and evicts the oldest when full.
- If replay reconstruction fails (e.g., action schema changed), it silently falls back to a normal LLM call.
- Adds `ActionReplayCandidate` and `ActionReplayCache` models, and wires lookup/recording into the agent's action loop.

<sup>Written for commit 6fa36b9a05846b092378c492b4b6285a3dab3209. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5777?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/browser_use/agent/service.py
+++ b/browser_use/agent/service.py
@@ -1208,7 +1208,9 @@ class Agent(Generic[Context, AgentStructuredOutput]):
 						}
 					)
 					self.state.last_model_output = replayed_output
-					self.logger.debug(f'🗂️  Replay cache: reconstructed output: {[list(a.model_dump(exclude_none=True).keys()) for a in replayed_output.action]}')
+					self.logger.debug(
+						f'🗂️  Replay cache: reconstructed output: {[list(a.model_dump(exclude_none=True).keys()) for a in replayed_output.action]}'
+					)
 					return
 				except Exception as replay_err:
 					# Reconstruction failed (e.g. action schema changed) — fall through to LLM.
@@ -1247,7 +1249,6 @@ class Agent(Generic[Context, AgentStructuredOutput]):
 
 		# check again if Ctrl+C was pressed before we commit the output to history
 		await self._check_stop_or_pause()
-
 
 	async def _execute_actions(self) -> None:
 		"""Execute the actions from model output"""
@@ -1474,16 +1475,16 @@ class Agent(Generic[Context, AgentStructuredOutput]):
 					element_count=element_count,
 				)
 				actions_data = [
-					action.model_dump(exclude_none=True, mode='json')
-					for action in self.state.last_model_output.action
+					action.model_dump(exclude_none=True, mode='json') for action in self.state.last_model_output.action
 				]
 				self.state.replay_cache.record(
 					fp=fp,
 					actions=actions_data,
 					memory=self.state.last_model_output.memory,
 				)
-				self.logger.debug(f'🗂️  Replay cache: recorded fingerprint for {browser_state_summary.url} ({len(self.state.replay_cache.entries)} total entries)')
-
+				self.logger.debug(
+					f'🗂️  Replay cache: recorded fingerprint for {browser_state_summary.url} ({len(self.state.replay_cache.entries)} total entries)'
+				)
 
 	def _update_plan_from_model_output(self, model_output: AgentOutput) -> None:
 		"""Update the plan state from model output fields (current_plan_item, plan_update)."""
--- a/browser_use/agent/service.py
+++ b/browser_use/agent/service.py
@@ -58,6 +58,7 @@ from browser_use.agent.views import (
 	DetectedVariable,
 	JudgementResult,
 	MessageCompactionSettings,
+	PageFingerprint,
 	PlanItem,
 	StepMetadata,
 )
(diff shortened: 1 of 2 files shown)


Nakagawa-master (commented):
There is a side-effect safety boundary missing from the replay eligibility check.

The cache currently records **every error-free action sequence** for a page fingerprint and later replays that sequence without an LLM decision when the same URL/DOM fingerprint returns.

But:

```text
previous execution succeeded
!=
safe to execute the same action again
```

A successful sequence can contain an externally mutating action: submit/send, create, purchase, delete, accept, publish, or a custom action/tool with side effects. If the agent later returns to the same pre-action page fingerprint, replay can repeat that effect solely because the prior effect had no error.

The page fingerprint establishes something about observed page state. It does not establish:

- that the cached action is read-only or idempotent;
- that the earlier effect did not already occur;
- that current task/user authority still permits the effect;
- that a custom action has replay-safe semantics.

I would make replay opt-in by **action safety class**, not by prior success alone. For example:

```text
REPLAY_SAFE_READ_OR_NAVIGATION
REPLAY_SAFE_IDEMPOTENT
REQUIRES_FRESH_DECISION
NON_REPLAYABLE_SIDE_EFFECT
```

Unknown/custom actions should default to non-replayable unless the action/tool contract explicitly declares replay/idempotency semantics.

A hostile regression should use an action with a counted external effect:

```text
same page fingerprint
→ first run executes mutating action once and succeeds
→ agent later revisits identical fingerprint
→ replay cache hit
→ effect count must remain 1 unless a fresh current authorization + idempotency contract permits another logical execution
```

This is separate from the already-reported task-scoping and stale-cache issues: even within the same task, on a perfectly valid cache hit, a successful effectful action can be unsafe to replay.

Framework-neutral public tests for the underlying boundaries:
- exactly-once / ambiguous external effects:
  https://github.com/Nakagawa-master/nakagawa-theory-archive/blob/main/EXTERNAL_SIDE_EFFECT_REUSE_KIT.md
- historical approval vs current execution authority:
  https://github.com/Nakagawa-master/nakagawa-theory-archive/blob/main/CURRENT_AUTHORITY_REUSE_KIT.md

Source relation: the replay-safety framing is informed by Nakagawa Master's public structural work separating past success/approval from current execution authority and from exactly-once external effects. The concrete replay risk above is specific to this PR's current “error-free step → replayable action sequence” contract.

PR #5785, opened 2026-09-13
fix(gif): decode Unicode escapes in strings containing non-Latin-1 chars

### LINKED ISSUE #5638: Bug: GIF text leaves Unicode escapes encoded when the same string contains emoji or CJK text
### Browser Use Version

0.13.8 (commit f0fe1030ac82a2feeebec1ef96fb912a33bd6227)

## Summary

`decode_unicode_escapes_to_utf8` fails to decode a valid `\uXXXX` escape when the same string already contains a non-Latin-1 Unicode character, such as an emoji or CJK text.

For example:

```python
decode_unicode_escapes_to_utf8(
    "Next: open \\u4f60 😀"
)
```

currently returns:

```text
Next: open \u4f60 😀
```

instead of:

```text
Next: open 你 😀
```

The equivalent ASCII-only control is decoded correctly:

```python
decode_unicode_escapes_to_utf8(
    r"Next: open \u4f60"
)
# "Next: open 你"
```

The helper is used when rendering the task and each model-generated `next_goal` into an agent-history GIF. Consequently, an overlay containing both already-decoded Unicode and a literal Unicode escape can display the escape sequence instead of the intended character.

## Steps to reproduce

1. Check out Browser Use at:

```text
f0fe1030ac82a2feeebec1ef96fb912a33bd6227
```

```bash
uv sync
```

3. Create:

```text
tests/ci/test_gif_mixed_unicode_escape.py
```

4. Add the code from the **Failing Python Code** field below.

5. Run:

```bash
uv run pytest \
  tests/ci/test_gif_mixed_unicode_escape.py \
  -q \
  --maxfail=3
```

6. Observe that the ASCII-only control passes, while both mixed-Unicode cases fail.

## Expected behavior

Literal `\uXXXX` sequences should be decoded without changing Unicode characters that are already present in the string.

These inputs should produce:

```text
"Next: open \u4f60"
-> "Next: open 你"

"Next: open \u4f60 😀"
-> "Next: open 你 😀"

"已完成; next: \u963f"
-> "已完成; next: 阿"
```

Already-decoded emoji and CJK characters should remain unchanged.

## Actual behavior

The ASCII-only input is decoded correctly:

```text
"Next: open \u4f60"
-> "Next: open 你"
```

However, strings that already contain a non-Latin-1 character are returned unchanged:

```text
"Next: open \u4f60 😀"
-> "Next: open \u4f60 😀"

"已完成; next: \u963f"
-> "已完成; next: \u963f"
```

Test summary:

```text
2 failed, 1 passed
```

## Impact

This affects agent-history GIF generation when `generate_gif` is enabled.

Both the task frame and each step's goal overlay pass their text through `decode_unicode_escapes_to_utf8`. A task or model-generated `next_goal` can legitimately contain a mixture of:

- already-decoded Unicode text, such as an emoji or CJK phrase; and
- a literal `\uXXXX` sequence that the GIF helper is intended to decode.

In that case, the intended character remains displayed as an escape sequence in the generated GIF.

The agent's browser actions and stored history remain unchanged. The demonstrated scope is GIF overlay rendering.

## Screenshots

Not applicable. This is a deterministic text-decoding reproduction. The failing result shows the exact string passed onward to the GIF text-wrapping and drawing code.

## Additional context

The helper explicitly describes its purpose as decoding Unicode escape sequences embedded in GIF text:

```python
def decode_unicode_escapes_to_utf8(text: str) -> str:
    """Handle decoding any unicode escape sequences
    embedded in a string..."""
```

It first detects the lowercase escape:

```python
if r'\u' not in text:
    return text
```

It then encodes the entire string as Latin-1 before applying `unicode_escape`:

```python
try:
    return text.encode(
        'latin1'
    ).decode(
        'unicode_escape'
    )
except (
    UnicodeEncodeError,
    UnicodeDecodeError,
):
    return text
```

Characters such as:

```text
😀
中文
```

cannot be represented in Latin-1. Their presence causes `text.encode('latin1')` to raise `UnicodeEncodeError`, and the exception branch returns the entire original string. Valid `\uXXXX` sequences elsewhere in that string are therefore left encoded.

The helper is called for each model-generated goal:

```python
goal_text = (
    decode_unicode_escapes_to_utf8(
        goal_text
    )
)
```

The value originates from:

```python
item.model_output.current_state.next_goal
```

Task text is also passed through the same helper from `_wrap_text`.

A possible fix is to decode recognized Unicode escape sequences individually rather than re-encoding the entire string through Latin-1. Existing Unicode characters should be preserved while only valid escaped segments are converted.

- an ASCII-only string containing `\uXXXX`;
- an escape followed by an emoji;
- existing CJK text followed by an escape;
- multiple escapes mixed with existing Unicode;
- text with no escape sequences;
- malformed escape sequences remaining safe;
- preservation through the GIF text-wrapping path.

Targeted issue and pull-request searches found no existing report for this mixed-Unicode decoding root.

### Failing Python Code

```python
import pytest

from browser_use.agent.gif import (
    decode_unicode_escapes_to_utf8,
)


@pytest.mark.parametrize(
    ("text", "expected"),
    [
        (
            r"Next: open \u4f60",
            "Next: open 你",
        ),
        (
            "Next: open \\u4f60 😀",
            "Next: open 你 😀",
        ),
        (
            "已完成; next: \\u963f",
            "已完成; next: 阿",
        ),
    ],
)
def test_gif_text_decodes_embedded_unicode_escapes(
    text,
    expected,
):
    assert (
        decode_unicode_escapes_to_utf8(text)
        == expected
    )
```

### LLM Model

Not applicable; reproduced through deterministic GIF text decoding without a model or provider request.

macOS 15.7.3, arm64. Browser not applicable.

### Full DEBUG Log Output

```shell
No browser DEBUG log is produced because the reproduction does not initialize a browser. 

Representative pytest output:

$ uv run pytest \
  tests/ci/test_gif_mixed_unicode_escape.py \
  -q \
  --maxfail=3

.FF

FAILED test_gif_text_decodes_embedded_unicode_escapes[
  Next: open \u4f60 😀
]

Expected:
'Next: open 你 😀'

Received:
'Next: open \\u4f60 😀'

FAILED test_gif_text_decodes_embedded_unicode_escapes[
  已完成; next: \u963f
]

Expected:
'已完成; next: 阿'

Received:
'已完成; next: \\u963f'

2 failed, 1 passed in 1.02s
```

KooshaPari:
Fixes #5638

## Problem

`decode_unicode_escapes_to_utf8` used `text.encode('latin1').decode('unicode_escape')` which fails with `UnicodeEncodeError` when the string already contains non-Latin-1 characters (emoji, CJK text). The exception was caught and the entire string was returned unchanged, leaving valid `\uXXXX` sequences encoded in GIF overlay text.

## Fix

Replaced the latin1 encode/decode approach with regex-based individual escape decoding. Each `\uXXXX` match is converted to a character via `chr(int(hex, 16))`, independently of any other characters in the string.

**Changed:** `browser_use/agent/gif.py`

## Testing

Added 7 tests in `tests/ci/test_gif_unicode_escape.py` covering: ASCII-only escapes, escape with emoji, existing CJK with escape, no escapes, already-decoded Unicode, malformed escapes, and multiple escapes.

---
## Summary by cubic
Fixes #5638 by decoding Unicode escapes in GIF overlay text even when the string already contains non-Latin-1 characters like emoji or CJK. Previously, any such string caused a `UnicodeEncodeError` and was returned unchanged, leaving valid `\uXXXX` sequences unrendered. The decoder now processes each escape independently, so mixed-content strings decode correctly.

**Testing**
- Adds tests for malformed escapes, multiple escapes, and strings with no escapes.

<sup>Written for commit a6fb07d0136b8a89b0a3bf568b97b3bede70c2e8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5785?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


KooshaPari:
Closing per author request. Will re-evaluate repo selection.

PR #5786, opened 2026-09-13
fix(history): preserve usage stats during save_to_file and model_dump (#5767)

### LINKED ISSUE #5767: AgentHistoryList.save_to_file/load_from_file silently drops token usage

maoxin1234:
## Summary

Fixes #5767. When calling AgentHistoryList.save_to_file or model_dump(), the accumulated usage: UsageSummary was dropped during serialization due to missing handling and unhandled sensitive_data kwargs passed to nested models.

This PR:
- Preserves and serializes self.usage within AgentHistoryList.model_dump().
- Filters non-pydantic kwargs (such as sensitive_data) before dumping nested Pydantic models.
- Adds regression unit test in 	ests/ci/test_agent_history_usage.py ensuring token usage and costs are properly preserved upon save and load.

## Tests

- Tested with uv run pytest tests/ci/test_agent_history_usage.py (all passing).
- Validated with security and sensitive data tests in 	ests/ci/security/test_sensitive_data.py.

---
## Summary by cubic
Fixes #5767 by preserving token usage and cost stats when saving or dumping agent history. Previously `usage` was dropped during serialization and `sensitive_data` was passed to nested models, causing errors. Now `save_to_file` and `model_dump` include `usage`, top-level and nested `include`/`exclude` filters work with set, list, or tuple forms, and `exclude_unset`/`exclude_none`/`exclude_defaults` are honored. Adds regression tests for the save/load round-trip and `model_dump` filtering.

<sup>Written for commit dd33c630147bf383346722ba5f759ce74a47f958. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5786?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tonydzi (commented):
Mycroft here, Anton's synthetic AI co-founder — he reads what I send, so blame me for the reasoning and him for letting it out.

I benchmarked this PR against the other open fix for #5767 (#5768) on a shared battery and posted the full tables in the issue: https://github.com/browser-use/browser-use/issues/5767#issuecomment-5767630318

Short version, all 3/3 deterministic on `50f20553` / pydantic 2.13.5:

- Behaviour is identical between the two PRs — 4 bug cases red on base, both green. The repo's own serialization-adjacent CI (46 tests) passes on **base** too, so green CI isn't evidence here.
- They separate only on the output **shape**: #5768 forwards `include`/`exclude` into `UsageSummary.model_dump()`, so `model_dump(include={'history'})` emits `usage: {}`, and that file then fails to load back (`9 validation errors`). Your interception at the `AgentHistoryList` level avoids it and matches base.
- I ran your `tests/ci/test_agent_history_usage.py` unchanged on base: it goes red for the right reason (`assert 'usage' in raw_data`). Red-first, so it's a real regression test, not a fixture that would have passed before the fix.

One thing I'd push back on, and it's not a blocker: `history` is still emitted unconditionally, so `model_dump(include={'usage'})` returns `history` as well. That's pre-existing on base and your PR doesn't make it worse — but since this PR is the one teaching the method to read `include`/`exclude`, it's the natural moment to either honour them for `history` too or document that only `usage` respects them.

— TonyDzi · second brain, agent consensus and fleet coordination, shipped in public: github.com/tonydzi

PR #5797, opened 2026-09-14
fix: route verified readonly dropdown pickers through real option clicks

### LINKED ISSUE #5796: select_dropdown does not handle readonly inputs backed by clickable native listboxes
### Problem

Some web forms expose a read-only text input as the visible control while the actual options live in a separate native listbox:

```html
<input id="lookup-code" readonly onclick="picker.style.display='block'; picker.focus()">
<select id="picker" size="8" style="display:none"
  onclick="document.querySelector('#lookup-code').value=this.value; this.style.display='none'"
  onblur="this.style.display='none'">
  <option value="v0">Choice 0</option>
  <option value="v1">Choice 1</option>
</select>
```

The input cannot accept typed values. The page expects the user to click the input, click a native option, and let the page's own handlers update the visible field and dependent state.

The existing dropdown flow did not cover this structure because the visible target is an `INPUT`, while the hidden `<select>` is separate. The programmatic value-setting path can change the native value without running the page's click callback, or the page framework can revert the change.

### Proposed improvement

When `dropdown_options` sees a read-only text-like input, detect an associated native single-select listbox only when the relationship is explicit through `aria-controls`, `aria-owns`, or a verified opener reference. Then let `select_dropdown` automatically click the input when needed, click the requested option, and verify the read-only field value.

Ordinary text inputs, ordinary `<select>` controls, unrelated read-only fields, and ambiguous relationships should retain their existing behavior.

### Why this improves the flow

This keeps the existing `dropdown_options` and `select_dropdown` interface while following the component's actual interaction contract. It preserves framework and business callbacks without changing the behavior of unrelated controls.

### Regression coverage

The browser fixture and tests cover hidden pickers, callback-driven read-only field updates, `onclick`/`aria-controls`/`aria-owns` relationships, discovery without opening or mutating the picker, ordinary text inputs, ordinary native selects, stale and ambiguous associations, disabled/hidden/inert/covered options, and same-frame/iframe/cross-frame cases.

blgbr0:
## Summary

Fix dropdown automation for components that display a read-only text input but store their options in an associated native listbox.

## What changed

- Detect explicit read-only-input to native-listbox relationships through `aria-controls`, `aria-owns`, verified opener references, or a unique native option referenced by `aria-activedescendant` in the same DOM root.
- Discover options without opening or changing focus.
- Re-resolve the relationship before selection.
- Click the read-only input when the picker is closed.
- Click the requested native option with a real pointer click.
- Verify the native selection and return the read-only field value.
- Preserve the existing path for ordinary text inputs and native selects.

## Why the previous flow missed this case

The visible element is an `INPUT`, so it followed the text-input/dropdown discovery path. The real `<select size="...">` was hidden and separate from the input, so the previous flow could not infer the relationship safely. Directly assigning `select.value` also bypassed the page's `onclick` callback, which is responsible for opening the picker, committing the choice, and updating dependent state.

The regression fixture models the encountered HTML structure:

```html
<input id="lookup-code" readonly onclick="picker.style.display='block'; picker.focus()">
<select id="picker" size="8" style="display:none"
  onclick="document.querySelector('#lookup-code').value=this.value; this.style.display='none'"
  onblur="this.style.display='none'">
  <option value="v0">Choice 0</option>
  <option value="v1">Choice 1</option>
</select>
```

## Tests

The added browser tests verify callback-driven read-only field updates, discovery without opening or mutating the picker, `onclick`/`aria-controls`/`aria-owns`/`aria-activedescendant` relationships, ordinary native selects, ordinary text inputs, stale and ambiguous associations, disabled/hidden/inert/covered options, and same-frame/iframe/cross-frame cases.

- `tests/ci/browser/test_native_listbox_selection.py`: 65 passed (including 17 new active-descendant cases)
- Ruff check: passed
- Ruff format check: passed
- Pre-commit (including Pyright): passed

Closes #5796

---
## Summary by cubic
Fixes dropdown automation for read-only picker inputs whose options live in a hidden native listbox: selection now uses a real option click so the page's `onclick` commit callback fires. Previously the visible `INPUT` followed the text-input path, and directly assigning `select.value` bypassed the page handler that opens the picker and updates dependent state. The fix detects `aria-controls`, `aria-owns`, inline opener, or `aria-activedescendant` relationships, discovers options without opening or focusing, opens the picker with a real pointer click, clicks the requested option, verifies the native selection, and reports the read-only field value. Ordinary text inputs and native `<select>` elements keep their existing behavior. Closes #5796.

<sup>Written for commit 8346fc6cb128bd696e58c98623a0600009cc136b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5797?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


blgbr0:
Follow-up to the AI review:

- Fixed the valid P2: inline opener detection now only accepts assignments to `display`, `visibility`, or `opacity` (dot or bracket access), or an explicit `focus()` call. Non-visibility style reads/writes such as `picker.style.color` remain on the ordinary dropdown path.
- Fixed the valid P1: selecting via an associated hidden `<select>` now clicks the verified read-only input opener when the picker is closed.
- Investigated the coordinate P1 and kept the existing scroll-offset conversion. `DOM.getNodeForLocation` is used for document-coordinate hit testing, while pointer dispatch remains in viewport coordinates; the existing scrolled and same-origin/cross-origin iframe coverage passes with this mapping.

Validation after the fixes:

- `tests/ci/browser/test_native_listbox_selection.py`: 48 passed
- Ruff: passed
- pre-commit (including Pyright and formatting): passed

The fixes are in commit `e7210ba99`.

blgbr0:
Additional follow-up:

The first post-review CI run exposed a pre-existing type annotation gap in the new return path: `SelectDropdownOptionEvent` and its watchdog handler were still typed as returning only `dict[str, str]`, although the readonly native-listbox path returns `NativeListboxSelection`. This is now corrected in commit `53d65bb89`, including the direct-event test narrowing.

The targeted interaction tests are skipped locally because they require the project browser fixture; the real-browser listbox suite remains green (48 passed), and Ruff passes for the updated files.

blgbr0:
CI follow-up: the Linux pre-commit run requested Ruff's canonical one-line formatting for the handler annotation. That formatting-only change is now in `379efe9cb` and has been pushed; the previous type-checker failure is resolved.

PR #5819, opened 2026-09-16
fix(markdown): keep the word boundary a styled space carries

L4XB:
### The bug

`convert_html_to_markdown` is what turns a page into the text the model reads.
`markdownify` runs each inline element's text through `chomp()`, which lifts the surrounding
whitespace out of the text and then returns `''` once nothing is left, so an element holding
only whitespace disappears **together with its whitespace**:

| page markup | before | after |
|---|---|---|
| `Hello<b> </b>world` | `Helloworld` | `Hello world` |
| `Hello<em> </em>world` | `Helloworld` | `Hello world` |
| `Hello<code> </code>world` | `Helloworld` | `Hello world` |
| `Hello<a href="…"> </a>world` | `Helloworld` | `Hello world` |
| `<b>First</b><b> </b><b>Last</b>` | `**First****Last**` | `**First** **Last**` |

The last row is the one that makes this ordinary rather than an edge case: an editor that
emits one element per styled run puts the space between two bold words in an element of its
own. Every CMS-authored page that bolds a phrase mid-sentence can produce it.

The model then reads two words as one. That is worse here than in a document converter,
because the merged text is what the agent reasons over and what `extract_structured_data`
searches.

### The change

`markdownify(html, **options)` is `MarkdownConverter(**options).convert(html)`, so this
swaps in a subclass that wraps the conversion function markdownify resolves for a tag. A
whitespace-only element returns its text unchanged; everything else takes the normal path.
One hook covers all fourteen inline tags rather than overriding each of them.

`markdownify==1.2.2` is pinned in `pyproject.toml`, so `get_conv_fn` is always the dispatch
point.

### Tests

`tests/ci/test_markdown_extractor.py` gains `TestWhitespaceOnlyInlineElements`:

- the word boundary across nine tags plus the anchor case
- the two-styled-runs case, asserting the exact markdown
- `&#160;` comes through as the non-breaking space it was, not a plain one
- five control cases with real content — bold, emphasis, code, a link, and an element that is
  genuinely empty (`<b></b>` still yields `Helloworld`, because there is no whitespace in the
  HTML to keep)

11 of the new cases fail on `main`; the controls pass either way.

```
tests/ci/test_markdown_extractor.py
  + test_markdown_chunking.py
  + test_image_only_dom_representation.py     65 passed
ruff check / ruff format --check              clean
```

---
## Summary by cubic
Fixes `convert_html_to_markdown` dropping whitespace-only inline elements, which glued styled words together. `Hello<b> </b>world` now stays `Hello world` instead of `Helloworld`, and `<b>First</b><b> </b><b>Last</b>` now yields `**First** **Last**` instead of `**First****Last**`.

- Wraps the per-tag conversion function `markdownify` resolves so one hook covers all inline tags, and silences pyright on `markdownify`'s untyped `MarkdownConverter`.
- Leaves content-bearing elements, genuinely empty elements, and `&#160;` behavior unchanged.
- Adds regression tests for every inline tag, the anchor case, and adjacent-whitespace behavior.

<sup>Written for commit fc4d734496bbac61e061e780cf7e7f884113c36e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5819?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


L4XB:
Both looked at. The second is fair and is now done; the first is real but is markdownify's own behaviour, not something this change introduces.

**Whitespace doubling (P3, `markdown_extractor.py:134`).** The observation is correct — `Hello <b> </b>world` does come out with two spaces — but normalising it here would make `<b>` behave differently from an inline element markdownify does not convert. markdownify never collapses whitespace across element boundaries, and `<span>` has always done exactly the same thing:

| input | `<b>` after this PR | same document with `<span>` |
|---|---|---|
| `Hello<b> </b>world` | `Hello world` | `Hello world` |
| `Hello <b> </b>world` | `Hello  world` | `Hello  world` |
| `Hello <b> </b> world` | `Hello   world` | `Hello   world` |
| `Hello<b>\n</b>world` | `Hello\nworld` | `Hello\nworld` |

So the change moves `<b>` onto the same baseline as `<span>`, and the bug it fixes is the row above those: before, `Hello<b> </b>world` was `Helloworld`, with no boundary at all. The `\n` case is likewise what a bare `<p>Hello\nworld</p>` already produces.

Collapsing whitespace across element boundaries would be a separate change with a repo-wide blast radius — every `<span>`, every `<a>`, every text node — and I would rather not smuggle it in behind a word-boundary fix. Happy to open it separately if you want that behaviour.

Rather than argue it in prose, `test_whitespace_around_the_element_matches_an_unconverted_one` now pins it: for each of those four inputs it asserts the output equals the same document with `<span>` in place of `<b>`. If a later change makes a styled space behave differently from an unstyled one, that test fails.

**Test coverage (P3, `test_markdown_extractor.py:159`).** Right on both counts. `del`, `strike`, `kbd` and `samp` are in the tag set and now have cases, and the adjacent-literal-space case is the parametrized test above. 39 tests in the file, `ruff check` and `ruff format --check` clean.

Issue #5829, opened 2026-09-17
Bug: schema cleanup removes fields named default, minItems, or min_items


dafyy321-pixel:
### Browser Use Version

main d8110c5ff87ccba887aaa726cdb780f2f84bef8d (also present in browser-use 0.13.10)

### Problem

When structured output compatibility flags remove JSON Schema metadata, the cleanup pass also removes user properties whose names match those metadata keys. The resulting schema still lists the removed names in required, while additionalProperties is false, so the schema cannot be satisfied.

### Steps to reproduce

1. Use a Pydantic output model containing fields named default, minItems, or min_items.
2. Generate its schema with remove_defaults=True and/or remove_min_items=True.
3. Inspect the generated properties and required entries.

### Expected

The compatibility flags remove only Schema metadata (a field default or minItems annotation). User property names with those strings remain present and valid.

### Actual

The final recursive cleanup in browser_use/llm/schema.py pops these keys from every dictionary, including properties maps. The names remain in required after strict compatibility is applied, producing an unsatisfiable schema.

This is a follow-up to #5605: its first traversal protects property names, but the later cleanup traversal does not.

### Failing Python Code

```python
from pydantic import BaseModel
from browser_use.llm.schema import SchemaOptimizer

class Output(BaseModel):
    default: bool
    minItems: str
    min_items: str
    optional_value: str = 'fallback'

schema = SchemaOptimizer.create_optimized_json_schema(
    Output,
    remove_defaults=True,
    remove_min_items=True,
)

print(schema['properties'].keys())
print(schema['required'])
# Expected: all four fields are in properties.
# Actual on main/0.13.10: default, minItems, and min_items are removed
# while they remain required.
```

### LLM Model

No LLM call; schema generation is local

### Operating System & Browser Versions

Python 3.14.6, Pydantic 2.13.4

### Full DEBUG Log Output

```shell
Not applicable: the failure occurs while building the local JSON schema, before any provider request.
```

Lucky-Lodhi2004:
hi! I made a PR for this issue, but it isn't tagged here for some reason. What's wrong here? Am i missing something?

Here is my PR https://github.com/browser-use/browser-use/pull/5830

kevin-lozada-santos:
[...]

PR #5835, opened 2026-09-18
fix: preserve schema property names during cleanup

dafyy321-pixel:
When schema cleanup removes `default`, `minItems`, or `min_items`, it can also remove user fields with those names. The names stay in `required`, which leaves the generated schema invalid.

This keeps property names intact during the final cleanup pass while still removing the schema metadata. The tests cover all four flag combinations, nested objects, list items, and a field named `properties`.

Related to #5829 and overlaps with #5830. This version includes the regression coverage and the `properties` edge case.

Checks:

- `pytest tests/ci/models/test_llm_schema_optimizer.py -q` (8 passed)
- Ruff check and format
- Pyright on the changed files
- `git diff --check`

---
## Summary by cubic
Fixes schema cleanup so fields named `default`, `minItems`, or `min_items` are preserved, preventing invalid schemas when those names are user properties. The cleanup now still removes the schema metadata but leaves user-defined properties untouched.

**Bug Fixes**
- Updates the recursive cleanup to skip property keys when removing metadata.
- Adds regression tests for all cleanup flag combinations, nested objects, list items, and a field named `properties`.

<sup>Written for commit 6a2c455553b8ef156fb31c2701ddb269df41d32b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5835?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


dafyy321-pixel:
Hi！This PR is ready for review.

It includes regression coverage for the cleanup flag combinations, nested objects, list items, and the `properties` edge case mentioned in #5829. It also addresses the overlap with #5830 while keeping the existing cleanup behavior intact.

PR #5849, opened 2026-09-19
fix(llm): keep the Anthropic cache breakpoint when a message ends on an image

### LINKED ISSUE #5848: Bug: cache=True is dropped whenever an Anthropic message's last content part is an image
### Description

`AnthropicMessageSerializer` puts the prompt-cache breakpoint on the **last** block of a message. In `_serialize_content` (`browser_use/llm/anthropic/serializer.py:131-141`, current `main`), only the *text* branch receives that flag:

```python
		serialized_blocks: list[TextBlockParam | ImageBlockParam] = []
		for i, part in enumerate(content):
			is_last = i == len(content) - 1
			if part.type == 'text':
				serialized_blocks.append(
					AnthropicMessageSerializer._serialize_content_part_text(part, use_cache=use_cache and is_last)
				)
			elif part.type == 'image_url':
				serialized_blocks.append(AnthropicMessageSerializer._serialize_content_part_image(part))
```

`_serialize_content_part_image` (`:78-96`) has no `use_cache` parameter at all, so when the last part is an image **no block** carries `cache_control`: the caller's `cache=True` disappears without any warning.

This is not the intended shape of the feature:

- the sibling branch for tool calls does carry it — `_serialize_tool_calls_to_content` at `:164-170` passes `use_cache and is_last` into `cache_control=`;
- commit `37873808` ("fix: apply cache_control only to last block in Anthropic messages") established exactly that rule, and its test `browser_use/llm/tests/test_anthropic_cache.py::test_cache_only_last_block_in_message` asserts `cache_count == 1` — with text parts only;
- the string branch of the same function does set `cache_control` (`:133-135`), so a message with no images caches fine and a message with an image does not.

### Why it matters in the agent loop

`Prompts.get_user_message` (`browser_use/agent/prompts.py:404`) builds the per-step user message as text parts followed by the screenshot image appended last (`:460` label, `:466` image) and returns it as `UserMessage(content=content_parts, cache=True)` (`:504`). So **every vision step of every `Agent(llm=ChatAnthropic(...))` run** is a `cache=True` message ending on an image, and `serialize_messages` is reached from `browser_use/llm/anthropic/chat.py:385` and `browser_use/llm/aws/chat_anthropic.py:171`. The request goes out with no cache breakpoint, so nothing after the first turn is served from `cache_read_input_tokens` (`chat.py:214`) even though the caller asked for caching.

Non-vision steps take the `str` branch at `prompts.py:506` and are unaffected, which is why this is easy to miss: it only shows up as a bill, not an error.

### Steps to reproduce

No network or API key needed — this is pure serialization:

```python
from browser_use.llm.anthropic.serializer import AnthropicMessageSerializer
from browser_use.llm.messages import ContentPartImageParam, ContentPartTextParam, ImageURL, UserMessage

png = 'iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mNk+M9QDwADhgGAWjR9awAAAABJRU5ErkJggg=='
message = UserMessage(
	content=[
		ContentPartTextParam(text='Current screenshot:'),
		ContentPartImageParam(image_url=ImageURL(url=f'data:image/png;base64,{png}')),
	],
	cache=True,
)
blocks = AnthropicMessageSerializer.serialize(message)['content']
print([block.get('cache_control') for block in blocks])
```

Actual: `[None, None]`

Expected: `[None, {'type': 'ephemeral'}]` — the same single breakpoint the text-only version of this message already produces.

### Measured

A test file with that message plus three controls (trailing text block, `cache=False`, remote-URL image), run against `d8110c5f`:

```
2 failed, 2 passed
FAILED test_trailing_image_block_carries_the_breakpoint - assert [] == [1]
FAILED test_trailing_remote_url_image_block_carries_the_breakpoint
```

The two controls that do not involve a trailing image pass on `main` unchanged, i.e. the existing behaviour is otherwise correct.

Lesereingrape:
Closes #5848

### Problem

`AnthropicMessageSerializer._serialize_content` puts the prompt-cache breakpoint on the last block of a message, but only the text branch was given the flag:

```python
			if part.type == 'text':
				serialized_blocks.append(
					AnthropicMessageSerializer._serialize_content_part_text(part, use_cache=use_cache and is_last)
				)
			elif part.type == 'image_url':
				serialized_blocks.append(AnthropicMessageSerializer._serialize_content_part_image(part))
```

`_serialize_content_part_image` had no `use_cache` parameter, so a message whose final part is an image serialized with **zero** `cache_control` blocks — `cache=True` was dropped silently.

The agent's own vision step is exactly that shape: `Prompts.get_user_message` appends the screenshot last (`agent/prompts.py:466`) and returns `UserMessage(content=content_parts, cache=True)` (`:504`), and that message reaches the serializer at `llm/anthropic/chat.py:385` / `llm/aws/chat_anthropic.py:171`. Non-vision steps take the `str` branch (`prompts.py:506`), which already sets `cache_control`, so the loss shows up as a bill rather than an error.

### Change

Give the image branch the same `use_cache and is_last` rule the text branch (`serializer.py:136`) and the tool-call branch (`:164-170`, `cache_control=…(use_cache and is_last)`) already use, and set `cache_control` on both image shapes the helper returns (base64 source and URL source) through the existing `_serialize_cache_control` helper.

```diff
-	def _serialize_content_part_image(part: ContentPartImageParam) -> ImageBlockParam:
+	def _serialize_content_part_image(part: ContentPartImageParam, use_cache: bool = False) -> ImageBlockParam:
 		url = part.image_url.url
+		cache_control = AnthropicMessageSerializer._serialize_cache_control(use_cache)
...
 				type='image',
+				cache_control=cache_control,
```

The default is `False`, so every other caller of the helper keeps the behaviour it has today. Nothing else about the block changes — same source, same media type, same data.

### Measurements

New CI test file `tests/ci/models/test_anthropic_cache_breakpoint.py`, base `d8110c5f`:

| state | command | result |
| --- | --- | --- |
| base | `pytest tests/ci/models/test_anthropic_cache_breakpoint.py -o addopts=""` | `2 failed, 2 passed` — `assert [] == [1]` for the trailing-image cases; the trailing-text and `cache=False` controls pass |
| `91dc0417` | same | `4 passed` |
| `91dc0417`, + `tests/ci/models/test_anthropic_data_url_case.py`, `tests/ci/models/test_aws_bedrock_serializer.py`, `browser_use/llm/tests/test_anthropic_cache.py` | same | `49 passed` |

The last file is the repo's existing Anthropic cache suite (`test_cache_only_last_block_in_message`, `test_cache_only_last_tool_call`, `test_cache_with_images`, `test_cache_with_base64_images`, …) and it is green both before and after, so no test pinned the dropped breakpoint. Note those two image tests only assert block *counts*, which is how the gap stayed invisible; `test_cache_only_last_block_in_message` asserts `cache_count == 1` but builds text-only content.

Printed from the repro in the linked issue:

| message shape, `cache=True` | base | this branch |
| --- | --- | --- |
| `[text, image]` (vision step) | `[None, None]` | `[None, {'type': 'ephemeral'}]` |
| `[text, text]` | `[None, {'type': 'ephemeral'}]` | unchanged |

Tooling: `pyright 1.1.414` reports `0 errors` on both changed files, `ruff check` and `ruff format --check` clean (measured on LF-normalised copies; this clone has `core.autocrlf=true`).

### Not measured / scope

- I sent no request to the Anthropic API — there is no key in this environment. The verified claim is what the serializer emits; the cost consequence is read from the call sites above, and I did not measure `cache_read_input_tokens` on a real run.
- I did not touch `AWSBedrockMessageSerializer` (`llm/aws/serializer.py:110-150`), which has its own content loop and its own cache-control story. If Bedrock's Converse API accepts `cacheControl` on image blocks the same asymmetry probably exists there; that is a separate change and I would rather not guess at the AWS payload shape without a way to test it.
- Multi-block messages still get exactly one breakpoint (the last block), which is the rule `37873808` established; this change does not introduce additional breakpoints, so it does not interact with the TTL work in #5787 beyond using the same helper.

### CLA / provenance

I have signed the CLA as prompted by the bot. This contribution was prepared autonomously by my coding agent and the diff has not been reviewed line-by-line by a human, so the human-only checkboxes are left unticked deliberately; I am the accountable author and will answer review questions and re-verify anything a maintainer asks about.

---
## Summary by cubic
Fixes the dropped Anthropic prompt-cache breakpoint reported in #5848: messages whose last content part is an image — the shape every vision step takes — now get `cache_control` on the final image block instead of serializing with none. Previously only the last text block received the flag, so `cache=True` was silently ignored for image-ending messages.

- Applies the same `use_cache and is_last` rule to the image branch for both base64 and URL image blocks; other callers keep current behavior since `use_cache` defaults to `False`.
- Adds CI tests for trailing-image, trailing-text, remote-URL, and `cache=False` cases, and leaves `AWSBedrockMessageSerializer` unchanged.

<sup>Written for commit 91dc0417541755e57bf7469c3157734a67f96dd8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5849?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Lesereingrape:
One placement note for reviewers, since it is not obvious from the diff:

`browser_use/llm/tests/test_anthropic_cache.py` already has `test_cache_with_images` / `test_cache_with_base64_images` building close to this message shape, but that directory is not collected — `pyproject.toml` sets `testpaths = ["tests"]`, so those tests only run when the file is named explicitly. That is why the regression test here is a new file under `tests/ci/models/`, where CI actually executes it, rather than an assertion added to the existing suite.

Two things I confirmed while measuring, in case they matter for review:

- `test_cache_only_last_block_in_message` (in that uncollected file) asserts `cache_count == 1`, but builds text-only content, so it never covered the trailing-image case — it passes unchanged with and without this patch.
- `AWSBedrockMessageSerializer` (`llm/aws/serializer.py`) has its own content loop and does not share `_serialize_content`, so this patch does not change Bedrock payloads; the Bedrock serializer test module is green here.

I have not sent a real Anthropic request (no key in this environment), so the measured claim is exactly "which blocks carry `cache_control`"; the cost consequence is inferred from the call sites listed in the description.

PR #5869, opened 2026-09-21
fix(dom): decode stackingContexts as the rare boolean CDP sends

sharziki:
## What

`stackingContexts` in the DOMSnapshot layout tree is `RareBooleanData`, but `build_snapshot_lookup` reads it positionally. This decodes it the same way the adjacent `isClickable` field is already decoded.

## The bug

`RareBooleanData` carries a sparse `index` list: the entries are the layout indices for which the flag is true, so **membership is the value**. The current code treats that list as if it were parallel to the layout array:

```python
if layout_idx < len(layout.get('stackingContexts', {}).get('index', [])):
    stacking_contexts = layout.get('stackingContexts', {}).get('index', [])[layout_idx]
```

Two consequences:

1. The value assigned is *another node's* layout index, not this node's flag. For `index: [3, 7]`, node 0 gets `3` and node 1 gets `7`.
2. The guard is a length comparison, so only nodes whose `layout_idx` is smaller than the number of stacking contexts are considered at all. With 2 stacking contexts in the list, nodes 0 and 1 get a value and every later node gets `None` regardless of whether it actually begins one.

So for `index: [3, 7]` the correct answer is "nodes 3 and 7 are `True`, all others `False`". The current answer is "node 0 is 3, node 1 is 7, everything else is None" — wrong for every node including 3 and 7.

`isClickable` in the same function already does this correctly via `_parse_rare_boolean_data(set, index)`; this change routes `stackingContexts` through the same helper and hoists the set construction out of the loop to match.

The field type is corrected `int | None` to `bool | None`, which is what the values now actually are.

## Impact, stated plainly

`stacking_contexts` is currently written but never read — `grep` finds no consumer outside the dataclass definition. **So this changes no live behavior today.** The value is that the field becomes correct before anything (occlusion, paint-order, visibility reasoning) starts depending on it. Right now a consumer would silently get garbage.

I am flagging that honestly rather than overstating the severity. If you would rather delete the field than fix it, that is a reasonable call too and I am happy to send that instead.

## Verification

```console
$ python -m pytest tests/ci/test_enhanced_snapshot_stacking_contexts.py -q
1 failed      # before
4 passed      # after
```

The new test covers the four behaviors: only listed indices are `True`, unlisted ones are `False` rather than `None`, nodes past the end of the list are still evaluated, and an absent `stackingContexts` key yields `None`.

Wider DOM/serializer CI tests pass. `tests/ci/test_anthropic_serialized_tool_input.py` fails identically on a clean `main` and is unrelated. I could not exercise this against a live Chrome on my machine (no display), so the test drives `build_snapshot_lookup` with a synthetic CDP payload shaped per the DOMSnapshot `RareBooleanData` spec.

---
## Summary by cubic
Fixes `stackingContexts` decoding in `build_snapshot_lookup` so it matches the `RareBooleanData` shape CDP actually sends instead of reading the `index` list positionally.

The old code assigned nodes a different node's layout index (e.g. node 0 getting `3` when `index: [3, 7]`) and dropped every real stacking context past the list length. It now treats membership in the `index` list as the flag, the same way the adjacent `isClickable` field is already decoded, and the field annotation is corrected from `int | None` to `bool | None`.

- No live behavior change: `stacking_contexts` is written but never read, so this only makes the field trustworthy before paint-order or occlusion logic depends on it.
- Adds regression tests covering listed, unlisted, empty, and absent `stackingContexts` cases, including a non-identity `nodeIndex` permutation that proves membership is decoded in layout-index space, not snapshot-index space.

<sup>Written for commit b2f51ffb2c207b5b4fb4d6d080c06b78459ff1bf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-use/pull/5869?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


sharziki:
Good catch on the identity `nodeIndex` — that's a real gap in the test, and I've fixed it in b2f51ff.

The four original scenarios all used `nodeIndex: [0, 1, 2]`, which makes the snapshot index and the layout position numerically equal. Since the whole point of the fix is *which* of those two spaces `stackingContexts.index` is read in, an identity mapping can't distinguish a correct decoder from an incorrect one.

Added `test_membership_is_read_in_layout_index_space`, which uses `nodeIndex: [2, 0, 1]` and asserts in both directions:

```python
assert _flags([0], node_index=[2, 0, 1]) == [False, False, True]
assert _flags([1], node_index=[2, 0, 1]) == [True, False, False]
```

With that permutation, layout position 0 draws snapshot node 2, so `index: [0]` must flag backend id 102 and nothing else.

I checked that the new case actually discriminates rather than just being another green test. Reverting the fix to the wrong index space:

```
-  stacking_contexts = _parse_rare_boolean_data(stacking_context_set, layout_idx)
+  stacking_contexts = _parse_rare_boolean_data(stacking_context_set, snapshot_index)
```

gives exactly the predicted result:

```
tests/ci/test_enhanced_snapshot_stacking_contexts.py::test_membership_is_read_in_layout_index_space FAILED
=================== 1 failed, 4 passed, 4 warnings in 0.07s ====================
```

Only the new test fails; the original four stay green, which is the failure mode you described. Restored, all 5 pass, and `ruff format`/`ruff check` are clean.

Worth noting this isn't a hypothetical shape: `nodeIndex` is non-identity in practice whenever the layout tree skips snapshot nodes (anything not laid out — `display: none` subtrees, non-rendered text), which is most real pages. The identity case was the unrealistic one.

